PR #19, opened 2015-01-05
Update the default db config file.

angaither:
Without an isset this causes undefined index errors if there is no .env file.

@jonuy


jonuy:
@angaither I did a similar thing in a4e89e9

For the sake of consistency do you want to either change this PR to use `getenv` or the other file to use `isset` like you do? I'm not particularly opinionated one way or another.

angaither:
@jonuy updated :+1:

jonuy:
:+1:

Issue #16, opened 2014-12-18
Modify behavior of PUT /campaigns/:nid/reportback


jonuy:
- Should return the `updated_at` value, instead of the `created_at` value.
- Should just update non-null values provided in the request. (ie - if "why_participated" already has a value in the doc, don't set it to null if it's not in the PUT request)

jonuy:
Actually, #20 took care of all of this too.

PR #20, opened 2015-01-06
Updates to CampaignController

### LINKED ISSUE #15: Return timestamps from GET /users/campaigns results
Currently the request returns everything but the timestamps.

### LINKED ISSUE #17: Save and return sid and rbid as ints
When doing a GET on /users/campaigns, the `rbid` and `sid` should be returned as ints instead of strings.

### LINKED ISSUE #18: Campaign timestamps should use accessors and mutators
Use setCreatedAt and getCreatedAt instead of `$campaign->created_at->format('Y-m-d H:i:s')`

https://github.com/DoSomething/user-api/blob/dev/app/controllers/CampaignController.php#L113

jonuy:
#### What's this PR do?
Modifies PUT requests for Campaigns. Like the User model, it no longer requires the full resource to be included in the request.

Also doing some additional validating and formatting like rbid, sid to ints and dates being formatted to `Y-m-d H:i:s`.

###### CampaignController::updateReportback()
Changes here make it so that only non-null values would get updated on the resource.

###### CampaignController::reportback()
Has similar logic, but additionally creates a campaign entry in the user's document if one doesn't yet exist. The changes here also add some more validation and error handling than before.

###### Campaign.php
Removes the timestamps from the hidden property so that they can be returned in queries. Rules added to enforce sid, rbid, and quantity as integers, and file_url as a url. And then timestamp accessors that format the dates (which in this case can be MongoDate objects) to `Y-m-d H:i:s`.

Unit tests: `$ vender/bin/phpunit`

You can also submit queries using this Postman collection: https://gist.github.com/jonuy/5ef93fa434db6ce87ba7
The relevant ones here are Submit Sign Up, Submit Report Back, Update Report Back, and Get User's Campaign Data.

Closes #15 
Closes #17 
Closes #18 

Fixes the PUT part of #16

### DIFF
--- a/app/controllers/CampaignController.php
+++ b/app/controllers/CampaignController.php
@@ -76,7 +76,7 @@ class CampaignController extends \BaseController {
     $campaign->sid = $sid;
     $campaign = $user->campaigns()->save($campaign);
     $response = array(
-      'created_at' => $campaign->created_at->format('Y-m-d H:i:s'),
+      CAMPAIGN_RESPONSE::created_at => $campaign->created_at,
       SIGNUP_ATTRIBUTE::sid => $campaign->sid
     );
 
@@ -123,7 +123,7 @@ class CampaignController extends \BaseController {
       $campaign = $user->campaigns()->save($campaign);
 
       $response = array(
-        'created_at' => $campaign->created_at->format('Y-m-d H:i:s'),
+        CAMPAIGN_RESPONSE::created_at => $campaign->created_at,
       );
 
       $statusCode = 201;
@@ -146,7 +146,7 @@ class CampaignController extends \BaseController {
       $campaign = $user->campaigns()->save($campaign);
 
       $response = array(
-        'updated_at' => $campaign->updated_at->format('Y-m-d H:i:s'),
+        CAMPAIGN_RESPONSE::updated_at => $campaign->updated_at,
       );
     }
 
@@ -196,11 +196,19 @@ class CampaignController extends \BaseController {
     $campaign = $user->campaigns()->save($campaign);
 
     $response = array(
-      'updated_at' => $campaign->updated_at->format('Y-m-d H:i:s'),
+      CAMPAIGN_RESPONSE::updated_at => $campaign->updated_at,
       REPORTBACK_ATTRIBUTE::rbid => $campaign->rbid
     );
 
     return Response::json($response, 200);
   }
 
+}
+
+/**
+ * Abstract class defining string values for response properties.
+ */
+abstract class CAMPAIGN_RESPONSE {
+  const created_at = 'created_at';
+  const updated_at = 'updated_at';
 }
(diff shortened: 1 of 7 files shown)


app/controllers/CampaignController.php
```diff
@@ -84,42 +92,67 @@ public function signup($id)
...
+        $campaign->quantity = Input::get(REPORTBACK_ATTRIBUTE::quantity);
+      }
+
+      if (!is_null(Input::get(REPORTBACK_ATTRIBUTE::why_participated))) {
+        $campaign->why_participated = Input::get(REPORTBACK_ATTRIBUTE::why_participated);
+      }
+
+      if (!is_null(Input::get(REPORTBACK_ATTRIBUTE::file_url))) {
+        $campaign->file_url = Input::get(REPORTBACK_ATTRIBUTE::file_url);
+      }
```
angaither: any reason you're grabbing each input individually instead of getting them all & mass assigning here?

jonuy: because I didn't know such a thing existed

app/controllers/CampaignController.php
```diff
@@ -84,42 +92,67 @@ public function signup($id)
...
+
+      $campaign = $user->campaigns()->save($campaign);
+
+      $response = array(
+        CAMPAIGN_RESPONSE::updated_at => $campaign->updated_at,
```
angaither: I had trouble having the correct values save here since laravel provided them, maybe it's different in mongo, but I remember having to use `Laravel::unguard()`

app/controllers/CampaignController.php
angaither: iz nice.

app/controllers/CampaignController.php
jonuy: I double checked this on my tests and the timestamps seemed to update fine when it'd call `save`.

jonuy:
@angaither Changed it so that it uses mass assigning instead of individually getting each input. Also changed the query in `updateReportback` to query for both rbid and nid instead of just rbid. I think that plus the way I was testing was why I was running into problems yesterday.

app/controllers/CampaignController.php
```diff
@@ -84,42 +92,58 @@ public function signup($id)
...
+    $campaign = $user->campaigns()->where('nid', '=', $nid)->first();
+
+    if (!($campaign instanceof Campaign)) {
+      $campaign = new Campaign;
+      $campaign->nid = $nid;
+      $campaign->rbid = $rbid;
+      $campaign->quantity = Input::get('quantity');
+      $campaign->why_participated = Input::get('why_participated');
+      $campaign->file_url = Input::get('file_url');
+      $campaign = $user->campaigns()->save($campaign);
```
angaither: could mass assign these here as well....

angaither:
Other than that, looking :+1: 
Glad you figured out the strange `Route::put` issue.... oy....

app/controllers/CampaignController.php
jonuy: Good point. Will do that too.

PR #21, opened 2015-01-08
Update local config.

angaither:
fixes `Config should be a string not array` error


jonuy:
Not sure why it's still showing a diff here, but I fixed this in #20 too.

:+1: if I'm not seeing something correctly and this still needs to get merged in. Otherwise, I think this is already resolved.

angaither:
@jonuy yup, not sure what's going on. I didn't see you already resolved it when I was making it...whoops

PR #23, opened 2015-01-15
Added a step after deploy to trigger runscope tests

jonuy:
There's a bucket of tests that are setup in Runscope to test Northstar on production. These tests act as additional validation of deployed code.

The curl request alone is enough to trigger the tests, and the `RUNSCOPE_TRIGGER_URL` has been setup in wercker.

cc: @blisteringherb


blisteringherb:
:+1:

angaither:
so fetch

Issue #22, opened 2015-01-13
Return all users


angaither:
User controller index, if there are no params sent, return all users with a paginatior

PR #24, opened 2015-01-16
Added return all users json.

angaither:
This is a demo for the aurora app, will need to be paginated.
Refs #22

@jonuy

### DIFF
--- a/app/controllers/UserController.php
+++ b/app/controllers/UserController.php
@@ -32,6 +32,11 @@ class UserController extends \BaseController {
     if($user instanceof User) {
       return Response::json($user, 200);
     }
+    else {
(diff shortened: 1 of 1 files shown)


jonuy:
I think it'd make more sense to only return all users if no resource identifiers are given. As this is right now, I could put in a wrong email and still get a 200 returned.

angaither:
Ah yes, good point.... well I think ideally this would accept a page argument, so it's not returning ALL users... that's why it's so lo-fi.
Do you mind if I merge this so I can show users in the demo and fix it up later?

jonuy:
oh yea, of course. :+1:

PR #27, opened 2015-01-22
Send paginated users data as response.

### LINKED ISSUE #22: Return all users

angaither:
Fixes #22 

@jonuy 
turns out laravel sends the info with the json response already, so there was no need for Fractal. 
oohhh @taylorotwell Thank you.

### DIFF
--- a/app/controllers/UserController.php
+++ b/app/controllers/UserController.php
@@ -1,5 +1,6 @@
 <?php
 
+
 class UserController extends \BaseController {
 
   /**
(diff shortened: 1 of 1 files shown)


jonuy:
:+1:

Nice. I tested different combinations of `limit` and `page` queries. Are there any others I should try out @angaither ? I'll also update our wiki with the changes you've been making.

angaither:
Those are the only two!

Issue #34, opened 2015-01-29
Allow searching for users by source & id


angaither:
The `users.show` method should take additional, optional parameters to allow searching for users by source and ID

angaither:
Closing this for #38

We are going to add sources at the top level of a user document, because embedded collections cannot be queried without the parent object. 
(You cannot get a user by knowing from the us & the drupal uid is 123456, in the sources array structure)

PR #35, opened 2015-01-29
Change drupalyman to source

### LINKED ISSUE #29: Add source and id
Create a source  array on the user object
This will store the app the user came from and the user id in that particular app. 
User can have many sources

```
source {
  app : cgg
  user_id : 1232
}
```

angaither:
Fixes #29 

This removes `drupal_uid` from the code and DB...
Adding in a source array to the user document.

N.B. this removes the ability to search by `drupal_uid`
Will be fixed with #34

@jonuy

### DIFF
--- a/app/controllers/CampaignController.php
+++ b/app/controllers/CampaignController.php
@@ -9,17 +9,13 @@ class CampaignController extends \BaseController {
    * @return Response
   */
   public function index()
-  {   
+  {
     $user = '';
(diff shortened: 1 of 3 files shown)


jonuy:
This looks fine. :+1: 

It doesn't allow for a way to submit a source through the API though, right?

Any thoughts on how we should do it? Allow for it to be set in both the POST `/users` and PUT `/users/:id` requests? Should source be required in the POST request?

angaither:
Yes, I think that makes sense. 
Made an issue for it #36

PR #41, opened 2015-02-10
Sources at top level

angaither:
- Added sources back into the top level of the user document. 
- Allow for searching by a user by `drupal_id` or `cgg_id`
- Updated user GET requests, returns as many users as found.

- Updated `drupal_uid` --> `drupal_id` for consistency with other systems.

@jonuy


jonuy:
I'm sorta hesitant about storing the `drupal_id` value as a concatenation of country and number id. Do you think it might be faster and cleaner to modify the queries instead?
```
$country = Session::get("country") ?: "us"; 
$user = User::where("drupal_id", $id)->where("country", $country)->first();
```

jonuy:
:+1:

PR #42, opened 2015-02-11
Update campaign show get

angaither:
Simplified logic in getting a user campaign to match user GET request. 

@jonuy


jonuy:
Same comment on casting `_id` to int as what I left in #42.

:+1: otherwise

PR #44, opened 2015-02-11
Typecast Id fields as int to get results.

### LINKED ISSUE #43: Search by ID fields returns no results
Id fields are stored as ints, they need to be typecast to an int for any results to return.

angaither:
Apparently, some people do need labels.
Fixes #43

**You see us as you want to see us - in the simplest terms, in the most convenient definitions. You see us as a brain, an athlete, a basket case, a princess and a criminal. **

@jonuy

### DIFF
--- a/app/controllers/UserController.php
+++ b/app/controllers/UserController.php
@@ -76,7 +76,7 @@ class UserController extends \BaseController {
     $user = '';
 
     // Type cast id fields as ints.
-    if (strpos($term,'_id') !== false) {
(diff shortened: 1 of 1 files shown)


jonuy:
Isn't __GET__ `/users/_id/hash123abc` how we'd try to get a user based on the mongo id? Does this break that query?

angaither:
Yeah. :information_desk_person:

jonuy:
:+1:

PR #46, opened 2015-02-12
Use isEmpty before sending user response.

### LINKED ISSUE #45: User GET requests will return an empty object.
Found this when I found #43 
When searching and it found no results it will return an empty object back, the endpoint should return an error.

angaither:
Fixes #45 

@jonuy


jonuy:
:+1:

PR #54, opened 2015-02-20
Don't save null keys, check for value, not key.

### LINKED ISSUE #53: Setting NULL values on user object
When migrating data over, if you send a new user, it will grab all `editableKeys` even if there is no value in that key, and create that on the user object and set the value as null.

angaither:
Fixes #53 

@jonuy


jonuy:
:+1:

PR #56, opened 2015-02-20
Make it sparse

angaither:
The `email` and `mobile` fields should be sparse, not unique. 
This will allow people to have null, or no email/mobile set and still be 'unique'

Refs dosomething/voting-app#274

@jonuy


jonuy:
`sparse`! I have learned something new today.

:+1:

angaither:
thank @mshmsh5000 

because, so did i!

angaither:
Actually, this might need to be another migration... I'm not sure if that will work though......

PR #57, opened 2015-02-24
Fixes failing phpunit and runscope tests

jonuy:
#### What's this PR do?
Fixes failing phpunit and runscope tests

- `UserController.php`: previous logic would only save values that were null
- `CampaignTest.php` and `UserTest.php`: route to GET user data was previously modified to `/users/{term}/{id}`. This updates the tests to match that.

- `$ vendor/bin/phpunit`
- and Runscope tests should pass

@angaither


angaither:
:+1: oy. :grimacing:

PR #59, opened 2015-02-26
Api key generation

### LINKED ISSUE #48: Add new api keys/tokens
To the production database, for safety.

angaither:
Creates api key generation endpoint.
Must first be authenticated to create this (i.e. this will be done through Aurora)

@jonuy 

Fixes #48 
Refs DoSomething/mbc-userAPI-registration#10

### DIFF
--- a/app/controllers/KeyController.php
+++ b/app/controllers/KeyController.php
@@ -24,7 +24,14 @@ class KeyController extends \BaseController {
    */
   public function store()
   {
+    $app_name = Input::get('app_name');
+    $key = new ApiKey();
+    $key->app_id = snake_case(str_replace(' ', '', $app_name));
+    $key->api_key = str_random(40);
 
+    $key->save();
+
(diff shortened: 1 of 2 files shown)


app/controllers/KeyController.php
```diff
@@ -0,0 +1,66 @@
...
+  /**
+   * Store a newly created resource in storage.
+   * POST /keys
+   *
+   * @return Response
+   */
+  public function store()
+  {
+    // Get the app name from submission.
+    $app_name = Input::get('app_name');
```
jonuy: Do you think app_name should be required? While testing I just created one where no app_name was provided.

jonuy:
Just the one question above about maybe requiring app_name. Otherwise :+1:

angaither:
@jonuy updated.... not sure how specific errors should be. meeeh

jonuy:
I think that's sufficient :+1:

PR #60, opened 2015-03-05
Ssh tunnel

blisteringherb:
Adding a gateway to allow Northstar to be deploy via the AWS admin box.  Eventually this will mean that I can pass all traffic to Northstar through a load balancer which will eliminate the need for using public IPs on all Wercker deployments.

@angaither 

@mshmsh5000 just an FYI


angaither:
:+1: yay

blisteringherb:
Realized that if the gateway wasn't set this would probably fail.  Fixing that.

PR #64, opened 2015-04-02
firstOrCreate users.

### LINKED ISSUE #50: First or Create user store
The `user.store` method should first attempt to search for the user and update if needed.

Imports with users who already are in the system will fail if the user is already registered from another system.

angaither:
On user registration, if this user with email or mobile already exists, update the user, don't fail out of registration.
Otherwise, create the new user.

Fixes #50


jonuy:
:+1:

PR #69, opened 2015-04-08
Guzzle and droops

angaither:
Initial set up of an example request to drupal from Northstar.
Adding guzzle library

Also currently pointing the local api to staging, because connecting two vagrant boxes is really tough. 

We might need to look into this. 
http://docs.vagrantup.com/v2/virtualbox/networking.html


jonuy:
:tropical_drink: :+1:

PR #70, opened 2015-04-09
Kill this editable params stuff

angaither:
Removed editable params for [laravel fillable stuffs](http://laravel.com/docs/4.2/eloquent#mass-assignment)
Also added interests and agg_id fields


jonuy:
Thanks! :+1:

PR #71, opened 2015-04-09
Connect to local vagrant box instead of staging.

angaither:
This along with [this](https://github.com/DoSomething/ds-homestead/pull/5) allows for connecting to your local dev environment instead of to staging.


jonuy:
:computer: :+1:

Issue #63, opened 2015-04-01
Update user fields to also store values that message broker accepts


jonuy:
This is an example user doc from message broker. Northstar should be updated to support any of the fields here that we don't currently store - particularly because message broker will start funneling its users towards Northstar as well.

```
{
    "_id" : ObjectId("540be248a593a11d23d073f4"),
    "address1" : "123 Sesame St.",
    "birthdate" : ISODate("1000-00-00T05:00:00Z"),
    "campaigns" : [
        {
            "nid" : 2461,
            "signup" : ISODate("2014-09-07T04:42:50Z"),
            "_id" : ObjectId("540be249216e3b24533e51b6")
        }
    ],
    "city" : "Westminster",
    "college_name" : "Chapman University",
    "degree_type" : "Master",
    "drupal_register_date" : ISODate("2015-03-31T21:42:35Z"),
    "drupal_uid" : 1234,
    "email" : "xxxx@gmail.com",
    "first_name" : "Big",
    "gpa" : 1.xx,
    "hs_gradyear" : 2015,
    "hs_name" : "Bolsa Grande High School",
    "last_name" : "Bird",
    "major_name" : "Economics",
    "mobile" : "1234567890",
    "race" : "A",
    "religion" : "Buddhist",
    "role" : "HSstudent",
    "sat_math" : 000,
    "sat_verbal" : 000,
    "sat_writing" : 000,
    "source" : "niche",
    "state" : "CA",
    "subscribed" : 1,
    "zip" : "99999"
}
```

angaither:
@jonuy @DeeZone I'm not sure if we would want to store all of these fields, here's a couple I have questions about 
`role` : maybe I just don't know what the options could be here, confused by this field
`subscribed` :what does this field mean? is that mailchimp? or mobile commons? is it a list of things they are subscribed to or is it a binary option?
`source` : what is this? we have a souces id section on the user object.

DeeZone:
Most of the additional fields are coming from niche.com user imports.
- `role` : is niche.com specific to scholarship application. I get the details if interested?
- `subscribed` : the default use email subscription preference. This is replaced / updated with:
```
subscriptions {
  mailchimp: 0 or 1
  digest: 0 or 1
  banned: 0 or 1
}
```
when one of these settings are set.
- `source` : the source of the user, typically from imports. Current possible settings are `nice`, `hercampus` or `att-ichannel`.

angaither:
@jonuy @DeeZone 
[...]

PR #72, opened 2015-04-09
Added support for mb fields.

angaither:
this adds support for the easier mb fields, but ignores things we may still have questions about.
Refs #63


jonuy:
:horse_racing: :mortar_board: :school: ok... I can't find enough emoticons to actually fit the new fields... :+1:

PR #74, opened 2015-04-15
Interests array

### LINKED ISSUE #73: Interest field
make sure it can save, new items, and it's a comma separated list when things get added

```
 interests: basketball, space jam, Michael Jordon 
```

angaither:
save interests as an array, instead of a comma separated list

maybe this could be done with a mutator?? but this works for now.

Fixes #73

### DIFF
--- a/app/controllers/UserController.php
+++ b/app/controllers/UserController.php
@@ -54,6 +54,11 @@ class UserController extends \BaseController {
         Session::flash('country', $input['country']);
       }
       foreach($input as $key => $value) {
+
+        if ($key == 'interests'){
+          $interests = explode(',', trim($value));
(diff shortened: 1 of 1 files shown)


app/controllers/UserController.php
```diff
@@ -54,7 +54,11 @@ public function store()
         Session::flash('country', $input['country']);
       }
       foreach($input as $key => $value) {
-        if(!empty($value)) {
+        if ($key == 'interests'){
+          $interests = array_map('trim', explode(',', $value));
```
jonuy: ah good ole `array_map`. nice

jonuy:
:+1:

PR #76, opened 2015-04-16
Forward user registration events to droops.

### LINKED ISSUE #61: Forward registration events
Once a user registration event occurs, Norhtstar should be responsible for sending that data over to drupal (only if the user registration event has an email attached to it)

angaither:
This forwards all users who have emails but no drupal id into the drooops

Unclear how we handle drupal registration failures right now....

Fixes #61

### DIFF
--- a/app/Northstar/Services/Drupal.php
+++ b/app/Northstar/Services/Drupal.php
@@ -42,6 +42,7 @@ class DrupalAPI {
   public function register($user)
   {
     $user->birthdate = date('Y-m-d', strtotime($user->birthdate));
+    $user->user_registration_source = $user->source;
     $response = $this->client->post('users', [
       'body' => json_encode($user),
       ]);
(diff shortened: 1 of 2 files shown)


jonuy:
Cool, this seems good. I'm semi-concerned about the blocking nature of this call, but we can see if it's even a prob in the first place. With them all being in the same private cloud, hopefully talking back and forth will be super fast.

Only question I have is around the `source` property in Northstar. In Drupal it looks like it's expecting a client to pass that value into `user_registration_source` instead. How would you suggest handling this?

angaither:
yeah, I should add that `source` question as an issue. 
It should be pretty easy, we just add a fake property to the user before sending it.
The only question is if we send the string directly or if we should also send 'agg' or 'cgg' if it has an id from another site, that might get tricky, (and it's unclear if those people should even be in drupal...)

jonuy:
Sounds good. :+1: 

And my guess would've just been to send the string alone. Additionally sending ids from other sites like agg and cgg to Drupal, I'd generally vote no - though it could be something we have to consider on a case by case basis. But default point of view I think is for new things to use Northstar alone as its user db. And to methodically find ways of moving already existing things to use Northstar too.

PR #78, opened 2015-04-17
orWhere was causing false negative matches in the login method.

### LINKED ISSUE #77: Login with an incorrect email still succeeds
This was seen with a test account: `touchlab-dev@example.com`.

An attempt to login with `touchlab-dev@example.commmm` and the correct password also works.

angaither:
When a user doesn't have a mobile number submitted to the login, the `orWhere` was returning a user if there was no mobile and allowing people with incorrect emails to login
Fixes #77


jonuy:
:trophy:

Issue #82, opened 2015-04-23
User registration return user


angaither:
Currently registration returns a `created_at` and an `_id` 
we could/should return the whole user object

angaither:
Fixed with #86

PR #86, opened 2015-04-24
Return user response for login/registration

### LINKED ISSUE #83: Return user on user login
Currently only a subset of the user's object is returned on login:
```
{
  email,
  mobile,
  created_at,
  updated_at,
  _id,
  session_token
}
```

We should return the entire user object.

angaither:
fixes #83, #82


jonuy:
:+1:

PR #88, opened 2015-04-27
When a user registers, return an authenticated user.

### LINKED ISSUE #81: User registration logs user in
Upon registration, the should should also be logged in

angaither:
Fixes #81


jonuy:
:+1:

PR #89, opened 2015-04-27
Added wercker build step to update composer

jonuy:
This should fix wercker builds.


angaither:
:+1: yay! the build finally worked!

PR #90, opened 2015-04-28
Hello, Northstar.

DFurnes:
# Changes
 - Fixes broken unit tests, and configures Wercker to run unit tests on build. :traffic_light: 
 - Minor cleanup – documentation, code style, etc.

:construction:

### DIFF
--- a/app/controllers/CampaignController.php
+++ b/app/controllers/CampaignController.php
@@ -50,13 +50,13 @@ class CampaignController extends \BaseController {
   */
   public function signup($id)
   {
-    $input = Input::only(SIGNUP_ATTRIBUTE::sid);
+    $input = Input::only('sid');
     $campaign = new Campaign;
     if (!$campaign->validate($input)) {
       return Response::json($campaign->getValidationMessages(), 401);
     }
     else {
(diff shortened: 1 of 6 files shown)


angaither:
AWESOME. :+1: 

ALSO. one thing maybe we should consider is removing the drupal-y id
maybe instead of `sid` and `rbid` it's `signup_id` &  `reportback_id`

cc @jonuy

DFurnes:
That'd make a ton of sense to me. :+1:

jonuy:
Yea, I like that idea. @weerd even started going in that direction too with the reportbacks work he was doing on the Drupal Services side.

angaither:
:cool: :cool: :cool:

weerd:
Yep yep! I'm working on doing some `POST` calls and to pass through data I'm using properties like `user_id`, `reportback_item_id` etc. It just feels like it would be much clearer for devs posting to the API and that are not familiar with Drupal :)

DFurnes:
Removed steps that tried to run PHPUnit tests on Wercker. Seems like something needs to be configured on the box first for MongoDB to work there. :cry: A task for another day!

PR #91, opened 2015-04-30
Campaign signup endpoint

DFurnes:
Initial attempt at a `/campaigns/{id}/signup` endpoint. :construction: 

### Open Questions
 - How does Northstar authenticate to the Drupal API? Where do we handle permissions for things?
 - We're checking for an existing `Campaign` resource on Northstar, but some users may have signed up directly through Drupal. Is it bad that there are two error conditions for the same thing?
 - Is it unexpected/weird to have different behavior based on presence of the `user` parameter? I wonder if it makes more sense to have that always/never required.

/cc @angaither @jonuy

### DIFF
--- a/app/Northstar/Services/Drupal.php
+++ b/app/Northstar/Services/Drupal.php
@@ -47,7 +47,7 @@ class DrupalAPI {
         'body' => json_encode($user),
         ]);
       return $response->json();
-    } catch (Exception $e) {
+    } catch (\Exception $e) {
       // whatever.
       return;
     }
--- a/app/Northstar/Services/Drupal.php
+++ b/app/Northstar/Services/Drupal.php
@@ -1,27 +1,26 @@
 <?php namespace Northstar\Services\Drupal;
 
+use GuzzleHttp\Client;
+use Config;
+
 class DrupalAPI {
 
   protected $client;
 
   public function __construct()
   {
-    $base_url = \Config::get('services.drupal.url');
-    if (\App::environment('local')) {
-      $base_url .=  ":" . \Config::get('services.drupal.port');
-    }
-    $base_url .= '/api';
-    $version = \Config::get('services.drupal.version');
-    $client = new \GuzzleHttp\Client([
-      'base_url' => [$base_url . '/{version}/', ['version' => $version]],
-      'defaults' => array(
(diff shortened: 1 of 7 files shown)


app/Northstar/Services/Drupal.php
```diff
@@ -53,4 +52,28 @@ public function register($d_user)
...
+    $payload = [
+      'uid' => $drupal_id,
+      'source' => $source
+    ];
+
+    $response = $this->client->post('campaigns/' . $campaign_id . '/signup', [
+      'body' => json_encode($payload)
+    ]);
+
+    dd($response);
```
angaither: pewwww

angaither:
:cool:

app/Northstar/Services/Drupal.php
DFurnes: Whoopsie, but yeah totally not ready to merge. Just wanted to get the code out there so we could talk about it.

jonuy:
> How does Northstar authenticate to the Drupal API? Where do we handle permissions for things?

As far as I know, we haven't created a way for Northstar to auth with Drupal. Two possible solutions:
- We do a login to get the session and token info and save those as ... environment variables? ... to include in all NS requests to Drupal.
- A more robust solution where Northstar actually does a login request to get that info, stores it in persistent storage, and logs in again or refreshes the token in the case that these things can expire.
- Or something in between.

> We're checking for an existing Campaign resource on Northstar, but some users may have signed up directly through Drupal. Is it bad that there are two error conditions for the same thing?

- Not sure I understand fully what you're talking about here...

> Is it unexpected/weird to have different behavior based on presence of the user parameter? I wonder if it makes more sense to have that always/never required.

- It's a bit of hidden logic that feels kinda iffy. I was wondering if it'd make sense to have different named endpoints. One that's meant for general submission of a signup or reportback, and one that's just meant for Drupal to update ids with.

DFurnes:
Cleaned some things up, and going to merge this branch so it doesn't get too long-lived. The ` /campaigns/:id/signup` currently returns a `501 Not Implemented` error since it's... well, not implemented. :sweat_smile:

angaither:
:+1:

PR #95, opened 2015-05-06
Drupal auth

DFurnes:
# Changes
 - Moves authentication into new `AuthController`. This fits standard conventions in other Laravel projects, and separates authentication concerns from user resource concerns.
 - Add authentication to Drupal API class. The first request for an authentication cookie/token will make a login request against the Drupal API, and then cache that session for the following 30 minutes.
 - Other minor cleanups.

For review: @angaither @jonuy


app/Northstar/Services/Drupal.php
```diff
@@ -35,44 +94,62 @@ public function campaigns($id = NULL)
...
-   * Forward registration to drupal.
+   * Forward registration to Drupal.
+   * @see: https://github.com/DoSomething/dosomething/wiki/API#create-a-user
+   *
+   * @param \User $user - User to be registered on Drupal site
+   *
+   * @return int - Created Drupal user UID
    */
-  public function register($d_user)
+  public function register($user)
```
angaither: so, the reason this guy had a different var name and was't `user` is it was saving the `user_registration_source` to the northstar user

though, I guess if we used `fill->input` instead, that would fix this.

app/Northstar/Services/Drupal.php
```diff
@@ -35,44 +94,62 @@ public function campaigns($id = NULL)
...
+      'body' => json_encode($payload),
+      'cookies' => $this->getAuthenticationCookie(),
+      'headers' => [
+        'X-CSRF-Token' => $this->getAuthenticationToken()
+      ]
     ]);
 
-    return $response->sid;
+    $body = $response->json();
+    $sid = $body[0];
```
angaither: is this `sid` because it's coming back from drupal? can we call it `signup_id`?

app/controllers/UserController.php
```diff
@@ -64,8 +64,8 @@ public function store()
       if ($user->email && !$user->drupal_id) {
         try {
           $drupal = new Northstar\Services\Drupal\DrupalAPI;
-          $response = $drupal->register($user);
-          $user->drupal_id = $response['uid'];
+          $uid = $drupal->register($user);
```
angaither: same here `user_id`

app/Northstar/Services/Drupal.php
DFurnes: Yep, `sid` because droops. I think it could make sense to just call it a signup ID anyways though.

app/Northstar/Services/Drupal.php
DFurnes: It's an instance of the `User` model that's being passed though, right? If so, I'd vote for just naming it that even though its just being used to prepare the request.

angaither: yes it is. but it was adding an extra field to the Northstar user when it was called that.

app/Northstar/Services/Drupal.php
```diff
@@ -35,44 +94,65 @@ public function campaigns($id = NULL)
...
-      $response = $this->client->post('users', [
-        'body' => json_encode($user),
-        ]);
-      return $response->json();
-    } catch (\Exception $e) {
-      // whatever.
-      return;
-    }
+    $payload = $user->toArray();
+    dd($payload);
```
jonuy: Remove before merge?

app/Northstar/Services/Drupal.php
DFurnes: Updated this function to make an array from the model using `toArray()` which it modifies to set the payload for the request.

app/Northstar/Services/Drupal.php
```diff
@@ -35,44 +94,65 @@ public function campaigns($id = NULL)
...
+      'headers' => [
+        'X-CSRF-Token' => $this->getAuthenticationToken()
+      ]
     ]);
 
-    return $response->sid;
+    $body = $response->json();
+    $signup_id = $body[0];
+
+    if(!$signup_id) {
```
jonuy: :grimacing: It might just be better if we cleaned up the Drupal API response. For resources that already exist, we still return the sid. And the response instead looks something like...
```
{
  sid: 123,
  message: "Already exists"
}
```
... with a status code of ... 200? 304? 409?

DFurnes: Totally agreed. That will also allow us to gradually sync Northstar with existing signups.

angaither:
:+1:

PR #97, opened 2015-05-07
Auth tests

DFurnes:
# Changes
 - Move auth tests into their own test suite, following changes in #95.
 - Add an [EditorConfig](http://editorconfig.org).

For review: @angaither

### DIFF
--- a/app/tests/AuthTest.php
+++ b/app/tests/AuthTest.php
@@ -50,6 +50,10 @@ class AuthTest extends TestCase {
     // Response should include user ID & session token
     $this->assertArrayHasKey('_id', $data);
     $this->assertArrayHasKey('session_token', $data);
+
+    // Assert token exists in database
+    $tokenCount = Token::where('key', '=', $data['session_token'])->count();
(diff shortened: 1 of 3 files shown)


app/tests/AuthTest.php
```diff
@@ -0,0 +1,76 @@
...
+    parent::setUp();
+    Route::enableFilters();
+
+    Artisan::call('migrate');
+    $this->seed();
+
+    $this->server = array(
+      'CONTENT_TYPE' => 'application/json',
+      'HTTP_X-DS-Application-Id' => '456',
+      'HTTP_X-DS-REST-API-Key' => 'abc4324',
```
angaither: can we get this stuff from the config files?

angaither:
lookin :white_check_mark: :ok_hand: :ok_woman:

app/tests/AuthTest.php
DFurnes: Not sure, could be worth looking into in the future though.

PR #105, opened 2015-05-11
Reportbacks

DFurnes:
# Changes
 - Adds an initial stab at a `POST campaigns/{id}/reportback` endpoint, which posts a new reportback to the Drupal API. Comments out tests for now, since they have nasty external dependencies on Dr00ps API.

# Discussion
 - Drupal API just provides one create/update endpoint. We can sorta "spoof" `200` vs `201` responses based on whether a reportback exists in Northstar, and same with separate `POST` and `PUT` routes, but it might make sense to just merge them?

# Screenshots
screen shot 2015-05-11 at 10 51 24 am

screen shot 2015-05-11 at 10 44 58 am

I _think_ this is actually working fab, since all my other images on the reportback form in Drupal are broken too (so I'm assuming that's a random Drupal environment issue). But will continue investigating.

For review: @angaither @jonuy

### DIFF
--- a/app/controllers/CampaignController.php
+++ b/app/controllers/CampaignController.php
@@ -67,6 +67,11 @@ class CampaignController extends \BaseController {
     // Get the currently authenticated Northstar user.
     $user = User::current();
 
+    // Return an error if the user doesn't exist.
+    if(!$user->drupal_id) {
+      return Response::json('The user must have a Drupal ID to sign up for a campaign.', 401);
+    }
+
     // Check if campaign signup already exists.
     $campaign = $user->campaigns()->where('drupal_id', $campaign_id)->first();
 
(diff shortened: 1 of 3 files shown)


app/controllers/CampaignController.php
```diff
@@ -99,24 +99,65 @@ public function signup($campaign_id)
...
+
+    if($validator->fails()) {
+      return Response::json($validator->messages(), 401);
+    }
+
+    // Get the currently authenticated Northstar user.
+    $user = User::current();
+
+    // Return an error if the user doesn't exist.
+    if(!$user->drupal_id) {
```
angaither: should this be part of the validator?

app/tests/CampaignTest.php
```diff
@@ -79,24 +79,25 @@ public function testSubmitCampaignSignup()
...
-    $content = $response->getContent();
-    $data = json_decode($content, true);
-
-    // The response should return a 201 Created status code
-    $this->assertEquals(501, $response->getStatusCode());
-
-    // Response should be valid JSON
-    $this->assertJson($content);
+  {
+    // @TODO: Test has external dependency... need to mock DrupalAPI!
```
angaither: maybe we can send these requests to stage?

app/controllers/CampaignController.php
DFurnes: I don't think so, since it's not user input.

angaither:
hey this is :cool: :cool: :cool: :+1:

app/tests/CampaignTest.php
DFurnes: I think the proper solution is to mock `DrupalAPI` with something like [Mockery](), since anything using the real Drupal API would be brittle & leave a trail of test signups in its wake. Haven't gotten a chance to yet.

angaither: ooooo, :+1:

DFurnes:
Okay, can mergey merge then. Any thoughts on the discussion Q posted above?

angaither:
I think, for now we should spoof the endpoints with different results, this could help us with logging/ updating images and quantity counts

DFurnes:
Alright, will hook up the `PUT` endpoint in another PR. :+1:

Issue #100, opened 2015-05-08
Upgrade to Laravel 5


jonuy:
I'd like to be able to do Observer pattern type stuff like in https://laracasts.com/lessons/the-observer-design-pattern-in-php

:100:

DFurnes:
Closed with #111! :hammer:

PR #106, opened 2015-05-11
PSR-4

DFurnes:
# Changes
 - Use PSR-4 namespacing for models & controllers. This'll make for less work in #100.

For review: @angaither


angaither:
:elephant: :white_check_mark:

PR #107, opened 2015-05-11
Mock Drupal API for signup & reportback tests.

DFurnes:
# Changes
 - Test coverage restored & green! :traffic_light: Added tests for updated signup & reportback endpoints.
 - Mocks `DrupalAPI` class, so unit tests do not have external dependency on another app.

For review: @angaither @jonuy


angaither:
mockery ++

:+1:

PR #109, opened 2015-05-11
Remove hardcoded placeholder encryption key.

DFurnes:
# Changes
 - Removes hardcoded placeholder encryption key.

For review: @angaither @jonuy


weerd:
:+1:

PR #113, opened 2015-05-12
Fix broken middleware.

DFurnes:
# Changes
 - Accidentally broke middleware in Laravel 5 upgrade, whoopsie. Fixed now. :cactus: 
 - Restored `/users/` test to check for `404`, since that now behaves as it was supposed to.

For review: @angaither @jonuy


jonuy:
Seems fine to me :+1:

PR #117, opened 2015-05-13
Fixes drupal_id not being saved on user register

jonuy:
#### What's this PR do?
The `drupal_id` wasn't getting saved on a register because the response needed to be json decoded first. This PR fixes that and also handles the case where a Northstar register fails because the user already exists on Drupal. In that case we send an authenticated request to get the user UID based on the user's email and then save that to the user's Northstar document.

#### Discussion
- We can't create a Drupal user without an email address. Previously on Drupal, we would just create a dummy one by appending `@mobile` ... is that what we do again? Is there a smarter solution we can implement now that we've got Northstar?
- Data in the GET params, like `create_drupal_user` are also getting added to the user doc. Can we make that not happen?

cc: @angaither @DFurnes

### DIFF
--- a/app/Services/DrupalAPI.php
+++ b/app/Services/DrupalAPI.php
@@ -115,7 +115,34 @@ class DrupalAPI
             'body' => json_encode($payload),
         ]);
 
-        return $response->uid;
+        $json = $response->json();
+        return $json['uid'];
+    }
+
+    /**
+     * Get a user uid by email.
+     * @see: https://github.com/DoSomething/dosomething/wiki/API#find-a-user
+     *
+     * @param String $email - Email of user to search for
+     * @return String - Drupal User ID
+     * @throws \Exception
+     */
+    public function getUidByEmail($email)
+    {
+        $response = $this->client->get('users?parameters[email]=' . urlencode($email), [
+            'cookies' => $this->getAuthenticationCookie(),
+            'headers' => [
+                'X-CSRF-Token' => $this->getAuthenticationToken()
+            ]
+        ]);
+
+        $json = $response->json();
+        if (sizeof($json) > 0) {
+            return $json[0]['uid'];
+        }
+        else {
+            throw new \Exception('Drupal user not found.', $response->getStatusCode());
+        }
     }
 
     /**
--- a/app/Services/DrupalAPI.php
+++ b/app/Services/DrupalAPI.php
@@ -129,11 +129,14 @@ class DrupalAPI
      */
     public function getUidByEmail($email)
     {
-        $response = $this->client->get('users?parameters[email]=' . urlencode($email), [
+        $response = $this->client->get('users', [
+            'query' => [
+                'parameters[email]' => $email,
+            ],
             'cookies' => $this->getAuthenticationCookie(),
             'headers' => [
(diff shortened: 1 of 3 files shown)


app/Services/DrupalAPI.php
```diff
@@ -115,7 +115,34 @@ public function register($user)
...
+     * Get a user uid by email.
+     * @see: https://github.com/DoSomething/dosomething/wiki/API#find-a-user
+     *
+     * @param String $email - Email of user to search for
+     * @return String - Drupal User ID
+     * @throws \Exception
+     */
+    public function getUidByEmail($email)
+    {
+        $response = $this->client->get('users?parameters[email]=' . urlencode($email), [
```
DFurnes: You can set query strings in Guzzle with an option: `'query' => ['parameters[email]' => $email]`

DFurnes:
One comment on setting query string, but otherwise :+1: :+1:!

aaronschachter:
It could be possible to create a Drupal user without an email address.. there is a contrib module called Optional mail which alters the User registration form: http://cgit.drupalcode.org/optional_mail/tree/optional_mail_on_register.module?id=8942b67 although I haven't downloaded and tested to verify it actually doesn't populate the `mail` field.

A big reason to require email is to provide users with the ability to reset their passwords... has there ever been discussion to move this functionality out of Drupal and into Northstar ?

DFurnes:
> We can't create a Drupal user without an email address. Previously on Drupal, we would just create a dummy one by appending `@mobile` ... is that what we do again? Is there a smarter solution we can implement now that we've got Northstar?

That makes sense to me... maybe `{NORTHSTAR_ID}@northstar.dosomething.org`?

> Data in the GET params, like `create_drupal_user` are also getting added to the user doc. Can we make that not happen?

We should switch to using `$fillable` to whitelist these. Right now, we're just blindly adding anything that's passed directly to the user object which could cause trouble.

DFurnes:
@aaronschachter Yeah, that sounds smart to me. I think perhaps Northstar handles reset requests & keys, and something in Message Broker to handle actually sending the emails?

angaither:
> Data in the GET params, like create_drupal_user are also getting added to the user doc. Can we make that not happen?

@jonuy @DFurnes yersssh, we need to change this to mass assignment [instead of the for loop](https://github.com/DoSomething/northstar/blob/dev/app/Http/Controllers/UserController.php#L61)

jonuy:
@angaither @DFurnes Changed to mass assignment and moved the 'interests' transform logic into a model mutator. Also set the query param in the `client->get` param.

@aaronschachter I haven't heard about any recent password reset talks, but given that message broker and Northstar are becoming more mature, going in a direction that uses those two alone makes sense. In the meantime, trying out the contrib module could be a good idea we float to maybe Drewish to look into?

angaither:
so into this :+1: :+1: :+1: :+1: :+1: :+1:

aaronschachter:
@jonuy Could make sense. The real need is just to save Users without email right? The contrib module will just make the field unmandatory on registration -- which is what we don't want, but we could pull the module down to and test to see whether or not it actually saves Users in the DB with a NULL email field.. (I can't tell from looking at the code)

PR #118, opened 2015-05-13
PUT reportbacks

### LINKED ISSUE #62: Signup/reportback forward to drupal
When northstar gets a signup or reportback, it should forward that info along to drupal

DFurnes:
# Changes
 - Route PUT to same reportback method, and update corresponding tests.

Routing them both to the same method because none of the logic was different between the two. The _only_ difference is the status code returned – all validation, Drupal API calls, database queries, etc. are shared between the two endpoints.

This PR also (finally) closes #62.

For review: @angaither @jonuy


angaither:
:+1:

PR #120, opened 2015-05-14
Attribute casting & validation

DFurnes:
# Changes
 - Uses [Eloquent attribute casting](https://mattstauffer.co/blog/laravel-5.0-eloquent-attribute-casting) to simplify setters and getters on models.
 - Use controller validation & request object to move request logic into the controller.

Verified that same JSON response is generated, and test suite still passes after refactoring.

For review: @angaither @jonuy


angaither:
:+1: 
clean it up

DFurnes:
:dancers:

PR #121, opened 2015-05-14
Including password in register call to Drupal

jonuy:
Drupal users were getting created without a password being included. So I found that when I registered through Northstar, I wouldn't be able to login to the Drupal site. This passes that password on to the Drupal register call.

cc: @angaither @DFurnes


DFurnes:
:+1:

PR #123, opened 2015-05-14
Adding support for mongo replica sets

blisteringherb:
The current settings work fine if the settings are pointed directly to the primary node in the Mongo replica set, but they'll break otherwise.  I've added the additional replica set parameter in 'options' to make sure that the app can always connect to the primary.  I also updated the hostnames in the settings on the server to reflect these changes.

@jonuy @angaither


jonuy:
:+1: thanks, Morgan!

PR #124, opened 2015-05-14
Fix crash on local because of missing replica DB

DFurnes:
# Changes
 - Only set options array if environment variable is set. 
 - Use new `env()` helper for other variables.

For review: @blisteringherb


angaither:
:+1: :100: :dancer:

Issue #84, opened 2015-04-23
Standardize responses to always return a dictionary/object


jonuy:
Some responses return an array, some return an empty string, and some do return a dictionary. Makes it difficult on the client side to handle the response when it could one of multiple things.

- Login ex:
  - login with no parameters, returns an array of errors
  - login with an account that doesn't exist: 200 status code and a string with the error
- User logout
- User delete

---

Responses that previously sent an empty body should look something like this on error...
```
{
  error: {
    message: "Reportback does not exist."
  }
}
```

And this on success...
```
{
  success: {
    message: "Booyakasha. Welcome to the Thunderdome!"
  }
}
```

PR #125, opened 2015-05-14
Structuring login, logout, and user delete responses

jonuy:
#### What's this PR do?
The intent here was to start standardizing the responses from Northstar. In particular, the focus was on a handful of responses that just replied back with a string as opposed to an object or array like some other responses did.

Responses affected:
- login
- logout
- user delete

Where a response would previously just reply back with a string, now it'll be structured like:
```
{
  status: {
    "message": "This is the error/success message"
  }
}
```

Where a response would've replied back with an object, now it'll assign it to a "data" field in an object. The only response affected by this PR right now is on a successful login.
```
{
  "data": <User object>
}
```

#84

#### Questions
This is a breaking change for any client code that uses these endpoints right now. So for example, login on the mobile apps will be broken until they get updated to support the new structure. This is sorta why I held off on updating _all_ the responses.

That being said, should we just go ahead and update all of them right now? This might be more of a question for the iOS and Android devs.

cc: @DFurnes @aaronschachter

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,9 @@
+app/storage
 /vendor
 /node_modules
 .env
 
 .DS_Store
+
+# PhpStorm
+.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-app/storage
 /vendor
 /node_modules
 .env
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -4,6 +4,7 @@ use Northstar\Models\User;
 use Northstar\Models\Token;
 use Illuminate\Http\Request;
 use Hash;
+use Symfony\Component\HttpKernel\Exception\HttpException;
 
 class AuthController extends Controller
 {
@@ -13,6 +14,7 @@ class AuthController extends Controller
      *
      * @param Request $request
      * @return Response
+     * @throws HttpException
      */
     public function login(Request $request)
     {
@@ -29,12 +31,8 @@ class AuthController extends Controller
             $user = User::where('mobile', '=', $input['mobile'])->first();
         }
 
-        $code = 200;
-        $status = '';
-
         if (!($user instanceof User)) {
-            $data = 'User is not registered.';
-            $status = 'error';
+            throw new HttpException(404, 'User is not registered.');
         } elseif (Hash::check($input['password'], $user->password)) {
             $token = $user->login();
             $token->user = $user->toArray();
@@ -42,25 +40,22 @@ class AuthController extends Controller
             // Return the session token with the user.
             $user->session_token = $token->key;
             $data = $user;
+            return $this->respond($user);
         } else {
-            $data = 'Incorrect password.';
-            $status = 'error';
-            $code = 412;
+            throw new HttpException(412, 'Incorrect password.');
         }
-
-        return $this->respond($data, $code, $status);
-
 
     }
 
     /**
      * Logout the current user by invalidating their session token.
      * @return Response
+     * @throws HttpException
      */
     public function logout(Request $request)
     {
         if (!$request->header('Session')) {
-            return $this->respond('No token given.', 200, 'error');
+            throw new HttpException(422, 'No token given.');
         }
 
         $input_token = $request->header('Session');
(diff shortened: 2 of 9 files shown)


app/Http/Controllers/AuthController.php
```diff
@@ -52,7 +60,7 @@ public function login(Request $request)
     public function logout(Request $request)
     {
         if (!$request->header('Session')) {
-            return response()->json('No token given.');
+            return $this->respond('No token given.', 200, 'error');
```
DFurnes: Should this be `422`?

DFurnes: I think I'd also suggest throwing errors with `HttpExceptions`, since they all run through the same error reporter in that case.

app/Http/Controllers/AuthController.php
```diff
@@ -28,21 +28,29 @@ public function login(Request $request)
...
         }
-        if (!($user instanceof User)) {
-            return response()->json("User is not registered.");
-        }
 
-        if (Hash::check($input['password'], $user->password)) {
+        $code = 200;
+        $status = '';
+
+        if (!($user instanceof User)) {
```
DFurnes: Is `!$user` sufficient here?

.gitignore
```diff
@@ -1,5 +1,9 @@
+app/storage
```
DFurnes: You should be able to just delete this folder. It's left over from Laravel 4.2.

aaronschachter:
My vote would be to just change while you're in it. I should be able to figure out how to fix Lets Do this iOS (the Reviewer doesn't hit northstar at all).  Not sure how you want to handle timing with TouchLab, although they go on vacation in a day...

app/Http/Controllers/AuthController.php
jonuy: I think so? It's also the logic that was there before and I didn't think to change it.

jonuy:
@DFurnes @aaronschachter Changed it now so that error responses were sent by throwing HttpExceptions. And I guess I'm leaning towards changing all the responses to adhere to this structure now instead of later. I'll make sure to talk to TouchLab about this tomorrow and I'll maybe just go in and change up the Android code while they're gone.

DFurnes:
:+1:! And I agree on standardizing them all now.

jonuy:
@angaither Will the changes to the response format here break Aurora?

angaither:
@jonuy it doesn't _look_ like it will.  lemme testttt

angaither:
@jonuy yeah, I'll fix aurora, thanks for the heads up!

angaither:
@jonuy i'm not getting any response from 
`http://northstar.dev:8000/v1/users` locally, it should just be returning all users, but nothing is in the data array. is it broken for you?

jonuy:
@angaither hm yup, I'm not getting anything returned either. Looking now.

jonuy:
@angaither @DFurnes This isn't too hacky, is it? This seems to fix the empty responses we were getting back from `/users`:

```
} elseif ((is_object($data) && !is_a($data, 'Illuminate\Pagination\AbstractPaginator'))
          || is_array($data)) {
```

Which returns the paginated data as it did before...


**ex:**
```
{
    "total": 5,
    "per_page": 20,
    "current_page": 1,
    "last_page": 1,
    "next_page_url": null,
    "prev_page_url": null,
    "from": 1,
    "to": 5,
    "data": [
        {
            "_id": "5430e850dt8hbc541c37tt3d",
            "email": "test@dosomething.org",
            ...
        },
        ...
```

DFurnes:
Would it be helpful to make a `$this->respondPaginated($eloquent)` method that takes an Eloquent query object, so that we can set common pagination settings in one place?

angaither:
agree. :+1:

jonuy:
@angaither @DFurnes So what if it's something like this then...

In UserController:
```
public function index()
{
    return $this->respondPaginated('Northstar\Models\User');
}
```

And in Controller:
```
protected function respondPaginated($className) {
    $obj = new $className();
    if (is_a($obj, 'Illuminate\Database\Eloquent\Model')) {
        $limit = Input::get('limit') ?: 20;
        $response = $obj->paginate($limit);
        return response()->json($response);
    }
}
```

I'm starting to feel the weakness in my PHP more and more. Can reflection be used here somehow? Do I need to instantiate a new object in order to determine if it's an Eloquent model? in order to call `paginate` which before was just called as a static method?

angaither:
@jonuy let me know when you deploy this so I can fix up/deploy all the issues in aurora
DoSomething/aurora/issues/20

Issue #98, opened 2015-05-07
Implement GET /campaigns/:nid/signup


jonuy:
The primary purpose of this is to allow the user to get either the `sid` or `source` from her signup resource that will be used as an invite code on the mobile app.

Expected response should look something like...
```
{
  nid: 123,
  sid: 456,
  signup_source: 789,
  timestamp: 123456789
}
```

- [x] finish up #62
- [x] add `signup_source` to the campaign model
- [x] implement me

cc: @angaither

jonuy:
Hm, probably also relies on work that'd be done here: https://github.com/DoSomething/dosomething/issues/4465

DFurnes:
Added as `GET /campaigns/:campaign_id` per comments in #126.

PR #126, opened 2015-05-15
Add route to retrieve campaign details.

DFurnes:
# Changes
 - Adds a route to show campaign details. References #98.

I renamed the previous `show` method to `index`, since it's listing a bunch of campaigns, and implemented this as a `campaign` route rather than a signup object, since that's the REST noun we're using to store this data. Totally up for discussion if anyone thinks there's a better way of handling.

For review: @angaither @jonuy

### DIFF
--- a/app/Http/Controllers/CampaignController.php
+++ b/app/Http/Controllers/CampaignController.php
@@ -28,9 +28,9 @@ class CampaignController extends Controller
      * @param $term string - Term to search by (eg. mobile, drupal_id, id, etc)
      * @param $id   string - The value to search for
      *
-     * @return Response
+     * @return \Illuminate\Http\Response
      */
-    public function show($term, $id)
+    public function index($term, $id)
     {
         // Find the user.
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/CampaignController.php
```diff
@@ -43,6 +43,28 @@ public function show($term, $id)
...
+     *
+     * @return \Illuminate\Http\Response
+     */
+    public function show($campaign_id)
+    {
+        $user = User::current();
+
+        $campaign = $user->campaigns()->where('drupal_id', $campaign_id)->first();
+
+        if(!$campaign) {
```
jonuy: :space_invader: `if (`

jonuy:
I think that makes sense. :+1:

app/Http/Controllers/CampaignController.php
DFurnes: (╯°□°）╯︵ ┻━┻

PR #127, opened 2015-05-15
add avatar controller

chloealee:


chloealee:
wait no I didn't mean to do this! Meant to update my branch so I could pull it down on my new computer - disregard!

PR #129, opened 2015-05-19
Bug fixes

jonuy:
:bug: :beetle: 

- The `paginate` function on an `Eloquent\Builder` object requires the parameter to be a number. So I'm casting to an `int` there.
- Apparently `response()->json($response, $code)` doesn't play nice when `$response` is a string and `$code` is something that's not 200. Like, if I set it to 204, the response body would come back empty. So I'm just letting the `DELETE /users` return a 200.

cc: @angaither @DFurnes


angaither:
shouldn't the limit automatically be cast an an int??

jonuy:
When I set `limit` in the query params, it gives me a string. :frowning:

angaither:
¯\_(ツ)_/¯  ok :+1:

PR #132, opened 2015-05-21
Signup groups

### LINKED ISSUE #98: Implement GET /campaigns/:nid/signup

### LINKED ISSUE #99: Implement endpoint to get users in my Signup Group
To support the friend/invite feature in the mobile app, this endpoint shows me who else signed up for a campaign with a shared invite code. We'd essentially query for all users who share a particular `sid` value (which is what we're using to tie groups together) in either their `sid` or `signup_source` property.

**The endpoint:**

`GET /signup-group/:sid`

**The mongo query:**

I'm basing this off of [this](http://docs.mongodb.org/manual/reference/method/db.collection.find/#query-an-array-of-documents). This probably isn't exactly right, but I think the query will look something like...
```
db.users.find({
  campaigns: {
    $elemMatch: {
      $or: [
        sid: <:sid>,
        signup_source: <:sid>
      ]
    }
  }
})
```

**The response:**

```
{
  users: [
    {
      _id,
      first_name,
      profile_image,
      sid,
      signup_source,
      rbid
    }
  ]
}
```

**Why in Northstar instead of Drupal?**

The information we need is already all here. `sid` and `signup_source` should all be here. For the facepile-like feature, user profile images will be here. For the feed of report backs from a group, each user's rbid will be here.

**Related issues:** #62, #93, #98

cc: @angaither

jonuy:
#### What's this PR do?
- Saves `source` param on signup to `signup_source`
- Adds the endpoint `GET /signup-group/:group_id`.

  It returns the users where the `:group_id` is equal to their own `signup_id` or `signup_source`. This gives us the users who have been invited to and joined a campaign with an invite code.

  If no users are found, it returns a 404.
- Adds test for `GET /signup-group`
- Fixes `DELETE /users` test
- Changes mobile numbers seeded in the DB since those combinations can actually be real numbers.

Closes #98 
Closes #99

cc: @angaither @DFurnes @aaronschachter

### DIFF
--- a/app/Http/Controllers/CampaignController.php
+++ b/app/Http/Controllers/CampaignController.php
@@ -106,6 +106,7 @@ class CampaignController extends Controller
         $campaign = new Campaign;
         $campaign->drupal_id = $campaign_id;
         $campaign->signup_id = $signup_id;
+        $campaign->signup_source = $request->input('source');
         $campaign = $user->campaigns()->save($campaign);
 
         $response = array(
--- a/app/Http/Controllers/SignupGroupController.php
+++ b/app/Http/Controllers/SignupGroupController.php
@@ -23,9 +23,22 @@ class SignupGroupController extends Controller
 
         if (count($group) == 0) {
             throw new NotFoundHttpException("No users found for the group ID.");
-        }
-        else {
-            return $this->respond($group);
+        } else {
+            // Get the campaign id associated with the signup group ID
+            for ($i = 0; $i < count($group[0]->campaigns); $i++) {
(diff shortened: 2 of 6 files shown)


DFurnes:
Looks good to me! :+1:

PR #133, opened 2015-05-28
User profile image upload to AWS

### LINKED ISSUE #93: User profile image
Add user profile image to the user registration/login stuffs

This photo should probably be base64 encoded, and then stored up on S3

chloealee:
Adds functionality to save the user profile photo to AWS

fixes #93

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,9 +1,0 @@
-APP_ENV=local
-APP_DEBUG=true
-APP_KEY=SomeRandomString
-
-DRUPAL_API_URL=http://dev.dosomething.org:8888
-DRUPAL_API_USERNAME=username
-DRUPAL_API_PASSWORD=password
-
-
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,7 @@
+-APP_ENV=local
+-APP_DEBUG=true
+-APP_KEY=SomeRandomString
+-
+-DRUPAL_API_URL=http://dev.dosomething.org:8888
+-DRUPAL_API_USERNAME=username
+-DRUPAL_API_PASSWORD=password
--- a/app/Http/Controllers/AvatarController.php
+++ b/app/Http/Controllers/AvatarController.php
@@ -36,7 +36,7 @@ class AvatarController extends Controller
     $user->save();
 
     // Respond to user with success
-    $this->respond();
+    return $this->respond('Photo Uploaded!');
   }
 }
 
--- a/app/Http/Controllers/AvatarController.php
+++ b/app/Http/Controllers/AvatarController.php
@@ -20,23 +20,23 @@ class AvatarController extends Controller
  * POST northstar.com/users/{id}/avatar
  */
 
-  public function store(Request $request, $id) 
+  public function store(Request $request, $id)
   {
     if(!$request->hasFile('photo'))
       return Response::json(['error' => 'No Photo Sent']);
- 
+
     if(!$request->file('photo')->isValid())
       return Response::json(['error' => 'Photo is not valid']);
- 
+
     $file = $request->file('photo');
 
     $v = Validator::make(
       $request->all(),
-      ['photo' => 'required|image|mimes:jpeg,jpg|max:8000']  
-    );  
+      ['photo' => 'required|image|mimes:jpeg,jpg|max:8000']
+    );
 
     if($v->fails())
-      return Response::json(['error' => $v->errors()]);         
+      return Response::json(['error' => $v->errors()]);
 
     // $filename = $this->aws->storeImage('avatars', $file);
     $filename = $this->aws->storeImage('avatars', $id, $file);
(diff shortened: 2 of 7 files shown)


.env.example
```diff
@@ -1,9 +0,0 @@
-APP_ENV=local
-APP_DEBUG=true
-APP_KEY=SomeRandomString
-
-DRUPAL_API_URL=http://dev.dosomething.org:8888
-DRUPAL_API_USERNAME=username
-DRUPAL_API_PASSWORD=password
-
```
angaither: I don't think we want to delete this one, was it b/c it's been a bit since a rebase?

chloealee: Oh yes, that is my mistake!

app/Http/Controllers/AvatarController.php
```diff
@@ -19,29 +19,35 @@ public function __construct(AWS $aws)
...
     $v = Validator::make(
       $request->all(),
-      ['photo' => 'required|image|mimes:jpeg,jpg|max:8000']
-    );
+      ['photo' => 'required|image|mimes:jpeg,jpg|max:8000']  
+    );  
 
     if($v->fails())
-      return Response::json(['error' => $v->errors()]);
+      return Response::json(['error' => $v->errors()]);
```
angaither: the extra spaces in here should be fixed if you re-save/commit with the editor config package installed :)

chloealee: installed it and re-committed - should be all set now!

app/Http/Controllers/AvatarController.php
```diff
@@ -19,11 +19,14 @@ public function __construct(AWS $aws)
  * Store a new avatar for a user.
  * POST northstar.com/users/{id}/avatar
  */
-  public function store(User $user, Request $request)
+
+  public function store(Request $request, $id)
   {
-    // dd($request->all());
+    if(!$request->hasFile('photo'))
+      return Response::json(['error' => 'No Photo Sent']);
```
DFurnes: You can actually remove these two `hasFile` and `isValid` checks, since your validator below handles those for you. Using the validator object is also nice because it automatically writes & formats error messages for you.

app/Http/Controllers/AvatarController.php
```diff
@@ -35,13 +38,16 @@ public function store(User $user, Request $request)
     if($v->fails())
```
DFurnes: This could be simplified a bit using [controller validation](http://laravel.com/docs/5.0/validation#controller-validation), which is a nice shortcut that handles making the validator and reporting errors for you. (Not that this is the wrong way of doing things – Laravel has an abundance of different "correct" ways of doing everything.)

```php
$this->validate($request, [
    'photo' => 'required|image|mimes:jpeg,jpg|max:8000'
]);
```

app/Http/Controllers/AvatarController.php
```diff
@@ -35,13 +38,16 @@ public function store(User $user, Request $request)
...
+    $filename = $this->aws->storeImage('avatars', $id, $file);
 
     // Save filename to User model
+    $user = User::where($id)->first();
     $user->avatar = $filename;
     $user->save();
 
     // Respond to user with success
-    return response()->json('Great!', 200);
+    return response()->json('Photo uploaded!', 200);
```
DFurnes: Uy made a nice [$this->respond()](https://github.com/DoSomething/northstar/blob/16163cdff72945639f94bf140eac7222ead0852d/app/Http/Controllers/Controller.php#L14-L34) method on the `Controller` base class just a little while ago, which would make sense to use here. It just helps make sure everything is formatted similarly so API clients know what to expect.

app/Services/AWS.php
```diff
@@ -8,16 +10,17 @@ class AWS
    * @param string $bucket
    * @param File $file
    */
-  public function storeImage($bucket, $file)
+  // public function storeImage($bucket, $file)
```
DFurnes: Looks like there is a stray comment left over here! :)

DFurnes:
A few teeny code style suggestions, but looks great! :+1:

app/Http/Controllers/AvatarController.php
chloealee: Oh right! You said this before but I forgot why we deleted so put back in bc I thought I accidentally got rid of - thank you!

chloealee:
Awesome, thanks so much Dave! Making these changes and will re-commit :)

chloealee:
Edits in most recent commit :)

app/Http/Controllers/AvatarController.php
```diff
@@ -19,29 +19,24 @@ public function __construct(AWS $aws)
...
+    $filename = $this->aws->storeImage('avatars', $id, $file);
 
     // Save filename to User model
+    $user = User::where($id)->first();
     $user->avatar = $filename;
     $user->save();
 
     // Respond to user with success
-    return response()->json('Great!', 200);
+    $this->respond();
```
DFurnes: One thing! You still need a message here, and a return statement: `return $this->respond('Avatar stored.');`

chloealee: ah duh! Thank you :)

chloealee:
fix success response

DFurnes:
Ship it! :shipit:

chloealee:
Closes #93

PR #139, opened 2015-06-02
rename /campaigns to users/campaigns

### LINKED ISSUE #138: Rename /campaigns
This route overlaps with `/campaigns` on drupal
Since we aren't asking for a list of campaigns with this route in NS, let's rename this to `user/campaigns`

chloealee:
Fixes #138 

Rename `/campaigns` to `users/campaigns`

### DIFF
--- a/app/Http/routes.php
+++ b/app/Http/routes.php
@@ -19,10 +19,10 @@ Route::get('/', function () {
 Route::group(['prefix' => 'v1', 'middleware' => 'auth.api'], function () {
     // Campaigns.
     Route::group(['middleware' => 'auth.token'], function () {
-        Route::get('campaigns/{campaign_id}', 'CampaignController@show');
-        Route::post('campaigns/{campaign_id}/signup', 'CampaignController@signup');
-        Route::post('campaigns/{campaign_id}/reportback', 'CampaignController@reportback');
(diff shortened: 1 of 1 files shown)


jonuy:
I slept on the naming of this route and I still feel kinda iffy about it. What about `user-campaigns/`?

With `users/` without a resource id following it, I think it suggests that it should return info for all users. I also wonder if there are any scenarios where this path could be technically problematic since it shares the same name as the main `users` path.

aaronschachter:
`campaign-activity`?  or `activity/campaigns` ?  (later we could have `activity/games`, etc)

aaronschachter:
or just `campaigns` `games` etc

DFurnes:
Another suggestion would be using `user/` as a "resource" for the current authenticated user, and `user/campaigns` for the list of all campaigns. This is how GitHub handles [listing the current user repos vs. a specific user's repos](https://developer.github.com/v3/repos/#list-your-repositories).

jonuy:
Ah, probably why Andrea suggested `user/campaigns` in the first place. oops. That also works if we want to do `user/games` and the like later on.

chloealee:
awesome, thanks! Just changed it to user/campaigns

jonuy:
:+1:

DFurnes:
Makes sense to me! :+1:

PR #140, opened 2015-06-03
Save Parse IDs to the user doc

### LINKED ISSUE #101: Update User model to store identifiers for GCM and APNS pushes
- `apns_token`: for users who are subscribed for Apple push notifications
- `gcm_registration_id`: for users who are subscribed for Android push notifications

jonuy:
#### What's this PR do?
Parse will be used to handle push notifications for our app on both Android and iOS. In preparation of Northstar or other services to send notifications to users, we need to know each user's Parse installation ID. This PR:

- Allows the ids to be saved as an array under `parse_installation_ids` in case there are multiple devices per user
- Removes the id on logout if specified. This can effectively ensure a user won't be notified on that device.
- Unit tests

##### User.php
Adds `parse_installation_ids` to the fillable array and a mutator to convert it from a comma-separated string to be saved as an array on the DB.

##### AuthController.php
If `parse_installation_ids` is included in the logout request, we remove that ID from the user doc.

Closes #101 

cc: @DFurnes @sbsmith86

### DIFF
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -68,6 +68,13 @@ class AuthController extends Controller
         } elseif ($token->user_id !== $user->_id) {
             throw new HttpException(403, 'You do not own this token.');
         } elseif ($token->delete()) {
+            // Remove Parse installation ID. Disables push notifications.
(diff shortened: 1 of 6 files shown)


sbsmith86:
Looks good to me :+1:

PR #145, opened 2015-06-11
Update on Avatar Controller to accept files and base64 strings of photo

### LINKED ISSUE #141: Return Profile avatar in GET users/identifier/[uid]
Refs #93  -- Confirmed I can POST to the avatar endpoint, but the photo URL needs to be returned in the GET user response.

chloealee:
add functionality to upload both file and base64 string avatar photos. Fixes #141


app/Http/Controllers/Controller.php
```diff
@@ -30,8 +30,7 @@ protected function respond($data, $code = 200, $status = 'success')
             $response = $data;
         }
 
-        return response()->json($response, $code);
-    }
+        return response()->json($response, $code, array(), JSON_UNESCAPED_SLASHES);    }
```
jonuy: The `}` looks like it accidentally got taken off its new line.

app/Services/AWS.php
```diff
@@ -10,21 +10,21 @@ class AWS
...
+            $data = base64_decode($file);
+            $filename = 'uploads/' . $folder . '/' . $id;
+            Storage::disk('s3')->put($filename, $data);
+        } else {
+            $extension = $file->guessExtension();
+            $filename = 'uploads/' . $folder . '/' . $id . '.' . $extension;
+            Storage::disk('s3')->put($filename, file_get_contents($file));
+        }
+
+        return 'https://s3.amazonaws.com/profiles-dosomething-org/' . $filename;
```
jonuy: We might want to consider making this hardcoded string an environment variable instead.

jonuy:
I only looked at the photo stuff since I know you've got the rest of the kudos coming in a later PR.

Just a couple comments, but otherwise :+1:

app/Http/Controllers/Controller.php
chloealee: @jonuy I think the spacing got messed up but I think it's actually still there, just appears on the same line? Let me know if I'm not right on this though!

jonuy: @chloealee Yea, it's there, but I'm saying it should be on its own line.

chloealee: @jonuy ah I see what you mean. When I open it up in sublime text, it appears to be on it's own line - maybe it'll update when I commit again?

chloealee: @jonuy ah I was looking at the wrong version - my bad! You are right I'll update this :)

chloealee:
@jonuy  just made all edits - let me know if I can merge!

Issue #103, opened 2015-05-08
Create observers to listen for and do the actual submitting of push notifications to Parse


jonuy:
While the observers in #102 determine whether or not a push notification should be sent, the idea is that those observers will then fire another event to actually send the messages to Parse. Even though Parse alone should help with sending notifications to both iOS and Android, I'm not sure if we'll want to create two separate observers to send notifications to the two separate platforms.

Parse documentation on sending pushes with their REST API:
- https://parse.com/docs/rest/guide#push-notifications-sending-pushes

Code from prototype that sent pushes to Parse:
- https://github.com/DoSomething/parse-prototype/blob/master/server/server.js

cc: @sbsmith86 @DFurnes

sbsmith86:
Closing this. I have set up a service that connects Northstar to the Parse API and created Handlers that hold the logic for sending a push notifications when an event fires.

PR #146, opened 2015-06-12
Group Sign Up Push Notification

sbsmith86:
### What's this PR do?

* Creates a `UserSignedUp` event that fires on every signup. This allows us to create handlers for anything that we would want to fire when this happens. This PR creates a handler that sends a push notification to other users in someone's group.
* Sets up a Parse service class that connects to the Parse API 

### Where should the reviewer start?

UserSignedUp.php

Creates a new event class that takes in the user that signed up and provides information about that user to the handler.

SendSignupPushNotification.php

Get the users group and send push notifications to the other users in that group that the user signed up.

### What are the relevant tickets?

Addresses #103 

cc @DFurnes


app/Events/UserSignedUp.php
```diff
@@ -0,0 +1,24 @@
...
+
+	/**
+	 * Create a new event instance.
+	 *
+	 * @return void
+	 */
+	public function __construct(User $user)
+	{
+		$this->drupal_id = $user->drupal_id;
+		$this->signup_id = $user->campaigns[0]->signup_id;
```
DFurnes: It might make sense to pass the `Campaign $campaign` object here too, since I'm not sure it will always be the first element in the array?

DFurnes: The other nice thing with that is you can let the `SerializesModels` trait handle "freezing" the state of each model at the point the event is fired, and just pass the model directly `$this->user = $user` and `$this->campaign = $campaign`. That way you don't have to change the event class if we add other handlers that need other properties.

.env.example
```diff
@@ -1,7 +0,0 @@
--APP_ENV=local
```
DFurnes: I think this file got accidentally deleted?

app/Events/UserSignedUp.php
sbsmith86: Ahh yea, I was wondering if I should take that route or if being more specific was better. But this makes sense to make it more flexible.

.env.example
sbsmith86: @DFurnes I think because I renamed it to `.env`?

DFurnes: Yup, but we need this file in the repo as an example for other people making their `.env` files!

sbsmith86: @DFurnes ahh of course of course

DFurnes:
Also looks like a `.DS_Store` snuck in somehow in the `app/` directory. :grimacing: 

Otherwise all looks good to me! :+1:

sbsmith86:
@DFurnes removed! Thanks!

jonuy:
Nice work @sbsmith86 :+1:

Issue #102, opened 2015-05-08
Create observer and subjects for handling events for push notifications


jonuy:
The notifications that I'm thinking we can handle in Northstar are for when a user signs up for a campaign with an invite code and when a user in a group submits a reportback.

Naming and all are just suggestions. I'm really not sure what the best way to do this in Laravel is.

**Subject:** Signed up with invite
- Event: "OnSignup"
- Fired in: `CampaignController@signup`

**Subject:** Reportback submitted by user in a group
- Event: "OnReportback"
- Fired in: `CampaignController@reportback`

**Subject:** Kudos submitted for a reportback item
- Event: "OnKudos"?
- Fired in: `KudosController@store` ... I think? Work on this one's being done in #137.

**Observer:** `PushNotificationsNotifier`
- `PushNotificationsNotifier@OnSignup`: check if the signup has a source and if that source maps to an existing group and if so, fire a push notification
- `PushNotificationsNotifier@OnReportback`:  check if the user who submitted a reportback is in a group and if so, fire a push notification
- `PushNotificationsNotifier@OnKudos`: determine who the user is who owns the reportback, then fire a push notification to that user

cc: @sbsmith86 @DFurnes 

The details above were merely a suggestion. It sounds like there might be a different and better way of structuring it all from what you two are talking about right next to me. You're free to go with whatever works.

sbsmith86:
@jonuy ok to close this? the event listeners and handlers are set up in Northstar now

jonuy:
@sbsmith86 :) yea I think so. We'll find out what, if anything, remains once we get some more robust testing setup on actual phones.

PR #148, opened 2015-06-19
Kudo Push Notification

sbsmith86:
This PR creates a new event `UserGotKudo.php` and handler `SendKudoPushNotification.php` that will send a push notification to the user that received the kudo.

This mainly sets up the logic and structure for the event as a lot of the kudos work is still being done. 

References #102 

/cc @DFurnes


jonuy:
:+1:

PR #151, opened 2015-06-22
fix comments in Kudos Controller and Avatar Controller

chloealee:
add more detail to comments and added more info regarding `@param` and `@return`

### DIFF
--- a/app/Http/Controllers/AvatarController.php
+++ b/app/Http/Controllers/AvatarController.php
@@ -14,7 +14,7 @@ class AvatarController extends Controller
 
    /**
    * Store a new avatar for a user.
-   * POST northstar.com/users/{id}/avatar
+   * POST /users/{id}/avatar
    *
    * @param Request $request
    * @param $id - User ID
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/AvatarController.php
```diff
@@ -15,6 +15,10 @@ public function __construct(AWS $aws)
    /**
    * Store a new avatar for a user.
    * POST northstar.com/users/{id}/avatar
```
jonuy: I'd suggest removing the `northstar.com` part here so that this line is then only `POST /users/{id}/avatar

app/Http/Controllers/KudosController.php
```diff
@@ -13,6 +13,15 @@ public function __construct(DrupalAPI $drupal)
         $this->drupal = $drupal;
     }
 
+   /**
+   * Store a new kudos from a user.
+   * Kudos request made from mobile app and forwarded to Northstar.
+   * Northstar finds the drupal user and sends request on to Drupal.
+   * POST northstar.com/kudos
```
jonuy: Same deal here with removing the `northstar.com` part.

jonuy:
@chloealee just the two comments about `northstar.com`, but afterwards feel free to merge

PR #152, opened 2015-06-24
Reportback Pushnotifications

sbsmith86:
### What does this PR do?

This PR creates new Event/Handler classes for the reportback event, allowing us to fire an event when a user reports back. 

In the event handler, we look for the users group and send a notification to everyone else in the group. 

**Some organizational changes:**
- I moved the query that gets a user's group into a query scope function that accepts a sign up id, so we can easily reuse this query around the codebase.

- Found a issue in my set up of the Kudo event class, where the `KudosController` wasn't pulling in the event class so it wouldn't work. This PR fixes that.

### DIFF
--- a/app/Events/UserReportedBack.php
+++ b/app/Events/UserReportedBack.php
@@ -1,0 +1,25 @@
+<?php namespace Northstar\Events;
+
+use Illuminate\Queue\SerializesModels;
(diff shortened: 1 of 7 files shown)


jonuy:
:+1: 

I like the use of the query scope.

PR #159, opened 2015-06-29
401 status code on invalid or missing token

### LINKED ISSUE #156: Return a 401 instead of 200 on token mismatch
If you submit a request with a fake token, you'll get a `200` response with an error `"Token mismatched"`.  Instead, return a 401. Maybe instead of sending the response using `Response::json`, we can do a `throw new HttpException(401, 'Token mismatched');`?

jonuy:
#### What's this PR do?
When an auth token is either invalid or missing, we now return a 401 status code instead of a 200.

This also adds unit tests to check both of those cases.

- `AuthenticateToken.php`: By throwing `HttpException`s, we allow the exception handler to deal with the response in its structured way. This is also how we've been handling other non-200 types of errors elsewhere.
- `AuthTest.php`: Hello unit tests.

Closes  #156

### DIFF
--- a/app/Http/Middleware/AuthenticateToken.php
+++ b/app/Http/Middleware/AuthenticateToken.php
@@ -3,6 +3,7 @@
 use Northstar\Models\Token;
 use Closure;
 use Response;
+use Symfony\Component\HttpKernel\Exception\HttpException;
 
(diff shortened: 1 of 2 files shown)


angaither:
:+1:

PR #160, opened 2015-06-29
Duplicate signup returns campaign object instead of 401

### LINKED ISSUE #157: Send campaign action data in response to a campaign signup that already exists
Submitting a signup for the same campaign currently results in a `401` and `Campaign signup already exists.` error message.

Instead, let's respond back with a `200` and the campaign info:

```
{
  signup_id: Number,
  signup_source: String,
  signup_group: Number or null,
  reportback_id: Number or null,
  created_at: Date
}
```

Note: `signup_group` doesn't exist as of the writing of this issue, but is on deck and can be tracked with this issue: https://github.com/DoSomething/northstar/issues/149

jonuy:
#### What's this PR do?
On a duplicate signup, instead of responding with a 401 and an error code, respond with the existing campaign object. And while creating a new signup will still respond with a 201, on duplicates we'll respond with a 200.

This also adds a unit test to check the duplicate scenario.

- `CampaignController.php`: The only real new thing happening here is:
```
if ($campaign) {
  // Campaign already signed up for. Return existing campaign object.
  return $this->respond($campaign, 200);
} else {
  ...
```
  Everything else from before is just getting thrown inside the `else` block.

- `CampaignTest.php`: The new test takes advantage of the `$this->signedUpServer` which was configured with a user who had already signed up for campaign `123`.

Closes #157

### DIFF
--- a/app/Http/Controllers/CampaignController.php
+++ b/app/Http/Controllers/CampaignController.php
@@ -99,7 +99,7 @@ class CampaignController extends Controller
 
         if ($campaign) {
             // Campaign already signed up for. Return existing campaign object.
-            $response = $campaign;
+            return $this->respond($campaign, 200);
         } else {
             // Create a Drupal signup via Drupal API, and store signup ID in Northstar.
             $signup_id = $this->drupal->campaignSignup($user->drupal_id, $campaign_id, $request->input('source'));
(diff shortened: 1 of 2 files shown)


angaither:
:+1:

Issue #158, opened 2015-06-29
testSubmitCampaignSignup unit test is failing


jonuy:
The unit test fails saying it's getting a 500 error code. The response is...
```
{
  error: {
    code: 500,
    message: "You must call Parse::initialize() before making any requests."
  },
  debug: {
    file: "/home/vagrant/sites/northstar/vendor/parse/php-sdk/src/Parse/ParseClient.php",
    line: 342
  }
}
```

I suspect this is because a signup event is triggered and we have Parse push notification code that responds to that event.

@sbsmith86 Any ideas on how this could be fixed?

sbsmith86:
@jonuy hmm it was passing on my local. I'm calling the initialize function in the Parse service I set up: 

`ParseClient::initialize($parse_app_id, $parse_api_key, $parse_master_key);`

I wonder if I didn't set the keys correctly. Is this happening the production server?

DFurnes:
You can mock the `ParseClient` so there aren't any external dependencies on the unit test. See the tests that mock `DrupalAPI` for an example.

sbsmith86:
@DFurnes Ah cool, I will take a look.

jonuy:
@sbsmith86 ah oops, I could _totally_ be doing something wrong here. If you're unable to reproduce, let me know and we can figure out what I'm doing wrong.

jonuy:
This doesn't seem to be happening anymore. ¯\_(ツ)_/¯

PR #164, opened 2015-06-30
Added signup_group parameter for storing invite code

### LINKED ISSUE #149: Change parameter name from signup_source to signup_group
`signup_source` can remain and serve to store things like `android-letsdothis` or `ios-letsdothis` or other fields like it's done on Drupal. 

`signup_group` can be explicitly used to store shared identifiers between members who are participating in a campaign together.

jonuy:
#### What's this PR do?
Previously we used `signup_source` to store both source info (like, this signup is from web or SMS, etc) or an invite code if the user was invited to sign up to the campaign by another. This was done out of a desire to not create a new column on the Drupal backend when we were thinking of storing this info there. But now that we're just storing it here in Northstar, let's just give it its own parameter.

In practice, for app devs signing up a user via an invite, this means that instead of sending the invite code in the `source` parameter, it'll need to be sent in the `group` parameter.

- `CampaignController.php` - This is the bulk of the logic that needed to change. `signup_source` stays and can be used strictly for tracking source. `signup_group` is populated by either a `group` value provided in the request, or the user's own `signup_id`.
- `SignupGroupController.php` and `User.php` - Logic updated to look at `signup_group` to determine if users are in the same group.
- `Campaign.php` - `signup_group` is an integer, like `signup_id`
- `UserTableSeeder.php` - Updating test doc to make unit tests happy

#### Testing done?
Verified phpunit tests were working. Though one seems to be failing due to a separate issue (#158)

#### Things to do before the merge
Update documentation on the wiki: https://github.com/DoSomething/api/wiki/Mobile-App-Usage

Closes #149

### DIFF
--- a/app/Http/Controllers/SignupGroupController.php
+++ b/app/Http/Controllers/SignupGroupController.php
@@ -17,9 +17,9 @@ class SignupGroupController extends Controller
      */
     public function show($id)
     {
-        // signup_id is saved as a number and signup_source is saved as a string
+        // signup_id and signup_group are saved as numbers
         $group = User::where('campaigns', 'elemMatch', ['signup_id' => (int)$id])
-                        ->orWhere('campaigns', 'elemMatch', ['signup_source' => $id])->get();
+                        ->orWhere('campaigns', 'elemMatch', ['signup_group' => (int)$id])->get();
 
         if (count($group) == 0) {
             throw new NotFoundHttpException("No users found for the group ID.");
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/SignupGroupController.php
```diff
@@ -17,9 +17,9 @@ class SignupGroupController extends Controller
      */
     public function show($id)
     {
-        // signup_id is saved as a number and signup_source is saved as a string
+        // signup_id and signup_group are saved as numbers
         $group = User::where('campaigns', 'elemMatch', ['signup_id' => (int)$id])
-                        ->orWhere('campaigns', 'elemMatch', ['signup_source' => $id])->get();
+                        ->orWhere('campaigns', 'elemMatch', ['signup_group' => (int)$id])->get();
```
angaither: do you need to cast this here when you're casting it campaign model?

jonuy: Yea, I think you do in the same way we do above with `::where('campaigns', 'elemMatch', ['signup_id' => (int)$id])`.

The `$id` value is coming in from the URL as a string. Just tried running unit tests with and without the casting, and it'll only pass if it gets casted to an `int`.

jonuy:
Wiki docs have been updated

angaither:
:+1:

Issue #153, opened 2015-06-25
Make responses static and predictable


jonuy:
In practice, I think this means we return all possible properties in our responses, at least providing a default empty response if there is no value.

For example, currently `GET /users/_id/abc123` could return us something like...
```
{
  "data": [
    {
      "_id": "abc123",
      "email": "test@example.com"
    }
  ]
}
```

But because `avatar` is a property that some users might have set, the suggestion is to also include that in the response even if it's not set for this user:
```
{
  "data": [
    {
      "_id": "abc123",
      "email": "test@example.com",
      "avatar": ""
    }
  ]
}
```

**Pros:**
- Will play nicer in Android dev with Retrofit/GSON conversion. Sounds like it'll make things easier on the iOS dev too.

**Cons:**
- Our Mongo backend intentionally doesn't store empty values for fields that aren't set. So if this were something we'd do, we either have to set empty values for these fields, or hardcode some application logic in Northstar to make sure it sends em back

Thoughts? Other pros and cons? @weerd @angaither @aaronschachter @eoji @toidiu @DFurnes

weerd:
So the only thing I'd say against adding empty responses is that we could end up with objects with lots of empty items (I'm thinking the Campaigns response object specifically) and bloat the response. Particularly if it's an index listing with lots of campaigns or reportbacks, that's a lot of data with potentially lots of empty items that aren't used). 

My default has been to keep object responses nice and trimmed down, but if it's not ideal open to change it up.

I feel an empty string is going to complicate things.... I'd suggest `NULL` which is a valid data type and easier to test against than having to check for example, if the value `isset()` and `!empty()`...

weerd:
Would it be possible as an alternative when for the Android app you need an empty array response, is that if the API responds with `NULL` or doesn't add the property, to check for that, and then if it's `NULL`or is not set, then transform it and change it to an empty array for the sake of the Java library that needs it that way.

I'm just debating if it's worth making changing how the response outputs so we can make a very specific library work with the data... just some food for thought.

toidiu:
So the place where NULL falls apart is the following: If the user is NULL then on the client side we now have to do null checks before we access name.  if(user != null) { .... }
This really becomes a problem with ugly code when there are several nested items.


{
  "data": [
    {
      "_id": "abc123",
      "user": 
          { 
              "name" : "test@example.com" 
          }
      "avatar": ""
    }
  ]
}

weerd:
[...]

PR #168, opened 2015-07-07
Returning null values for user and campaigns properties not set

### LINKED ISSUE #155: Update user campaigns response object to return all properties
Off the top of my head, I think it's just...
```
{
  signup_id,
  signup_source,
  signup_group,
  reportback_id,
  campaign_id
}
```

And if there is no value set, it should return `null`.

Ref: https://github.com/DoSomething/northstar/issues/153

jonuy:
Sorta feels hacky? But other than modifying all documents already in the db to store null values for properties, there doesn't seem to be a better way to handle this other than with middleware.

#### What's this PR do?

Modifies the responses that include user and campaign objects to return all possible properties for them even if they're not set. Values default to null.

This is all the result of discussion that happened here: https://github.com/DoSomething/northstar/issues/153

Endpoints affected:

- `GET user/campaigns/{campaign_id}`
- `POST user/campaigns/{campaign_id}/signup`
- `POST user/campaigns/{campaign_id}/reportback`
- `PUT user/campaigns/{campaign_id}/reportback`
- `GET users/{term}/{id}`
- `POST users/{id}/avatar`
- `GET users/{term}/{id}/campaigns`

The majority of the changes are in the new middleware files. Basically it's intercepting the response before it goes back to the user and fills in User and Campaign properties with null values for keys that aren't set.

#### Any background context?
https://github.com/DoSomething/northstar/issues/153

Closes #155

### DIFF
--- a/app/Http/Middleware/CampaignResponseMiddleware.php
+++ b/app/Http/Middleware/CampaignResponseMiddleware.php
@@ -10,17 +10,22 @@ class CampaignResponseMiddleware {
     public function handle($request, Closure $next)
     {
         $response = $next($request);
+        if (!is_object($response)) {
+            return $response;
+        }
+
+        $statusCode = $response->getStatusCode();
         $response = $response->getData();
 
         if (is_array($response->data)) {
             foreach ($response->data as $campaign) {
-                self::fillCampaign($campaign);
+                $this->fillCampaign($campaign);
             }
         } elseif (is_object($response->data)) {
-            self::fillCampaign($response->data);
+            $this->fillCampaign($response->data);
         }
 
-        return json_encode($response);
+        return response()->json($response, $statusCode, array(), JSON_UNESCAPED_SLASHES);
     }
 
     /**
--- a/app/Http/Middleware/CampaignResponseMiddleware.php
+++ b/app/Http/Middleware/CampaignResponseMiddleware.php
@@ -1,0 +1,42 @@
+<?php
+
+namespace Northstar\Http\Middleware;
+
+use Closure;
+use Northstar\Models\Campaign;
+
+class CampaignResponseMiddleware {
+
+    public function handle($request, Closure $next)
+    {
+        $response = $next($request);
+        $response = $response->getData();
+
+        if (is_array($response->data)) {
+            foreach ($response->data as $campaign) {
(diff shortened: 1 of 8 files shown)


angaither:
:+1: looks ok, I would love if @DFurnes could give any additional middleware feelings, I haven't used it yet!

app/Http/Middleware/CampaignResponseMiddleware.php
```diff
@@ -0,0 +1,42 @@
...
+class CampaignResponseMiddleware {
+
+    public function handle($request, Closure $next)
+    {
+        $response = $next($request);
+        $response = $response->getData();
+
+        if (is_array($response->data)) {
+            foreach ($response->data as $campaign) {
+                self::fillCampaign($campaign);
```
DFurnes: Should this be `$this->fillCampaign($campaign)` since the method isn't static?

DFurnes:
Seems okay. I wonder if it might make more sense to add this as responsibility of the `respond()` base controller method? I feel like it seems kinda hidden away as a middleware.

app/Http/Middleware/CampaignResponseMiddleware.php
jonuy: oh yup, you right

jonuy:
@DFurnes I thought about putting this in the base controller, but it felt meh because auth, signup-group, kudos, and api key controller code goes through that base controller too. With the middleware, I could at least pick and choose which routes to apply this to.

DFurnes:
Fair, although I could see us wanting to do similar things for those items as well one day & having to make a middleware for each type of object could get tedious.

jonuy:
True. Placing the logic in a single location in the base controller can make it less tedious.

We would however I think lose the decoupling of resources that are otherwise unrelated - the base controller being made aware of and working on the User, Campaign, ApiKey, and Token models.

Hm, even in the two middleware scenarios here, the code between the two, although pretty similar, still has slight variations in its logic. Something about keeping things decoupled still feels like the route to go here... but I can be swayed.

DFurnes:
Hmm, that's fair. The nested resources do make it more complicated.

Seems fine to merge for now & just keep an eye on maybe? :+1: from me.

PR #170, opened 2015-07-09
Save created_at if passed.

### LINKED ISSUE #169: Add support to save a created_at timestamp
When creating a user this attribute can't be mass assigned, but in the case of DoSomething/dosomething/pull/4640 we don't want to loose the signup date we have from old users, so we need to allow this value to be set if sent in the NS registration request

angaither:
Fixes #169 

After this is merged, we can test again in stage, and run this in prod to get all users into NS!
woo!


jonuy:
:beers:

PR #171, opened 2015-07-10
GET /signup-group fixes and reportback data

### LINKED ISSUE #163: Campaign content included in the GET /signup-group response
Currently this endpoint returns an array of user docs. While it does include IDs that could be used to then query for the relevant campaign content (ie - reportbacks), we can query that from the Drupal API ourselves within Northstar.

- Similar to #162, return campaign content in the response
- Maybe also take this chance to only return relevant campaign info? Like if a user's signed up for a campaign that's not relevant to the group, then don't include it in the response.

jonuy:
#### What's this PR do?

- Gets reportback data from Drupal for each user in a group who has reported back on the campaign associated with the group.
- Fix where `GET users/{term}/{id}/campaigns` was incorrectly being pushed through the `user` middleware.

I think it's sorta straightforward? But let me know if there are any questions. I'll just pick out a couple of the more interesting changes:

-  I think I was wrong about the need to cast these ids in the query:
```
$group = User::where('campaigns', 'elemMatch', ['signup_id' => $id])
                       ->orWhere('campaigns', 'elemMatch', ['signup_group' => $id])->get();
```
  The casting only seems to happen in responses, but the data saved in the backend still ends up being a string because that's how we get it back from Drupal.
- Culling campaigns out of the response that aren't associated with the group. I was thinking it'd be weird if someone reported back for 10 campaigns, but only one of the campaign objects returned had the full reportback data. Or on the flip side, that it'd be unnecessary to get the reportback for all the other campaigns.
```
if ($user->campaigns[$i]->drupal_id != $campaign_id)
```

Closes #163

### DIFF
--- a/app/Http/Controllers/SignupGroupController.php
+++ b/app/Http/Controllers/SignupGroupController.php
@@ -8,6 +8,11 @@ class SignupGroupController extends Controller
 {
     protected $drupal;
 
+    public function __construct(DrupalAPI $drupal)
+    {
+        $this->drupal = $drupal;
+    }
+
     /**
      * Display the users who share the specified signup group id.
      * GET /signup-group/:id
@@ -36,9 +41,6 @@ class SignupGroupController extends Controller
             }
         }
 
-        // Drupal API client
-        $drupal = new DrupalAPI;
-
         foreach ($group as &$user) {
             if (isset($user->campaigns)) {
                 for ($i = 0; $i < count($user->campaigns); $i++) {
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/SignupGroupController.php
```diff
@@ -18,28 +20,49 @@ class SignupGroupController extends Controller
...
+        for ($i = 0; $i < count($group[0]->campaigns); $i++) {
+            $campaign = $group[0]->campaigns[$i];
+            if ($campaign->signup_id == $id || $campaign->signup_group == $id) {
+                $campaign_id = $campaign->drupal_id;
+                break;
             }
+        }
+
+        // Drupal API client
+        $drupal = new DrupalAPI;
```
DFurnes: This would make sense to use constructor injection for, so we can mock it during tests.

jonuy: Good call

angaither:
:+1:

weerd:
Were you trying to cast IDs?

jonuy:
@weerd ... maybe. :grin: The casting may have been a byproduct of some of the decisions made earlier in the project. Could be worth adding an issue to remove the casting if there isn't a strong use-case for it.

weerd:
The stuff from Drupal should always be returning IDs as strings. IDs as ints doesn't make as much sense cause it's not a numerical property but more of an identifier :) But yeah, we should aim to standardize it across the repos!

PR #174, opened 2015-07-14
Support to query for multiple signup groups

### LINKED ISSUE #173: Query parameter for GET /signup-group to get multiple groups in a single call
ex: `GET /signup-group?ids=123,456,789`

This'll help reduce the number of requests that need to be made by the mobile app and other client apps.

jonuy:
#### What's this PR do?

- Support for `GET /signup-group?ids=123,456,789`: Allows us to get multiple signup group info in a single call for app-side performance.
- (This one I feel less good about) If `signup_group` was saved as null or is not set, which is the case for signups done before PR #171, then it fills it in with the value of `signup_id`. This is logic that would otherwise have to happen client-side. So figured if done server-side, maybe it just stays in a single place. I'm pretty iffy on this one though, so if you give me one reason that I shouldn't do this, I'll yank it out of this PR.

- **SignupGroupController.php**: The logic for getting signup group data was moved to private helper `getGroup` function. The `index()` function was added to account for the cases where we want to retrieve multiple group info. And both `index` and `show` essentially reuse the same logic in `getGroup` to do what they need to do.
- **Campaign.php** and __*ResponseMiddleware.php__: Logic for `fillCampaign` got moved to a singled location in the Campaign model - `populateAllAttributes()` - I'm open to a better name. Also adds the logic that responds with `signup_group = signup_id` if `signup_group` is not set.

Closes #173

cc: @angaither @DFurnes @weerd

### DIFF
--- a/app/Http/Controllers/SignupGroupController.php
+++ b/app/Http/Controllers/SignupGroupController.php
@@ -27,16 +27,15 @@ class SignupGroupController extends Controller
         $response = [];
 
         if ($request->has('ids')) {
-            $groupIds = explode(',', $request->input('ids'));
+            $group_ids = explode(',', $request->input('ids'));
 
-            foreach ($groupIds as $groupId) {
-                $group = $this->getGroup($groupId);
+            foreach ($group_ids as $group_id) {
+                $group = $this->getGroup($group_id);
                 if (!empty($group)) {
                     $response[] = $group;
                 }
             }
-        }
-        else {
+        } else {
             throw new BadRequestHttpException("Missing ids query parameter.");
         }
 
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/SignupGroupController.php
```diff
@@ -14,26 +15,65 @@ public function __construct(DrupalAPI $drupal)
...
+     */
+    public function index(Request $request)
+    {
+        $response = [];
+
+        if ($request->has('ids')) {
+            $groupIds = explode(',', $request->input('ids'));
+
+            foreach ($groupIds as $groupId) {
+                $group = $this->getGroup($groupId);
```
angaither: some js style variable names here :smirk:

angaither:
:+1:

PR #175, opened 2015-07-15
Fix for crash that came from trying to access an undefined property

jonuy:
#### What's this PR do?

Fixes crash that comes from attempting to access `signup_id` when it's undefined. Can particularly be seen from sending a GET request to the `/users` endpoint. Also fixes unit tests that failed due to the aforementioned bug (oops :bow:)


angaither:
:+1:

PR #181, opened 2015-07-22
Customized push notification for reportback event

### LINKED ISSUE #178: Push Notifications: Add data to reportback notification
We need to add a JSON object to each notification that the app can parse to figure out where to send the user next. When a user gets a notification that someone in their group reported back for a campaign, we need to attach the reportback and group info to the notification. See #177  and #176 for what that data looks like.

jonuy:
#### What's this PR do?
Packs additional info into the Parse pushes so that apps can deal with them appropriately.
- The message is customized to include username and campaign name
- Group info and reportback item data also included so the receiver of the push notification can know what the reportback item is and what group it's a part of

Includes a new unit test to verify the structure of the data we're sending to Parse.

- Most of the magic happens in `SendReportbackPushNotification::createPushData`. We get the username, campaign name and who to send pushes to here. The function will return array that we can then loop through to send pushes to the individual users who need it.
- **PushNotificationTest.php**: Unit test to verify the data coming out of `createPushData`
- **SignupGroupController.php**: this was just a realization that there was another scope function already doing this same logic
- **User.php** and **Campaign.php**: primary thing to note here is the removal of properties that are getting casted to ints. See the **Potential problem** section below for more info.

Running the PHPUnit tests would be cool if you wanna do taht.

#### How was this tested?
- Unit tests
- Plus local/staging tests that sent data to Parse. You can see that the data sent looks like we expect it to here: https://www.parse.com/apps/lets-do-this--6/push_notifications/TnW42fCIOj

Closes #178 

#### Potential problem
`drupal_id`, `reportback_id`, `signup_id` and `signup_group`. I've removed the integer casting we were doing on these. Drupal returns these values as strings and so instead of juggling between two different data types, seems to make sense to just keep them as strings.

This will probably temporarily break at least the Android app that expects these values to be returned as ints. @aaronschachter will it similarly cause problems on the iOS side?

Does this seem like the right thing to do?

#### Potential todo
If the above is ok, then:
- updating documentation to indicate `drupal_id`, `reportback_id`, `signup_id` and `signup_group` are strings
- updating relevant app code to support that
cc: @angaither

### DIFF
--- a/tests/PushNotificationTest.php
+++ b/tests/PushNotificationTest.php
@@ -5,7 +5,8 @@ use Northstar\Handlers\Events\SendReportbackPushNotification;
 use Northstar\Models\User;
 use Northstar\Models\Campaign;
 
-class PushNotificationTest extends TestCase {
+class PushNotificationTest extends TestCase
+{
 
     public function setUp()
     {
@@ -53,8 +54,11 @@ class PushNotificationTest extends TestCase {
                             'created_at' => '1234567890',
                             'status' => 'approved'
                         ],
-                    ]
-                ]
+                    ],
+                ],
+                'campaign' => [
+                    'title' => 'Test Campaign',
+                ],
             ]
         ];
 
@@ -75,6 +79,7 @@ class PushNotificationTest extends TestCase {
         $this->assertArrayHasKey('extras', $push_data);
         $this->assertArrayHasKey('completion', $push_data['extras']);
         $completion = $push_data['extras']['completion'];
+        $this->assertEquals($push_data['alert'], 'Push U. shared a photo in your Test Campaign group.');
         $this->assertEquals($push_data['alert'], $completion['message']);
         $this->assertArrayHasKey('message', $completion);
         $this->assertArrayHasKey('group', $completion);
@@ -84,4 +89,5 @@ class PushNotificationTest extends TestCase {
         $this->assertCount(1, $completion['reportback_items']['data']);
         $this->assertEquals($campaign->reportback_id, $completion['reportback_items']['data'][0]['id']);
     }
+
 }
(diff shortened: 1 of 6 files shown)


angaither:
@jonuy not sure how this will work if we already have values in the db as ints... I think we will have to [update the types of existing fields](http://stackoverflow.com/questions/4973095/mongodb-how-to-change-the-type-of-a-field)
Another potential thing (that maybe doesn't matter much...) is the `where`s might be faster on the int fields vs string fields. Personally, I would favor transforming the data before it's returned on the drupal side vs storing ints as strings in mongo?

tests/PushNotificationTest.php
```diff
@@ -0,0 +1,91 @@
...
+        $this->assertEquals($push_data['alert'], $completion['message']);
+        $this->assertArrayHasKey('message', $completion);
+        $this->assertArrayHasKey('group', $completion);
+        $this->assertArrayHasKey('reportback_items', $completion);
+
+        // Verify reportback_item in push data matches the one in the event.
+        $this->assertCount(1, $completion['reportback_items']['data']);
+        $this->assertEquals($campaign->reportback_id, $completion['reportback_items']['data'][0]['id']);
+    }
+}
```
sbsmith86: New line? This is really cool!! thanks for setting this up :)

sbsmith86:
@jonuy The parse logic looks great to me! +1 for the unit test.

jonuy:
@angaither So I ran some queries through the db and it looks like all the id fields I mentioned above are actually saved in the db as strings already. The casting to an integer I guess only happens as the field gets returned from the app?

angaither:
otay :+1:

PR #185, opened 2015-07-23
add base_url to reduce lag time for mobile app

chloealee:


jonuy:
Just curious, what's the theory behind why this might help reduce lag time?

chloealee:
Makes sure that it provides read-after-write consistency - is that correct, @sheydari-ds ?

sheydari-ds:
Yes!

jonuy:
I couldn't find the code that this actually affects. Where does `base_url` get used?

Also, this feels nitpicky, but the newline is extraneous.

chloealee:
thanks for catching that, @jonuy ! Just deleted the extra line. 

Using this file sets up all of the credentials needed to get into S3 I believe. Setting the `base_url` here should ensure the read-after-write is enabled and thinking that it should reduce lag time. 

@sheydari-ds let me know if I'm incorrect about anything above or missing anything!

sheydari-ds:
@chloealee Yes, that's correct. @jonuy Essentially, Amazon guarantees read-on-write consistency if we use a specific endpoint for S3, instead of the default one.

jonuy:
I was mostly just looking to see where in Northstar code it's being used, and it looks like it's actually in the aws-sdk-php library which is why I didn't see it initially when going through code just on :octocat: 

:+1:

PR #189, opened 2015-07-28
updates error message

### LINKED ISSUE #188: Fix error message on failed login for user that doesn't exist
Currently, when the user doesn't exist and you try to login, you get a 404 error with the message "User is not registered."

Quoting Sergii here, this is a "user enumeration" security vulnerability.

We should make the response more generic to something along the lines of "your user name and password are not correct." Or whatever it is that we already send when a login fails for a user that _does_ actually exist.

chloealee:
Fixes #188 

Changes error message to read "Invalid username or password" instead of "User is not registered."

### DIFF
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -33,7 +33,7 @@ class AuthController extends Controller
         }
 
         if (!($user instanceof User)) {
-            throw new NotFoundHttpException('User is not registered.');
+            throw new NotFoundHttpException('Invalid username or password.');
         } elseif (Hash::check($input['password'], $user->password)) {
             $token = $user->login();
             $token->user = $user->toArray();
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -43,7 +43,7 @@ class AuthController extends Controller
             $data = $user;
             return $this->respond($user);
         } else {
-            throw new HttpException(412, 'Incorrect password.');
+            throw new HttpException(412, 'Invalid username or password.');
         }
 
     }


jonuy:
@chloealee Could you also change the copy on line 46 (`throw new HttpException(412, 'Incorrect password.');) to be the same "Invalid username or password"? This is based off of Aaron's suggestion in #188.

After that, feel free to merge.

chloealee:
@jonuy ah yes - sorry I missed that! Just changed :)

app/Http/Controllers/AuthController.php
```diff
@@ -33,7 +33,7 @@ public function login(Request $request)
         }
 
         if (!($user instanceof User)) {
-            throw new NotFoundHttpException('User is not registered.');
+            throw new NotFoundHttpException('Invalid username or password.');
```
aaronschachter: @chloealee @jonuy -- Should this still be a 404, or a 412?

jonuy: Oh yup, you're right. This should definitely not be a 404. @chloealee go ahead and also change this to throw a regular HttpException.

As for the status code to use here and below on line 46, I'm thinking maybe we go with a 400 instead. 
- http://stackoverflow.com/a/3290198
- http://stackoverflow.com/a/6110891

I'm not sure if there are things on the Android side yet that specifically expect a 412. But I can fix up if need be. @aaronschachter anything on the iOS side specifically listening for a 412?

aaronschachter: @jonuy Nothing specific for 412, just currently displaying the error message (and code, while we're in development).

PR #192, opened 2015-07-29
makes login case insensitive

### LINKED ISSUE #184: Make login case insensitive
Currently get a 404 when attempting to login when you use different capitalization for an email that exists.

chloealee:
Fixes #184

### DIFF
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -27,9 +27,9 @@ class AuthController extends Controller
         ]);
 
         if ($request->has('email')) {
-            $user = User::where('email', '=', $input['email'])->first();
+            $user = User::where('email', 'LIKE', $input['email'])->first();
         } elseif ($request->has('mobile')) {
-            $user = User::where('mobile', '=', $input['mobile'])->first();
+            $user = User::where('mobile', 'LIKE', $input['mobile'])->first();
         }
 
         if (!($user instanceof User)) {
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -27,9 +27,10 @@ class AuthController extends Controller
         ]);
 
         if ($request->has('email')) {
-            $user = User::where('email', 'LIKE', $input['email'])->first();
+            $email = strtolower($request->email);
+            $user = User::where('email', '=', $email)->first();
         } elseif ($request->has('mobile')) {
-            $user = User::where('mobile', 'LIKE', $input['mobile'])->first();
+            $user = User::where('mobile', '=', $input['mobile'])->first();
         }
 
         if (!($user instanceof User)) {


jonuy:
@chloealee By using `LIKE` does that also allow for logins where the email/mobile are different lengths than what they should be?

Like, if my account is `test@dosomething.org`, would this change also allow me to login with `test@dosomething.orgishness`?

chloealee:
@jonuy nope - it shouldn't! Should just account for case.

angaither:
you might also just be able to `strtolower()`before running `=`

chloealee:
@angaither should I use `strtolower()` instead of of `LIKE`?

angaither:
@chloealee yeah, I think @jonuy is right, it does pattern matching and it would allow non-matches to pass.

chloealee:
ah got it! I tried to test with the example that @jonuy provided above and it threw the error when i typed in `test@dosomething.orgishness` but didn't when I did a mix of upper case and lower case for `test@dosomething.org` so thought that could be the solution. I can do more digging though to make sure. 

@jonuy and I were chatting yesterday and he said he wanted to avoid converting all to lowercase before searching because some emails might not be stored in the database as such.

jonuy:
It looks like `LIKE` is essentially doing a [regex query](https://github.com/jenssegers/laravel-mongodb/blob/e2bc57c374845ed18ce55e4d8df4b676ae8bac73/src/Jenssegers/Mongodb/Query/Builder.php#L918). So yea, maybe this does work. But then this seems like it'd open the door for someone to input regex in place of a phone or number and possible wreak some havoc?

Doing a regex query also negates any benefits we'd have from indexing on the email or mobile field. So maybe what we do actually end up doing is a `strtolower` before executing a query? Which would mean either lowercasing the values before putting them in the db, or holding a lowercased version in a separate field explicitly for querying against, and also running a script through the db that modifies any docs that don't have lowercased values.

@angaither @chloealee what do you think we should do?

angaither:
I think it would be a pretty simple validation rule to lowercase them before we save them & run a script to update any thing we have in caps currently. 

I really don't think the regex query is a good idea for saftey/indexing

chloealee:
Great that makes sense! Would the script be a separate issue to create?

chloealee:
Found this on lines 63-70 in the `User` model - does this mean that emails are already being stored in lowercase? 

```
    /**
     * Email address mutator that converts the email value to lowercase
     *
     */
    public function setEmailAttribute($value)
    {
        $this->attributes['email'] = strtolower($value);
    }
```

angaither:
yeah! awesome, I thought so, but wasn't sure ;)

chloealee:
great, thanks for confirming, @angaither ! I'll push the changes I made now then for review :)

angaither:
:+1:

PR #194, opened 2015-08-05
User Dedupe Script

### LINKED ISSUE #187: Remove duplicate users
Refs #182 

Need to go into production and write a script that searches all users, looks to see if there is another user with that email/mobile
then removes the user with the most recent `updated_at` attribute (as that is the second time the user was added from the script) 

This should be a [custom artisan command](http://laravel.com/docs/5.1/artisan#writing-commands) that first grabs all users, loops through each one, and searches for a user with the same email or phone number.

chloealee:
#### What does this PR do?
Adds script to remove duplicate users and adds mobile number conversion in User model to only save numbers (no spaces, -, ., etc.). 

Run `php artisan users:dedupe` 
@sheyd and I also tested in Thor and seems to be working correctly, except for one error (Sena will elaborate on this in comments below). 

Fixes #187


DFurnes:
@chloealee If you rebase from `dev` you should get much more reasonable feedback from StyleCI. I can sit down with you sometime today or tomorrow to look through that if you want! :)

chloealee:
that would be awesome, thanks @DFurnes !

sheyd:
@DFurnes @blisteringherb @angaither Quick notes from the testing @chloealee and I did on Thor Northstar:

1. Looks like it gets through the end of all records (user records scrubbed for security):
```
user deleted: zzz@example.com 555xxx555xxx5x555x55555xxxx5x5x5x
user deleted: zzz_1234@example.com.uk 666xxx666xxx6x666x66666xxxx6x6x6x
```

2. This error pops up at the end:
```
[MongoResultException]                                                            
  mongo1:27017: exception: aggregation result exceeds maximum document size (16MB) 
```

Before we roll into Prod on this, I'd like to do a Thor Northstar re-fresh with updated backend values (the ones in Thor are much smaller than Prod), and re-run the test. Thoughts?

angaither:
yeah, let's retest!

sheyd:
We hit another error, this time: 

```
 [ErrorException]     
  Undefined offset: 0  
```

I'll re-sync with @chloealee on this next week when she's back.

chloealee:
@sheyd did a little research and realized that I didn't need the `[0]` index in one of the loops. I think this triggered the ```[ErrorException]     
  Undefined offset: 0  ``` so should be good now! Can we test again?

DFurnes:
OMG THERE ARE SO MANY COMMITS IN HERE :scream:

chloealee:
soon they're be crushed along with all my little green boxes! :sob:

chloealee:
@angaither thanks to @jonuy 's help and refactored code, the script is now working! You have to run the script a few times until only the message `Script successful. Re-run script until no users are deleted and yields only this message. If only this message, deduplication complete!` is returned (with no `echo`s of users being deleted beforehand). 

I have the Jenkins jobs set up on my computer to test if you want to test together to go through - just let me know!

angaither:
:+1: cool, we can get this merged, I'm not sure if we want to run it now before break or wait till next year. but awesome job with this!

chloealee:
@angaither added another line of code I was missing to properly switch users and make sure that the `master_doc` is always the original user. 

Just ran it again in Thor and from the users I saw (three random emails and three random mobile numbers) it worked correctly! Now I’m just trying to find a user that was created after the original that has new fields the original didn’t to see if that merged correctly but am having trouble finding examples. I saw that it worked in tinker when creating this scenario. Is it worth it to keep looking for a real example in Thor?

angaither:
:+1: from here, @DFurnes wanna take a look as well?

chloealee:
@DFurnes added the conditional and changed the regex to match voting app for consistency!

DFurnes:
Looks good to me! :ship: :tada:

Issue #183, opened 2015-07-23
Non-standard error message when registering with a mobile number that's already taken


jonuy:
Current error response is something like...
```
mobile =     (
       "The mobile has already been taken."
)
```

The expectation is the error would be something like...
```
{
  error: {
    code: 412,
    message: "Incorrect password."
  }
}
```

Maybe we throw a try/catch block around the validate code in `UserController::store`? I'm assuming the error response is getting triggered by an exception coming out of there

DFurnes:
We should be able to just override the error handler [`buildFailedValidationResponse`](https://github.com/laravel/framework/blob/16f59c2d86eee2e67c7947fd5c7a1a8fafc95ad4/src/Illuminate/Foundation/Validation/ValidatesRequests.php#L74-L90) that's added from the `ValidatesRequests` trait in our `Controller` base class.

aaronschachter:
Any chance who ever takes this on can close out  #172 also?

chloealee:
@jonuy @aaronschachter looks like there is no error message popping up if the mobile number exists already and instead just let's the user overwrite the password like #172 says. I can hop on #172 as well and fix that before this issue/wrap this issue into the fix for #172.

PR #195, opened 2015-08-06
adds password check so user cannot overwrite existing data

### LINKED ISSUE #172: Submitting a user register when the user already exists will overwrite existing data
This means it can be used maliciously to allow one person to change another's password and have access to their account.

ref: https://github.com/DoSomething/LetsDoThis-iOS/issues/126

chloealee:
Fixes #172

### DIFF
--- a/app/Http/Controllers/UserController.php
+++ b/app/Http/Controllers/UserController.php
@@ -7,6 +7,7 @@ use Input;
 use Response;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 
+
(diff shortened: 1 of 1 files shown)


jonuy:
What happens when you try to send a POST request to an existing account but with an incorrect password?

chloealee:
It throws this error: 

```
{
  "email": [
    "The email has already been taken."
  ]
}
```

I'll need to update this error message though to tackle and close out #183

jonuy:
:sunglasses: :+1:

Issue #180, opened 2015-07-20
Add index on User source field.


angaither:
@jonuy looking at returning specific users in aurora (DoSomething/aurora/issues/66)

We were thinking about creating an enpdoint that returns all users from a given source
Does `users?source=agg` make sense as the endpoint

jonuy:
@angaither yea, I think that'd be fine. Do you think we'd also want to support querying against multiple sources? Something like `/users?source=agg,bgg,cgg,`?

angaither:
yes! @jonuy great ideas!

DFurnes:
This is something that'd be super useful for Aurora. We just need to add a MongoDB index on the `source` field, and then we can use this with existing `?filter[source]=agg,cgg` functionality.

DeeZone:
Adding to list of requestors for this feature:

> There's a need to create MailChimp entries from mobileapp user registrations to ensure they're getting the "Weekly Do" email newsletter. This will be taken care of automagically once the `quicksilver-api` `/user/register` endpoint is in production. Until then, the plan is to create a daily batch job to collect `mobileapp` user registrations from Northstar and add them to MailChimp.

> I think I'll need a functional GET `/v1/users?filter[source]=mobileapp` type endpoint on Northstar to make this happen.

It might also be helpful to also filter by a timeframe as suggested by @DFurnes :
> I guess we’d also want to add a way to query by timeframe as well then?

Issue #179, opened 2015-07-20
Seed dummer users with a source


angaither:
Refs DoSomething/aurora/issues/66

PR #196, opened 2015-08-10
Get Users by sources endpoint

axton21:
# What's this PR doing?
This is also to address an Aurora issue DoSomething/aurora#66 
We need an endpoint to get users by source.
We're not sure if this is the right way to go about this, please let me know if any changes are needed.

# Any involved tickets?
* #179 Seed dummer user with a source
* #180 Endpoint that returns all users with a source

### DIFF
--- a/app/Http/Controllers/UserController.php
+++ b/app/Http/Controllers/UserController.php
@@ -6,7 +6,6 @@ use Northstar\Models\User;
 use Input;
 use Response;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
-// use Illuminate\Database\Eloquent;
 
 class UserController extends Controller
 {
@@ -16,15 +15,13 @@ class UserController extends Controller
      * GET /users
      * Get /users?attr1=value1&attr2=value2&...
      *
-     * @TODO ?page= in next_page_url and last_page_url has to be changed to
-     *       &page= in the end.
      * @return Response
      */
     public function index()
     {
          $inputs = Input::except('page');
          $users = User::where($inputs);
-         $response = $this->respondPaginated($users);
+         $response = $this->respondPaginated($users, $inputs);
          if (!empty($response->getData()->data)) {
            return $response;
          }
--- a/app/Http/Controllers/UserController.php
+++ b/app/Http/Controllers/UserController.php
@@ -25,12 +25,11 @@ class UserController extends Controller
         }else{
             $inputs = Input::all();
             $users = User::where($inputs)->get();
-            return $this->respond($users);
-        }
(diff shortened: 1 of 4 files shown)


axton21:
@jonuy

jonuy:
@axton21 The code itself looks fine. Question though - why a new endpoint as opposed to allowing for filtering by a query parameter instead? Like it was mentioned in #180, why not something along the lines of `/users?source=agg,bgg,cgg`?

axton21:
@jonuy Sounds like a good idea, we got excited since it works in first try. thanks! We'll do some research on that and push commit here.

axton21:
@jonuy using `/users?source=agg,bgg,cgg` so the endpoint will be able to handle multiple queries as well.
So far aurora only needs single source query, but I think we will build it this way to future proof the code.

axton21:
@angaither @jonuy 
### This function accept requests:
* `/users` to return all users (same as before).
* `/users?term1=value1&term2=value2` to return API response of more specific search.

app/Http/Controllers/UserController.php
```diff
@@ -14,13 +14,23 @@ class UserController extends Controller
...
+     * Get /users?attr1=value1&attr2=value2&...
      *
      * @return Response
      */
     public function index()
     {
-        $query = User::query();
-        return $this->respondPaginated($query);
+        $inputs = Input::all();
+        if(empty($inputs)){
```
jonuy: Spacing needs to be cleaned up here to follow our code style...
```
if (...) {
  ...
} else {
  ...
}
```

app/Http/Controllers/UserController.php
```diff
@@ -14,13 +14,23 @@ class UserController extends Controller
...
+            $query = User::query();
+            return $this->respondPaginated($query);
+        }else{
+            $users = User::where($inputs)->paginate(20);
```
jonuy: ok, so I'm not entirely sure, but does `User::where()` return a type query `Builder`? In which case you could do something like...
```
return $this->respondPaginated(User::where($inputs));
```

It'd simplify the logic a little, would allow us to still use the `respondPaginated` function to do the pagination work, and avoids us having to throw an Exception.

axton21:
@jonuy Apparantely this code makes the pagination url doesnt work.  I ll do some research and changes again.

jonuy:
Hm :confused: ok, thanks. The main thing really is just getting paginated responses to be sent by the `respondPaginated` function to make it easy to standardize things.

axton21:
@jonuy

axton21:
@jonuy

jonuy:
:+1:

PR #197, opened 2015-08-13
Standardizing error response on failed login

### LINKED ISSUE #183: Non-standard error message when registering with a mobile number that's already taken

jonuy:
#### What's this PR do?
Standardizes the error message response when credentials fail to be like...
```
{
  "error": {
    "code": 401,
    "message": "Invalid email or password."
  }
}
```

When a user tries to register with an email or mobile that already exists, the response is...
```
{
  "error": {
    "code": 401,
    "message": "The mobile has already been taken."
  }
}
```

- **AuthController.php**: The exception thrown when the auth is incorrect now has a status code of 401, and it specifies `email` or `mobile` in the error message as opposed to `username`.
- **UserController.php**:
  - Simplifies the password check logic so that we're not executing a `User::where` query multiple times.
  - Overrides the `buildFailedValidationResponse` function to deliver a custom error response.
- **Controller.php**: Overrides the `buildFailedValidationResponse` function from the `ValidatesRequests` trait

Closes #183

cc: @DFurnes - Am I doing this ValidatesRequests trait thing right?

### DIFF
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -6,6 +6,7 @@ use Illuminate\Http\Request;
 use Hash;
 use Symfony\Component\HttpKernel\Exception\HttpException;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+use Symfony\Component\HttpKernel\Exception\UnauthorizedHttpException;
 
 class AuthController extends Controller
 {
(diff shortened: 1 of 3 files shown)


DFurnes:
@jonuy Looks good to me!

PR #198, opened 2015-08-13
Reportback Source Property

jonuy:
#### What's this PR do?
Adds support for a `source` property for campaign reportbacks. We're already both saving a `source` property to Northstar and forwarding it along to Phoenix for user registrations and campaign signups.

**CampaignController.php**

- A `source` string passed to the reportback endpoint will be saved in the campaign doc under `reportback_source`.
- The response from the reportback endpoint will now give the entirety of the campaign activity doc. Previously it would give something like...
```
{
        "data": {
          "reportback_id": "1622",
          "created_at": {
            "date": "2015-07-15 15:29:10",
            "timezone_type": 3,
            "timezone": "UTC"
          },
          "drupal_id": null,
          "signup_group": null,
          "signup_id": null,
          "signup_source": null
        }
      }
```
... where a bunch of fields were incorrectly `null`.

cc: @DFurnes @angaither

### DIFF
--- a/app/Http/Controllers/CampaignController.php
+++ b/app/Http/Controllers/CampaignController.php
@@ -1,6 +1,7 @@
 <?php namespace Northstar\Http\Controllers;
 
 use Illuminate\Http\Request;
+use Input;
 use Northstar\Events\UserSignedUp;
 use Northstar\Events\UserReportedBack;
 use Northstar\Models\Campaign;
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/CampaignController.php
```diff
@@ -180,12 +182,17 @@ public function reportback($campaign_id, Request $request)
         }
 
         $campaign->reportback_id = $reportback_id;
+
+        if (Input::has('source')) {
```
DFurnes: I'd suggest using the controller's `$request` object here: `$request->has('source')`. I think thats the preferred way of doing things in Laravel 5.

PR #199, opened 2015-08-14
updates error message

### LINKED ISSUE #183: Non-standard error message when registering with a mobile number that's already taken

chloealee:
Fixes #183

### DIFF
--- a/app/Console/Commands/RemoveDuplicateUsersCommand.php
+++ b/app/Console/Commands/RemoveDuplicateUsersCommand.php
@@ -1,0 +1,77 @@
+<?php namespace Northstar\Console\Commands;
+
+use Illuminate\Console\Command;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Input\InputArgument;
+use Northstar\Models\User;
+
+class RemoveDuplicateUsersCommand extends Command {
+
+	/**
+	 * The console command name.
+	 *
+	 * @var string
+	 */
+	protected $name = 'users:dedupe';
+
+	/**
+	 * The console command description.
+	 *
+	 * @var string
+	 */
+	protected $description = 'Script to remove duplicate users.';
+
+	/**
+	 * Create a new command instance.
+	 *
+	 * @return void
+	 */
+	public function __construct()
+	{
+		parent::__construct();
+	}
+
+	/**
+	 * Execute the console command.
+	 *
+	 * @return mixed
+	 */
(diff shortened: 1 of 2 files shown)


chloealee:
Hi @jonuy ! I accidentally added the other changes I'm making for a script I'm working on. If this solution in `UserController.php` gets approved, can I just add the changes in that one file?

chloealee:
oh shoot - I just realized that you already closed that issue! Sorry I thought it would automatically update but looks like it didn't because it wasn't refreshed. Should I just close this @jonuy ?

app/Http/Controllers/UserController.php
```diff
@@ -53,6 +53,8 @@ public function store(Request $request)
 
         if ($found_user && password_verify($input['password'], $found_user->password)) {
             $user = $found_user;
+        } else {
+            throw new HttpException(412, 'Incorrect password.');
```
jonuy: What I'd done in a previous commit was let the `validate` function handle catching the error and throwing the exception instead of it happening here. The error message there specifies if a phone number or email is already taken. I _think_ that's ok, considering this endpoint's meant to be used for user registration. Let me know if you haves cases where we'd want to throw the exception here though.

app/Console/Commands/RemoveDuplicateUsersCommand.php
```diff
@@ -0,0 +1,77 @@
...
+	 *
+	 * @return mixed
+	 */
+	public function fire()
+	{
+        // Get all the users.
+		$users = User::all();
+
+        $duplicates = User::select('email')->havingRaw('count(*) > 1')->get()->toArray();
+        dd($duplicates);
```
jonuy: Might not want to commit the `dd()`

jonuy:
Oh oops, you said the Command you were working on is still in progress. Should I not review that then?

And then with your UserController change, yea I made a change that should close it out, but there might be some use-case I'm not thinking about where your solution also makes sense to have. If you can think of one, let me know. And then I think it might just be better bringing that change in from its own new branch, separate from the Command work you're doing.

PR #201, opened 2015-08-18
Profile photo upload fix

### LINKED ISSUE #200: Updating avatar photo for any user on staging only updates `test@dosomething.org` account
Using the `POST /users/:user_id/avatar` call on staging only updates the `test@dosomething.org` account with the ID of: `POST /users/:user_id/avatar`.

jonuy:
#### What's this PR do?
- Fixes the query that would find the user to update with an avatar submission
- Fixes the response to put the url in the `photo` param instead of an `avatar` param

#### Any background context?
The bug would manifest itself in updating the profile of a user that was not the user specified in the originating call.

Fixes #200


chloealee:
:+1:

PR #203, opened 2015-08-27
Fix on GET /users/:term/:id/campaigns when reportbacks are deleted on Drupal

jonuy:
#### What's this PR do?

When reportbacks get deleted on Drupal, they're not removed from Northstar. So when Northstar queries for them, there is no 'data' object that comes back in the response and it dies. This PR just does an `isset` check as a quick fix to prevent the process from terminating.

#### Questions

Thoughts on if we'll ultimately want to remove the reportback id from the user's document on Northstar too? Potentially could happen here - which feels iffy. Or could happen as an event triggered from Drupal when a reportback is deleted there.


angaither:
:+1: 
we should probably handle removing the reportback/signups from drupal.

jonuy:
Cool. I'll create an issue for it.

PR #204, opened 2015-09-11
changes drupal_id to an integer

chloealee:
#### What does this PR do? 
Changes the `drupal_id` to an integer to fix the `GET /users/drupal_id/<drupal_id>` endpoint


DFurnes:
:hammer: :ship:

PR #206, opened 2015-09-15
Fixes /users PUT response to include actual user data

### LINKED ISSUE #205: PUT /users response returns user data with null values
Sending a PUT request to the /users endpoint to update a user document does work. It returns a 202 and the document is actually updated.

But the response will look something like...
```
{
  "data": {
    "updated_at": {
      "date": "2015-08-18 17:59:13",
      "timezone_type": 3,
      "timezone": "UTC"
    },
    "email": null,
    "mobile": null,
    "first_name": null,
    "last_name": null,
    "birthdate": null,
    "photo": null,
    "interests": null,
    "race": null,
    "religion": null,
    "school_id": null,
    "college_name": null,
    "degree_type": null,
    "major_name": null,
    "hs_gradyear": null,
    "hs_name": null,
    "sat_math": null,
    "sat_verbal": null,
    "sat_writing": null,
    "addr_street1": null,
    "addr_street2": null,
    "addr_city": null,
    "addr_state": null,
    "addr_zip": null,
    "country": null,
    "cgg_id": null,
    "drupal_id": null,
    "agg_id": null,
    "source": null,
    "parse_installation_ids": null
  }
}
```

Even if those values have been set on the user, they'll just return null.

jonuy:
#### What's this PR do?
The response to a PUT /users request used to only have a valid `updated_at` field. All other user data was null. This fixes that.

This also updates a phpunit test to verify that.

#### Any background context?
Came across this while doing work on the Android app. The app would update its cache of user information based on what'd get returned from the /users PUT response. And the previous response basically seemed like it would null everything out.

Fixes #205

### DIFF
--- a/app/Http/Controllers/UserController.php
+++ b/app/Http/Controllers/UserController.php
@@ -161,9 +161,7 @@ class UserController extends Controller
 
             $user->save();
 
-            $response = array('updated_at' => $user->updated_at);
-
-            return $this->respond($response, 202);
+            return $this->respond($user, 202);
         }
 
         throw new NotFoundHttpException('The resource does not exist.');
--- a/tests/UserTest.php
+++ b/tests/UserTest.php
@@ -112,8 +112,9 @@ class UserTest extends TestCase
         // Response should be valid JSON
         $this->assertJson($content);
 
-        // Response should return updated at and id columns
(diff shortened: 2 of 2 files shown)


angaither:
:+1:

PR #209, opened 2015-10-06
Return a 422 on signup requests where a signup already exists on Phoenix

jonuy:
#### What's this PR do?

On the mobile app, I need to distinguish between a failed signup because of a server error and a failed signup because it actually already exists.

Phoenix responds with a `200` status code but an empty body in cases of signups that already exist. When that happens, I return a `422` response instead of a `500` like it did previously.

#### More background info

These scenarios where Phoenix has the signup already but Northstar doesn't shouldn't really happen in practice. But it has been encountered in some test scenarios where NS signup data is removed but hasn't been on Phoenix. And it is at least slightly more of a graceful handling of the problem than the `500` we'd sent before.

### DIFF
--- a/app/Services/DrupalAPI.php
+++ b/app/Services/DrupalAPI.php
@@ -3,6 +3,7 @@
 use GuzzleHttp\Client;
 use Config;
 use Cache;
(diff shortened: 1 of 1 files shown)


chloealee:
:+1:

Issue #215, opened 2015-10-20
Upgrade to Laravel 5.1


DFurnes:
Let's get this bugger updated. LTS, built-in authorization, testing niceties, HHVM support, … :computer:

PR #214, opened 2015-10-19
adds  to update function and updates routes

chloealee:
#### What does this PR do? 
Updates `PUT /users/:id` to `PUT /users/:term/:id` so it's possible to query with more than just `:mongo_id` e.g.:

`PUT /users/drupal_id/:drupal_id`
`PUT /users/_id/:mongo_id`

#### Where should the reviewer start? 
`app/Http/routes.php` line 42
`app/Http/Controllers/UserController.php` lines 135-149

#### How should this be manually tested? 
Using Postman or Paw, hit endpoint with `PUT /users/drupal_id/:id` and submit` college_name` in body. When request goes through, make sure `college_name` is updated.

#### What are the relevant tickets?
Fixes #213

PR #216, opened 2015-10-20
Quick fixes

DFurnes:
#### Changes
I wanted to make sure everything was working okay before [trying to upgrade to Laravel 5.1](https://github.com/DoSomething/northstar/issues/215), and ran into a failing test! Fixed the test to reference the new URL structure changed in #214, and also removed the broken route so it `404`s instead of blowing up.

I'd recommend just reading each commit.

`vendor/bin/phpunit` :traffic_light: 

---
For review: @jonuy @chloealee

### DIFF
--- a/app/Http/routes.php
+++ b/app/Http/routes.php
@@ -36,7 +36,7 @@ Route::group(['prefix' => 'v1', 'middleware' => 'auth.api'], function () {
 
     // Users.
     Route::group(['middleware' => 'user'], function () {
-        Route::resource('users', 'UserController');
+        Route::resource('users', 'UserController', ['except' => ['show', 'update']]);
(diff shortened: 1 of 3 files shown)


app/Http/routes.php
```diff
@@ -36,7 +36,7 @@
 
     // Users.
     Route::group(['middleware' => 'user'], function () {
-        Route::resource('users', 'UserController');
+        Route::resource('users', 'UserController', ['except' => ['show', 'update']]);
```
jonuy: ooo I didn't know about this `except` business. Thanks!

jonuy:
:+1:

app/Http/routes.php
chloealee: me neither, so cool! Thanks!!

PR #219, opened 2015-10-22
Multiple users can be retrieved by _id and drupal_id

### LINKED ISSUE #212: Add support for getting multiple users' information by _id in a single call
#### The request

We currently have a way to [get multiple users](https://github.com/DoSomething/northstar/wiki/Spec#retrieving-all-users) but only based on a single shared attribute. What we'd like to add is support for something like...

`GET /users?_id=1,2,3,4,5`

#### Problem this should help fix

When we query Phoenix for [reportback-items](https://github.com/DoSomething/phoenix/wiki/API#retrieve-a-collection-of-reportback-items), it also queries Northstar for user info for each. So N reportback-items by N different users turns into N additional requests that are actually being made. Offering support for getting multiple users' by _id turns those N additional requests into just a single one.

#### Also up for debate

Do we name the param `_id` as it is in the database and other parts of the API, or start changing things to `id` which might seem to ... be more familiar? make more sense?

jonuy:
#### What's this PR do?
Primarily this allows for multiple identifiers to be used for the `_id` and `drupal_id` params on the `GET /users` endpoint. Additional attributes can still be included like before.

ex:
- GET /users?_id=100,101,102
- GET /users?drupal_id=100,101,102
- GET /users?_id=100,101,102&mobile=100

This PR also includes a unit test for the new functionality plus some unit test fixes for a plain ole `GET /users` and a test for a nonexistent user.

The new stuff in UserController basically turns the comma-delimited values of the `_id` and `drupal_id` params into arrays. Those arrays then get added to the Builder query via a whereIn, but only if there was anything passed to `_id` or `drupal_id`.

#### How was this tested?
I manually tested the examples above with real values and ran the PHPUnit tests.

#### Any background context?
This should help with the slowness from the reportback-items endpoint. Ref https://github.com/DoSomething/phoenix/pull/5408

Closes #212

#### Questions
I ran into an issue when using the `XDEBUG_SESSION_START=PHPSTORM` flag while debugging. Because of line 25 where it goes... `$users = User::where($inputs);`, it'd try to find user documents that had the `XDEBUG_SESSION_START` property and ultimately wouldn't find any. I added that string to the `except()` array, but it felt weird to commit that to the repository.

Should I have included it anyway for future debugging?

cc: @DFurnes @angaither @weerd 

I really don't know who I should be assigning Northstar stuff too.

### DIFF
--- a/app/Http/Controllers/UserController.php
+++ b/app/Http/Controllers/UserController.php
@@ -21,13 +21,14 @@ class UserController extends Controller
      */
     public function index()
     {
-        $inputs = Input::except(['page', '_id', 'drupal_id']);
+        $except_list = User::$indexes;
+        array_push($except_list, 'page');
+        $inputs = Input::except($except_list);
         $users = User::where($inputs);
 
         // Query for multiple ids
         $query_ids = [];
-        $id_keys = ['_id', 'drupal_id'];
-        foreach ($id_keys as $id_key) {
+        foreach (User::$indexes as $id_key) {
             if (Input::has($id_key)) {
                 $str_ids = Input::get($id_key);
                 $arr_ids = explode(',', $str_ids);
--- a/app/Http/Controllers/UserController.php
+++ b/app/Http/Controllers/UserController.php
@@ -5,7 +5,6 @@ use Northstar\Services\DrupalAPI;
 use Northstar\Models\User;
 use Input;
 use Response;
-use Symfony\Component\HttpKernel\Exception\HttpException;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Symfony\Component\HttpKernel\Exception\UnauthorizedHttpException;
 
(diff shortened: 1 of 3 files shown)


jonuy:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;odf9VIQt&#x2F;22-northstar-get-users-endpoint-updated-to-support-queries-for-multiple-users"> Northstar GET &#x2F;users endpoint updated to support queries for multiple users</a>

app/Http/Controllers/UserController.php
```diff
@@ -22,8 +21,24 @@ class UserController extends Controller
      */
     public function index()
     {
-        $inputs = Input::except('page');
+        $inputs = Input::except(['page', '_id', 'drupal_id']);
```
DFurnes: It might make sense to whitelist this instead? Perhaps you could limit it to things in the `User`'s `$fillable` array?

app/Http/Controllers/UserController.php
```diff
@@ -22,8 +21,24 @@ class UserController extends Controller
...
         $users = User::where($inputs);
+
+        // Query for multiple ids
+        $query_ids = [];
+        $id_keys = ['_id', 'drupal_id'];
```
DFurnes: It might be smart to put these as an `$indexes` property on the User, and then grab them from anywhere that we'd want to know what's safely query-able.

chloealee:
:+1: so cool to see how you did this, Uy!

DFurnes:
Looks good to me! :boat: Added two thoughts on potential ways to avoid `XDEBUG_SESSION_START=PHPSTORM` messin' with things, but definitely not blockers to merge.

jonuy:
@DFurnes I made a change in 3cd51ba in response to your suggestions. Does this seem more okay?

And then I ended up not using the `$fillable` array in the `Input::except` line primarily because `_id` wasn't in that list and shouldn't be.

DFurnes:
Cool! :shipit:

PR #217, opened 2015-10-20
Update to Laravel 5.1

DFurnes:
#### Changes
Updates Northstar to Laravel 5.1. (Closes #215.) For details on specific changes, see the [Upgrade Guide](http://laravel.com/docs/5.1/upgrade#upgrade-5.1.0).

#### How should the reviewer start?
Each change relates to an item in the [Upgrade Guide](http://laravel.com/docs/5.1/upgrade#upgrade-5.1.0), so I'd recommend skimming through that before reviewing the code changes.

#### How should this be manually tested?
Double-check that endpoints still function as expected. The PHPUnit test suite remains green. :traffic_light: 

---
For review: @jonuy

PR #220, opened 2015-10-23
Fixes error that was ocurring during Artisan optimize.

DFurnes:
#### Changes

I forgot to remove these from the providers array when I was [updating things to Laravel 5.1](https://github.com/DoSomething/northstar/pull/217), so it was getting all angry that it couldn't load them (because the files don't exist anymore). Whoops!

Either run `php artisan optimize --force`, or just verify the Wercker build is happy. :cactus: 

---

For review: @sheyd

### DIFF
--- a/config/compile.php
+++ b/config/compile.php
@@ -16,8 +16,6 @@ return [
     'files' => [
 
         realpath(__DIR__ . '/../app/Providers/AppServiceProvider.php'),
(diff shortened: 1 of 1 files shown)


DFurnes:
And [it works](https://app.wercker.com/#buildstep/562a4bd53b5372590b01c350)! Although, we're not getting the little "✓" in Github which is a bummer.

sheyd:
:chart:

PR #224, opened 2015-10-26
Add an index to `drupal_id` column.

DFurnes:
#### Changes
Adds a [sparse index](https://docs.mongodb.org/manual/core/index-sparse/) on the `drupal_id` field for the `users` collection. This should fix the really slow performance we're seeing on the the UserController@show endpoint.

#### Screenshots
Indexes before the migration:
screen shot 2015-10-26 at 3 40 16 pm

Indexes after migration:
screen shot 2015-10-26 at 3 40 59 pm

---
For review: @jonuy


jonuy:
All the :+1:s

Issue #222, opened 2015-10-23
Wercker git hook broken


DFurnes:
It looks like this broke after we renamed this repository from `user-api`. Builds are still happening, but Wercker is trying to notify the old repository of changes so they're not showing up on PRs or commits.

Apparently, Wercker [don't have a way to change the GitHub repository](https://github.com/wercker/support/issues/78) in their interface, so I think its attempts to "notify" the old repo and just getting swallowed by the redirect. The recommended fix (which worked for Forge when I had to fix that) was just to delete the Wercker app and re-create. :grimacing: 

/cc @blisteringherb

DFurnes:
Little follow-up on this: it looks like we can [contact their support](http://devcenter.wercker.com/docs/faq/how-can-i-update-a-repo-url.html) to change the repo URL.

DFurnes:
I submitted a help ticket, will update when I hear back.

DFurnes:
Fixed! Closing.

PR #227, opened 2015-11-06
Run PHPUnit test suite on builds.

DFurnes:
#### Changes
Run our PHPUnit test suite on all pull requests using Wercker. This is useful because now we can't accidentally merge code in that breaks the build. Automate all the things! :traffic_light: 

Like DoSomething/voting-app#435, this requires some modifications to the Wercker box that we're doing on every build, but we can clean that up later. In particular, it installs MongoDB and the PHP MongoDB extension.

The Wercker build [already did](https://app.wercker.com/#buildstep/563d24726709dc217b045c52)! ~~Well, almost – the web hook on the Wercker app for this repository broke when we changed repository names, and [needs to be fixed](https://github.com/DoSomething/northstar/issues/222).~~ And webhooks are now working again!

---
For review: @jonuy @angaither

### DIFF
--- a/wercker.yml
+++ b/wercker.yml
@@ -4,15 +4,27 @@ build:
   # The steps that will be executed on build
   steps:
     - script:
+        name: Install MongoDB
+        code: |-
+          sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 7F0CEB10
+          echo "deb http://repo.mongodb.org/apt/ubuntu "$(lsb_release -sc)"/mongodb-org/3.0 multiverse" | sudo tee /etc/apt/sources.list.d/mongodb-org-3.0.list
+          sudo apt-get update
+          sudo apt-get install -y mongodb-org
+          sudo service mongod status
+    - script:
         name: Update Composer
         code: sudo composer self-update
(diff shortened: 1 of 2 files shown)


wercker.yml
```diff
@@ -4,15 +4,27 @@ build:
...
     - script:
-        name: update packages list
-        code: sudo apt-get update
-    - script:
         name: install php mongo extension
         code: sudo apt-get install php5-mongo
+    - script:
+        name: phpunit
+        code: |-
+          cp .env.example .env
```
jonuy: Is there any concern about the .env.example file being modified in some way that it'll actually break tests when run on wercker?

DFurnes: Ideally that'd always be a "clean slate" that you could just copy to get a new environment running. If it does break a build though, that's a good reminder for us to update instructions in the README or update the `.env.example`!

DFurnes:
Wercker support [was able to update the repository name for our app](https://github.com/DoSomething/northstar/issues/222), and builds are now hooked up properly and showing up on pull requests again. :green_heart:

jonuy:
:rocket:

PR #229, opened 2015-11-12
drupal password checker 

### LINKED ISSUE #226: Drupal password checker
### The Problem

We have a lot of users in Northstar that have a `drupal_password`, but no `password`.

**Number of accounts with a drupal_password, but no password:** 2,640,924

`db.users.find({drupal_password:{$exists:true}, password:{$exists: false}}).count()`

---

_Not sure if this information actually adds anything to the discussion, but I ran these queries anyway, so here it is..._

**Number of accounts with a drupal_password:** 2,756,471

`db.users.find({drupal_password:{$exists:true}}).count()`

**Number of accounts without a password:** 235,599

`db.users.find({password:{$exists:true}}).count()`

---

So we could feasibly run into a lot of scenarios where someone tries to login on the mobile app with the credentials they already had on the Drupal site and get an "invalid login" error.

### A Solution?

Create something in the Northstar login flow that authenticates against the user's Drupal password.

1. User sends a login request
2. If no `password` is found on the user doc, attempt to authenticate against the `drupal_password`
3. If it succeeds (or maybe even if it doesn't), copy `drupal_password` into the user's `password` field. So then for future requests, the user will have their `password` field properly filled in.
4. Allow or deny authentication like normal.

@marahml Could this be a launch blocker to the mobile app?

cc: @angaither @chloealee @DFurnes

chloealee:
#### What does this PR do?
adds DrupalPasswordChecker, Password, and Registrar services to log in a user in with drupal password if there is no password. It then updates the users by saving the password to `$user->password` and deletes `$user->drupal_password`.

AuthController -> Registrar -> DrupalPasswordChecker -> Password

1. Find a user that only has a `drupal_password` (or delete a user's `password` by going into mongo and running `db.users.update({email: "user_email@email.com"}), {$unset: {password:1}})` in the 
userapi db). 

2. In Postman/Paw, hit the `/login` endpoint. Body should have `email` and `password`: 
```
{
  "email": "user_email@email.com", 
  "password": "<drupal-pw>"
}
```

3. Even though the user did not have a `password`, it should log the user in with the `drupal_password` and return the user object like so: 
```
{
  "data": {
    "_id": "563d251cbffebc482d8b4568",
    "email": "clee@dosomething.org",
    "drupal_id": "1704953",
    "birthdate": "1995-04-12 00:00:00",
    "first_name": "Chloe",
    "country": "US",
    "updated_at": "2015-11-12 21:06:57",
    "created_at": "2015-11-06 22:09:32",
    "session_token": "c2dIdWE2UEF0SzBNNTZsSEN5TzI1bmdSSDBKTEQxb3p0eGk5R2c0Rk9VPQ=="
  }
}
```
4. Check the user in mongo and make sure that `drupal_password` = `null` and there is now a stored `password` in the user object by using `db.users.find({email: "user_email@email.com"}).pretty()`

Fixes #226


app/Http/Controllers/AuthController.php
```diff
@@ -3,14 +3,23 @@
...
 use Symfony\Component\HttpKernel\Exception\UnauthorizedHttpException;
+use Northstar\Services\Registrar;
+
 
 class AuthController extends Controller
 {
 
+    public function __construct(DrupalAPI $drupal, Registrar $registrar)
+    {
+        $this->drupal = $drupal;
```
DFurnes: I think `DrupalAPI` isn't being used here, so should be safe to remove.

app/Http/Controllers/AuthController.php
```diff
@@ -37,17 +46,8 @@ public function login(Request $request)
...
-            $token->user = $user->toArray();
-
-            // Return the session token with the user.
-            $user->session_token = $token->key;
-            $data = $user;
-            return $this->respond($user);
-        } else {
-            throw new UnauthorizedHttpException(null, 'Invalid ' . $login_type . ' or password.');
-        }
+        $user = $this->registrar->login($user, $input, $login_type);
```
DFurnes: Not necessary for merging this, but I think the `login` method here should accept a similar signature to Laravel's built in `Auth::attempt`. That way we can place the `$login_type` checks above in the registrar as well.

```
 $user = $this->registrar->login($input)
```

app/Services/Password.php
```diff
@@ -0,0 +1,116 @@
...
+* The maximum allowed log2 number of iterations for password stretching.
+*/
+define('DRUPAL_MAX_HASH_COUNT', 30);
+
+/**
+* The expected (and maximum) number of characters in a hashed password.
+*/
+define('DRUPAL_HASH_LENGTH', 55);
+
+class Password
```
DFurnes: I think I'd recommend keeping all of these functions inside the `DrupalPasswordChecker` class. Without the context of this pull request, it's not 100% clear that this is only used for Drupal password hashing.

app/Services/Registrar.php
```diff
@@ -0,0 +1,48 @@
...
+
+    public function login($user, $input, $login_type)
+    {
+        if (($user instanceof User) && Hash::check($input['password'], $user->password)) {
+            $token = $user->login();
+            $token->user = $user->toArray();
+
+            // Return the session token with the user.
+            $user->session_token = $token->key;
+            $data = $user;
```
DFurnes: Is this variable being used anywhere?

app/Services/Registrar.php
```diff
@@ -0,0 +1,48 @@
...
+                $user->password = $input['password'];
+                $user->drupal_password = null; // @TODO
+                $user->save();
+
+                $token = $user->login();
+                $token->user = $user->toArray();
+
+                // Return the session token with the user.
+                $user->session_token = $token->key;
+                $data = $user;
```
DFurnes: Same question – I think this might be unused.

app/Services/Registrar.php
```diff
@@ -0,0 +1,48 @@
...
+            $data = $user;
+            return $user;
+        } else if (($user instanceof User) && !($user->password)) {
+
+            // check to see if $input['password'] equals user's drupal password
+            if ($this->drupal_password_checker->user_check_password($input['password'], $user->drupal_password)) {
+
+                // if they're the same, make $input['password'] into a hash and save it to the user.
+                $user->password = $input['password'];
+                $user->drupal_password = null; // @TODO
```
DFurnes: ^ `@TODO`!

DFurnes:
Dropped some comments/suggestions! I think it would also make sense to pull the `$user->login()` method into the `Registrar` (or perhaps rename `Auth`) so we keep all our authentication logic in one place, but again definitely not necessary before merging this.

DFurnes:
@chloealee It'd also be great to get a quick unit test added to the test suite based on the testing instructions in your PR. I'd be happy to sit down and walk through this with you if you want!

app/Http/Controllers/AuthController.php
chloealee: you're so right - added this in at the beginning and forgot to delete!

app/Http/Controllers/AuthController.php
chloealee: much cleaner - updated this!

app/Services/Password.php
chloealee: true! I was just scared it was getting too long but totally makes sense! I've combined these files :)

app/Services/Registrar.php
chloealee: I don't think so - just had it in the original code so kept it there in case. Should I just remove?

app/Services/Registrar.php
chloealee: Removed!

app/Services/Registrar.php
chloealee: Should I delete this all together so the user doesn't even have a `drupal_password`?

chloealee:
awesome, thanks so much for all this feedback @DFurnes !  Re: moving `$user->login()` into the `Registrar`, do you mean moving the rest of the `login` function there and just replacing it with one line that reads `$user = $this->registrar->login(Request $request)` and just keeping the `logout` function in the `AuthController`? 

Also, I'd love to walk through building a unit test if you have time tomorrow! Would it be okay if we merged this and then built out the unit test separately? I know that @uy wants to test this functionality tomorrow on staging/production so we can have it for the launch of the mobile app.

app/Http/Controllers/AuthController.php
```diff
@@ -3,14 +3,22 @@
 use Northstar\Models\User;
 use Northstar\Models\Token;
 use Illuminate\Http\Request;
+use Northstar\Services\DrupalAPI;
```
DFurnes: One more! This `use` statement isn't doing anything now that you removed the `DrupalAPI` from the constructor.

app/Services/Registrar.php
DFurnes: I think that makes sense. You can use the `unset` method on the model to remove that field from the user document.

```php
$user->unset('drupal_password');
```

DFurnes:
Looking great! One more little nitpick, and responded to your question on unsetting the `drupal_password` field. Otherwise 100% good to merge!

DFurnes:
Made an issue for creating a unit test. Happy to walk through that process with you later today! :smiley:

app/Services/Registrar.php
chloealee: gotcha, updated!

app/Http/Controllers/AuthController.php
chloealee: oh yes! Deleted :)

chloealee:
great, thanks @DFurnes ! Made those two changes - to double double confirm, cool to merge?

Also, thanks so much for offering to help! Let me know when you're free today to chat about the unit test!

DFurnes:
Yeah, ship it!! :ship:

chloealee:
woooo!

PR #231, opened 2015-11-16
Tidying things up

DFurnes:
#### Changes
As part of [kicking off Data 2.0](https://github.com/DoSomething/quasar/issues/5) (oooOOOooh!), I'm doing a quick once-over to see what needs documenting or cleaning up in Northstar. I made a few changes, broken up into individual commits by change. Feel free to argue against any changes that don't make sense!

Highlights:
 - Renamed `DrupalAPI` to `Phoenix`. This seems a bit more straightforward (it's an API wrapped _specifically_ for Phoenix, not all Drupal APIs), and it was our only "API" class with the `API` suffix on the class name. (db0bba2)
 - Use Laravel 5's `Request` object rather than the `Input` facade. This reads a bit better, since rather than grabbing the input "out of thin air", it's just an object passed to the controller method. The API is exactly the same. (056fe9d)
 - Update documentation for the `AWS` class (had fallen out of date). (b8ab83a)
 - Reference Drupal's API documentation for `DrupalPasswordChecker` methods, so it's clear that those are not our code and can be treated as a "black box". (c6450b7)
 - Fixed indentation and removed some unnecessary class aliases. (4a97fdd, 08899c6, c59ba50)

#### How do I test this?
Unit tests should continue to pass, so that's good. :traffic_light:

---
For review: @jonuy 
/cc @chloealee @weerd

### DIFF
--- a/app/Http/Controllers/AvatarController.php
+++ b/app/Http/Controllers/AvatarController.php
@@ -12,14 +12,14 @@ class AvatarController extends Controller
         $this->aws = $aws;
     }
 
-   /**
-   * Store a new avatar for a user.
-   * POST /users/{id}/avatar
-   *
-   * @param Request $request
-   * @param $id - User ID
-   * @return Response
-   */
+    /**
+     * Store a new avatar for a user.
+     * POST /users/{id}/avatar
+     *
+     * @param Request $request
+     * @param $id - User ID
+     * @return \Illuminate\Http\Response
+     */
     public function store(Request $request, $id)
     {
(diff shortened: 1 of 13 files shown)


DFurnes:
__Note:__ I added StyleCI to this repo... I'll apply a patch to fix those failures in another PR.

jonuy:
:+1: Looks good. And once it's up on staging and eventually production, our Runscope tests should also let us know if we missed anything and it breaks.

DFurnes:
:+1: :boom:

Issue #228, opened 2015-11-09
API Key scopes


DFurnes:
Currently, all API keys are implicitly trusted. This is a reasonable assumption when all the keys are stored server-side (such as in Phoenix or Aurora). It's potentially dangerous now that we'll have keys that are stored in client-side code (like the mobile app) and traveling over public networks.

For example, any user that finds the application ID & key by either inspecting the application bundle or sniffing HTTP traffic would be able to spoof any request that the mobile app is able to do, which includes things like creating new API keys.

We should consider adding scopes to differentiate trusted clients and "untrusted" clients.

/cc @jonuy @angaither

jonuy:
+1 to scoping keys. But don't bundles get obfuscated in some sort of way? Though I guess anything could get decompiled and inspected. Same deal with https, I thought we'd be safe from general sniffing of packets since the data should be encrypted when being sent. But I guess there's probably something someone could do to act as a secure proxy and get the info they need.

DFurnes:
Just for documentation, @jonuy did some digging and it turns out Android application bundles are [pretty easy](http://www.decompileandroid.com) :sweat_smile: to decompile. Luckily it shouldn't be too hard to make a quick fix, and if we get around to using a standard OAuth implementation we should get this out of the box.

DFurnes:
And #239 is deployed to Thor now. Hopefully no issues should crop up and we can deploy this to production sometime this week. :boat:

Some other things we should lock down to `admin` scoped keys:

- [ ] Deleting users
- [ ] Seeing full profile details of other users

(Anything else that comes to mind for anyone?)

DFurnes:
This has been running on Thor's Northstar for a while now, with no reported errors. I think we should be safe to deploy to production, and then update our `aurora` and `phoenix` API keys to have `admin` scoped privileges.

(I'd like to open an endpoint for modifying key scopes as well, but obviously would want to limit that to only be available to `admin` keys so it's a bit of a chicken-and-egg problem. :hatching_chick:)

angaither:
:+1:

jonuy:
:+1: As long as it's only the keys endpoint getting locked down for now (since the mobile apps use all the other endpoints), this sounds good to deploy.

DFurnes:
> :+1: As long as it's only the keys endpoint getting locked down for now (since the mobile apps use all the other endpoints), this sounds good to deploy.

[...]

Issue #202, opened 2015-08-25
Use OAuth/JWT for authentication tokens


DFurnes:
Now that we're hooking more things into Northstar (yay!), we should probably start thinking about how we're handling authentication among multiple services.

At the moment, Northstar is able to authenticate users to itself (by providing a session token), but in order to perform authenticated requests to other services it must act as a proxy. For example, there are quite a few endpoints that use the user's Northstar session to do something on the Drupal site. As we consider adding more different services (such as a separate campaign activity API for signups & reportbacks), this gets more and more unwieldy.

From doing a bit of research, one solution seems to be granting a signed token that can be verified by any of our other services. [JSON Web Tokens](http://jwt.io) seem to be a standard way to provide a signed token (that indicates an authenticated session) which any of our internal backend services can verify. This avoids Northstar needing to act as a gatekeeper or each service needing to continually ping Northstar to check if a session is still active.

Example payload and signed token:

<img width="990" alt="screen shot 2015-08-25 at 11 39 23 am" src="https://cloud.githubusercontent.com/assets/583202/9471369/f3e4e484-4b1d-11e5-8910-14d790f59549.png">

Thoughts? @uy @angaither

DFurnes:
This would also help with role-based authorization, since we could store user roles (like "admin" or "staff") in Northstar where they belong, but still easily make use of them elsewhere (like Aurora or the campaign activity API service).

DFurnes:
For future reference: single sign-on for Zendesk, which is a requested feature for our Help Center, also [relies on JSON Web Tokens](https://support.zendesk.com/hc/en-us/articles/203663816-Setting-up-single-sign-on-with-JWT-JSON-Web-Token-).

jonuy:
This sounds pretty cool. I'm struggling to grasp what the whole scope of the work here is. I'm assuming this affects at the very least Northstar and any client apps that interface with it in an authorized way. Will it need Phoenix work too? Will we eventually be pulling out the code where Northstar forwards requests onto Phoeni?

DFurnes:
Ideally not. I think we could continue to use the same API for the mobile app with an OAuth password grant, and use a different grant and more privileged scope on products like Aurora or Phoenix that need it. (Still trying to totally understand the best way to handle this but I think that's the basic idea!)

[...]

Issue #108, opened 2015-05-11
Revisit how we handle authentication tokens.


DFurnes:
The way we're currently handling authentication tokens breaks a lot of built-in Laravel goodies. We should be handling authentication through the Auth facade, with a [custom provider](http://laravel.com/docs/5.1/authentication#adding-custom-authentication-drivers).

PR #235, opened 2015-11-19
Make DrupalPasswordHash static

DFurnes:
#### Changes
This just occurred to me while looking into [other](https://github.com/DoSomething/northstar/issues/228) [authentication](https://github.com/DoSomething/northstar/issues/202) [improvements](https://github.com/DoSomething/northstar/issues/108) & I think it seems like a nice little touchup. Renamed `DrupalPasswordChecker` to `DrupalPasswordHash` and made `check` a static method, since that makes its API read identically to Laravel's built-in `Hash::check` and removes the overhead of instantiating and dependency-injecting the class since it doesn't have any internal state.

### How should this be tested?
Automated tests should still pass. :traffic_light: 

---
For review: @jonuy @angaither

### DIFF
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -2,7 +2,6 @@
 
 namespace Northstar\Http\Controllers;
 
-use Northstar\Models\User;
 use Northstar\Models\Token;
 use Illuminate\Http\Request;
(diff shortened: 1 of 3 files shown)


jonuy:
Seems like a good idea to me

PR #236, opened 2015-11-19
updates inline comments 

### LINKED ISSUE #234: update inline comments
just noticed these comments were off for `GET` user requests

https://github.com/DoSomething/northstar/blob/67ea51f5fba7fb777c3117c8a3326c8686b36194/app/Http/Controllers/UserController.php#L27

@chloealee @DFurnes can you take a look at these comments and others when working in NS

chloealee:
#### What does this PR do?
Fixes comment in UserController.php construct function to have GET all capitalized. Updates comment in AvatarController.php store function to remain consistent with other endpoints. 

Fixes #234

### DIFF
--- a/app/Http/Controllers/UserController.php
+++ b/app/Http/Controllers/UserController.php
@@ -23,8 +23,7 @@ class UserController extends Controller
 
     /**
      * Display a listing of the resource.
-     * GET /users
-     * Get /users?attr1=value1&attr2=value2&...
+     * GET /users/:term/:identifier
      *
      * @param Request $request
      * @return \Illuminate\Http\Response
--- a/app/Http/Controllers/UserController.php
+++ b/app/Http/Controllers/UserController.php
@@ -23,7 +23,8 @@ class UserController extends Controller
 
     /**
      * Display a listing of the resource.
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/UserController.php
```diff
@@ -23,8 +23,7 @@ public function __construct(Phoenix $phoenix)
 
     /**
      * Display a listing of the resource.
-     * GET /users
-     * Get /users?attr1=value1&attr2=value2&...
+     * GET /users/:term/:identifier
```
DFurnes: Wait, this is the index method... I think `/users/:term/:identifier` is for the `show` method below.

DFurnes: I'm pretty sure the two examples here are correct, although `Get` should be all-caps.

chloealee: ohhhh I see now. And the second just gives an example of how to use the queries? Changed!

chloealee:
@DFurnes / @angaither whenever you get a sec to take a look!

DFurnes:
Looks good to me! :+1:

PR #239, opened 2015-12-07
Scopes

DFurnes:
#### Changes
Adds `scope` property to API keys, to separate "user" versus "administrator" privileges. This fixes the problem where a malicious user could use the key publicly bundled with the mobile app to view other keys or create their own keys. In the future, we could easily add more scopes to lock down additional privileges we don't want every app to have. References #228.

Sorry, this ended up being a hefty PR. :grimacing: It's much more readable if you read it commit-by-commit, so I'd recommend that approach. All tests, as usual, should continue to be green. :traffic_light: 

#### Any required steps to deploy this?
We will need to update Phoenix & Aurora's API keys to have the `admin` scope. For the time being, the only thing that's "locked down" is the API key endpoints which only Aurora accesses, but I'm sure more and more will be restricted as we lock things down. :closed_lock_with_key: 

---
For review: @angaither 
/cc @jonuy @chloealee


angaither:
kewl, let's deploy this to thor NS and test :tada: :+1:

PR #241, opened 2015-12-07
Validate email is <=  60 characters on registration.

### LINKED ISSUE #240: Drupal registration fails on > 60 char email address
Found via testing on the mobile app. After registering users with extremely long email addresses, the user accounts were successfully created (we were able to log into the mobile app and successfully authenticate via northstar.) 

However, after attempting to sign up for a campaign, we receive the following error, perhaps suggesting that the user wasn't successfully created on Phoenix: 
simulator screen shot dec 7 2015 3 22 30 pm

Discovered behavior on both staging and production while testing with email usernames that are 61 and 65 chars. A 56 char username works. 

**Accounts which fail:**
tongxiangtongxiangtongxiangtongxiangtongxiang@dosomething.org 
tongxiangtongxiangtongxiangtongxiang1234567890tongxiang@gmail.com

**Accounts which succeed:**
 tongxiangtongxiangtongxiangtongxiang1234567890@gmail.com 

Password: < the typical summer ball pasttime of DoSomething.org >

DFurnes:
#### Changes
Fixes #240. 

Drupal was failing to register users with longer email addresses, leading to unexpected behavior with campaigns/signups. This seems like a suitable quick fix, and we can get a better longer term fix in when we update Drupal to authenticate against Northstar sometime in the (hopefully near) future.

#### How should I test this?
Registering a user with a very long email address should give an error response, like so:
```
{
  "error": {
    "code": 401,
    "message": "The email may not be greater than 60 characters."
  }
}
```

Unit tests, as always, should continue to pass. :traffic_light: 

---
For review: @angaither 
/cc @tongxiang

### DIFF
--- a/app/Http/Controllers/UserController.php
+++ b/app/Http/Controllers/UserController.php
@@ -95,7 +95,7 @@ class UserController extends Controller
             // is for unique email or phone numbers, and that should return a user
             // from the query above.
             $this->validate($request, [
-                'email' => 'email|unique:users|required_without:mobile',
+                'email' => 'email|max:60|unique:users|required_without:mobile',
                 'mobile' => 'unique:users|required_without:email',
             ]);
         }


DFurnes:
The nitpick from StyleCI was fixed in #239, so not going to address that here.

angaither:
:+1:

PR #244, opened 2015-12-08
Scope issues

DFurnes:
#### Changes
We were getting some `500` errors when testing #239 on Thor, because we tried to do stuff with the array index before checking if it was really a thing or not.

This PR also restricts the ability to delete users to keys with the `admin` scope, and removes the `auth` middleware from the user controller (since it does not require or do anything with an active user session).

:bug: :boom:

---
For review: @chloealee @angaither

### DIFF
--- a/app/Http/Controllers/UserController.php
+++ b/app/Http/Controllers/UserController.php
@@ -20,7 +20,8 @@ class UserController extends Controller
     {
         $this->phoenix = $phoenix;
 
-        $this->middleware('key:user');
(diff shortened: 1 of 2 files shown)


DFurnes:
For further context on 003bbdf, this wasn't failing on the test suite since we seed `scope` values when creating our API keys, but the existing keys on production don't have anything set for that attribute.

angaither:
:+1:

PR #245, opened 2015-12-08
Don't trigger Runscope tests on deploy.

DFurnes:
#### Changes
Let's not trigger Runscope tests on deploys (at least for the moment), since they add a lot of noise to chat & are not testing the same environment that the deploy is sent to. :grimacing: 

---
For review: @angaither @jonuy


jonuy:
:+1:

PR #247, opened 2015-12-17
writes test for drupal password checker

### LINKED ISSUE #230: Build unit test for Drupal password hasher
The instructions under "How should this be manually tested?" in #229 are a perfect blueprint for building a unit test! Let's get one set up so we can ensure this functionality continues to work. :traffic_light:

chloealee:
#### What does this PR do?
Builds a unit test for drupal password checker (#229). The drupal password checker allows a user to sign into the mobile app if they have a `drupal_password` but no `password`. It checks if the inputted `password` is the same as their `drupal_password` and if so, signs them in, deletes the `drupal_password`, hashes the `password` and saves it to their user object. 

tests/AuthTest.php lines 168-195

In line 186, I have this assertion: `$this->assertArrayHasKey('password', $user['attributes']);` 
I tried to do `$user->toArray();` but this didn't output the user with the new `password` attribute. When I just did `dd($user)` however, the object did contain the `password` attribute so went the above route instead. 

Does it matter if I use `$user['attributes']` or `$user['original']`? They're both the same array.   

SSH into northstar vagrant environment. Run `vendor/bin/phpunit`

Fixes #230

### DIFF
--- a/app/Services/DrupalPasswordHash.php
+++ b/app/Services/DrupalPasswordHash.php
@@ -49,6 +49,11 @@ class DrupalPasswordHash
             return ($hash && $stored_hash == $hash);
     }
 
+    private static function hash($password)
+    {
+        return self::_password_crypt('sha512', $password, $stored_hash);
+    }
+
     // @see: https://api.drupal.org/api/drupal/includes%21password.inc/function/_password_crypt/7
     private static function _password_crypt($algo, $password, $setting)
     {
--- a/app/Services/Registrar.php
+++ b/app/Services/Registrar.php
@@ -46,8 +46,8 @@ class Registrar
 
                 return $user;
             }
-        } else {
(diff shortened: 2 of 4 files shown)


DFurnes:
> In line 186, I have this assertion: $this->assertArrayHasKey('password', $user['attributes']); 
I tried to do $user->toArray(); but this didn't output the user with the new password attribute. When I just did dd($user) however, the object did contain the password attribute so went the above route instead.

I'm not sure tbh. The `$attributes` property is the Model's attributes before any mutators are run and should be what you want in this case, but I'm not sure what the `$original` property is for... there's like zero documentation on it.

DFurnes:
Looks good to me! :shipit:

chloealee:
awesome, thanks @DFurnes !

PR #248, opened 2015-12-17
Signup group should require a user scoped key.

DFurnes:
#### Changes
I incorrectly scoped signup groups to require an `admin` scoped key in #239. This fixes that. :hammer: 

Signup groups should be accessible from non-admin API keys.

#### Questions
...do we use signup groups anymore?

---
For review: @angaither @jonuy

### DIFF
--- a/app/Http/Controllers/SignupGroupController.php
+++ b/app/Http/Controllers/SignupGroupController.php
@@ -19,7 +19,7 @@ class SignupGroupController extends Controller
     {
         $this->phoenix = $phoenix;
 
-        $this->middleware('key:admin');
+        $this->middleware('key:user');
     }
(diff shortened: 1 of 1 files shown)


jonuy:
:+1: to :ship: if you'd like.

Signup groups ended up getting cut from the mobile app. And there aren't any immediate future plans to bring them back. I'm guessing it's something we can just cut if we wanna do some clean up.

To double check, @marahml we don't have any plans to bring back invites and groups in the near future for the mobile app, do we?

DFurnes:
@jonuy Good to know. I'll make a "refactor" issue to remove signup groups from the API once we confirm they're not on the roadmap.

PR #249, opened 2015-12-18
GET /users/{term}/{id} endpoint test

chloealee:
#### What does this PR do?
Writes a test to ensure that `GET /users/{term}/{id}` endpoint works correctly with the following terms: `_id`, `mobile`, `email`, and `drupal_id`

tests/AuthTest.php lines 195-236

`vagrant ssh` into your northstar environment and run `vendor/bin/phpunit` to run all phpunit tests

### DIFF
--- a/tests/AuthTest.php
+++ b/tests/AuthTest.php
@@ -196,12 +196,13 @@ class AuthTest extends TestCase
     /**
      * Tests GET /users/{term}/{id} endpoint works correctly
      */
-    public function testRetrieveUser() {
+    public function testRetrieveUser()
+    {
         // User info
         $user = User::find('5430e850dt8hbc541c37tt3d');
 
         // GET /users/_id/<user_id>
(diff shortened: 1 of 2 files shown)


tests/AuthTest.php
```diff
@@ -192,4 +192,46 @@ public function testDrupalPasswordChecker()
         $tokenCount = Token::where('key', '=', $data['data']['session_token'])->count();
         $this->assertEquals($tokenCount, 1);
     }
+
+    /**
+     * Tests GET /users/{term}/{id} endpoint works correctly
+     */
+    public function testRetrieveUser() {
+        // User info
+        $user = User::find('5430e850dt8hbc541c37tt3d');
```
angaither: I don't think this `_id` will work across multiple environments

chloealee: @angaither testing it based on the first user in the UserTableSeeder.php file! Does that make sense to do/will work for the testing purposes?

chloealee:
@DFurnes mind taking a look when you get the chance?

DFurnes:
Looks good to me! :boat:

PR #252, opened 2016-01-04
Avatar test

chloealee:
#### What does this PR do?
Writes a test to ensure `POST /users/:user_id/avatar` endpoint works correctly, accepting both an avatar file or Base64 string in the data request.

tests/UserTest.php lines 181-232

vagrant ssh into your northstar environment and run vendor/bin/phpunit to run all phpunit tests

### DIFF
--- a/tests/UserTest.php
+++ b/tests/UserTest.php
@@ -231,6 +231,32 @@ class UserTest extends TestCase
     }
 
     /**
+     * Test for creating a user's profile image with a file
+     * POST /users/:user_id/avatar
+     *
+     * @return void
+     */
+    public function testCreateUserAvatarWithFile()
+    {
+        $payload = [
+            'photo' => 'profile_pic.jpeg',
+        ];
+
+        $response = $this->call('POST', 'v1/users/5480c950bffebc651c8b456f/avatar', [], [], [], $this->server, json_encode($payload));
+        $content = $response->getContent();
+        $data = json_decode($content, true);
+
+        // The response should return a 200 status code
+        $this->assertEquals(200, $response->getStatusCode());
+
+        // Response should be valid JSON
+        $this->assertJson($content);
+
+        // Response should return avatar's url
+        $this->assertNotEmpty($data['data']['photo']);
+    }
+
+    /**
      * Test for deleting an existing user
      * DELETE /users
      *
--- a/tests/UserTest.php
+++ b/tests/UserTest.php
@@ -24,16 +24,14 @@ class UserTest extends TestCase
             'HTTP_Session' => 'S0FyZmlRNmVpMzVsSzJMNUFreEFWa3g0RHBMWlJRd0tiQmhSRUNxWXh6cz0=',
         ];
 
-<<<<<<< HEAD
         $this->serverRetrieveUser = [
             'HTTP_Accept' => 'application/json',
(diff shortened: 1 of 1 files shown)


DFurnes:
@chloealee Looks like these tests are failing on Wercker! :grimacing: 

screen shot 2016-01-05 at 10 40 07 am

I _think_ you might just need to mock the AWS service so it isn't trying to actually upload a photo when it runs tests, but haven't looked super closely.

chloealee:
ah, thanks @DFurnes ! Is this the right way to go/is it okay to add this new file to mock for the test? http://www.podpea.co.uk/blog/mock-third-party-apis-with-phpunit/

DFurnes:
You can create a mock specifically for this unit test using [Mockery](https://github.com/padraic/mockery), which allows you to create the mock and specify it's "expected" response on demand. Take a look at [CampaignTest](https://github.com/DoSomething/northstar/blob/188a5bbd788569cb79d944ac4c9a14d990459355/tests/CampaignTest.php) for a working example using the Drupal API. Here's the basic idea:

```php
public function setUp()
{
  // ...
  // Mock AWS API class
  $this->awsMock = $this->mock('Northstar\Services\AWS');
}

// ...

public function testCreateUserAvatarWithFile()
{
  // Mock successful response from AWS API
  $this->awsMock->shouldReceive('storeImage')->once()->andReturn('http://bucket.s3.amazonaws.com/example.jpg');
  // ...
}
```

chloealee:
@DFurnes so awesome, thanks for this! Looks like it's passing now!

tests/UserTest.php
```diff
@@ -179,6 +182,64 @@ public function testUpdateUser()
...
+     * @return void
+     */
+    public function testCreateUserAvatarWithBase64()
+    {
+        $payload = [
+            'photo' => '123456789',
+        ];
+
+        // Mock successful response from AWS API
+        $this->awsMock->shouldReceive('storeImage')->once()->andReturn('http://bucket.s3.amazonaws.com/123456789');
```
DFurnes: This should also return a file URL, like the mock above.

DFurnes: From looking a bit more closely at the source for [AvatarController](https://github.com/DoSomething/northstar/blob/188a5bbd788569cb79d944ac4c9a14d990459355/app/Http/Controllers/AvatarController.php#L41) it looks like this should always end up being named by the user ID. So, for example `5480c950bffebc651c8b456f.jpg` in this case.

tests/UserTest.php
```diff
@@ -179,6 +182,64 @@ public function testUpdateUser()
...
+     * @return void
+     */
+    public function testCreateUserAvatarWithFile()
+    {
+        $payload = [
+            'photo' => 'example.jpeg',
+        ];
+
+        // Mock successful response from AWS API
+        $this->awsMock->shouldReceive('storeImage')->once()->andReturn('http://bucket.s3.amazonaws.com/example.jpg');
```
DFurnes: Same comment on the filename here.

DFurnes:
Two tiny issues, and then it looks like you need to resolve some merge conflicts. :confounded: Otherwise looks good to me!

chloealee:
thanks, @DFurnes ! Resolved those and fixed merge conflicts. Will merge now :)

PR #253, opened 2016-01-05
Initial proof-of-concept of using Fractal to format API responses.

DFurnes:
# Changes
This was a little proof-of-concept I had started working on before the holidays to use [Fractal](http://fractal.thephpleague.com) as a transformer for Northstar's API responses. This allows us to set a standardized schema for responses that doesn't necessarily need to match our database schema.

Currently, I've only updated the `User::show` endpoint as an example but it shouldn't be much work to use this for all endpoints if we decide to move forward with this approach. Two primary benefits:

1. This allows us to declaratively specify what a response should look like (similar to how you would with templates for front-end views, and the great work @weerd is doing in the Phoenix app). Another great comp is [JBuilder](https://github.com/rails/jbuilder) from Rails.
2. The transformer is (I think) a smart place to add authorization logic to limit what fields are accessible based on a  user's permissions... so for example, we could limit fields based on API key scope, or to _only_ be accessible by the user who owns that profile.

If we decide to go with this approach, this would also remove the need for the [user](https://github.com/DoSomething/northstar/blob/188a5bbd788569cb79d944ac4c9a14d990459355/app/Http/Middleware/UserResponseMiddleware.php) and [campaigns](https://github.com/DoSomething/northstar/blob/188a5bbd788569cb79d944ac4c9a14d990459355/app/Http/Middleware/CampaignResponseMiddleware.php) middleware that add `null` fields to responses.

/cc @angaither @weerd @jonuy


weerd:
oooooo :)

DFurnes:
This could also make the inevitable versioning of the API (e.g. separate locked-in-stone `v1` and `v2` response formatting) easier, since we can easily swap out response transformers via URL.

DFurnes:
I removed fields that shouldn't be accessible except by the currently authenticated user from the response in c52adf7. I'd like to add these back in _only_ for admin-scoped keys or sessions that are authenticated as that particular user, but will save that for another PR.

I don't think any of those fields are actively being used but wanted to give a heads up for API consumer folks just in case. @jonuy @aaronschachter

angaither:
> I don't think any of those fields are actively being used but wanted to give a heads up for API consumer folks just in case.

[this list of fields](https://github.com/DoSomething/northstar/pull/253/files#diff-429e3c9dd0c7f0b71a94986411034c76R36)? they are being used, I think I'm confused.

DFurnes:
Hmm, okay. Which of those fields are being used by the mobile app? I'll re-add any that are actively being used, but presumably none of these are things we'd want on a "public profile".

angaither:
OK. I was under the impression the fields were removed, and couldn't be set. But this is just removing them from user gets. 
Thanks for clarifying via Slack @DFurnes 


I like this. But we will need to update mobile app/phoenix accordingly

DFurnes:
@aaronschachter @jonuy So by default, Fractal's built-in `DataArraySerializer` uses an array `data` object for collections of items (just like our current responses), but gives single item responses as a plain object in the `data` field (rather than an object inside a one-item array). So `$response['data']['first_name']` instead of `$response['data'][0]['first_name']`.

Since the "show" endpoint is always going to return a single resource, this seems like it makes more sense but I wanted to get your thoughts on how this would affect the app. (We could also make our own custom serializer or just always wrap items in a collection before serializing, so this isn't a _requirement_ for using Fractal btw.)

aaronschachter:
Sounds ok to me - but the only change that I can tell will be the object instead of array as you mention. I believe this should keep it consistent with how Phoenix returns single item responses, right?  That's always a good thing.

app/Http/Transformers/UserTransformer.php
```diff
@@ -0,0 +1,50 @@
...
+            '_id' => $user->_id, // @DEPRECATED: Will be removed.
+
+            'first_name' => $user->first_name,
+            'photo' => $user->photo,
+            'interests' => $user->interests,
+
+            // @TODO: Restrict these, based on API key scope or authenticated user.
+            'email' => $user->email,
+            'mobile' => $user->mobile,
+            // 'last_name' => $user->last_name,
```
jonuy: @DFurnes On Android, at least I use `last_name` so that we can display it as a last initial in the app UI. I also see that we save `birthdate` but I'm not sure if it's used for anything. And `country` _I think_ could also be used. From what I remember, Phoenix gets this field from the user's data in Northstar when building its /reportback-items response.

jonuy:
@DFurnes On your note about `DataArraySerializer`, does this mean that on the client-side we'll have to check if `data` is an object or an array before processing/casting it?

For example, a **GET /users/email/something@example.com** call, most of the time there will only be one user doc. But we could also run into scenarios where there are multiple documents for this one user - mainly if the same user registered and updated information on different platforms.

DFurnes:
~~Now that we've run the de-duplication script, email and mobile should both be unique keys I think. That's definitely a problem if we're still creating duplicates.~~ Oh, I see. We're not validating against creating duplicates in our [`update` method](https://github.com/DoSomething/northstar/blob/ed7374548e41e0cf7f08c0d59a45b4812959ffe3/app/Http/Controllers/UserController.php#L176-L199) are we? :confounded:

app/Http/Transformers/UserTransformer.php
DFurnes: Hmm, `country` seems like a reasonable public profile field. I'm a bit more wary about `last_name` and `birthdate`. For last name, would it make sense to add a computed `last_initial` public field?

aaronschachter:
@DFurnes Can you alert the `#mobile-app` Slack room when this is deployed to production? It'd be good to get a Kiwi 1.0.1 release out into the App Store once it's live (and presumably breaking the live app)

DFurnes:
@aaronschachter Will do!

jonuy:
Hm thoughts about versioning the API so that potentially breaking changes like this would be part of v2? Then the apps which are pointing to v1 won't break and we can gradually deprecate and delete v1 once the apps are updated.

As an aside, the number of users this affects right now is pretty much 0, but it might be a good candidate to practice the versioning process with.

angaither:
+1 @jonuy

DFurnes:
> As an aside, the number of users this affects right now is pretty much 0, but it might be a good candidate to practice the versioning process with.

That's the way I was looking at it, but definitely agreed on versioning the API in the future, and transformers should make that much easier when we do. My gut is that we'd be overengineering by versioning the API when there still aren't any "stable" releases relying on it, but I'm open to giving it a go if people feel strongly otherwise.

app/Http/Transformers/UserTransformer.php
DFurnes: @jonuy I added `country` and a computed `last_initial` field to the public profile response. Let me know if that works for you!

jonuy: @DFurnes yes! ty ty :bow:

DFurnes:
Alright, a few more updates! Paginated API responses are now run through the transformers as well. In addition to the newly "transformed" item results, pagination is now listed under the `meta` object (shown below):

pagination response

I've also tweaked how filters are queried. Instead of using the raw query parameter (which could have conflicts against other parameters like pagination & caused us to do a bit of a song-and-dance to reconcile), all filters are now namespaced under a `filter` parameter. This follows the [recommendation](http://jsonapi.org/recommendations/#filtering) made in JSON-API spec. For example, to get the following Drupal IDs and mobile number:

```
GET v1/users?filter[drupal_id]=100001,100002,100003&filter[mobile]=5555550100
```

/cc @jonuy @aaronschachter

jonuy:
All makes sense and all will likely break the mobile apps in one way or another (which I think we're ok with). I'd want @aaronschachter to chime in too, but this could get merged into the dev branch and deployed at least to staging for us to be able dev against. But then hold off on a deploy to production until we're closer to being ready to update the apps?

... or do we feel ok just pushing the API changes all the way through to production and also breaking the apps until we're able to push out a fix?

/cc @marahml

DFurnes:
I think pushing to staging or Thor (whichever you guys QA against) first is definitely a good idea.

jonuy:
Yea, let's def do that. The question I had was more so if we also deploy to production soon after - or just remember not to do it until mobile apps are ready for the change, which could take more coordination than it's worth. Maybe a conversation to have outside the GitHubs.

DFurnes:
Alright, the last piece is done! Campaigns are now :sparkles: transformed :sparkles: as well. I'll take a pass over the documentation this afternoon and make sure everything is updated.

@angaither @jonuy Mind giving this another look before I merge?

angaither:
I feel great.

PR #255, opened 2016-01-07
Scope gotchas

DFurnes:
#### Changes
Fixes some little _gotchas_ that I ran into when trying to add scopes to existing API keys (using the Artisan REPL). Check out each commit for details, but the basic gist is that the [laravel-mongodb](https://github.com/jenssegers/laravel-mongodb) seems to have some unexpected behavior with casting and arrays that removing this code seems to clean up. :innocent: 

For review: @angaither

### DIFF
--- a/app/Models/ApiKey.php
+++ b/app/Models/ApiKey.php
@@ -23,15 +23,6 @@ class ApiKey extends Model
     ];
 
     /**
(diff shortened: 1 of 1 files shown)


angaither:
ok :+1:

PR #257, opened 2016-01-21
Add search parameter for loosely finding matching results.

DFurnes:
#### Changes
Adds a `search` parameter for loosely finding matching results. This is used to power [Aurora's search functionality](https://github.com/DoSomething/aurora/pull/122). For example, to find all users with emails containing `example.com`:

```
GET /v1/users/?search[email]=example.com
```

The Laravel MongoDB `like` operator looks like it is a shortcut [for performing a regex query on the MongoDB database](https://github.com/jenssegers/laravel-mongodb/blob/4526972eb65ea47399df18e42f843b0b733c28d3/src/Jenssegers/Mongodb/Query/Builder.php#L929-L940). I'm not sure exactly how performance on this kind of query is, but it looks like [it still uses any indexes on the relevant fields](https://docs.mongodb.org/manual/reference/operator/query/regex/#index-use) so hopefully not too bad.

#### Open Questions
Should this parameter be limited to `admin` scoped keys?

I added a unit test for the new search parameter in `UserTest.php`. (a250a4b)

---
For review: @angaither

### DIFF
--- a/app/Http/Controllers/Controller.php
+++ b/app/Http/Controllers/Controller.php
@@ -185,7 +185,7 @@ abstract class Controller extends BaseController
         }
 
         // Searches may only be performed on indexed fields.
-        $filters = array_intersect_key($searches, array_flip($indexes));
+        $searches = array_intersect_key($searches, array_flip($indexes));
 
         // For the first `where` query, we want to limit results... from then on,
         // we want to append (e.g. `SELECT * WHERE _ OR WHERE _ OR WHERE _`)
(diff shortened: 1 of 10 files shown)


angaither:
:+1: please update the [spec](https://github.com/DoSomething/northstar/wiki/Spec)

also, yes? probably should be scoped

angaither:
oh, there's also a package of postman routes that I think are very out of date...
https://github.com/DoSomething/northstar/wiki/Postman-dev-routes-collection

DFurnes:
Okay, searches are now limited to `admin` scoped keys! :lock: 

I'll update the spec this afternoon.

PR #258, opened 2016-01-25
Fix inability to search for an ID.

DFurnes:
#### Changes
The MongoDB Object ID doesn’t seem to be queryable by regex. Luckily, I don’t think you’d ever really *want* to, so let’s just force that field to only search by exact match.

References DoSomething/aurora#125.

---
For review: @angaither


angaither:
:+1:

PR #259, opened 2016-01-25
Let's just not use regex at all. :ghost:

DFurnes:
#### Changes
This isn’t a huge surprise, but regex queries have a pretty huge performance penalty. I think the majority of cases where the search function will be used are for looking up an exact user, so lets keep it simple.

---

For review: @angaither


angaither:
:+1:

PR #260, opened 2016-01-25
Fix search tests.

DFurnes:
#### Changes
Updates tests for the changes in #259, since searches now only check for an exact match.

I was too hasty! :runner: 

---
For review: @angaither


angaither:
pew pew pew pew :ship:

PR #261, opened 2016-01-26
Response permissions

DFurnes:
#### Changes
This pull request allows admin-scoped keys to view full user profile details. This allows tools like Aurora to access a user's full profile (including last name, address, etc.), while still keeping it under lock and key :closed_lock_with_key: for normal "public" API responses.

#### What about letting users see their own full profiles?
Yes! I've started on that work (which involves #108) but am going to put it up as its own branch later, so as not to make another monster PR. :japanese_ogre: 

Automated tests added in f71cbca should pass. :traffic_light: 

---
For review: @angaither

### DIFF
--- a/app/Http/Transformers/UserTransformer.php
+++ b/app/Http/Transformers/UserTransformer.php
@@ -2,6 +2,7 @@
 
 namespace Northstar\Http\Transformers;
 
+use Northstar\Models\ApiKey;
 use Northstar\Models\User;
 use League\Fractal\TransformerAbstract;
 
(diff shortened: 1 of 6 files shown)


app/Http/Transformers/UserTransformer.php
```diff
@@ -13,39 +14,51 @@ class UserTransformer extends TransformerAbstract
...
+        if (ApiKey::current()->hasScope('admin')) {
+            $response['last_name'] = $user->last_name;
+        }
+
+        $response['last_initial'] = $user->last_initial;
+
+        $response['photo'] = $user->photo;
+        $response['interests'] = $user->interests;
+
+        if (ApiKey::current()->hasScope('admin')) {
```
angaither: you test for this scope, do we need to or can it be all in the same section?

app/Http/Transformers/UserTransformer.php
```diff
@@ -13,39 +14,51 @@ class UserTransformer extends TransformerAbstract
...
+
+        // References to app-specific user IDs.
+        $response['drupal_id'] = $user->drupal_id;
+        $response['cgg_id'] = $user->cgg_id;
+        $response['agg_id'] = $user->agg_id;
+
+        $response['parse_installation_ids'] = $user->parse_installation_ids;
+
+        $response['updated_at'] = $user->updated_at->toISO8601String();
+        $response['created_at'] = $user->created_at->toISO8601String();
```
angaither: accessors/mutators?

tests/UserTest.php
```diff
@@ -43,12 +43,38 @@ public function setUp()
...
-     * Test for retrieving a user
-     * GET /users
+     * Test for retrieving a user with a non-admin key.
+     * GET /users/:term/:id
+     *
+     * @return void
+     */
+    public function testGetPublicDataFromUser()
+    {
+        $response = $this->call('GET', 'v1/users/email/test@dosomething.org', [], [], [], $this->userScope);
```
angaither: I know this is just in the tests and it doesn't _really_ matter but it would be cool to have app version be a variable so you don't have to update it on all the tests. just a thought

app/Http/Transformers/UserTransformer.php
DFurnes: I prefer breaking it out into two sections so the full response is ordered logically (just for debugging/reading responses in Paw). Not something I feel super strongly about though, so could be combined.

app/Http/Transformers/UserTransformer.php
DFurnes: I think this makes sense to do in the transformer so that we can still access the full date object from the model if we want to format it differently in different contexts.

tests/UserTest.php
DFurnes: Yeah, agreed. I think there's a lot of cleanup we could do in the test suite... setting a session token or API key is especially cumbersome.

angaither:
:+1:

PR #262, opened 2016-01-26
Key scope should always return an array.

DFurnes:
#### Changes

When I removed the "default" `user` value for scope, we started returning `null` for some key scopes that had been created before keys were given an empty array default value. This broke a couple of API keys, and the API key admin interface in Aurora.

---
For review: @angaither

### DIFF
--- a/app/Models/ApiKey.php
+++ b/app/Models/ApiKey.php
@@ -60,11 +60,25 @@ class ApiKey extends Model
     }
 
     /**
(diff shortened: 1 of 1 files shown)


angaither:
:+1:

DFurnes:
Amended the commit to fix some StyleCI issues. :pencil2: Once tests pass, will merge!

PR #263, opened 2016-01-26
Laravel 5.2

DFurnes:
#### Changes
Updates Northstar to Laravel 5.2! Not only do we get all the latest _fanciest_ new goodies, Laravel 5.2 also allows us to easily [override the Guard class](https://mattstauffer.co/blog/multiple-authentication-guard-drivers-including-api-in-laravel-5-2) which [would have involved some hackery](https://laracasts.com/discuss/channels/general-discussion/l5-extending-guard-object-in-authentication-framework?page=1) before. This will help us clean up some of the custom authentication work we've been doing in a future PR.

---
For review: @angaither

### DIFF
--- a/app/Providers/AuthServiceProvider.php
+++ b/app/Providers/AuthServiceProvider.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace Northstar\Providers;
+
+use Illuminate\Contracts\Auth\Access\Gate as GateContract;
+use Illuminate\Foundation\Support\Providers\AuthServiceProvider as ServiceProvider;
+
+class AuthServiceProvider extends ServiceProvider
+{
+    /**
+     * The policy mappings for the application.
+     *
+     * @var array
+     */
+    protected $policies = [
+        // ...
+    ];
+
+    /**
+     * Register any application authentication / authorization services.
+     *
+     * @param  \Illuminate\Contracts\Auth\Access\Gate  $gate
+     * @return void
+     */
+    public function boot(GateContract $gate)
+    {
+        parent::registerPolicies($gate);
+    }
+}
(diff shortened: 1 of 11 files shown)


app/Providers/AuthServiceProvider.php
```diff
@@ -0,0 +1,29 @@
...
+use Illuminate\Foundation\Support\Providers\AuthServiceProvider as ServiceProvider;
+
+class AuthServiceProvider extends ServiceProvider
+{
+    /**
+     * The policy mappings for the application.
+     *
+     * @var array
+     */
+    protected $policies = [
```
angaither: FUN what's a policy!?

DFurnes: [New authorization stuff](https://laravel.com/docs/5.2/authorization) added in Laravel 5.1. :lock:

config/app.php
```diff
@@ -78,7 +91,7 @@
     |
     */
 
-    'key' => env('APP_KEY', 'SomeRandomString'),
+    'key' => env('APP_KEY'),
```
angaither: this is something I remember from 100 years ago, but I think when setting up without a default if gives errors...but maybe that was an older v of laravel?

DFurnes: I think that's still true. I imagine the reason it's removed is so that it'd fail on production if you hadn't set the environment variable (the `.env.example` has `SomeRandomString` in it).

angaither: so do you have to set that up on a local instance?

angaither: in order to get the app to work i mean

DFurnes: Yep.

angaither:
looks kewl

DFurnes:
:sunglasses:

PR #265, opened 2016-01-27
Use custom Guard for authentication.

### LINKED ISSUE #108: Revisit how we handle authentication tokens.

DFurnes:
##### Custom Guard & Token Authentication
By registering a [custom guard](https://laravel.com/docs/5.2/authentication#adding-custom-guards) that understands how our `tokens` collection is structured, we can use Laravel’s built-in Auth facade (and Gate, etc.) to manage user sessions and permissions. (56cb07d)

There are no "user-facing" behavior changes to token authentication in this pull request, although it _does_ introduce two alternative ways to provide a User's session token. I'd like to remove support for the `Session` header in the future (since the API is stateless and so calling it a session is a bit of a misnomer).

```sh
# quick n' easy (as a query string)
GET /v1/users?token=xxxxxxx

# preferred (Authorization header)
GET /v1/users
Authorization: Bearer xxxxxxx

# legacy (as Session header)
GET /v1/users
Session: xxxxxxx
```

Closes #108.

##### Authorization
Now that we're using Laravel's built-in authentication code, we can also take advantage of Laravel 5.2's excellent [new authorization tools](https://laravel.com/docs/5.2/authorization)! We use the `Gate` facade (and an associated policy class) to determine whether to show full profile details for any given user. (bc0afa9)

This doesn't break any existing behavior, but will provide additional fields when querying the current user's profile data from the mobile app (or other non-privileged clients).

### Where should the reviewer start?
I'd recommend reviewing commit-by-commit.

### How should this be tested?
Automated tests should all pass! :traffic_light: 

---
For review: @angaither @weerd
/cc @jonuy @aaronschachter

### DIFF
--- a/app/Auth/DrupalPasswordHash.php
+++ b/app/Auth/DrupalPasswordHash.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace Northstar\Services;
+namespace Northstar\Auth;
 
 define('DRUPAL_MIN_HASH_COUNT', 7);
 define('DRUPAL_MAX_HASH_COUNT', 30);
--- a/app/Auth/NorthstarTokenGuard.php
+++ b/app/Auth/NorthstarTokenGuard.php
@@ -1,0 +1,124 @@
+<?php
+
+namespace Northstar\Auth;
+
+use Illuminate\Auth\TokenGuard;
+use Illuminate\Http\Request;
+use Illuminate\Contracts\Auth\Guard;
+use Illuminate\Contracts\Auth\UserProvider;
+use Mockery\CountValidator\Exception;
+use Northstar\Models\Token;
+use Northstar\Models\User;
+
+class NorthstarTokenGuard extends TokenGuard implements Guard
+{
+    /**
+     * Create a new authentication guard.
+     *
+     * @param  \Illuminate\Contracts\Auth\UserProvider  $provider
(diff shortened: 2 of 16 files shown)


DFurnes:
Had to make a change in 2f09df8 to check for the API key scope _separately_ than user's privileges (rather than treating it as a "superuser" override), since the `before` callback on the policy class is skipped if there isn't a currently authenticated user.

DFurnes:
Okay, _one_ more thing – I noticed that validation errors were no longer being returned after updating to Laravel 5.2. Fixed in 8106c57, and slightly cleaned up how we override the trait. I also took the opportunity to make error responses match our standard formatting:

```json
{
  "code": 422,
  "message": "Failed validation.",
  "errors": {
    "email": [
      "The email has already been taken."
    ]
  }
}
```

@jonuy @aaronschachter Heads up that if you're using validation responses anywhere in the apps, the response will now be nested under the `errors` property rather than just tossed as the root-level response. No other changes. :)

weerd:
dang! looks quite sophisticated ;) nice work!

:+1:

PR #266, opened 2016-01-28
Unique random keys

DFurnes:
### Changes
Three changes, none of which are breaking!

##### Ensure API keys & Tokens are unique
We used to have a _small_ chance of key collisions (and also used two different methods for generating keys, whaaaa) – this PR standardizes to use Laravel's battle-tested `Str::random` and ensures the key is unique before saving.

##### Deprecates `X-DS-Application-Id` header
API keys are 40 characters long, so I don't think an extra (relatively easily guessable) 5-10 characters increases our security too much. No changes are needed, but this header will just be silently ignored from now on.

##### Use Auth facade for Authentication middleware
I forgot to update the `auth` middleware when making updates in #265, so it was not using the same logic to get the key from the request. This removes hardcoded references to the `Session` header. (Although it will still continue to function identically, if set!)

---

For review: @angaither @weerd 
/cc @aaronschachter @jonuy

### DIFF
--- a/app/Http/Middleware/Authenticate.php
+++ b/app/Http/Middleware/Authenticate.php
@@ -2,6 +2,7 @@
 
 namespace Northstar\Http\Middleware;
 
+use Auth;
 use Northstar\Models\Token;
 use Closure;
 use Symfony\Component\HttpKernel\Exception\HttpException;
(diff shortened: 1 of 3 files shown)


weerd:
:+1:

PR #268, opened 2016-02-03
Auth and users and profiles oh my

### LINKED ISSUE #94: Upon registration and login, password required
@DeeZone was getting errors that the password field was required upon registration
I think might be caused by #88

So I guess, a solution could be only login a user if the user is registered with a password?
But, for CGG, we logged in a user if they had the correct birthdate. 
this issue should probably be open for discussion

### LINKED ISSUE #131: Ensure unique mobile and email on user registration or update
There is this in `UserController@store`:

```
if (Input::has('email')) {
    $user = User::where('email', '=', $check['email'])->first();
} elseif (Input::has('mobile')) {
    $user = User::where('mobile', '=', $check['mobile'])->first();
}
```

but doesn't account for a case where I would POST my normal email with a mobile number that another user already owns.

There doesn't appear to be any check in `UserController@update`.

### LINKED ISSUE #238: Cannot update user profile without password
Due to security concerns, we locked down updating user records in #195 to require a user's password. Unfortunately this means we can't attach any profile data when existing users vote for a voting campaign like CGG or AGG though. :cry: 

Once we [add better API scoping](https://github.com/DoSomething/northstar/issues/228), we should revisit allowing "admin" scoped API Keys to update profile data without providing a password, since there are quite a few trusted internal apps that would legitimately want to do this.

/cc @angaither @jonuy

DFurnes:
#### Updates authentication endpoints
This PR includes some changes to authentication endpoints, aimed at simplifying "permission" logic and separating the use cases for user-facing tokens and internal service "administrative" tasks.

:warning: __Breaking Change:__ Login requests should now post to `auth/token`, and will return Tokens (with a nested user object), rather than a "hacked" user object. The associated user will be given under a `user` property in the JSON response.

In addition to being more "true" to the real underlying process (users don't have a `session_token` property in the database), this also allows us to use our transformer classes so the user response will be consistent with that given elsewhere.

:warning: __Breaking Change:__ Logout requests should now post to `auth/invalidate` (but otherwise are functionally identical to the old `/logout`).

:white_check_mark: __New Feature:__ The `auth/verify` endpoint lets us verify user credentials without creating a new user authentication token. This will be useful for testing credentials when users log into Phoenix (without creating bajillions of unused tokens). References DoSomething/phoenix#4526.

:warning: __Breaking Change:__ Users should now be registered via `POST auth/register`. In contrast to the `POST /users` endpoint, this _requires_ a password & will create and return an authentication token as well.

There are also some nice behind-the-scenes changes in this PR: authentication "methods" are now shared within the `Registrar` class, which helps clean up shared registration behaviors that occur conditionally in multiple places.

#### Updates user endpoints
:warning: __Breaking Change:__ User modification endpoints (`POST /users`, `PUT /users/:term/:id`, and `DELETE /users/:id`) are now _only_ accessible to admin-scoped keys. Because of this, they no longer require a user's password to change their profile!

 This lets us differentiate "user-facing" endpoints from internal endpoints (e.g. forwarding a user from CGG or Message Broker) which should neither require a password or create a new session. This also means that admin-scoped keys will no longer require a password to change user records! Closes #238. Closes #94.

:white_check_mark: __New Feature:__ The `POST /profile` endpoint allows user-scoped keys to update the currently authenticated user's profile. `GET /profile` is a convenient shortcut for getting the currently authenticated user's profile, and is equivalent to `GET /users/_id/:id`.

:hammer: __Bug Fix:__ User profiles can no longer be modified so that they contain duplicate emails or mobile numbers. Input validation, yeah! Fixes #131.

## Where should the reviewer start?
I would recommend looking at things commit-by-commit, as usual. It might also be helpful to look at the before and after for the `Registrar` ([before](https://github.com/DoSomething/northstar/blob/aaf40d5024250b076c78374e5daf287c8d4f57c5/app/Auth/Registrar.php), [after](https://github.com/DFurnes/northstar/blob/74e46152fa507f34ab25d9579af54b5ea488f682/app/Auth/Registrar.php)), `AuthController` ([before](https://github.com/DoSomething/northstar/blob/aaf40d5024250b076c78374e5daf287c8d4f57c5/app/Http/Controllers/AuthController.php), [after](https://github.com/DFurnes/northstar/blob/74e46152fa507f34ab25d9579af54b5ea488f682/app/Http/Controllers/AuthController.php)) and `UserController` ([before](https://github.com/DoSomething/northstar/blob/aaf40d5024250b076c78374e5daf287c8d4f57c5/app/Http/Controllers/UserController.php), [after](https://github.com/DFurnes/northstar/blob/74e46152fa507f34ab25d9579af54b5ea488f682/app/Http/Controllers/UserController.php)).

Sorry, I know this is a huge PR but it didn't seem to make sense to do piecemeal. :grimacing: 

## How should this be tested?
Automated tests should continue to pass. :traffic_light: I'll add tests for the profile endpoints and admin key restriction on the user-modifying endpoints in just a bit!

I'll also update documentation once this merges! :memo: 

---
/cc @aaronschachter @jonuy

### DIFF
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -3,51 +3,159 @@
 namespace Northstar\Auth;
 
 use Hash;
+use Illuminate\Contracts\Auth\Guard;
+use Northstar\Models\Token;
 use Northstar\Models\User;
+use Northstar\Services\Phoenix;
 use Symfony\Component\HttpKernel\Exception\UnauthorizedHttpException;
 
 class Registrar
 {
-    public function login($input)
+    /**
+     * @var Guard
+     */
+    protected $guard;
+
+    /**
+     * @var Phoenix
+     */
+    protected $phoenix;
+
+    /**
+     * Registrar constructor.
+     * @param Guard $guard
+     * @param Phoenix $phoenix
+     */
+    public function __construct(Guard $guard, Phoenix $phoenix)
     {
-        $login_type = 'username';
-        if ($input['email']) {
-            $email = strtolower($input['email']);
-            $user = User::where('email', '=', $email)->first();
-            $login_type = 'email';
-        } elseif ($input['mobile']) {
-            $user = User::where('mobile', '=', $input['mobile'])->first();
-            $login_type = 'mobile';
+        $this->guard = $guard;
+        $this->phoenix = $phoenix;
+    }
+
+    /**
+     * Authenticate a user based on the given credentials,
+     * and create a new session token.
+     *
+     * @param array $credentials
+     * @return mixed
+     */
+    public function login($credentials)
+    {
+        $user = $this->resolve($credentials);
+
+        if(! $this->verify($user, $credentials)) {
+            throw new UnauthorizedHttpException(null, 'Invalid credentials.');
         }
 
-        if (($user instanceof User) && Hash::check($input['password'], $user->password)) {
-            $token = $user->login();
-            $token->user = $user->toArray();
+        return $this->createToken($user);
+    }
 
-            // Return the session token with the user.
-            $user->session_token = $token->key;
+    /**
+     * Resolve a user account from the given credentials.
+     *
+     * @param array $credentials
+     * @return User|null
+     */
+    public function resolve($credentials)
+    {
+        if ($credentials['email']) {
+            $email = strtolower($credentials['email']);
+            return User::where('email', $email)->first();
+        }
 
-            return $user;
-        } elseif (($user instanceof User) && ! ($user->password)) {
+        if ($credentials['mobile']) {
+            return User::where('mobile', $credentials['mobile'])->first();
+        }
 
-            // check to see if $input['password'] equals user's drupal password
-            if (DrupalPasswordHash::check($input['password'], $user->drupal_password)) {
+        return null;
+    }
 
-                // if they're the same, make $input['password'] into a hash and save it to the user.
-                $user->password = $input['password'];
-                $user->unset('drupal_password');
-                $user->save();
+    /**
+     * Verify the given user and credentials. If the user has a Drupal
+     * password & it matches, re-hash and save to the user document.
+     *
+     * @param User $user
+     * @param array $credentials
+     * @return bool
+     */
+    public function verify($user, $credentials)
+    {
+        if(!$user) return false;
 
-                $token = $user->login();
-                $token->user = $user->toArray();
+        if (Hash::check($credentials['password'], $user->password)) {
+            return true;
+        }
 
-                // Return the session token with the user.
-                $user->session_token = $token->key;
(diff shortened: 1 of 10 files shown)


app/Auth/Registrar.php
```diff
@@ -3,51 +3,164 @@
...
+     * @return User|null
+     */
+    public function register($input, $user = null)
+    {
+        // If there is no user provided, create a new one.
+        if (! $user) {
+            $user = new User;
+        }
+
+        $user->fill($input);
```
angaither: I know we didn't care so much about this before, but is there anything we should block from mass updating? things we don't want the user to change? are those things guarded? 

i know they are, you're smart.

app/Http/Controllers/ProfileController.php
```diff
@@ -0,0 +1,59 @@
...
+     * @param Request $request
+     * @return \Illuminate\Http\Response
+     */
+    public function update(Request $request)
+    {
+        $user = Auth::user();
+
+        $this->validate($request, [
+            'email' => 'email|max:60|unique:users,email,'.$user->id.',_id',
+            'mobile' => 'unique:users,mobile,'.$user->id.',_id',
```
angaither: do we need to validate this stuff in each request? can this be shared across update & register?

app/Auth/Registrar.php
DFurnes: This will still use the [`$fillable`](https://github.com/DoSomething/northstar/blob/aaf40d5024250b076c78374e5daf287c8d4f57c5/app/Models/User.php#L23-L42) array on the user model. I think everything in there is decently safe, but it's a good thing to keep an eye on. Things like `drupal_id` or `agg_id` could definitely be worth thinking about limiting to admin keys...

> i know they are, you're smart.

Hah! :stuck_out_tongue_winking_eye:

app/Http/Controllers/ProfileController.php
DFurnes: Yeah, I think it'd be worth pulling out validation into the `Registrar` as well, so we can conditionally change validation rules based on whether a user exists (or other factors)... that's what we're doing [in the voting app](https://github.com/DoSomething/voting-app/blob/42bddbd4bdf69a14b065b0dedcdde15937f702ba/app/Services/Registrar.php#L32-L51) and it seemed to work well.

angaither:
without sending a request to logout, do they invalidate ever?
> Logout requests should now post to auth/invalidate (but otherwise are functionally identical to the old /logout

angaither:
code looks :ok:

DFurnes:
> without sending a request to logout, do they invalidate ever?

Nope, which is probably something we should think about...

weerd:
Fancy :cocktail: :+1:

DFurnes:
Merging! :bow:

weerd:
:boom:

PR #271, opened 2016-02-04
Fixes fatal error on creating new API keys.

DFurnes:
#### Changes
Quick bug fix. I cleverly named two variables `$key` and expected Laravel to figure it out.

I added a test for this endpoint (which fails before applying the fix, and is green now). :traffic_light: 

---


angaither:
> and expected Laravel to figure it out.



:+1:

weerd:
cool! :+1:

PR #274, opened 2016-02-11
Add a super simple health check endpoint.

DFurnes:
#### Changes
Adds a dead simple `/status` endpoint that New Relic can ping to make sure the server is not dead.

---
For review: @mshmsh5000


mshmsh5000:
A thing of beauty is a joy forever. :+1:

PR #275, opened 2016-02-11
Query string kaboom

DFurnes:
#### Changes
I didn't test whether query strings were properly forwarded, and things were blowing up. Fixes the error by getting the query string as an array, and added a test just cuz.

---
For review: @angaither 
/cc @aaronschachter


angaither:
:question: :boom: :ship:

PR #276, opened 2016-02-11
Update reportback method to pass query strings correctly.

DFurnes:
#### Changes
In my haste to fix the issue in #275, I forgot to update the reportbacks index endpoint as well.

---

For review: @angaither


angaither:
HASTE :+1:

Issue #279, opened 2016-02-16
Internal server error on reporting back


DFurnes:
@aaronschachter reported a `500 Internal Server Error` when trying to hit the `POST reportbacks/` endpoint from the iOS app. From pointing my local Northstar instance to talk to the staging server for it's Phoenix endpoints, I'm getting the following error in the debug message:

```
Server error response [url] http:\/\/staging.beta.dosomething.org\/api\/v1\/campaigns\/1286\/reportback [status code] 500 [reason phrase] : Reportback file_url not found.
```

DFurnes:
From the documentation for the Phoenix endpoint, it looks like `file_url` is only required if `file` (the Base-64 encoded string) isn't provided. It definitely [looks like](https://github.com/DoSomething/northstar/blob/cbcca00bb4c84a6fb255408c17ba90bc26c95e8b/app/Services/Phoenix.php#L303) we're sending this along though...

DFurnes:
Ah, I think I got it. I [removed](https://github.com/DoSomething/northstar/pull/273/files#diff-eb9950cc3e4522e41dd68c31824f0497L220) the `'filename' => 'test123456.jpg'` parameter from the API reportback method in my changes in #273. From digging in the [Phoenix endpoint source](https://github.com/DoSomething/phoenix/blob/62106f609af0c415e97559b19b05e9394a600338/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc#L314) it looks like that might have been an undocumented property instructing Phoenix to save the data URL with that filename?

/cc @aaronschachter @weerd

aaronschachter:
Good investigating @DFurnes -- I think you're right -- it does look like this is (missing from the docs) [https://github.com/DoSomething/phoenix/wiki/api#campaign-reportback]. I have a vague recollection that we needed a filename into order to get the File API's save working.. (it's been well over a year since i've looked at this code :older_man: ) 

I was using Services default endpoints as reference when I created these endpoints, which gives you back just simply the new entity id upon a POST to create. Ideally we would alter this to return the relevant Signup and Reportback responses from Transformer land... but that's a different issue.

DFurnes:
@aaronschachter Mmm, makes sense. I put up a quick fix in #280 so this shouldn't be a blocker™.

One thing that could make sense would be having the Northstar API "automatically" perform a `GET /reportbacks/:id` on the returned string so we get the response we're expecting and then can transparently update the Phoenix endpoint later on?

[...]

PR #280, opened 2016-02-16
Reportback bugfixes

DFurnes:
#### Changes
Two quick bug fixes:

:bug: __Fixed bug:__ I made a typo when making the `getReportback` method, so fixed that.

:bug: __Fixed bug:__ Adds a randomly generated `filename` value to the `createReportback` payload, so it doesn't kaboom when trying to report back for a campaign. References #279.

---
For review: @weerd @aaronschachter


aaronschachter:
LGTM :+1:

Issue #281, opened 2016-02-16
Nicer API responses for targeted action proxies


DFurnes:
Discussed this earlier this morning in #279 and at sprint planning. Since the "targeted actions" Services endpoints return some pretty unhelpful responses (`["3872"]`, `[false]`, etc.), let's try to normalize those and provide more helpful responses in Northstar until things can be refactored in Phoenix. :construction_worker: 

/cc @aaronschachter @jonuy

weerd:
Yaaaaais please!

DFurnes:
Fixed by #282! :hammer:

PR #282, opened 2016-02-16
Return the full reportback object from the store endpoint.

DFurnes:
### Changes
References #281. This is a temporary measure until we can update the underlying API on Phoenix!

:warning: __Breaking Change:__ The `POST /reportbacks` endpoint will now return a full reportback object (same formatting as the `GET /reportbacks/:id` endpoint), rather than an array with a string ID.

:warning: __Breaking Change:__ The `POST /signups` endpoint will now return a full signup object (same formatting as the `GET /signups/:id` endpoint), rather than an array with a string ID.

### Discussion
~~For the signup endpoint, I'm not sure whether I _can_ fetch the existing signup... I tried fetching signups from the `signups/` endpoint with the proper `?user=xxx` and `&campaign=xxx` filter parameters, but it ended up returning like 5 campaigns. :disappointed: 

This might be a campaign-run issue...? Not 100% sure but would love thoughts if anyone has them.~~

Okay, never mind! I was passing a single `campaign=123` rather than `campaigns=123` and I am very dumb so never mind anything okay bye.

---
For review: @weerd @angaither 
/cc @jonuy @aaronschachter

### DIFF
--- a/app/Http/Controllers/ReportbackController.php
+++ b/app/Http/Controllers/ReportbackController.php
@@ -121,9 +121,10 @@ class ReportbackController extends Controller
 
         // HACK: Since the "create reportback" services endpoint returns a less-than-helpful
         // response (see above), we'll use that to fetch the full transformed response.
+        $reportbackResponse = $this->phoenix->getReportback($reportback[0]);
+
         // NOTE: Since we don't know whether a reportback is new or just updating an existing RB,
         // we'll always just return 200 OK. This will probably change when we update Phoenix.
-        $reportbackResponse = $this->phoenix->getReportback($reportback[0]);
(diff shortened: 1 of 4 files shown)


DFurnes:
Woohoo, with the changes in DoSomething/phoenix#6174 the `GET /reportbacks/:id` is now working for newly-submitted reportbacks! :tada:

weerd:
yay! :dancers:

weerd:
:+1:

DFurnes:
And was able to fix the issue I'd been having with fetching an existing signup. Hackzzzz! :muscle:

weerd:
:+1: :dancer:

PR #283, opened 2016-02-17
Clean up Avatar controller.

DFurnes:
#### Changes
Fixes issues parsing file extension from Data URLs (secret: we _weren't_ parsing file extensions) when uploading avatars. Also just tidies up logic so it’s easier to follow what’s going on. :leaves: 

Also locks down the avatar resource, so a user can only edit their own avatar. :lock: 

I ran through testing this on my local (with the `Storage::disk('s3')->put()` call commented out) and the paths & file data are now correctly formed now for both data URLs & multipart form submissions.

I'll test this for realsies once we have this up on QA.

---
/cc @aaronschachter @weerd

### DIFF
--- a/app/Http/Controllers/AvatarController.php
+++ b/app/Http/Controllers/AvatarController.php
@@ -2,9 +2,11 @@
 
 namespace Northstar\Http\Controllers;
 
+use Northstar\Http\Transformers\UserTransformer;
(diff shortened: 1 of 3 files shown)


weerd:
:+1: :sheep: It!

PR #284, opened 2016-02-17
Remove unused classes/routes/tests.

DFurnes:
#### Changes
Removes unused things. A lot of this stuff relies on the local Mongo `campaigns` subresource which is not something we're actively using anymore, so it's confusing to keep code that expects it to be in any way useful. :bomb: 

Tests, passing, y'know. :traffic_light: 

#### Discussion
Can we delete `campaigns` from each user? Should we? How does Mongo even? :elephant: 

---
For review: @angaither


angaither:
:skull: :coffin: :gun: 
RIP KUDOS 
and the misguided notions we would store :pancakes:like data in mongo

weerd:
kudos

:sob: :sob: :sob:

DFurnes:
:fire: :wave:

PR #290, opened 2016-02-29
Jeez why is this so slow i guess its because we never set an index

DFurnes:
#### Changs
Adds indexes to the `key` on authentication tokens, and the `api_key` on API keys. So that we can look them up super fast even when there's a super lot of them.

Things should work exactly the same, but not slow as molasses.

---
For review: @angaither


angaither:
ya :ship:

PR #291, opened 2016-02-29
Fix typo in down migration.

DFurnes:
#### Changes
Was "unmigrating" the wrong table. :cactus: 

---
For review: @angaither


angaither:
:memo:

PR #292, opened 2016-03-01
Fix POST upsert issue & add Mobile Commons fields to user.

DFurnes:
#### Changes

:bug: Fixes issue where the `POST users/` route was not [upserting](https://docs.mongodb.org/v2.6/reference/glossary/#term-upsert) because validation was not excluding the "existing" document when checking for uniques.

:white_check_mark: Adds `mobilecommons_id` and `mobilecommons_status` fields to the user document and exposes them through the API. Content for the two new fields is coming soon!

:warning: Remove `race` and `religion` from the user profile. These are never being set by the user (and not being used by the mobile app), so it seems silly to be returning them.

---

### DIFF
--- a/app/Http/Controllers/UserController.php
+++ b/app/Http/Controllers/UserController.php
@@ -72,17 +72,16 @@ class UserController extends Controller
      */
     public function store(Request $request)
     {
-        $user = null;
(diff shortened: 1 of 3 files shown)


angaither:
:+1:

PR #293, opened 2016-03-02
Add language to user object.

DFurnes:
#### Changes
Adds `language` field to the user, so we can collect that. References DoSomething/phoenix#6224.

---


angaither:
:globe_with_meridians:

PR #295, opened 2016-03-02
Source should be immutable

DFurnes:
#### Changes
:warning: Makes `source` an immutable property on the User document. This allows us to always include a value in `source` when making an "upsert" to `POST users/` but _only_ save that value if that's really the initial source of the user.

:book: Update docblocks on `User` mutators, and [tidy up listing of user fields](https://github.com/DFurnes/northstar/blob/source-should-be-immutable/documentation/endpoints/users.md#create-a-user) in docs. 

I tested in my local Paw manually, and added a simple functional test for this and the expected "upsert" behavior (from #292) to make sure this doesn't break down the road.

---

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -79,14 +79,6 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
     protected $dates = ['created_at', 'updated_at'];
 
     /**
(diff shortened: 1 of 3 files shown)


angaither:
:hammer:

weerd:
:+1:

PR #299, opened 2016-03-04
Add StatHat logging!

DFurnes:
#### Changes
Adds StatHat logging for each route & a couple of authentication/key errors. :chart_with_upwards_trend:  :tophat:

#### Questions
Is logging each time a route is matched overkill? I went back and forth, but it does give us nice visibility since there's no Google Analytics in the JSON API world...

---
For review: @angaither @weerd @mshmsh5000

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -15,3 +15,6 @@ S3_URL=https://s3.example.com
 PARSE_APP_ID=ParseAppId
 PARSE_API_KEY=ParseAPIKey
 PARSE_MASTER_KEY=ParseMasterKey
+
+STATHAT_EZ_KEY=puppet-sloth@example.com
+STATHAT_APP_NAME=northstar-dev
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -96,6 +96,7 @@ class AuthController extends Controller
         // Attempt to delete token.
         $deleted = $token->delete();
         if (! $deleted) {
+            app('stathat')->ezCount(env('STATHAT_APP_NAME', 'northstar').' - logout error');
             throw new HttpException(400, 'User could not log out. Please try again.');
         }
 
--- a/app/Providers/EventServiceProvider.php
+++ b/app/Providers/EventServiceProvider.php
@@ -4,6 +4,7 @@ namespace Northstar\Providers;
 
 use Illuminate\Contracts\Events\Dispatcher as DispatcherContract;
 use Illuminate\Foundation\Support\Providers\EventServiceProvider as ServiceProvider;
+use Illuminate\Routing\Events\RouteMatched;
 
 class EventServiceProvider extends ServiceProvider
 {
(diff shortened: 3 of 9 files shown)


.env.example
```diff
@@ -15,3 +15,6 @@ S3_URL=https://s3.example.com
 PARSE_APP_ID=ParseAppId
 PARSE_API_KEY=ParseAPIKey
 PARSE_MASTER_KEY=ParseMasterKey
+
+STATHAT_EZ_KEY=puppet-sloth@example.com
```
angaither: lovely lovely

app/Http/Controllers/AuthController.php
```diff
@@ -96,6 +96,7 @@ public function invalidateToken(Request $request)
         // Attempt to delete token.
         $deleted = $token->delete();
         if (! $deleted) {
+            app('stathat')->ezCount(env('STATHAT_APP_NAME', 'northstar').' - logout error');
```
angaither: does this make different buckets like
DEV - northstar --logout

it should probably be the other way around, right?
northstar -- dev

app/Providers/EventServiceProvider.php
```diff
@@ -34,6 +27,11 @@ public function boot(DispatcherContract $events)
     {
         parent::boot($events);
 
-        //
+        // Update count on StatHat every time a route is hit.
+        // e.g. will increment the "northstar - v1/users/{term}/{id}" stat each
+        // time a client attempts to view a user profile through that route.
+        $events->listen(RouteMatched::class, function (RouteMatched $match) {
```
angaither: woaaaaaaaahhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh

app/Http/Controllers/AuthController.php
DFurnes: `northstar` is the fallback given to `env()` for if the environment variable with an instance name isn't set. So you could set the `STATHAT_APP_NAME` to something like `northstar-dev` and it'd format the stat like so:

`"northstar-dev - logout error"`.

DFurnes: It does kinda suck to have to do this everywhere though, so thinking of adding a little "prefix" setting to the StatHat package that could just prepend this to everything.

angaither: right, but do we have the `app_name`s set to different things across multiple envs?
ideally they would all fall into the same `northstar` bucket, right? 

though I guess historically we don't separate environments in stathat, so maybe

DFurnes: Oh, I totally misunderstood. I could see it being useful to determine which environment is getting weird traffic patterns, since we care way more about errors or malicious activity on prod than QA for example.

We separate environments like this for Voting App so stats are prefixed (`agg -` or `cgg -`)... which yeah, idk if we've ever used but seems like a nice thing? :grin:

angaither: yeah, true
it wasn't voting-app with buckets for each instance.

angaither:
this is fineee :chart_with_upwards_trend:

app/Providers/EventServiceProvider.php
mshmsh5000: Does this mean a new stat is created for every user ID?

DFurnes: Nope! It will make a new state based on the route that is matched... so if the client requests `v1/users/_id/5430e850dt8hbc541c37tt3d`, it will increment the `northstar - v1/users/{term}/{id}` stat.

PR #300, opened 2016-03-07
Use v2 of StatHat package! See DoSomething/stathat-php#4.

DFurnes:
#### Changes
Use v2 of StatHat package! See DoSomething/stathat-php#4 for details.

The biggest change here is moving the app name prefix to the configuration array, rather than having to build it from the environment variable each time we trigger a stat.


angaither:
:+1:

PR #302, opened 2016-03-09
Tests tests tests

DFurnes:
#### Changes
Updates our test suite to use the [new testing methods](https://laravel.com/docs/5.1/testing) added in Laravel 5.1. It makes things like a _million_ times easier to read and write. :traffic_light: 

I also added `withScopes()` and `asUser()` helper methods. They will automatically make the requested key or tokens in the database and set the proper headers for the next request. :dancer: 

---
For review: @angaither @weerd

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -162,6 +162,10 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
      */
     public function setPasswordAttribute($value)
     {
+        // Since we're setting the password, we can discard any existing
+        // Drupal-hashed value copied into the `drupal_password` field.
+//        $this->unset('drupal_password');
+
         $this->attributes['password'] = Hash::make($value);
     }
 
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -162,10 +162,6 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
      */
     public function setPasswordAttribute($value)
     {
-        // Since we're setting the password, we can discard any existing
-        // Drupal-hashed value copied into the `drupal_password` field.
-//        $this->unset('drupal_password');
-
         $this->attributes['password'] = Hash::make($value);
     }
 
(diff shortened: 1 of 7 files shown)


tests/UserTest.php
```diff
@@ -100,18 +52,21 @@ public function testGetAllDataFromUser()
...
-        $data = json_decode($response->getContent());
-        $this->assertObjectHasAttribute('data', $data);
-        $this->assertObjectHasAttribute('meta', $data);
-        $this->assertObjectHasAttribute('pagination', $data->meta);
-        $this->assertObjectHasAttribute('total', $data->meta->pagination);
-        $this->assertObjectHasAttribute('count', $data->meta->pagination);
-        $this->assertObjectHasAttribute('per_page', $data->meta->pagination);
-        $this->assertObjectHasAttribute('current_page', $data->meta->pagination);
-        $this->assertObjectHasAttribute('links', $data->meta->pagination);
+        $this->get('v1/users');
```
angaither: not needed but it might be easier to have the version abstracted out so you don't have to have it all over the place

app/Models/User.php
```diff
@@ -162,6 +162,10 @@ public function setParseInstallationIdsAttribute($value)
      */
     public function setPasswordAttribute($value)
     {
+        // Since we're setting the password, we can discard any existing
+        // Drupal-hashed value copied into the `drupal_password` field.
+//        $this->unset('drupal_password');
```
weerd: not sure but weird comment spacing thing?

angaither:
okie :ship:

app/Models/User.php
DFurnes: oh gee, didn't mean to commit that :weary:

tests/UserTest.php
DFurnes: Mmm, not sure how much extra work that'd be but it would definitely be cleaner.

weerd:
Hokay! :+1:

PR #304, opened 2016-03-09
Cleanup and password tests

DFurnes:
#### Changes
I'm trying to figure out why some users aren't validating when we check Phoenix logins against Northstar... haven't found any big issues yet, but did figure out that we're not removing the `drupal_password` field when a user changes their password on Phoenix. (525627f)

I added some more unit tests with known good/bad Drupal password hashes as well (acc164c).

Finally, an unrelated change: I moved shared controller methods into traits, rather than having them all on the base `Controller` class. This makes it a bit easier to skim, and will make it easier to yank them out into a shared package at some point. (cd1e643)

I'd recommend going commit-by-commit. :1234: 

:traffic_light: 

---
For review: @angaither @weerd

### DIFF
--- a/tests/PasswordTest.php
+++ b/tests/PasswordTest.php
@@ -1,0 +1,43 @@
+<?php
+
+use Northstar\Models\User;
+
+class PasswordTest extends TestCase
+{
+    /**
+     * Tests that Drupal password hasher is working correctly.
+     */
+    public function testAuthenticatingWithDrupalPassword()
+    {
+        $user = User::create([
+            'email' => 'dries.buytaert@example.com',
+            'drupal_password' => '$S$DOQoztwlGzTeaobeBZKNzlDttbZscuCkkZPv8yeoEvrn26H/GN5b',
+        ]);
+
+        $this->withScopes(['user'])->json('POST', 'v1/auth/verify', [
+            'email' => 'dries.buytaert@example.com',
+            'password' => 'secret',
+        ]);
+
+        // Assert response is 200 OK and has expected data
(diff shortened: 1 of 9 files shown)


tests/PasswordTest.php
```diff
@@ -0,0 +1,92 @@
...
+        ]);
+
+        // Assert user has been updated in the database with a newly hashed password.
+        $user = $user->fresh();
+        $this->assertArrayNotHasKey('drupal_password', $user['attributes']);
+        $this->assertArrayHasKey('password', $user['attributes']);
+
+        // Finally, let's try logging in with the newly hashed password
+        $this->withScopes(['user'])->json('POST', 'v1/auth/verify', [
+            'email' => 'dries.buytaert@example.com',
```
angaither: totes profesh

tests/PasswordTest.php
```diff
@@ -0,0 +1,92 @@
...
+        $this->assertResponseStatus(200);
+    }
+
+    /**
+     * Test that updating a user's password will re-hash it.
+     */
+    public function testUpdatingUser()
+    {
+        $user = User::create([
+            'email' => 'acquia.consultant@example.com',
```
angaither: yikes

angaither:
SURE

tests/PasswordTest.php
DFurnes: :sweat_drops:

tests/PasswordTest.php
DFurnes: :wink:

PR #305, opened 2016-03-09
Error formatz

DFurnes:
#### Changes
Up for discussion, but two changes based on feedback by @aaronschachter:

1. Add a `code` to the `respond()` method on the controller, so it's consistent with error responses.
2. Wrap `422 Unprocessable entity` validation errors in `error` object, so it's consistent with other error codes. Since `error.errors` seems a bit strange, maybe we should rename that to `fields` since it's an array of errors per field?

I did a quick search through each of the app repositories and it doesn't look like this will _break break_ anything, but [any code that looks for those individual field errors](https://github.com/DoSomething/LetsDoThis-Android/blob/52422e7f8170d942cb756f2b9aed1f05be284a06/app/src/main/java/org/dosomething/letsdothis/tasks/BaseRegistrationTask.java#L85-L91) will have to be updated to check for the new "path" in the object.

---
For review: @angaither @weerd @jonuy
/cc @aaronschachter @jonuy

### DIFF
--- a/app/Http/Controllers/Controller.php
+++ b/app/Http/Controllers/Controller.php
@@ -24,9 +24,11 @@ abstract class Controller extends BaseController
     protected function buildFailedValidationResponse(Request $request, array $errors)
     {
         $response = [
-            'code' => 422,
-            'message' => 'Failed validation.',
-            'errors' => $errors,
+            'error' => [
+                'code' => 422,
+                'message' => 'Failed validation.',
+                'fields' => $errors,
+            ],
         ];
 
         return response()->json($response, 422);
--- a/app/Http/Controllers/Traits/TransformsResponses.php
+++ b/app/Http/Controllers/Traits/TransformsResponses.php
@@ -109,6 +109,7 @@ trait TransformsResponses
     {
         $response = [
             $status => [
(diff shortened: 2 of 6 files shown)


aaronschachter:
I'm into it -- found this all by diving deep into the network errors we're getting back from Crashlytics here: https://fabric.io/do-something/ios/apps/org.dosomething.letsdothis/issues/56d74325f5d3a7f76b4131c3

Helps figure out which ones to filter (don't need to track every time someone tries to register with existing email)

jonuy:
Including `code` in the message body always felt a little weird to me because you should already be getting that back in the `status` line. That being said though, I'm fine with these changes. From the Android app's standpoint, there isn't any specific parsing of error messages that we're doing there.

DFurnes:
Yeah, it's definitely a little more verbose. But it is nice to have everything be consistent.

DFurnes:
It also sounds like getting the HTTP status code is a lot more difficult in the iOS world, so if it's more convenient to pull it as a part of the object I'm all for that.

jonuy:
:+1:

weerd:
post-merge :+1:

PR #306, opened 2016-03-10
Fix registration and login issues.

### LINKED ISSUE #303: Run mutators before validating uniqueness
We should run mutators before validating uniqueness, since right now the `unique` validator will check if the exact string exists in the database, so registering `Example@DoSomething.org` would not trigger an error if `example@dosomething.org` exists in the database (and same with phone numbers).

References DoSomething/LetsDoThis-iOS#903.

DFurnes:
#### Changes
This PR adds two fixes for the login and registration flows:

:bug: Fixes logging in by mobile. Previously the user had to be a machine and strip out special characters themselves in order to resolve the right account. Now they can be human and just go wild with all the characters - we'll strip out non-numeric ones before trying to find it in the database.

:bug: Fixes #303. Previously, duplicate users could be created on registration if you used a different capitalization. To fix that, we run the mutators for unique fields (`email` and `mobile`) before running the validator, so it checks the thing it should be checking!

I tried a red-green-refactor thing for fun, so reviewing commit-by-commit will first add a failing test for things that are wrong, and then a fix for that failing test in the following commit. :traffic_light: 

---
For review: @angaither @weerd 
/cc @aaronschachter @jonuy

### DIFF
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -51,10 +51,10 @@ class Registrar
     }
 
     /**
-     * Normalize the given credentials (either before validating, or
-     * before saving to the database).
+     * Normalize the given credentials in the array or request (for example, before
+     * validating, or before saving to the database).
      *
-     * @param $credentials
+     * @param \ArrayAccess|array $credentials
      * @return mixed
      */
     public function normalize($credentials)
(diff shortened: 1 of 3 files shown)


weerd:
snazzy! :dancer: :+1:

angaither:
:heart_eyes: 
> I tried a red-green-refactor thing for fun, so reviewing commit-by-commit will first add a failing test for things that are wrong, and then a fix for that failing test in the following commit. :traffic_light:

PR #307, opened 2016-03-10
Limit results to 100 per page.

DFurnes:
#### Changes
Teeny thing that occurred to me the other night. A malicious user could have entered like 9001 users per page and we would have been all like “ok” and chugged along to process a monstrous response. Instead, let's set a reasonable limit of 100 per page and save ourselves the trouble.

---
For review: @angaither @weerd @jonuy


jonuy:
Quite reasonable :indeed mustache man isn't an emoji here: :+1:

PR #309, opened 2016-03-11
Fixes more dupe bugs...

DFurnes:
#### Changes
Found some more bugs that were allowing dupes to be created:

:bug: Fixes issue where you could create a duplicate user by "upserting" from the `POST users/` endpoint with one index that matched an existing user and one that did not.

:beetle: Fix issue where you could update a user's profile to have a duplicate ID if you used a non-normalized form of that index field (like with capital letters or symbols or stuff).

#### Test this, how do I?
Check 'em, commit-by-commit! Red ones are bugs I found, green ones are fixes or refactors. :traffic_light: 

---
For review: @angaither @weerd

### DIFF
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -80,13 +80,26 @@ class Registrar
     {
         $credentials = $this->normalize($credentials);
 
-        if (! empty($credentials['email'])) {
-            return User::where('email', $credentials['email'])->first();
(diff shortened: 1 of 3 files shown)


DFurnes:
I think we could probably fix this more elegantly by moving the validation step into the Registrar, but this gets everything back to green for now! :smile:

angaither:
:+1:

PR #310, opened 2016-03-11
Fixes #308.

### LINKED ISSUE #308: Query params aren't included in profile/signups and profile/reportbacks
This presents a problem if, for example, a user has more than 25 signups or reportbacks on their profile.

Doing some like this right now isn't supported: `GET /profile/signups?count=100`

DFurnes:
#### Changes
Adds support for other query params in `/profile/signups` and `/profile/reportbacks` endpoints.

Fixes #308.

---
For review: @angaither @weerd


DFurnes:
@jonuy Tested on my local and it works! :heavy_check_mark: :checkered_flag: :dancer:

PR #311, opened 2016-03-14
Fix verification when user includes a space in their email.

DFurnes:
#### Changes
Some users were being _SLOPPY_ and including a trailing space in their emails. The nerve! This normalizes those values before validating or checking for them in the database. :relieved: 

---
For review: @angaither @weerd

### DIFF
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -60,7 +60,7 @@ class Registrar
     public function normalize($credentials)
     {
         if (! empty($credentials['email'])) {
-            $credentials['email'] = strtolower($credentials['email']);
+            $credentials['email'] = trim(strtolower($credentials['email']));
         }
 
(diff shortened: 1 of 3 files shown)


app/Auth/Registrar.php
```diff
@@ -60,7 +60,7 @@ public function login($credentials)
     public function normalize($credentials)
     {
         if (! empty($credentials['email'])) {
-            $credentials['email'] = strtolower($credentials['email']);
+            $credentials['email'] = trim(strtolower($credentials['email']));
```
angaither: do we want this in mobile too?

angaither:
:+1:

weerd:
:+1:

app/Auth/Registrar.php
DFurnes: I think the regex should handle it, but should probs make a test for that too!

PR #314, opened 2016-03-16
Speed up test suite & expose Faker to test cases.

DFurnes:
#### Changes
One last round of touch-ups to the test suite: Speed Demon Edition™. :racehorse: 

* Reset the database between each test, without running the database seeders. This speeds things up considerably (like 2x faster!) and is nice because it makes each test completely self-contained (rather than relying on a seed user that could break the test if changed later). 
* Run tests on a separate `northstar-test` database. This means that running the test suite will no longer wipe your local environment and make manual testing harder! :innocent: 
* Adds a Faker instance to the base test case: this lets us easily generate dummy data within tests, e.g. `$this->faker->unique()->email`. This instance is reset between every test.
* Adds annotations to each model with the "magic" database properties. This silences PHPStorm warnings from magically resolved properties (and thus silences my OCD), but it’s also a nice place for documentation for each field!

---

For review: @angaither @weerd

### DIFF
--- a/app/Models/ApiKey.php
+++ b/app/Models/ApiKey.php
@@ -6,6 +6,16 @@ use Illuminate\Support\Str;
 use Jenssegers\Mongodb\Model;
 use Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException;
 
+/**
+ * The API key model. These identify the "client application" making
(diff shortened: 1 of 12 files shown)


angaither:
oooolala :+1:

PR #315, opened 2016-03-16
Everything and the kitchen sync

### LINKED ISSUE #312: Add unique index to email & mobile fields
Like, why haven't we done this yet?!?

Luckily, it seems like we've plugged the bugs that were letting this happen so we shouldn't see the issue anymore regardless, but if not it's better to fail with a constraint violation than save invalid data that we have to clean up later on! 

See [MongoDB Documentation](https://docs.mongodb.org/manual/tutorial/create-a-unique-index/).

DFurnes:
#### Changes
Two things before we [re-sync from Phoenix](https://github.com/DoSomething/phoenix/pull/6282): 

 - When a `drupal_password` is added to an account, remove their existing `password` so we don't continue trying to authenticate against that. These two fields should be mutually exclusive.
 - Closes #312 by adding unique indexes to... y'know, unique fields! Computer science. Great win.

#### Hmm, how are we gonna deploy this?
This is probably gonna be another bummer to deploy because it will need to delete and recreate two massive indexes, so I think we may have to manually run these migrations to get this in production.

---
For review: @angaither @weerd

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -207,6 +207,21 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
     }
 
     /**
+     * Mutator to remove any existing password if we migrate a hashed password.
+     * This is a one-time thing for syncing users from Phoenix and ensuring that
+     * we *only* keep their latest hashed Drupal password.
+     */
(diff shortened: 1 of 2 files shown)


angaither:
:+1:

Issue #316, opened 2016-03-16
Fix index conflict on empty strings


DFurnes:
We have some users that seem to have `""` saved as their email in Thor, which is causing the [unique index migration](https://github.com/DoSomething/northstar/blob/dev/database/migrations/2016_03_16_181119_AddUniqueIndexes.php) to fail when [deploying to Thor](https://app.wercker.com/#deploystep/56e9c2b6d19bbb306a0455ac). We should fix any existing accounts with this problem, and then make sure that it can't happen again! :no_bicycles:

angaither:
:japanese_ogre:

DFurnes:
It looks like we're also saving `null` values into these fields sometimes, and MongoDB sparse indexes only skip the unique constraint if the field flat-out doesn't exist (so two documents with `null` will conflict). Wahhh.

DFurnes:
This is fixed on QA and Thor after a little bit of [Mongo doctoring](https://gist.github.com/DFurnes/0aa32172d135aee96a47).

I'll do the same on production on Monday. :construction_worker:

DFurnes:
Alright, and fixed on production. :hammer:

PR #317, opened 2016-03-17
Add a test for unique constraint when creating users.

DFurnes:
#### Changes
Just making sure that we're not still creating the problem seen in #316...

:bug: Fix issue where indexed fields could be set to `null` or an empty string, which would cause the "unique" constraint violation added in #315 to trigger an error.

:beetle: Fix issue where a user could be updated to have _neither_ an email or mobile number.

:leaves: Moved user validation into the Registrar so we can use the same logic everywhere. 

:traffic_light: We didn't have tests for the profile endpoints. Whaaaa? Added some.

Is there anything I'm missing? Any other ways that you could possibly save a `null` or `''` value to an indexed field? Or otherwise get up to no good? :boom: :interrobang: 

---
For review: @angaither @weerd

### DIFF
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -4,6 +4,10 @@ namespace Northstar\Auth;
 
 use Hash;
 use Illuminate\Contracts\Auth\Guard;
+use Illuminate\Http\JsonResponse;
+use Illuminate\Validation\Factory as Validation;
+use Illuminate\Http\Request;
+use Illuminate\Validation\ValidationException;
 use Northstar\Models\Token;
 use Northstar\Models\User;
 use Northstar\Services\Phoenix;
@@ -22,14 +26,21 @@ class Registrar
     protected $phoenix;
 
     /**
+     * @var Validation
+     */
+    protected $validation;
+
+    /**
      * Registrar constructor.
      * @param Guard $guard
      * @param Phoenix $phoenix
+     * @param Validation $validation
      */
-    public function __construct(Guard $guard, Phoenix $phoenix)
+    public function __construct(Guard $guard, Phoenix $phoenix, Validation $validation)
     {
         $this->guard = $guard;
         $this->phoenix = $phoenix;
+        $this->validation = $validation;
     }
 
     /**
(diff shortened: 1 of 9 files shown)


app/Auth/Registrar.php
```diff
@@ -71,6 +82,45 @@ public function normalize($credentials)
...
+     * @param array $additionalRules
+     * @throws ValidationException
+     */
+    public function validate(Request $request, User $user = null, array $additionalRules = [])
+    {
+        $fields = $request->all();
+
+        $existingId = isset($user->id) ? $user->id : 'null';
+        $rules = [
+            'email' => 'email|max:60|unique:users,email,'.$existingId.',_id|required_without:mobile',
```
weerd: Looks like you're not using the new `.styleci.yml` :wink:

DFurnes: :information_desk_person:

app/Auth/Registrar.php
```diff
@@ -71,6 +82,45 @@ public function normalize($credentials)
...
+            'mobile' => 'unique:users,mobile,'.$existingId.',_id|required_without:email',
```
weerd: what does passing `$existingId` do in both these instances?... is that a field in the table?

DFurnes: That tells the [uniqueness validator](https://laravel.com/docs/5.2/validation#rule-unique) to ignore the given ID when checking the database for duplicates of that field, so you don't get a "unique" validation error if trying to update a user with their existing email or mobile number.

app/Models/User.php
```diff
@@ -145,8 +144,10 @@ public function getLastInitialAttribute()
      */
     public function setEmailAttribute($value)
     {
-        // Skip mutator if attribute is null.
+        // Drop field if attribute is empty string null.
```
weerd: typo, `...if empty string or null.`?

DFurnes: :pencil2: :heavy_check_mark:

tests/UserTest.php
```diff
@@ -262,6 +262,101 @@ public function testCreateUser()
...
+     * @return void
+     */
+    public function testCreateMultipleUsers()
+    {
+        // Create some new users
+        for ($i = 0; $i < 5; $i++) {
+            $this->withScopes(['admin'])->json('POST', 'v1/users', [
+                'email' => $this->faker->unique()->email,
+                'mobile' => '', // this should not save a `mobile` field on these users
+                'source' => 'phpunit',
```
weerd: hehe

app/Auth/Registrar.php
weerd: Oh man! That's cool :dancer:

weerd:
Hokay! Looks snazzy! :+1:

app/Models/User.php
```diff
@@ -228,10 +231,10 @@ public function setDrupalPasswordAttribute($value)
     public function setPasswordAttribute($value)
     {
         if (isset($this->drupal_password)) {
-            $this->unset('drupal_password');
+            $this->drop('drupal_password');
         }
 
-        $this->attributes['password'] = Hash::make($value);
+        $this->attributes['password'] = bcrypt($value);
```
angaither: BCRYPT! holy :blowfish:

angaither: in case that isn't clear IT'S A BLOWFISH

DFurnes: :laughing:

PR #319, opened 2016-03-18
Resolve users from the Registrar for show endpoint.

### LINKED ISSUE #318: Email@example.com returns user not found
When you try to get a user with Capitalfirst@email.com ns returns zero people.

Refs DoSomething/gladiator/issues/105

DFurnes:
#### Changes

:bug: Fixes #318. Resolve users from the Registrar when querying for the "show" endpoint. This lets us normalize fields before querying, and prevents querying by a non-indexed field.

:rocket: Because we're running all queries through the Registrar, we can now also easily map "id" to the `_id` Mongo field, so we don't need to expose database implementation details as part of our API. Yay! 

:relieved: No plans to remove the ability to query by `_id` anytime soon, by the way. But I am removing it from the documentation since we might remove it with a v2 down the line.

:dancer: Also runs filters and searches through the Registrar's normalization function, so they're no longer case-sensitive and also take advantage of the "id" to `_id` mapping.

---
For review: @angaither @weerd

### DIFF
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -144,7 +144,7 @@ class Registrar
         // For the first `where` query, we want to limit results... from then on,
         // we want to append (e.g. `SELECT * WHERE _ OR WHERE _ OR WHERE _`)
         $firstWhere = true;
-        foreach (['id', '_id', 'email', 'mobile', 'drupal_id'] as $type) {
+        foreach (User::$indexes as $type) {
             if (isset($credentials[$type])) {
                 $matches = $matches->where($type, '=', $credentials[$type], ($firstWhere ? 'and' : 'or'));
                 $firstWhere = false;
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -73,6 +73,12 @@ class Registrar
      */
     public function normalize($credentials)
     {
+        // Map id to Mongo's _id ObjectID field
+        if (! empty($credentials['id'])) {
+            $credentials['_id'] = $credentials['id'];
+            unset($credentials['id']);
+        }
+
         if (! empty($credentials['email'])) {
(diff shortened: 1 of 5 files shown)


angaither:
> This lets us normalize fields before querying, and prevents querying by a non-indexed field.

let's us search and find users with Email@example.com? or no

DFurnes:
@angaither Yep! :smile:

angaither:
@DFurnes :raised_hands:

app/Auth/Registrar.php
```diff
@@ -138,13 +144,18 @@ public function resolve($credentials)
         // For the first `where` query, we want to limit results... from then on,
         // we want to append (e.g. `SELECT * WHERE _ OR WHERE _ OR WHERE _`)
         $firstWhere = true;
-        foreach (['email', 'mobile'] as $type) {
+        foreach (User::$indexes as $type) {
```
weerd: :)

documentation/endpoints/keys.md
```diff
@@ -115,9 +115,12 @@ curl -X GET\
   "data": {
     "app_id": "testapplication",
     "api_key": "1laEQhhKtQEaPK0qpESdXHm2EbdLu5sRIRLcRtF8",
+    "scope": [
+      "admin",
+      "user"
+    ],
```
weerd: Style related... further above all in one line, but here separate lines? Either is fine, just go with one :)

weerd:
Coolio! Minor comment on style, but otherwise :+1:

documentation/endpoints/keys.md
DFurnes: :rage1:

DFurnes:
@weerd Tidied up in fb47a0e. :wink:

PR #324, opened 2016-04-01
Adds multi user support

deadlybutter:
#### What's this PR do?
Adds multi user support to the signups index 

Well thats kinda tricky, I had problems getting my local DB to march my drupals. BUT we do seed a user by default in northstar, so I used that user twice in the URL. From a code/logic perspective it doesnt matter that we used them twice in the URL.

screen shot 2016-04-01 at 2 42 18 pm
`http://northstar.app/v1/signups?competition=1&user=5430e850dt8hbc541c37tt3d,5430e850dt8hbc541c37tt3d&campaigns=1485`

### DIFF
--- a/app/Http/Controllers/SignupController.php
+++ b/app/Http/Controllers/SignupController.php
@@ -43,16 +43,17 @@ class SignupController extends Controller
     {
         $options = $request->query();
 
-        // If a user is specified, turn Northstar ID into Drupal ID
+        // Transform user into users.
         if (! empty($options['user'])) {
-            if (strpos($options['user'], ',') > 0) {
-                $usersQuery = explode(',', $options['user']);
-                $options['user'] = User::drupalIDForNorthstarId($usersQuery);
-            } else {
-                $options['user'] = User::drupalIDForNorthstarId($options['user']);
-            }
+            $options['users'] = $options['user'];
         }
-        
+
+        // If a user is specified, turn Northstar ID into Drupal ID
+        if (! empty($options['users'])) {
+            $usersQuery = explode(',', $options['users']);
+            $options['users'] = User::drupalIDForNorthstarId($usersQuery);
+        }
+
         return $this->phoenix->getSignupIndex($options);
     }
 
--- a/app/Http/Controllers/SignupController.php
+++ b/app/Http/Controllers/SignupController.php
@@ -45,7 +45,15 @@ class SignupController extends Controller
 
         // If a user is specified, turn Northstar ID into Drupal ID
         if (! empty($options['user'])) {
-            $options['user'] = User::drupalIDForNorthstarId($options['user']);
+            if (strpos($options['user'], ',') > 0) {
+                $usersQuery = explode(',', $options['user']);
+                $options['user'] = [];
+                foreach ($usersQuery as $user) {
+                    array_push($options['user'], User::drupalIDForNorthstarId($user));
+                }
+            } else {
+                $options['user'] = User::drupalIDForNorthstarId($options['user']);
+            }
         }
 
         return $this->phoenix->getSignupIndex($options);
--- a/app/Http/Controllers/SignupController.php
+++ b/app/Http/Controllers/SignupController.php
@@ -52,9 +52,7 @@ class SignupController extends Controller
                 $options['user'] = User::drupalIDForNorthstarId($options['user']);
             }
         }
-
-        dd($options['user']);
-
+        
         return $this->phoenix->getSignupIndex($options);
     }
 
--- a/app/Http/Controllers/SignupController.php
+++ b/app/Http/Controllers/SignupController.php
@@ -47,14 +47,13 @@ class SignupController extends Controller
         if (! empty($options['user'])) {
             if (strpos($options['user'], ',') > 0) {
                 $usersQuery = explode(',', $options['user']);
-                $options['user'] = [];
-                foreach ($usersQuery as $user) {
-                    array_push($options['user'], User::drupalIDForNorthstarId($user));
-                }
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/SignupController.php
```diff
@@ -45,7 +45,15 @@ public function index(Request $request)
 
         // If a user is specified, turn Northstar ID into Drupal ID
         if (! empty($options['user'])) {
-            $options['user'] = User::drupalIDForNorthstarId($options['user']);
+            if (strpos($options['user'], ',') > 0) {
+                $usersQuery = explode(',', $options['user']);
+                $options['user'] = [];
+                foreach ($usersQuery as $user) {
+                    array_push($options['user'], User::drupalIDForNorthstarId($user));
```
DFurnes: This will run N MongoDB queries for N users... I think it'd be a good idea to update the `User::drupalIDForNorthstarId` method to accept an array of Northstar IDs and form a single query to fetch all their equivalent Drupal IDs.

app/Http/Controllers/SignupController.php
```diff
@@ -45,7 +45,15 @@ public function index(Request $request)
 
         // If a user is specified, turn Northstar ID into Drupal ID
         if (! empty($options['user'])) {
-            $options['user'] = User::drupalIDForNorthstarId($options['user']);
+            if (strpos($options['user'], ',') > 0) {
+                $usersQuery = explode(',', $options['user']);
+                $options['user'] = [];
```
DFurnes: This seems a bit confusing at first glance... maybe using [`array_map`](http://php.net/manual/en/function.array-map.php) to do this instead of wiping the array and re-building from the temporary variable might be more clear.

DFurnes:
Left two comments. It'd also be good to add test coverage for this to [TestSignup.php](https://github.com/DoSomething/northstar/blob/dev/tests/SignupTest.php) before merging.

deadlybutter:
@codal-sara hows this?

deadlybutter:
@DFurnes hi. hows this update?

app/Http/Controllers/SignupController.php
```diff
@@ -47,15 +47,14 @@ public function index(Request $request)
...
-                foreach ($usersQuery as $user) {
-                    array_push($options['user'], User::drupalIDForNorthstarId($user));
-                }
+                $options['user'] = User::drupalIDForNorthstarId($usersQuery);
             } else {
                 $options['user'] = User::drupalIDForNorthstarId($options['user']);
             }
         }
 
+        dd($options['user']);
```
DFurnes: *cough*

deadlybutter: SHIT

deadlybutter:
look dave theres nothing to see here anymore see!

app/Http/Controllers/SignupController.php
```diff
@@ -45,9 +45,14 @@ public function index(Request $request)
 
         // If a user is specified, turn Northstar ID into Drupal ID
         if (! empty($options['user'])) {
```
DFurnes: This will only work if the client uses the `user` parameter, which I think is deprecated now. It might be good to do the same `user` → `users` transform we do in Phoenix here so we can support both.

deadlybutter: mmm yes yes

app/Http/Controllers/SignupController.php
```diff
@@ -45,9 +45,14 @@ public function index(Request $request)
...
-            $options['user'] = User::drupalIDForNorthstarId($options['user']);
+            if (strpos($options['user'], ',') > 0) {
```
DFurnes: Is this check necessary? I think you should be able to just use the first half of this conditional regardless of whether there's a `,` in the string.

DFurnes:
@deadlybutter You've also got two StyleCI errors to fix up! :pencil2: :x:

DFurnes:
Great! Can you also add a new test to  [TestSignup.php](https://github.com/DoSomething/northstar/blob/dev/tests/SignupTest.php) for this functionality?

deadlybutter:
@DFurnes any idea why the current test failed in the wercker thing?

DFurnes:
Not sure, I wish it gave more info on errors. :weary: Does that test not fail on your local machine?

tests/SignupTest.php
```diff
@@ -14,9 +14,10 @@ class SignupTest extends TestCase
     public function testSignupIndex()
     {
         $user = User::create(['drupal_id' => '100001']);
+        $user2 = User::create(['drupal_id' => '100002']);
 
         // For testing, we'll mock a successful Phoenix API response.
-        $this->mock(Phoenix::class)->shouldReceive('getSignupIndex')->with(['user' => '100001'])->once()->andReturn([
+        $this->mock(Phoenix::class)->shouldReceive('getSignupIndex')->with(['users' => '100001,100002'])->once()->andReturn([
```
DFurnes: This is being called with `['users' => ['10001, '10002']` since you separate it out into an array [in the controller](https://github.com/deadlybutter/northstar/blob/f038a99e6a2a154d695a67b10146a7b2c3eee22a/app/Http/Controllers/SignupController.php#L53-L54). Updating the mock to expect the array `->with(['users' => ['100001','100002']])` causes the tests to use your mock and pass.

deadlybutter:
@DFurnes updated test and it passes. 

also is northstar using the same style rules as gladiator?

DFurnes:
@deadlybutter I haven't had a chance to update the string concatenation rule to use spaces like we do elsewhere now, but otherwise everything should be the same.

:+1: to merge this!

PR #327, opened 2016-04-06
Allow username field when normalizing credentials.

### LINKED ISSUE #322: Add "username" to credentials flow
At the moment, we accept either an `email` or `mobile` and a `password` when authenticating users. This makes the query nice and simple, but it means each client application needs to either choose one or the other to prompt the user for, or handle "detecting" the input type before sending the auth request.

Since this is such a common use-case, we should allow a `username` parameter that handles this mapping within Northstar. For reference, here's [how we do it](https://github.com/DoSomething/phoenix/blob/bce57ab610a13428aabd71e8b6b8ce9d32846d76/lib/modules/dosomething/dosomething_user/dosomething_user.module#L942-L960) in Phoenix.

DFurnes:
#### Changes
Closes #322. Since the OAuth 2 password grant also [requires a username field](https://tools.ietf.org/html/rfc6749#section-4.3.2) (rather than `email` or `mobile` like we currently provide), this is also a pre-requisite to implementing #202.

I did a bit of cleanup along the way, but the important changes are at the top of the `normalize` method of the Registrar, which now checks for a "username" field in the user's credentials and transforms it into either an "email" or "mobile" if found.

---
For review: @angaither @weerd

### DIFF
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -80,40 +80,14 @@ class Registrar
         }
 
         if (! empty($credentials['email'])) {
-            $credentials['email'] = $this->normalizeEmail($credentials['email']);
+            $credentials['email'] = trim(strtolower($credentials['email']));
         }
 
         if (! empty($credentials['mobile'])) {
-            $credentials ['mobile'] = $this->normalizeMobile($credentials['mobile']);
+            $credentials ['mobile'] = preg_replace('/[^0-9]/', '', $credentials['mobile']);
         }
 
         return $credentials;
-    }
-
-    /**
-     * Sanitize an email address before verifying or saving to the database.
-     * This method will likely be called multiple times per user, so it *must*
-     * provide the same result if so.
-     *
-     * @param string $email
-     * @return string
-     */
-    public function normalizeEmail($email)
-    {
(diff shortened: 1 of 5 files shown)


app/Auth/Registrar.php
```diff
@@ -73,24 +73,68 @@ public function login($credentials)
...
+
+    /**
+     * Confirm that the given value is an e-mail address.
+     *
+     * @param string $value
+     * @return bool
+     */
+    protected function isEmail($value)
+    {
+        return filter_var(trim($value), FILTER_VALIDATE_EMAIL) !== false;
```
weerd: I'm a fan of PHP's `filter_var()` :)

app/Http/Controllers/AuthController.php
```diff
@@ -92,7 +107,7 @@ public function verify(Request $request)
      */
     public function invalidateToken(Request $request)
     {
-        $token = Auth::token();
+        $token = $this->auth->token();
```
weerd: Furnes hates facades, haha :japanese_ogre:

DFurnes: :smirk:

weerd:
Looks cool! :+1:

PR #328, opened 2016-04-07
Quick fix on how arrays are handled in signup controller

deadlybutter:
#### What's this PR do?
When requesting this URL via Gladiator, for some reason Northstar gets it already as an array & thus doesn't need to be exploded(). However when requesting from the browser, its a string that does need to be exploded.

`http://northstar.app/v1/signups?competition=1&users=5571df42a59dbf3b7a8b456d,5575e568a59dbf3b7a8b4572&campaigns=1485`

### DIFF
--- a/app/Http/Controllers/SignupController.php
+++ b/app/Http/Controllers/SignupController.php
@@ -50,7 +50,12 @@ class SignupController extends Controller
 
         // If a user is specified, turn Northstar ID into Drupal ID
         if (! empty($options['users'])) {
-            $usersQuery = explode(',', $options['users']);
+            if (is_array($options['users'])) {
+                $usersQuery = $options['users'];
+            } else {
+                $usersQuery = explode(',', $options['users']);
+            }
+
             $options['users'] = User::drupalIDForNorthstarId($usersQuery);
         }
 


DFurnes:
Huh, that's weird. What error do you see when making this request via Gladiator?

deadlybutter:
From gladiator 
screen shot 2016-04-07 at 10 46 11 am

From URL
screen shot 2016-04-07 at 10 46 21 am

If I point this back to NS QA, it returns a blank response like this.

screen shot 2016-04-07 at 10 47 26 am

DFurnes:
So this works correctly through Gladiator when pointed to your local Northstar, but not to Northstar QA?

deadlybutter:
Correct, after these changes it works locally

DFurnes:
Hmm, magic! :sparkles: If you can add a test case for when the request is giving an array that'd be great.

deadlybutter:
@DFurnes I think thats the current test case? Let me check

deadlybutter:
Yeah the current test case it two id's comma separated

DFurnes:
Right, I mean a failing test case for the existing code, since the current tests pass before your change. I'm having trouble understanding why ``$request->query()['users']`` would ever be an array to trigger that first conditional you added.

DFurnes:
@deadlybutter Did you find out whether this was an issue in Northstar?

deadlybutter:
oh wow i forgot about this, no this isnt needed!

PR #330, opened 2016-04-11
Rename ApiKey to Client.

DFurnes:
#### What's this PR do?
This PR renames the ApiKey model to "Client", with two aliased properties `client_id` and `client_secret` matching to the existing app ID and key. This language makes it more clear how it'll be used once we set up OAuth-based authentication for #202.

Check that everything looks okay in the diff! I renamed any classes, but will leave database migrations to rename the existing fields for a later PR. :smile: 

#### Checklist

---
For review: @weerd


app/Http/Controllers/ClientController.php
```diff
@@ -3,21 +3,21 @@
...
-     * @var ApiKeyTransformer
+     * @var ClientTransformer
      */
     protected $transformer;
 
-    public function __construct(ApiKeyTransformer $transformer)
+    public function __construct(ClientTransformer $transformer)
     {
-        $this->transformer = new ApiKeyTransformer();
+        $this->transformer = new ClientTransformer();
```
weerd: shouldn't this just be 

```php
 $this->transformer = $transformer
```

Otherwise what's the point of passing it into the construct?

DFurnes: Hmm, good point!

weerd:
¡OLÉ
:sheep: it!!

PR #333, opened 2016-04-14
Oauth Middleware

DFurnes:
#### What's this PR do?
This PR updates our custom Guard and scope middleware to accept new OAuth JWT tokens as well as our old database tokens. This will allow any protected route to be accessed using either form of token.

It also includes a tiny fix to maintain backwards compatibility with how the `v1/scopes` response is structured (since the backend array storing those was modified to support the OAuth implementation).

__Note:__ The `\Northstar\Http\Middleware\ParseOAuthHeader` middleware is commented out because it causes issues on PHP < 5.6 (see original comments below). Since this just means the request attributes will be unset, everything will continue to only check the for the "legacy" auth tokens.

I'd recommend taking a look commit-by-commit. Most of the changes are very small, although there are some bigger "re-shuffling" changes like moving scope methods/property into it's own class in cfdc684.

I'll add tests once the middleware can be uncommented (since right now it'd just fail fail fail).

#### Checklist

---
For review: @weerd @angaither

### DIFF
--- a/app/Auth/NorthstarTokenGuard.php
+++ b/app/Auth/NorthstarTokenGuard.php
@@ -59,13 +59,48 @@ class NorthstarTokenGuard extends TokenGuard implements Guard
      */
     public function getUserForToken($tokenKey)
     {
+        // If the provided token is 32 characters long, it's a legacy
+        // database token. Otherwise, it must be a JWT access token.
+        if (strlen($tokenKey) === 32) {
+            return $this->getUserFromLegacyToken($tokenKey);
+        }
+
+        return $this->getUserFromJWTAccessToken();
+    }
+
+    /**
+     * Fetch a legacy authentication token from the database to
+     * get it's corresponding user.
+     *
+     * @param $tokenKey
+     * @return User|null
+     */
+    public function getUserFromLegacyToken($tokenKey)
+    {
         $token = Token::where($this->storageKey, $tokenKey)->first();
 
         if (empty($token)) {
-            return;
+            return null;
         }
 
         return User::find($token->user_id);
+    }
+
+    /**
+     * Retrieve the user from a parsed JWT access token that
+     * was provided with the current request.
+     *
+     * @return User|null
+     */
+    public function getUserFromJWTAccessToken()
+    {
+        $id = request()->attributes->get('oauth_user_id');
+
+        if (empty($id)) {
+            return null;
+        }
+
+        return User::find($id);
     }
 
     /**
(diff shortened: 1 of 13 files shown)


DFurnes:
Hmm, so it looks like there is a problem using the PSR-7 request bridge with PHP < 5.6...

> LogicException: getContent() can only be called once when using the resource return type and PHP below 5.6.

The Wercker box and production servers are all using PHP 5.5.9, so it seems like this may be blocked from merging until we can update. References DoSomething/devops#26.

DFurnes:
Disabling the middleware resolves those issues, so I can continue working on this in the meantime.

DFurnes:
Alright, this is ready for review! If testing locally, the `\Northstar\Http\Middleware\ParseOAuthHeader` middleware can be un-commented to parse any provided JWT access tokens. Without that middleware, the OAuth request attributes will always be empty and so the "legacy" code will be used.

app/Auth/NorthstarTokenGuard.php
```diff
@@ -101,15 +136,22 @@ public function validate(array $credentials = [])
...
      */
     public function token()
     {
         $tokenKey = $this->getTokenForRequest();
 
-        return Token::where($this->storageKey, $tokenKey)->first();
+        // If the provided token is 32 characters long, it's a legacy
+        // database token. Otherwise, it must be a JWT access token,
+        // which does not have a corresponding database record.
+        if (strlen($tokenKey) === 32) {
```
angaither: I don't know the right laravel-y way to do this, but can it be a part of the model, or a helper to check if legacy token? I feel like this `if` will get repeated all over the damn house

app/Http/Controllers/Traits/FiltersRequests.php
```diff
@@ -67,7 +67,7 @@ public function search($query, $searches, $indexes)
         }
 
         // Only "admin" keys should be able to search
-        Client::gate('admin');
+        Scope::gate('admin');
```
angaither: `scope::gate` makes me chuckle

app/Http/Middleware/ParseOAuthHeader.php
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /**
+     * Is this a legacy database token? If so, we don't want to try to
+     * parse it as a JWT access token.
+     *
+     * @param $header
+     * @return bool
+     */
+    public function isLegacyToken($header)
```
angaither: oh look, it's here.

app/Auth/NorthstarTokenGuard.php
DFurnes: I don't think it will, since all the token "parsing" should take place inside the Guard or the OAuth parsing middleware. Everything else should just be using the Auth facade, like `Auth::user` and `Auth::check`, etc.

app/Http/Controllers/Traits/FiltersRequests.php
DFurnes: 🐐

app/Auth/NorthstarTokenGuard.php
angaither: it's here [three](https://github.com/DoSomething/northstar/pull/333/files#diff-a3de2283b08a5b37e8a882a4b4012b86R150) [times](https://github.com/DoSomething/northstar/pull/333/files#diff-a3de2283b08a5b37e8a882a4b4012b86R64) [already](https://github.com/DoSomething/northstar/pull/333/files#diff-3a9c47323ed8df2ce9f0370e9ee1e6d0R65)

DFurnes: Only in the two files I mentioned! I definitely don't think it's a bad idea to pull it into a helper function, but if we're using it outside of those two classes then I think we've made a mistake.

angaither:
👍

PR #331, opened 2016-04-11
Add OAuth endpoints using League oauth-server package.

DFurnes:
#### What's this PR do?
This PR introduces OAuth authorization endpoints using a v5 release candidate of the League's [oauth2-server](http://oauth2.thephpleague.com) package. It also introduces a new `v2` namespace, where the new OAuth versions of the authentication endpoints will live.

There's three major components things that are worth taking a peek at here:
 - The [__AuthServiceProvider__](https://github.com/DFurnes/northstar/blob/league-oauth-v5/app/Providers/AuthServiceProvider.php#L55-L79) now configures the OAuth server so it can be resolved from the service container elsewhere in our application. It's responsible for registering repositories (how the package knows how to store/retrieve data) and enabling & configuring which grants we're going to use.
 - The [__repositories__](https://github.com/DFurnes/northstar/tree/league-oauth-v5/app/Auth/Repositories) allow us to customize how the OAuth package talks to the database.
 - The [__entities__](https://github.com/DFurnes/northstar/tree/league-oauth-v5/app/Auth/Entities) are returned by the OAuth repositories... they're our Eloquent models translated to a format the OAuth package expects.

References #202.

#### How should this be reviewed?
screen shot 2016-04-13 at 12 01 55 pm

...mwahahah!

Okay, but really it's mostly scaffolding to glue together our models and the OAuth server! I'd recommend just skimming the whole diff. I also added tests and documentation for each of the new endpoints. This PR does _not_ handle validating JWTs or include middleware for these new tokens.

For the time being, the new endpoints are only available in debug mode so that we can safely continue developing these features without creating a massive feature branch. :construction_worker: 

#### Checklist
- [x] Documentation added for changed endpoints.
- [x] Tests added for new features/bug fixes.

---
For review: @weerd

PR #336, opened 2016-04-18
Database updates for OAuth clients table.

DFurnes:
#### What's this PR do?
This PR includes database updates for the changes made in #331 and #333.

It renames the `api_keys` table to `clients` since "API keys" are now being used as OAuth 2 clients, and renames the `app_id` and `api_key` fields on each of those documents to `client_id` and `client_secret` so that the corresponding database queries make more sense.

Since we need to maintain backwards compatibility for the "legacy" authentication flow, I added mutators to "map" the old `app_id` and `api_key` fields to the new database fields. I updated database calls (like seeding test clients and verifying database records) in the test suite, but otherwise left all the functional aspects of the tests unchanged.

I'd recommend checking things out commit-by-commit. There's a lot of little changes scattered about for renaming the database columns, but otherwise it's all small changes. 😇 

#### Checklist

---
For review: @angaither @weerd

### DIFF
--- a/app/Auth/Entities/ClientEntity.php
+++ b/app/Auth/Entities/ClientEntity.php
@@ -24,9 +24,9 @@ class ClientEntity implements ClientEntityInterface
      */
     public function __construct($client_id, $scopes)
     {
-        $this->name = $client_id;
+        $this->name = $client_id; // @TODO: If we store a human-readable client name, use here.
         $this->allowedScopes = $scopes;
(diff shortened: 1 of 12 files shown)


angaither:
clients still send `X-DS-REST-API-Key` & `X-DS-Application-Id`?

DFurnes:
Yep, old auth is still supported. 👌

angaither:
👍

Issue #334, opened 2016-04-18
Fix duplicate Drupal ID records.


DFurnes:
It seems like when a user changes their email in Phoenix, they end up with a new account created in Northstar (one with the old email and their Phoenix ID, and one with the new email and their Phoenix ID). This is a bit weird because we specifically [update the user](https://github.com/DoSomething/phoenix/blob/5e10294012f91ebbe7915f48f27db244810ae9fc/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module#L136-L158) by their Phoenix ID when sending the new fields to Northstar... but @weerd has found a few cases where it's occurred.

On the Northstar side, it's probably a good idea to add a unique index to this field & some better validation to throw a 422 if a client app tries to do this.

DFurnes:
This seems to have occurred for only 20 users on Northstar QA, and 19* users on Northstar production. 

*Northstar production _also_ has 126,957 users that are counted as having "duplicate" Drupal IDs because the field is saved with `null` for it's value. I'm not 100% sure why this is but interestingly it hasn't seem to occur on the QA server! 💭 

For reference, the dupes query:

```js
db.users.aggregate(
  [
    {
      $group: {
        _id: { drupal_id: "$drupal_id" },
        uniqueIds: { $addToSet: "$_id" },
        count: { $sum: 1 }
      }
    },
    { 
      $match: {
        count: { $gt: 1 }
      }
    }
  ],
  {
    allowDiskUse: true
  }
);

```

DFurnes:
Thinking more on the Northstar QA data, though, I think there's a good chance a lot of those records will be unreliable since all of our dev environments are hooked up to northstar-qa.dosomething.org (and so any users registered on our local environments would all share overlapping sequential Drupal UIDs, regardless of what the data "really" is on staging.dosomething.org)...

DFurnes:
Hmm, tried running the modified "dedupe" script on Northstar QA. It worked for a couple records but seems to have hit a brick wall where it somehow creates a new duplicate mobile number for a merged user and dies on a constraint violation. May have to do some Mongo diving with @angaither or @sheyd on Monday. 🏊

DFurnes:
Gonna try picking this back up today after we had a few more unsynced user accounts (due to having multiple "hidden" accounts with that Drupal ID). MongoDB sure doesn't make it easy. 🌾

DFurnes:
[...]

PR #337, opened 2016-04-18
Fix users getting duplicate Drupal IDs.

DFurnes:
#### What's this PR do?
This PR fixes an issue where users with duplicate `drupal_id`s could be created (for example, when changing an email on an account in Phoenix) because we weren't using those fields to resolve and update existing accounts when POSTing to `users/`.

It also includes two clean-up commits: fixing an incorrect docblock (365af08), and removing the "key" alias to the renamed "scope" middleware (365af08).

References #334.

#### Before deploying this...
We'll want to manually clean up any duplicate `drupal_id`s before merging this, since otherwise those accounts would become "unwritable" since validation would always fail. ~~Probably a good idea to repurpose @chloealee's dedupe script?~~

__Update:__ I updated the dedupe script in 2e67a77, and tested it on a few fake "dupes" I had made before applying these fixes. Should be safe to run again on QA/production once this merges!

I'd recommend skimming the whole diff, since I ended up changing my approach a bit partway through. The changes made in 365af08 add a bit of noise, but otherwise it's all pretty straightforward.

#### Checklist

---
For review: @angaither @weerd @chloealee

### DIFF
--- a/app/Console/Commands/RemoveDuplicateUsersCommand.php
+++ b/app/Console/Commands/RemoveDuplicateUsersCommand.php
@@ -44,9 +44,9 @@ class RemoveDuplicateUsersCommand extends Command
                 [
                     [
                         '$match' => [
-                                'email' => [
-                                    '$ne' => '',
-                                ],
+                            'email' => [
+                                '$ne' => '',
+                            ],
                         ],
                     ],
                     [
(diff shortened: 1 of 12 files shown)


angaither:
👍

app/Console/Commands/RemoveDuplicateUsersCommand.php
```diff
@@ -120,6 +120,49 @@ public function fire()
...
         $this->deduplicate($mobile_duplicates, 'mobile');
+
+        // Find all duplicate users by Drupal ID.
+        $drupal_duplicates = User::raw(function ($collection) {
+            return $collection->aggregate(
+                [
+                    [
+                        '$match' => [
+                            'drupal_id' => [
+                                '$ne' => '',
```
weerd: what's a `ne`?

DFurnes: I _think_ "not equal"? I'm not gonna lie, I just copy-pasted from the query Chloe had used above. 😇

weerd:
Coolio!
:sheep: it!

app/Console/Commands/RemoveDuplicateUsersCommand.php
weerd: 👌

PR #343, opened 2016-04-19
adds all information for user

### LINKED ISSUE #340: Add user property in the Northstar proxy to "append" it to each signup
To avoid multiple requests to northstar from phoenix, add user property in the Northstar proxy to "append" it to each signup. 

Refs conversation [here](https://github.com/DoSomething/phoenix/pull/6381).

chloealee:
#### What's this PR do?
Adds the user property in the Northstar signup proxy, in order to remove user properties (besides `id`) in [$this->user business in the phoenix](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_signup/includes/Signup.php#L111-L115) and then "append" it to each signup in the result in the Northstar proxy method (since then it could happen as a single database query, rather than multiple HTTP requests).

Refs phoenix updates [here](https://github.com/DoSomething/phoenix/pull/6381). 

On my local, added an extra line to test, and all information was returned as expected. 
```
        foreach ($results['data'] as $key => $result) {
            $drupal_id = array_get($result, 'user.drupal_id');
            $user = User::where('drupal_id', $drupal_id)->first();

            $results['data'][$key]['user'] = [
                'id' => $user ? $user->id : null,
                'first_name' => $user ? $user->first_name : null,
                'last_initial' => $user ? $user->last_initial : null,
                'photo' => $user ? $user->photo : null,
                'country' => $user ? $user->country : null,
                'test' => 'test',
            ];
        }
        print_r($results);
        die();
```
screen shot 2016-04-25 at 12 04 07 pm

All tests should pass when running `vendor/bin/phpunit`

#### Checklist

Do we need the above @DFurnes ?

Fixes #340

### DIFF
--- a/app/Http/Controllers/SignupController.php
+++ b/app/Http/Controllers/SignupController.php
@@ -56,21 +56,16 @@ class SignupController extends Controller
 
         $results = $this->phoenix->getSignupIndex($options);
 
-        foreach (array_slice($results, 1) as $result) {
-            $northstar_id = $result[0]['user']['id'];
-            $user = User::where('_id', $northstar_id)->first();
-            $last_initial = substr(($user->last_name), 0, 1);
+        foreach ($results['data'] as $key => $result) {
+            $drupal_id = $result['user']['drupal_id'];
+            $user = User::where('drupal_id', $drupal_id)->first();
 
-            if ($last_initial === false) {
-                $last_initial = null;
-            }
-
-            $result[0]['user'] = [
-                'first_name' => $user->first_name,
-                'last_initial' => $last_initial,
-                'photo' => isset($user) ? $user->photo : null,
-                'country' => $user->country,
-                'test' => 'test',
+            $results['data'][$key]['user'] = [
+                'id' => $user ? $user->id : null,
+                'first_name' => $user ? $user->first_name : null,
+                'last_initial' => $user ? $user->last_initial : null,
+                'photo' => $user ? $user->photo : null,
+                'country' => $user ? $user->country : null,
             ];
         }
 
--- a/app/Http/Controllers/SignupController.php
+++ b/app/Http/Controllers/SignupController.php
@@ -57,7 +57,7 @@ class SignupController extends Controller
             $options['users'] = User::drupalIDForNorthstarId($usersQuery);
 
             $query = $this->newQuery(User::class);
-            
+
             // For the first `where` query, we want to limit results... from then on,
             // we want to append (e.g. `SELECT * WHERE _ OR WHERE _ OR WHERE _`)
             $firstWhere = true;
@@ -71,11 +71,11 @@ class SignupController extends Controller
         }
 
         $results = $this->phoenix->getSignupIndex($options);
-        
+
         foreach ($results['data'] as $key => $result) {
             $drupal_id = array_get($result, 'user.drupal_id');
             $user = $users->get($drupal_id);
-            
+
             // If Phoenix gave the expected drupal_id in the user response, replace it with our own data.
             if (! empty($user)) {
                 $results['data'][$key]['user'] = [
@@ -86,7 +86,6 @@ class SignupController extends Controller
                     'country' => $user->country,
                 ];
             }
-            
         }
 
         return $results;
--- a/app/Http/Controllers/SignupController.php
+++ b/app/Http/Controllers/SignupController.php
@@ -54,7 +54,27 @@ class SignupController extends Controller
             $options['users'] = User::drupalIDForNorthstarId($usersQuery);
         }
 
-        return $this->phoenix->getSignupIndex($options);
+        $results = $this->phoenix->getSignupIndex($options);
+
+        foreach (array_slice($results, 1) as $result) {
+            $northstar_id = $result[0]['user']['id'];
+            $user = User::where('_id', $northstar_id)->first();
+            $last_initial = substr(($user->last_name), 0, 1);
+
+            if ($last_initial === false) {
+                $last_initial = null;
+            }
+
+            $result[0]['user'] = [
+                'first_name' => $user->first_name,
+                'last_initial' => $last_initial,
+                'photo' => isset($user) ? $user->photo : null,
+                'country' => $user->country,
+                'test' => 'test',
+            ];
+        }
+
+        return $results;
     }
 
     /**
--- a/app/Http/Controllers/SignupController.php
+++ b/app/Http/Controllers/SignupController.php
@@ -193,6 +193,8 @@ class SignupController extends Controller
     }
 
     /**
+     * Queries for all specified drupal users at once.
+     *
      * @param array $drupalIds
      * @return User
      */
--- a/app/Http/Controllers/SignupController.php
+++ b/app/Http/Controllers/SignupController.php
@@ -80,7 +80,7 @@ class SignupController extends Controller
 
         foreach ($results['data'] as $key => $result) {
             $drupal_id = array_get($result, 'user.drupal_id');
-            $user = $users->get($drupal_id);
+            $user = User::where('drupal_id', $drupal_id)->first();
 
             // If Phoenix gave the expected drupal_id in the user response, replace it with our own data.
             if (! empty($user)) {
--- a/app/Http/Controllers/SignupController.php
+++ b/app/Http/Controllers/SignupController.php
@@ -48,40 +48,30 @@ class SignupController extends Controller
             $options['users'] = $options['user'];
         }
 
+        
+
         // If a user is specified, turn Northstar ID into Drupal ID
         if (! empty($options['users'])) {
             // Update the '?users=xxx,xxx,xxx' option to be based on Drupal ID, rather than Northstar ID
             $usersQuery = explode(',', $options['users']);
             $options['users'] = User::drupalIDForNorthstarId($usersQuery);
 
-            $query = $this->newQuery(User::class);
-
-            // For the first `where` query, we want to limit results... from then on,
-            // we want to append (e.g. `SELECT * WHERE _ OR WHERE _ OR WHERE _`)
-            $firstWhere = true;
-            foreach ($options['users'] as $drupal_id) {
-                $query->where('drupal_id', '=', $drupal_id, ($firstWhere ? 'and' : 'or'));
-                $firstWhere = false;
-            }
+            $users = $this->usersForDrupalIds($options['users']);
 
             $results = $this->phoenix->getSignupIndex($options);
         } else {
             // Since we are not specifying a NS user, grab results from Phoenix first and get Drupal ID from response to use in query below to find user. 
             $results = $this->phoenix->getSignupIndex($options);
-            $query = $this->newQuery(User::class);
-            $users = collect($results['data'])->pluck('user.drupal_id');
 
-            foreach ($users as $key => $drupal_id) {
-                $query->where('drupal_id', '=', $drupal_id);
-            }
+            $drupalIds = collect($results['data'])->pluck('user.drupal_id');
+
+            $users = $this->usersForDrupalIds($drupalIds);
         }
 
-        // Make an array keyed by the Drupal ID... e.g. ['100010' => {User}, '10002' => {User}]
-        $users = $query->get()->keyBy('drupal_id');
-
+        // Now, fill in the 'user' object on the response before returning it.
         foreach ($results['data'] as $key => $result) {
             $drupal_id = array_get($result, 'user.drupal_id');
-            $user = User::where('drupal_id', $drupal_id)->first();
+            $user = $users->where('drupal_id', $drupal_id)->first();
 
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/SignupController.php
```diff
@@ -54,7 +54,27 @@ public function index(Request $request)
             $options['users'] = User::drupalIDForNorthstarId($usersQuery);
         }
 
-        return $this->phoenix->getSignupIndex($options);
+        $results = $this->phoenix->getSignupIndex($options);
+
+        foreach (array_slice($results, 1) as $result) {
+            $northstar_id = $result[0]['user']['id'];
```
chloealee: retesting this, a lot of the users are coming up with the error message that `the resource does not exist`. From the phoenix side though, this looks like the [northstar id](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_signup/includes/Signup.php#L110) so not sure why this would be sending incorrect data back (unless the two databases on my local are out of sync?).

chloealee: actually - I think that it does have something to do with the databases not synced up because I tried hitting both northstar's `/v1/users/drupal_id/:drupal_id` and `/v1/users/_id/:id` with some the same id and `resource does not exist` popped up for both.

DFurnes:
> - [ ] Documentation added for changed endpoints.
> - [ ] Tests added for new features/bug fixes.
> Do we need the above @DFurnes ?

Yes! There's some [existing tests](https://github.com/DoSomething/northstar/blob/caf65300ecd904fbc004c59f92d8e3a23b901096/tests/SignupTest.php#L8-L36) for these endpoints that we should update to include mocked `user.drupal_id` properties (and then validate that it produces the right changes to the response). Let me know if you want to sit down and walk through this together! 😄

It'd also be good to update the [documentation](https://github.com/DoSomething/northstar/blob/dev/documentation/endpoints/signups.md) to mention that it also builds a full `user` response for this endpoint, since that's not a part of the response that the "raw" Phoenix API would give.

chloealee:
thanks for explaining all of this, @DFurnes ! I'm working on the test right now and I think there might be something wrong with my configuration or something I'm missing on my local perhaps? 

Even when commenting out any new changes I've added, a few of the last tests are not passing with the first error `.................FFFFFPHP Fatal error:  Class 'Lcobucci\JWT\Parser' not found in /home/vagrant/DoSomething.org/northstar/tests/OAuthTest.php on line 130`

Have you seen this before?

DFurnes:
@chloealee Hmm, have you run `composer install` lately?

chloealee:
oh that seemed to be it, @DFurnes! Thank you!

app/Http/Controllers/SignupController.php
```diff
@@ -54,7 +54,22 @@ public function index(Request $request)
             $options['users'] = User::drupalIDForNorthstarId($usersQuery);
         }
 
-        return $this->phoenix->getSignupIndex($options);
+        $results = $this->phoenix->getSignupIndex($options);
+
+        foreach ($results['data'] as $key => $result) {
+            $drupal_id = array_get($result, 'user.drupal_id');
+            $user = User::where('drupal_id', $drupal_id)->first();
```
DFurnes: This will query the database N times for N users, which could get kinda slow. Let's change this to fetch all the users using a single query and then loop through those results to update the response. You should be able to do this pretty easily by using the "query" methods that built in to our base controller:

```php
$query = $this->newQuery(User::class);
$query = $this->filter($query, $options['users'], User::$indexes);
$users = $query->get();
```

DFurnes: Make sure you run those queries _before_ turning the Northstar IDs into Drupal IDs, btw!

DFurnes: If you run into problems with that, it also wouldn't be a bad idea to build the query manually [like we do in that method](https://github.com/DoSomething/northstar/blob/54f32acfc3e0d181fad9ce41b0f5c99c054d3c6c/app/Http/Controllers/Traits/FiltersRequests.php#L43-L49).

chloealee:
Thanks for the help, @DFurnes ! This is ready for final review.

chloealee:
I lied before, @DFurnes - _now_ it is ready for another review! 

When testing `/signups`, I received this error:
```
{
  "error": {
    "code": 500,
    "message": "Call to a member function get() on array"
  },
  "debug": {
    "file": "/home/vagrant/DoSomething.org/northstar/app/Http/Controllers/SignupController.php",
    "line": 77
  }
}
```

because `$users` only exists if you pass user id as params. I added 

```
        } else {
            // Get all drupal ids at once for index
            $query = $this->newQuery(User::class);
            $users = $query->select('drupal_id')->get()->keyBy('drupal_id');
        }
``` 
to only query for all users at once if `users` param is empty.

app/Http/Controllers/SignupController.php
```diff
@@ -48,13 +48,51 @@ public function index(Request $request)
...
+            $firstWhere = true;
+            foreach ($options['users'] as $drupal_id) {
+                $query->where('drupal_id', '=', $drupal_id, ($firstWhere ? 'and' : 'or'));
+                $firstWhere = false;
+            }
+
+            // Make an array keyed by the Drupal ID... e.g. ['100010' => {User}, '10002' => {User}]
+            $users = $query->get()->keyBy('drupal_id');
+        } else {
+            // Get all drupal ids at once for index
```
DFurnes: Ohhhh, yeah that's a good point. Let's move the user-fetching code _after_ we get the response from Phoenix so we can specifically query for those `drupal_id`s that are included in the response.

DFurnes: I _think_ this could be done like `$users = collect($results)->pluck('data.*.user.drupal_id')` down on line 78.

chloealee: @DFurnes instead of querying the database, do you mean pulling out all of the `drupal_ids` and putting it into an array, with drupal_id as the keys, and to search for when using `$user = $users->get($drupal_id);`?

chloealee: @DFurnes ah I see - sorry I didn't refresh!

chloealee: @DFurnes I might not be understanding correctly but I don't _think_ we can use something like `pluck` because that will just return an array of the `drupal_ids` when really, we want to grab all of the user _objects_ all at once, in able to fill the properties out in `results['data'][$key]['user']` - is this right?

DFurnes: I think `pluck` will work! 

If we're not querying for specific users, we'll want to get an array of the Drupal IDs that are returned in the Phoenix response so that we can then query for those specific IDs, so step 1 is to get the array of IDs, and then step 2 is to query the database based on that array.

chloealee: ahhh I see now - thanks @DFurnes ! I added the `pluck` in but also had to use `$results = $this->phoenix->getSignupIndex($options);` in two places - not sure if this makes it messy but would love your thoughts!

app/Http/Controllers/SignupController.php
```diff
@@ -50,11 +50,51 @@ public function index(Request $request)
...
+            // we want to append (e.g. `SELECT * WHERE _ OR WHERE _ OR WHERE _`)
+            $firstWhere = true;
+            foreach ($options['users'] as $drupal_id) {
+                $query->where('drupal_id', '=', $drupal_id, ($firstWhere ? 'and' : 'or'));
+                $firstWhere = false;
+            }
+
+            $results = $this->phoenix->getSignupIndex($options);
+        } else {
+            $results = $this->phoenix->getSignupIndex($options);
```
DFurnes: It'd be nice to add a comment here explaining why this case needs to fetch the Phoenix result first and then translate the returned Drupal IDs into Northstar IDs _afterwards_, whereas the first case translates the explicitly requested ones _before_ making the Phoenix request.

(I think the way this is handled is just fine btw, but it does read a little weird at first glance!)

app/Http/Controllers/SignupController.php
```diff
@@ -50,11 +50,51 @@ public function index(Request $request)
...
+            }
         }
 
-        return $this->phoenix->getSignupIndex($options);
+        // Make an array keyed by the Drupal ID... e.g. ['100010' => {User}, '10002' => {User}]
+        $users = $query->get()->keyBy('drupal_id');
+
+        foreach ($results['data'] as $key => $result) {
+            $drupal_id = array_get($result, 'user.drupal_id');
+            $user = User::where('drupal_id', $drupal_id)->first();
```
DFurnes: We should be able to re-use the original query that you made above when requesting an explicit group of users via the query string `'?users=xxx,xxx,xxx'`. For the other case, it'd be nice to do the same `SELECT * WHERE _ OR WHERE _ OR WHERE _` down here so we're not making a separate query for each user.

tests/SignupTest.php
```diff
@@ -13,26 +13,99 @@ class SignupTest extends TestCase
...
+                        'drupal_id' => '100003',
+                    ],
+                ],
+            ],
+        ]);
+
+        $response = $this->asUser($user)->withScopes(['user'])->get('v1/signups?users='.$user->_id);
+        $this->assertResponseStatus(200);
+        $this->seeJson();
+        $json_decoded_response = $this->decodeResponseJson();
```
DFurnes: This should be 🐫-case in Laravel! (Or you could inline it in the statement below since it's only used once!)

DFurnes:
Do we need to do the same logic for the `user` response on the "show" `/signup/:id` endpoint as well?

chloealee:
@DFurnes great call - added the logic in for `/signup/:id`!

app/Http/Controllers/SignupController.php
```diff
@@ -95,7 +124,22 @@ public function profile(Request $request)
      */
     public function show($signup_id)
     {
-        return $this->phoenix->getSignup($signup_id);
+        $result = $this->phoenix->getSignup($signup_id);
+
+        $user = User::where('drupal_id', $result['data']['user']['drupal_id'])->first();
```
DFurnes: What happens here if the Phoenix response 404s?

app/Http/Controllers/SignupController.php
```diff
@@ -147,4 +191,23 @@ public function store(Request $request)
         // If we successfully created signup, return "show" response w/ 201
         return response()->json($signupResponse, 201);
     }
+
+    /**
+     * @param array $drupalIds
```
DFurnes: Can you add a quick description to this docblock? 😇

app/Http/Controllers/SignupController.php
chloealee: ah good point! Would it then just return a 404 as the response since it returns the result of the phoenix call in line 142? Or are you saying more so what happens if the the call is unsuccessful to phoenix?

DFurnes: Right. I haven't tested this, but I think it would crash since the `$result` from the Phoenix API call wouldn't have a `['data']['user']['drupal_id']` index. Wrapping all this logic inside an `array_has($result, 'data.user.drupal_id')` check would keep this safe.

chloealee: I tested quickly on my local by hardcoding `$result = $this->phoenix->getSignup('z');` and got this error: 
```
{
  "error": {
    "code": 404,
    "message": "That signup could not be found."
  },
  "debug": {
    "file": "/home/vagrant/DoSomething.org/northstar/app/Services/Phoenix.php",
    "line": 200
  }
}
```
but I still think you're right that it doesn't hurt to add a check - better safe than sorry! I'll add now :)

DFurnes: Oh, sweet! 👍

In that case you're probably good to ship, but yeah no harm in adding a check!

DFurnes:
🚢 ⛵ 🚀

app/Http/Controllers/SignupController.php
chloealee: oh wait! I added this code to check and looks like it dies right away after the call to phoenix if a `404` is hit because the same error message is returned instead of `bye`
```
    public function show($signup_id)
    {
        $result = $this->phoenix->getSignup('z');

        if (array_has($result, 'data.user.drupal_id')) {
            print_r('hi');
            die();
            $user = User::where('drupal_id', $result['data']['user']['drupal_id'])->first();

            // If Phoenix gave the expected drupal_id in the user response, replace it with our own data.
            if (! empty($user)) {
                $result['data']['user'] = [
                    'id' => $user->id,
                    'first_name' => $user->first_name,
                    'last_initial' => $user->last_initial,
                    'photo' => $user->photo,
                    'country' => $user->country,
                ];
            }
        }
        print_r('bye');
        die();
        return $result;
    }
```
is it okay to ship without the `array_has($result, 'data.user.drupal_id')` check?

DFurnes: Yeah I think that sounds fine! 👍

PR #346, opened 2016-04-25
Fix issue with upsert changing records that it shouldn't be.

### LINKED ISSUE #345: User can be "upserted" with a changed email
Still investigating this, but I _seem_ to have been able to "upsert" my existing account to have a different email by registering on Four Legged Finishers with: my real phone number (which is attached to my account), but a _different_ email (in this case, my email with a `+1` on the email).

DFurnes:
#### What's this PR do?
Fixes #345. This pull request addresses a bug with the `POST users/` endpoint. Currently, clients that use the endpoint can upsert an existing index for an account, which is probably _not_ expected behavior.

For example, when submitting `{email: 'test+1@dosomething.org', mobile: '5551234455'}`, it would make sense to upsert a user with only `{mobile: '5551234455'}` and no `email` set (so they become `{email: 'test+1@dosomething.org', mobile: '5551234455'}`).

BUT it would be confusing and unexpected to _change_ the user's email if it resolves to a user that already has a different email set... so when there's a user with `{email: 'test@dosomething.org', mobile: '5551234455'}`, submitting `{email: 'test+1@dosomething.org', mobile: '5551234455'}` would CHANGE that existing user's email to the provided `test+1@dosomething.org`. 😕 

Luckily, this can only occur from "admin" clients but we should definitely fix this as soon as possible.

Check the failing test case added in the first commit, and the changes made to fix it.

#### Checklist

---
For review: @angaither

### DIFF
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -158,10 +158,17 @@ class Registrar
         if ($user) {
             $fields = array_merge($user->toArray(), $fields);
 
-            // Makes sure we can't "upsert" a record to have a changed index if set.
+            // Makes sure we can't "upsert" a record to have a changed index if already set.
             // @TODO: There must be a better way to do this...
             foreach (User::$indexes as $index) {
                 if ($request->has($index) && ! empty($user->{$index}) && $fields[$index] !== $user->{$index}) {
+                    app('stathat')->ezCount('upsert conflict');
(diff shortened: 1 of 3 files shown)


angaither:
👍  do we want to log these attempts?

DFurnes:
Yeah, I think that's a good idea.

angaither:
👍

PR #347, opened 2016-04-26
Fix some more issues with updating indexes...

DFurnes:
#### What's this PR do?
The changes made in #345 made it so indexes could never be changed once set, which was not the intended behavior. 😞 This PR adds a test to ensure we _can_ change an index when explicitly updating a record via `PUT users/:term/:id`, and then makes the test pass by moving that validation check into the `POST users/` controller endpoint so it only applies to the "upsert" action.

This PR also cleans up how we format _422 Unprocessable Entity_ responses because this was happening all over the place! Now anytime we want to report a validation error, we just create a `NorthstarValidationException` which then gets returned as a HTTP response in the `Handler`.

These are all relatively small independent changes, so I'd recommend reviewing things commit-by-commit for this pull request! 👓 

#### Checklist

---
For review: @angaither @weerd

### DIFF
--- a/app/Http/Controllers/UserController.php
+++ b/app/Http/Controllers/UserController.php
@@ -2,12 +2,12 @@
 
 namespace Northstar\Http\Controllers;
 
+use Illuminate\Http\JsonResponse;
 use Illuminate\Http\Request;
 use Northstar\Auth\Registrar;
 use Northstar\Http\Transformers\UserTransformer;
 use Northstar\Services\Phoenix;
 use Northstar\Models\User;
-use Symfony\Component\HttpKernel\Exception\HttpException;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/UserController.php
```diff
@@ -95,13 +96,7 @@ public function store(Request $request)
...
                 ]);
 
-                return new JsonResponse([
-                    'error' => [
-                        'code' => 422,
-                        'message' => 'Failed validation.',
-                        'fields' => [$index => 'Cannot upsert an existing index.'],
-                    ],
-                ], 422);
+                throw new NorthstarValidationException([$index => 'Cannot upsert an existing index.']);
```
weerd: Such DRY!

weerd:
👍

weerd:
still 👍

DFurnes:
🙇

PR #348, opened 2016-05-03
Don't seed fake Drupal IDs or sources.

DFurnes:
#### What's this PR do?
No longer create fake `drupal_id` or `source` values when seeding the database, since this just makes it confusing when those don't match up with real corresponding data in other services.

If you run the seeder, it shouldn't have errors & shouldn't seed random values for those fields. 💁 

#### Checklist

---
For review: @angaither

### DIFF
--- a/database/factories/ModelFactory.php
+++ b/database/factories/ModelFactory.php
@@ -21,7 +21,6 @@ $factory->define(Northstar\Models\User::class, function (Faker\Generator $faker)
         'addr_zip' => $faker->postcode,
         'country' => $faker->countryCode,
         'language' => $faker->languageCode,
-        'drupal_id' => $faker->numberBetween(1, 400000),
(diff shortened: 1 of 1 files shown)


angaither:
@DFurnes does this just set `drupal_id` to `null` for seeded users?

DFurnes:
Yep, it won't save a `drupal_id` to those documents since we don't include it in the attributes specified on the factory. (MongoDB is weird and treats `null` and "not having a field" differently, but yeah 😝)

angaither:
👍 

I wonder if we do need that in the seed though, like some people should have real phoenix ids?

DFurnes:
Yeah that's definitely useful for testing, but I don't think we can seed that without knowing the corresponding profiles in the associated Drupal instance.

DFurnes:
For context, I noticed this when testing the Phoenix export script on my local and having the "real" data from my Phoenix instance conflict with the existing "fake" IDs that Faker had created.

PR #356, opened 2016-05-13
Don't enforce arbitrary max-length for email.

### LINKED ISSUE #355: Revisit email length restriction
In response to #240, we added a length restriction on emails (specifically, that they have to be less than 60 characters long) because Drupal was refusing to register users with longer emails. Strangely, we're now getting validation errors when trying to migrate users from Drupal who have emails _longer_ than that!

I just tested registering a user with a whopping 76-character long email on staging and it seemed to work just fine, including signing up and reporting back for a campaign. This user, however, triggers a validation error when trying to forward it to Northstar.

```
loremipsumdolorsitametconsecteturadipiscingelitduisut1234567890b@example.com
```

Perhaps the 60-character restriction has been increased to 76 characters? 😕

DFurnes:
#### What's this PR do?
Fixes #355. We were arbitrarily limiting emails to 60 characters, even though it turns out Drupal uses the exact same email validation logic as Laravel. This PR removes that restriction so we can once again sync users who have really long emails.

👀 

#### Checklist

---
For review: @angaither

### DIFF
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -147,7 +147,7 @@ class Registrar
 
         $existingId = isset($user->id) ? $user->id : 'null';
         $rules = [
-            'email' => 'email|max:60|unique:users,email,'.$existingId.',_id|required_without:mobile',
+            'email' => 'email|unique:users,email,'.$existingId.',_id|required_without:mobile',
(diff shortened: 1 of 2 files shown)


angaither:
👍

PR #359, opened 2016-05-26
Updating Jenssegers\Mongo to version 3

blisteringherb:
#### What's this PR do?
This upgrades the Jenssegers\Mongodb package to version 3 so that it's compatible with PHP7.0. 

Make sure the files look reasonably sane, and if you have the resources, update your local env to PHP7 and test.

For review: …

* v2 of Jenssegers\Mongo doesn't support PHP 7.0, so updating those references according to the documentation here https://github.com/jenssegers/laravel-mongodb#upgrading

### DIFF
--- a/app/Models/Client.php
+++ b/app/Models/Client.php
@@ -3,7 +3,7 @@
 namespace Northstar\Models;
 
 use Illuminate\Support\Str;
-use Jenssegers\Mongodb\Model;
+use Jenssegers\Mongodb\Eloquent\Model;
 
 /**
  * The Client model. These identify the "client application" making
--- a/app/Models/RefreshToken.php
+++ b/app/Models/RefreshToken.php
@@ -2,7 +2,7 @@
 
 namespace Northstar\Models;
 
-use Jenssegers\Mongodb\Model;
+use Jenssegers\Mongodb\Eloquent\Model;
 
 /**
  * The OAuth refresh token model.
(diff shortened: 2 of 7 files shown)


DFurnes:
@blisteringherb It looks like we might need to update the Wercker box's Mongo extension?

DFurnes:
@blisteringherb I did a bit more digging this morning, and I think the build error may just be because the PHP Mongo extension is set to install _after_ the Composer step, so it's not able to satisfy that new dependency in the latest release of the package.

So I think a quick fix is probably to move that above the Composer step in the Wercker config (or move it in to the box itself). We should also probably update the box to be running PHP 7 if that's what's running on QA now, since otherwise the tests are not gonna be all that reliable.

blisteringherb:
I updated the box to have PHP 7 with mongo and Node 4.4. Unfortunately the PHPUnit tests are still failing. Do you know where the MongoDB\Database::execute() might be getting called?  https://app.wercker.com/#buildstep/5748af35af0b1c312f0255c1

DFurnes:
@blisteringherb Check out `2016_04_15_192326_RenameClientTable.php` on line 26.

DFurnes:
Going to close this since it has quite a few merge conflicts. The only required app changes now should just be updating the `laravel-mongodb` package to `3.0.*` (and updating the Wercker config to use the newer box you had set up here).

Let me know if I can be of any help! 😄

PR #360, opened 2016-05-27
Pass query param to opt-out of forwarding users back to ourselves.

DFurnes:
#### What's this PR do?
Since DoSomething/phoenix#6510, users created via the Services user endpoint are now automatically forwarded to Northstar unless the `?forward=false` query parameter is passed. This adds that query parameter to Northstar's Phoenix API call, so we don't make an extra "trip" back.

👀 

#### Checklist

---
For review: @angaither @weerd


angaither:
👍

PR #362, opened 2016-05-31
Return 200 or 201 status code.

DFurnes:
#### What's this PR do?
This pull request updates the `POST v1/users` endpoint to return a `201 Created` status code if the user is newly created from that request (since it was previously _just_ returning `200 Okay` for all requests).

This allows client applications to infer whether a user was newly created without doing any voodoo with the `created_at` timestamp.

Check out the code, then check out those tests. 👀 

#### Checklist

---
For review: @angaither


angaither:
👍

angaither:
hm, actually wait, do we check for specific codes in other apps? we might want to check that

DFurnes:
Oh, good catch. We do specifically check for `200` in Phoenix. 🐦

DFurnes:
Going to merge this to unblock @DeeZone, and we'll just have to make sure we don't deploy to Northstar production until after tomorrow's Phoenix deploy. ⌚

PR #363, opened 2016-06-01
Prevent editing internal fields from profile endpoint.

DFurnes:
#### What's this PR do?
This pull request prevents users from being able to edit "internal" fields on their profile, such as Drupal ID or Mobile Commons ID. These fields can now only be edited by "trusted" internal applications via the `users/` endpoints.

Review the code & updated test case.

#### Checklist

---
For review: @angaither

### DIFF
--- a/app/Http/Controllers/AuthController.php
+++ b/app/Http/Controllers/AuthController.php
@@ -5,6 +5,7 @@ namespace Northstar\Http\Controllers;
 use Illuminate\Http\Request;
 use Northstar\Http\Transformers\TokenTransformer;
 use Northstar\Http\Transformers\UserTransformer;
+use Northstar\Models\User;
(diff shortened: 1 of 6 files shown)


angaither:
should we also allow `source` to only be internal?

angaither:
oh and maybe `parse_installation_ids`?

DFurnes:
At the moment the mobile app explicitly sends a source & the device's Parse IDs when registering a user, so we need to leave those both accessible. Ideally we could set the source automatically based on the OAuth client ID making the request.

DFurnes:
@angaither Dropped in two more fixes: locked down those fields from the `auth/register` endpoint (35df797, fee579f) and a teeny capitalization fix in an unrelated test (17ff9be).

angaither:
🚢

PR #364, opened 2016-06-01
Enable OAuth by feature flag.

DFurnes:
#### What's this PR do?
This PR adds a feature flag so we can conditionally enable OAuth by environment, a unit test for using the OAuth middleware to access a protected route, and updates the Wercker box to use PHP 5.6.

#### How should this be reviewed?
👀

#### Checklist
- [ ] Documentation added for changed endpoints.
- [x] Tests added for new features/bug fixes.

---
For review: @angaither 
/cc @blisteringherb

PR #365, opened 2016-06-01
Fix bootstrap issue with conditionally enabling middleware.

DFurnes:
#### What's this PR do?
This fixes an issue introduced in #364: when checking the config object in the Kernel's constructor, the app hadn't been fully initialized yet. This _totally_ makes sense in hindsight, but didn't show up in the test suite (since then the tests bootstrap the app before any request is made). The more you know! 💫 

Aaaaaanyways... this moves that check to happen after the Kernel bootstraps the application.

This should not cause terrifying errors on every single request now.

#### Checklist

---
For review: @angaither


angaither:
sure thing 🚢

PR #366, opened 2016-06-02
Update those deps

DFurnes:
#### What's this PR do?
This PR unpins the Laravel framework dependency (originally pinned due to a method conflict between the MongoDB package & Laravel which was fixed [in laravel-mongodb v2.3.0](https://git.io/vrpzK)), and updates the OAuth2 Server package to the latest release (from 5.0.0-RC2 to 5.0.3).

This PR also includes two fixes for issues that were [preventing us from upgrading to the new MongoDB PHP extension](https://github.com/DoSomething/northstar/pull/359) for PHP 7:
 - I've removed the `$incrementing = false` property from the User model. This was causing issues because that property instructs Eloquent _not_ to load an automatically generated ID.
 - I've removed some uses of the old PHP extension's built-in classes (which were used to rename collections and fields in migrations). Since these were really only necessary for preserving existing production data (and have already run), this should have no negative side effects.

🐎 

Automated tests should continue to pass!

#### Checklist

---
For review: @angaither

### DIFF
--- a/app/Auth/Repositories/ClientRepository.php
+++ b/app/Auth/Repositories/ClientRepository.php
@@ -11,12 +11,15 @@ class ClientRepository implements ClientRepositoryInterface
     /**
      * Get a client.
      *
-     * @param string $clientIdentifier The client's identifier
-     * @param string $grantType The grant type used
-     * @param null|string $clientSecret The client's secret (if sent)
+     * @param string      $clientIdentifier   The client's identifier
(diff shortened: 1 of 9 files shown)


DFurnes:
@blisteringherb I've tested this branch against the new MongoDB extension on PHP 5.6 & all the tests are passing now, so this should unblock you on #359.

blisteringherb:
👍

angaither:
👍

Issue #361, opened 2016-05-27
Should we remove upsert behavior from the POST users endpoint?


DeeZone:
When a new user request results in a new versus updated user the response code should reflect the difference - `200` vs `201`.

DeeZone:
@DFurnes Any idea on the time frame? This is currently a blocker for Quicksilver user import implementation.

DFurnes:
I should be able to fit it in this sprint. :)

DeeZone:
Would it make sense to make a call to check if the user exists with **Retrieve A User** (https://github.com/DoSomething/northstar/blob/dev/documentation/endpoints/users.md#retrieve-a-user) to determine if the Niche user data is an existing user? If new, a **Create User** call could be made. 

Otherwise, as @DFurnes has pointed out, the concern of updating the user information without the user consent might happen. An example being an existing user with a known mobile number but an unknown / updated email. I don't think we want to update the email address with the Niche value with a **Create User** upsert?

DFurnes:
Yeah, I think that could be a potential concern. I wonder if instead of "upserting" for this endpoint, we should consider having it return an unmodified existing resource if found, and then requiring the client application to make a second call to explicitly update that resource.

Perhaps if an existing account is found, it returns a `307 Temporary redirect` status code?

DeeZone:
My understanding is `201` should be returned for "created" documents. Perhaps returning `200` with the existing user details would be a good response for **Create User** when the user already exists? Or does that create potential for confusion for the developer even when the name of the end point is very clear? 

`307` sounds kinda "funky" to me as this seems to have nothing to do with `Temporary redirect`.

DFurnes:
Yeah, I feel like a `2xx` status code would be a bit confusing since we don't actually modify the resource in that scenario (so the request isn't really successful). It's definitely a bit funky but I think it makes decent sense based on the description in the RFC:

> ##### 10.3.8 307 Temporary Redirect
> The requested resource resides temporarily under a different URI. Since the redirection MAY be altered on occasion, the client SHOULD continue to use the Request-URI for future requests...

DeeZone:
It feels like "*splitting hairs*" but a response of `200` suggests to me that the request (connection to the server) worked but is not a `201` as nothing was created. But that's the wonders of REST, to many interpretations. `307` works for me.

DFurnes:
[...]

PR #369, opened 2016-06-08
Toggle off upsert

DFurnes:
#### What's this PR do?
This PR adds an optional `?upsert=` parameter that allows clients to opt-out of the "upsert" functionality in the `POST v1/users/` endpoint. References #361.

If an existing user is found, it will be returned in the `context` object of the error. For example:

```js
// 422 Unprocessable Entity

{
  "error": {
    "code": 422,
    "message": "Failed validation.",
    "fields": {
      "id": [
        "A record matching one of the given indexes already exists."
      ]
    },
    "context": {
      "id": "575859099a892003e8586dd3",
      "_id": "575859099a892003e8586dd3",
      "first_name": "Dave",
      "last_name": null,
      "last_initial": "",
      "photo": null,
      "email": "dfurnes@dosomething.org",
      "mobile": null,
      "interests": null,
      "birthdate": null,
      "addr_street1": null,
      "addr_street2": null,
      "addr_city": null,
      "addr_state": null,
      "addr_zip": null,
      "source": null,
      "mobilecommons_id": null,
      "parse_installation_ids": null,
      "mobilecommons_status": null,
      "language": null,
      "country": null,
      "drupal_id": null,
      "updated_at": "2016-06-08T17:42:45+0000",
      "created_at": "2016-06-08T17:42:33+0000"
    }
  }
}
```

Check out the changes, and the updated tests! 👀 

There's a little hackery around transforming `User` models in the validation error response... I'll make an issue, but I think that highlights a need for some sort of central `transform()` helper that can transform something outside of a controller (especially once we have a `v1` and `v2` of different transformers! 😱).

#### Checklist

---
For review: @angaither @weerd @DeeZone

### DIFF
--- a/app/Http/Controllers/UserController.php
+++ b/app/Http/Controllers/UserController.php
@@ -77,10 +77,14 @@ class UserController extends Controller
      */
     public function store(Request $request)
     {
-        // This is an "upsert" endpoint (so it will either create a new user, or
+        // This endpoint will upsert by default (so it will either create a new user, or
         // update a user if one with a matching index field is found).
-        // So, does this user exist already?
         $existing = $this->registrar->resolve($request->only('id', 'email', 'mobile', 'drupal_id'));
+
+        // If `?upsert=false` and a record already exists, return a custom validation error.
+        if (! filter_var($request->query('upsert', 'true'), FILTER_VALIDATE_BOOLEAN) && $existing) {
+            throw new NorthstarValidationException(['id' => ['A record matching one of the given indexes already exists.']], $existing);
+        }
 
         // Normalize input and validate the request
         $request = $this->registrar->normalize($request);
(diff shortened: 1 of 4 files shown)


DeeZone:
Looks good to me 👍

app/Http/Controllers/UserController.php
```diff
@@ -77,32 +77,36 @@ public function index(Request $request)
      */
     public function store(Request $request)
     {
-        // This is an "upsert" endpoint (so it will either create a new user, or
+        // This endpoint will upsert by default (so it will either create a new user, or
         // update a user if one with a matching index field is found).
-        // So, does this user exist already?
-        $user = $this->registrar->resolve($request->only('id', 'email', 'mobile', 'drupal_id'));
-        $upserting = ! is_null($user);
+        $existing = $this->registrar->resolve($request->only('id', 'email', 'mobile', 'drupal_id'));
```
weerd: totally not needed for merge, but might help with understanding the code if `$existing` was renamed to something like `$existingUser`? Thought?

weerd:
Small comment, otherwise looks good! 👍

app/Http/Controllers/UserController.php
DFurnes: Agreed! Renamed. 😄

PR #377, opened 2016-06-30
Add 'role' field to user document.

DFurnes:
#### What's this PR do?
Users should have a `role` field that can be used to store their role (for now either `user`, `staff`, or `admin`) and assign permissions across services. This allows us to have a single source for where authorization is defined, rather than multiple individual stores.

Users are prohibited from changing this field when registering or updating their profile (via the `User::$internal` property).

Take a look and let me know if anything looks amiss. Is there any way a user could sneakily set this property for themselves and become an admin without permission?

#### Checklist

---
For review: @angaither @weerd

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -217,7 +217,7 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
         if (! in_array($value, ['user', 'staff', 'admin'])) {
             return;
         }
-        
+
         $this->attributes['role'] = $value;
     }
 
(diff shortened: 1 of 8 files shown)


app/Models/User.php
```diff
@@ -201,6 +202,26 @@ public function setSourceAttribute($value)
...
+     * Mutator for the `role` field.
+     */
+    public function getRoleAttribute()
+    {
+        return ! empty($this->attributes['role']) ? $this->attributes['role'] : 'user';
+    }
+
+    /**
+     * Mutator for the `role` field.
+     */
```
weerd: Maybe specify that the param here is an array in the docblock?

weerd:
Quick comment, otherwise 👍

app/Models/User.php
DFurnes: Good call, updated in 3cfc208!

PR #379, opened 2016-07-05
Add test helpers for OAuth.

DFurnes:
#### What's this PR do?
This PR updates Northstar's test suite to support using both the "legacy" auth & the new OAuth middleware to authorize resource requests:

👋 It moves `v1` endpoint tests into a new `tests/Legacy/` folder so we can keep them around for maintaining backwards compatibility, and similarly renames the test helpers for the old auth scheme.

🔑 It adds a new [`asUser`](https://github.com/DFurnes/northstar/blob/225687ded3b038c70b03df94f1a7ce56edae190f/tests/TestCase.php#L88-L121) test helper method that creates a JWT and sets it as the bearer token. This allows us to easily test routes with JWT authentication without having to go through the entire flow.

Tests should continue to pass. The only _new_ code is in the [`asUser`](https://github.com/DFurnes/northstar/blob/225687ded3b038c70b03df94f1a7ce56edae190f/tests/TestCase.php#L88-L121) helper.

I think there's probably a cleaner way to handle this helper method (without making the whole access token entity by hand), but [the implementation from the oauth2-server package](https://github.com/thephpleague/oauth2-server/blob/68e4b1d39064ba69af2d2461cb657e8aa053447f/src/Grant/AbstractGrant.php#L311-L349) is locked in a protected method so this seems decent enough for now.

#### Checklist

---
For review: @weerd @angaither


weerd:
Would naming the directory something like `/version1` or `/v1`. Thinking about what happens when there's a v3 of the api and where v2 tests go or does that not really matter?

weerd:
Quick question, otherwise 👍

DFurnes:
> Would naming the directory something like /version1 or /v1. Thinking about what happens when there's a v3 of the api and where v2 tests go or does that not really matter?

Yeah, that's not a bad idea. Kinda hoping we don't ever have to deal with it though. 😁

PR #380, opened 2016-07-05
Add index to the source field, so it can be filtered.

### LINKED ISSUE #180: Add index on User source field.

DFurnes:
#### What's this PR do?
Closes #180. This adds an index to the `source` field on the user collection, so it can be filtered against via the API without everything grinding to a halt.

👀

#### Checklist

---
For review: @angaither @weerd

### DIFF
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -180,7 +180,7 @@ class Registrar
         // For the first `where` query, we want to limit results... from then on,
         // we want to append (e.g. `SELECT * WHERE _ OR WHERE _ OR WHERE _`)
         $firstWhere = true;
-        foreach (User::$indexes as $type) {
+        foreach (User::$uniqueIndexes as $type) {
             if (isset($credentials[$type])) {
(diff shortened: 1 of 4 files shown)


weerd:
Wercker is failing, but code looks good. 👍

DFurnes:
Ah! There was a bug! 🐛 We were using the `$indexes` array [in UserController](https://github.com/DoSomething/northstar/blob/59151bf7b32d997b6907c48e44309953a88f18fa/app/Http/Controllers/UserController.php#L93-L106) to prevent people from upserting an existing "credential" (like an upsert changing an account's existing email, no way!) but that used the same array we whitelist "filterable" indexed fields from.

The fix was having separate arrays to track unique vs. indexed fields. Might be a better solution to think about, but this get's us back to green at least! 🚥

weerd:
👍

PR #382, opened 2016-07-05
Add role scopes & middleware.

DFurnes:
#### What's this PR do?
This PR adds new `role:staff` and `role:admin` middleware that grants a client the ability to act on those roles _only_ if the authorized user has that role assigned.

So for example, a normal user who logged in to Aurora would be granted the `role:staff` and `role:admin` scopes because the client requested them, but would still only have `role: "user"` on their token so they couldn't act on them.

(The existing `admin` scope will still grant admin privileges to _any_ user to maintain backwards compatibility, but we should consider phasing that out since it seems like we could easily shoot ourselves in the foot with it.)

Tests should continue to pass.

#### Checklist

---
For review: @angaither @weerd

### DIFF
--- a/app/Auth/Scope.php
+++ b/app/Auth/Scope.php
@@ -13,8 +13,15 @@ class Scope
      * @var array
      */
     protected static $scopes = [
+        'role:admin' => [
+            'description' => 'Allows this client to act as an administrator if the user has that role.',
+        ],
+        'role:staff' => [
+            'description' => 'Allows this client to act as a staff member if the user has that role.',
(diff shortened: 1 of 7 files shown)


DFurnes:
I also updated the names & secrets for the seeded clients in 12be51f so they're easier to remember. We should maybe consider updating the seed `.env.example` values elsewhere to match these.

angaither:
🚢 

thanks for the IRL chat about this! 
@_@ ➡️  😄

PR #384, opened 2016-07-07
Misc cleanup and fixes

DFurnes:
#### What's this PR do?
This PR includes a couple of miscellaneous fixes:

🌱 Renames the `ApiKeyTableSeeder` to `ClientTableSeeder` because that's the name now.

🍃 Tidies up the `Client` and `Scope` classes, and removes some methods from the Client model that were only used for the old `X-DS-REST-API-Key` authentication method, since it's confusing to keep them around if they aren't applicable for everything.

🔑 When checking if a user is allowed to view/edit a profile, we should also be checking their role since admins and staff should be allowed to view and edit profiles on Aurora. This is necessary for the OAuth reworking being done in DoSomething/aurora#135.

Tests should continue to pass!

#### Checklist

---
For review: @weerd @angaither

### DIFF
--- a/app/Auth/Scope.php
+++ b/app/Auth/Scope.php
@@ -54,8 +54,7 @@ class Scope
     }
 
     /**
-     * Return whether a properly scoped API key is provided
-     * with the current request.
+     * Return whether the current request includes the proper client scopes.
      *
      * @param $scope - Required scope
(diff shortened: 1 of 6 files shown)


weerd:
Coolio! 👍

PR #385, opened 2016-07-07
Add updated v2 clients endpoints.

DFurnes:
#### What's this PR do?
This PR adds a `v2/clients` resource that uses the new OAuth field names (e.g. `client_secret` instead of `api_key`) & removes deprecated fields from the transformer, and a corresponding test suite that adds test coverage for the new routes alongside OAuth middleware.

It also moves deprecated endpoints into a "Legacy" folder (like the `KeyController` for the old `v1/keys` endpoint), and adds a new `asNormalUser()` and `asAdminUser()` helpers that creates convenient admin and "normal" users for tests.

Check out the diff and new tests. 👀 🚥 

#### Checklist

---
For review: @weerd @angaither


weerd:
¡OLÉ!
:sheep: it!

PR #388, opened 2016-07-12
Add a Role class for handling role authorization.

DFurnes:
#### What's this PR do?
There's a couple changes in this PR, so I'd recommend checking things out commit-by-commit:

🚥 Splits up methods in `ClientTest` to make things less messy. This makes it easier to see what is causing an issue when something breaks, since it prevents side-effects from previous assertions from leaking into the following embedded "tests" when they're all in one method. (4e9e1b8)

🏭 Defines factories for users with the "staff" or "admin" roles. (4031d13)

🔏 Adds a `Role` class to handle authorization by user roles. This closely mirrors the [`Scope`](https://github.com/DoSomething/northstar/blob/dev/app/Auth/Scope.php) class that we already have, so they can be used more-or-less the same. (8b3ed5c)

💁 Allow users with the "staff" role to view and edit user records. This allows staff members to access the new OAuth-powered Aurora using their Northstar role. (770e1e5)

👓 Restrict looking up a user by username, email, or mobile to "admin" or "staff" user roles (in addition to allowing the `admin` scope to override this check, which was all we checked before). This is because we don't want non-trusted users to be able to look up profile details (even something as innocuous as first name) by a user's email or phone number. (ce66885)

I'd recommend reviewing commit-by-commit. Tests should continue to pass!

#### Checklist

---
For review: @angaither @weerd

### DIFF
--- a/app/Auth/Role.php
+++ b/app/Auth/Role.php
@@ -1,0 +1,98 @@
+<?php
+
+namespace Northstar\Auth;
+
+use League\OAuth2\Server\Exception\OAuthServerException;
+use Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException;
+
+class Role
+{
+    /**
+     * Available roles.
+     * @var array
+     */
+    protected static $roles = [
+        'admin' => [
+            'description' => 'This user is an administrator.',
+        ],
+        'staff' => [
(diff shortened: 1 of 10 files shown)


weerd:
huzzah! 
:sheep: it!

PR #389, opened 2016-07-12
Some additional admin functionality.

DFurnes:
#### What's this PR do?
🍡 Adds an index on the `user_id` and `client_id` fields for refresh tokens so we can easily look up all of a user or client's refresh tokens.

🔢 Adds a `refresh_tokens` field to the client responses that shows a count of how many refresh tokens that client has. We can display this in Aurora for an easy way to track how much active usage an client has. (This isn't done super efficiently right now, but I'm not too worried since we won't likely ever have all that many clients.)

🔑 Adds an endpoint to allow admins to request Northstar's public key. This is used by clients to verify JWT tokens (but _cannot_ be used to create tokens). This will be displayed in Aurora for easy access.

🍸 Updates scope & role "gate" methods to always use the OAuth-style error formatting when on a `v2/...` route (so for example, a request missing the `Authorization` header will still return a 401 rather than defaulting to the legacy 403).

Tests should continue to pass. It's probably easiest to review this PR commit-by-commit.

#### Checklist

---
For review: @weerd @angaither

### DIFF
--- a/app/Auth/Role.php
+++ b/app/Auth/Role.php
@@ -85,9 +85,9 @@ class Role
         if (! static::allows($allowedRoles)) {
             app('stathat')->ezCount('invalid role error');
 
-            // If scopes have been parsed from a provided JWT access token, use OAuth access
-            // denied exception to return a 401 error.
-            if (request()->attributes->has('oauth_scopes')) {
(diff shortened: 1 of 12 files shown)


weerd:
:sheep: it!

PR #391, opened 2016-07-14
Only allow admins to set user roles.

DFurnes:
#### What's this PR do?
This PR makes it so only admins (or clients with `admin` scope) can edit other user's roles. This prevents staff members from setting the admin role for themselves (which would render the two different roles moot!)

It also prevents setting the `role` field when creating or upserting a user, to limit the potential of doing this by accident. 🔒

Tests should pass. Are there loopholes I missed?

#### Checklist

---
For review: @weerd @angaither


angaither:
🚢

PR #393, opened 2016-07-18
Use json parameter for JSON body.

DFurnes:
### What's this PR do?
Hmm, getting some errors from the last pull request (which weren't caught by the test suite since we mock our API classes so they don't have outside dependencies):

```
Client error: `POST https://staging.beta.dosomething.org/api/v1/auth/login` resulted in a `401 Unauthorized : Missing required argument username` response:\n\n  \n[\"Missing required argument username\"]\n
```

I think this should fix the issue, but I don't have a working local Phoenix at the moment so the easiest option is probably just to merge and check on QA/Thor. 😁 

Check that the changes make sense!

### Checklist

---
For review: @angaither @weerd

### DIFF
--- a/app/Services/Phoenix.php
+++ b/app/Services/Phoenix.php
@@ -43,7 +43,7 @@ class Phoenix
             ];
 
             $response = $this->client->post('auth/login', [
-                'body' => json_encode($payload),
+                'json' => $payload,
             ]);
 
             $body = json_decode($response->getBody()->getContents(), true);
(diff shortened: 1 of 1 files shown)


DFurnes:
For context on why I think switching to `json` parameter will fix this: [Upgrade Guide: POST Requests](https://github.com/guzzle/guzzle/blob/master/UPGRADING.md#post-requests).

weerd:
interesting... 👍

PR #399, opened 2016-07-29
Remove admin scope requirement for searching.

DFurnes:
#### What's this PR do?
The `?search[<field>]=<value>` was previously gated to the `admin` key. We can remove this since the _entire_ index route is now gated to admins (especially because this check prevents search from working on Aurora because it no longer has the `admin` scope, only `role:admin` and the user's role).

We can verify that this is still working since [this test](https://github.com/DoSomething/northstar/blob/97070f4a567c7049fff3838f9b08817c80b8432f/tests/Legacy/UserTest.php#L295-L317) still passes. 🚥

#### Checklist

---
For review: @angaither

### DIFF
--- a/app/Http/Controllers/Traits/FiltersRequests.php
+++ b/app/Http/Controllers/Traits/FiltersRequests.php
@@ -2,8 +2,6 @@
 
 namespace Northstar\Http\Controllers\Traits;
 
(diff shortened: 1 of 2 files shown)


angaither:
🚢 :shipit: 🚀

PR #401, opened 2016-08-02
Revert the throttling pull request (for now?)

DFurnes:
### Changes
__Reverts DoSomething/northstar#400.__ So, unfortunately our production Laravel servers never get the IP from the actual request (since all the requests technically just come from HAProxy or Fastly or something and they aren't configured to forward the "real" IP of the request along).

It's the opposite of #400, so 🔂.

---
For review: @angaither


DFurnes:
It looks like there's a [Fastly option](https://docs.fastly.com/guides/basic-configuration/manipulating-the-x-forwarded-for-header) to forward the "real" IP along as the `X-Forwarded-For` header, and a [`forwardfor` option](http://serverfault.com/a/30336) to do the same in HAProxy. I'll make a devops issue to see if these are reasonable things to enable.

DFurnes:
References DoSomething/devops#161.

angaither:
ahhhhh. tricky. also, drats.

angaither:
oh and 🚢

PR #402, opened 2016-08-03
Add a script to remove users with duplicated Drupal IDs.

DFurnes:
#### What's this PR do?
This PR adds an Artisan `northstar:clean_drupal_ids` script which removes records with duplicated Drupal IDs (keeping the oldest, since that seems like a smart heuristic from spot-checking some production data).

References #334.

There's [a test](https://github.com/DFurnes/northstar/blob/0831c8380ca61960406780a15d844cf215b0d3c7/tests/Artisan/CleanDrupalIdsCommandTest.php)! Did I check all the right edge cases?

#### Checklist

---
For review: @angaither @weerd

### DIFF
--- a/tests/Artisan/CleanDrupalIdsCommandTest.php
+++ b/tests/Artisan/CleanDrupalIdsCommandTest.php
@@ -37,12 +37,21 @@ class CleanDrupalIdsCommandTest extends TestCase
             }
         }
 
+        // Make 5 users with explicitly null Drupal ID. Trouble-makers!
+        for ($i = 0; $i < 5; $i++) {
+            User::create([
+                'first_name' => $this->faker->firstName,
+                'email' => $this->faker->email,
+                'drupal_id' => null,
+            ]);
+        }
+
         // There should be 20 users to start with...
-        $this->assertEquals(20, User::all()->count(), 'created all the expected duplicates');
+        $this->assertEquals(25, User::all()->count(), 'created all the expected duplicates');
 
         // And after running the command, we should have only the 10 unique Drupal IDs.
         $this->artisan('northstar:clean_drupal_ids');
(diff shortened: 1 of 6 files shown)


DFurnes:
I'm sad because this test suite passes on my local. Will investigate some more in the morn' 😢

DFurnes:
Well, not 100% sure what's up with that, but re-installing my Composer dependencies got everything synced up with what Wercker was seeing. Still seeing some intermittent(?!?) assertion failures so thinking there might be an ordering issue in my query.

tests/Artisan/CleanDrupalIdsCommandTest.php
```diff
@@ -0,0 +1,52 @@
...
+        $tony->setCreatedAt(new Carbon('March 10 1963'))->save();
+
+        $steve = User::create(['first_name' => 'Steve', 'last_name' => 'Rogers', 'drupal_id' => '12346']);
+        $steve->setCreatedAt(new Carbon('July 4 1920'))->save();
+
+        // Make a user with a Drupal ID, but no duplicates.
+        $kamala = User::create(['first_name' => 'Kamala', 'last_name' => 'Khan', 'drupal_id' => '55555']);
+
+        // Make some users with no Drupal ID.
+        factory(User::class, 7)->create();
```
angaither: The only case I think is left out of here is if `drupal_id` is set to `null` the string. I'm not sure if that matters or is actually different.

DFurnes: Oooh, good point. Lemme look into it.

DFurnes: Yep, that would've been a kablooie. 💥 Fixed in 3d9c333.

angaither:
🚢

PR #403, opened 2016-08-04
Add little more info to the logs for Drupal ID cleanup command.

DFurnes:
#### What's this PR do?
This just adds a _wee_ little bit more information to the cleanup command: see [screencast](http://cl.ly/gt1y). 🎥 

Test results should not change!

#### Checklist

---
For review: @angaither


angaither:
🚢 

there's like LOTS AND LOTS of these right?

It might be kind of nice to add a confirm in for some people we aren't sure about, but then I guess that means we would have to approve every one... hrm

DFurnes:
Yeah. I think if we can figure out a better explanation for why these all exist, then maybe we could figure out a "confidence" in each dupe and prompt if unsure...

PR #404, opened 2016-08-05
Some improvements to Dr00ps script.

DFurnes:
#### What's this PR do?
Things this PR does:

😥 Fixes the wrong URL for Drupal profiles: it's `/user/123123`, not `/users/123123`.

🔢 Adds an `--aurora` option so the Aurora profile URL can be customized per environment.

☑️ If the user's first name isn't empty (most of the dupes on prod seem to be missing first names, but there's a few that aren't), then prompt before deleting so we can be a bit more cautious.

Test should still pass.

#### Checklist

---
For review: @angaither


angaither:
I like this method! 
🚢

PR #405, opened 2016-08-12
Validate mobiles

### LINKED ISSUE #357: Perform mobile validation when creating/updating users.
We currently accept _anything_ for the mobile field and then sanitize the provided value to remove any non-numeric characters (e.g. `(203) 555-1234` to `2035551234`). However, we don't trigger a validation error on obviously invalid phone numbers (for example, one user has the phone number of `3`).

We should consider better validation for this, perhaps using [libphonenumber](https://github.com/giggsey/libphonenumber-for-php) like on Jubilee.

DFurnes:
#### What's this PR do?
Fixes #357. This PR adds validation and better normalization for the `mobile` attribute, based on the rules that we apply for this field in Phoenix. This will prevent users from being created with silly mobile numbers.

This PR also applies the "unset if empty string" rule for all attributes, so we're not duplicating this for each of our unique indexed fields. I don't think there should be any downside to doing this.

Tests! 🚥 

#### Checklist

---
For review: @angaither @weerd

### DIFF
--- a/app/Models/Model.php
+++ b/app/Models/Model.php
@@ -14,5 +14,22 @@ use Jenssegers\Mongodb\Eloquent\Model as BaseModel;
  */
 class Model extends BaseModel
 {
-    // ...
+    /**
+     * Set a given attribute on the model.
+     *
+     * @param  string  $key
+     * @param  mixed   $value
+     * @return mixed
+     */
+    public function setAttribute($key, $value)
+    {
+        // Drop field if attribute is empty string or null.
+        if (empty($value)) {
+            $this->drop($key);
+
+            return null;
+        }
+
+        return parent::setAttribute($key, $value);
+    }
 }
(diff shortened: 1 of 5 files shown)


DFurnes:
Alright, this should be ready for review now. We probably also need to make a script to clean up any existing borked data in the Northstar database (remove phone numbers that are less than 9 digits or more than 12, and convert 11-digit US phone numbers to 10-digit).

app/Models/Model.php
```diff
@@ -14,5 +14,22 @@
...
 {
-    // ...
+    /**
+     * Set a given attribute on the model.
+     *
+     * @param  string  $key
+     * @param  mixed   $value
+     * @return mixed
+     */
+    public function setAttribute($key, $value)
```
angaither: zomg laravel magic?

angaither:
💯

app/Models/Model.php
DFurnes: Laravel magic! 🐰 🎩

PR #410, opened 2016-09-01
Fix issue with mobile number normalization.

### LINKED ISSUE #409: Check if sanitized phone is correct
See https://github.com/DoSomething/northstar/blob/e6255f2fbc5cd162587867adb2457c07bfd81b51/app/Models/User.php#L191

```php
        // Remove all non-numeric characters.
        $sanitizedValue = preg_replace('/[^0-9]/', '', $value);
        // If it's 11-digits and the leading digit is a 1, then remove country code.
        if (strlen($sanitizedValue) === 11 && $sanitizedValue[0] === 1) {
            $this->attributes['mobile'] = substr($sanitizedValue, 1);
            return;
        }
```

Shouldn't `$sanitizedValue[0]` be `=== '1'` (1 as string)?
As I understand, [`preg_replace()`](http://php.net/manual/en/function.preg-replace.php) returns string.

DFurnes:
#### What's this PR do?
This fixes a bug with normalizing mobile numbers (accidentally trying to strictly compare a `'1'` with `1`, and so always skipping the conditional). This means some accounts will still have been saved with leading 1s in the production database.

~~It also uses the Registrar's normalize methods for the User model's mutators to DRY things up.~~ Well, there seems to be some sort of strange error when trying to resolve the Registrar from the User model... keeping things WET for the moment, and will extract to a "Normalizer" in a future PR. 💦 

Fixes #409. References #357.

The failing test added in 382df67 should :sos:, ~~and then it should be ✓ in 8853a65~~... okay but then it should _really_ be ✅ in 10e7ec7. Great work.

#### Checklist

---
For review: @angaither @weerd 
/cc @sergii-tkachenko

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -7,6 +7,7 @@ use Illuminate\Foundation\Auth\Access\Authorizable;
 use Illuminate\Auth\Authenticatable;
 use Illuminate\Contracts\Auth\Authenticatable as AuthenticatableContract;
 use Illuminate\Contracts\Auth\Access\Authorizable as AuthorizableContract;
+use Northstar\Auth\Registrar;
 use Northstar\Auth\Role;
 
 /**
@@ -161,7 +162,7 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
      */
     public function setEmailAttribute($value)
     {
-        $this->attributes['email'] = strtolower($value);
+        $this->attributes['email'] = app(Registrar::class)->normalizeEmail($value);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


app/Models/User.php
```diff
@@ -188,10 +188,8 @@ public function setMobileAttribute($value)
         $sanitizedValue = preg_replace('/[^0-9]/', '', $value);
 
         // If it's 11-digits and the leading digit is a 1, then remove country code.
-        if (strlen($sanitizedValue) === 11 && $sanitizedValue[0] === 1) {
-            $this->attributes['mobile'] = substr($sanitizedValue, 1);
-
-            return;
+        if (strlen($sanitizedValue) === 11 && $sanitizedValue[0] === '1') {
```
weerd: Is this the WETness? Maybe add a `@TODO` note?

tests/UserTest.php
```diff
@@ -189,4 +189,26 @@ public function testUpdateProfileAsAdmin()
...
+    /**
+     * Test that creating a user results in saving normalized data.
+     * POST /users
+     *
+     * @return void
+     */
+    public function testFieldsAreNormalized()
+    {
+        $this->asAdminUser()->json('POST', 'v1/users', [
+            'first_name' => 'Batman',
```
weerd: Yaaaiiis! 

image

weerd:
One quick comment, but looks good! 👍

angaither:
💦

app/Models/User.php
DFurnes: Good call! Added TODOs to the relevant docblocks.

DFurnes:
I also ended up undoing the fix in the registrar (2fe6020), since that would prevent Sergii from being able to look up existing non-normalized records for his current work. Will re-apply once all the production data is cleaned up.

weerd:
👍

angaither:
👍

PR #411, opened 2016-09-01
Adds Facebook user ID

deadlybutter:
#### What's this PR do?
Adds Facebook user ID's as an indexed field to the users model. To go alongside that it also updates documentation, tests, validation rules, model factory and creates a migration.

In another PR I'll add the Facebook verification functionality which verifies that a given users Facebook auth token is legitimate.

Pieces of functionality that are the most important right now to continue prototyping social login...
- POST /v1/users 
 - body: `{facebook_id: 123456}`
- GET   /v1/users/facebook_id/<Facebook ID>
- PUT   /v1/users/email/<email>
 - body: `{facebook_id: 123456}`

#### Checklist


deadlybutter:
strange, also getting these User Test failures locally as well but unclear how my PR broke them..

app/Auth/Registrar.php
```diff
@@ -148,8 +148,9 @@ public function validate(Request $request, User $user = null, array $additionalR
 
         $existingId = isset($user->id) ? $user->id : 'null';
         $rules = [
-            'email' => 'email|unique:users,email,'.$existingId.',_id|required_without:mobile',
-            'mobile' => 'mobile|unique:users,mobile,'.$existingId.',_id|required_without:email',
+            'email' => 'email|unique:users,email,'.$existingId.',_id|required_without_all:mobile,facebook_id',
```
DFurnes: I think you still want [`required_without`](https://laravel.com/docs/5.3/validation#rule-required-without) here. Switching it to [`required_without_all`](https://laravel.com/docs/5.3/validation#rule-required-without-all) would mean that you'd need `email` unless _both_ the `mobile` and `facebook_id` fields were set on an account.

DFurnes:
I think switching out that validation rule should fix the test failures. If not, you can add a `dd($this->decodeJsonResponse())` before the status code assertion to see what the specific errors are.

app/Auth/Registrar.php
deadlybutter: ah, yes your right, I didn't see that required_without supported multiple options

DFurnes:
👍

PR #412, opened 2016-09-02
Add "allowed grants" field to clients. Remove old keys resource.

DFurnes:
#### What's this PR do?
🔢 This PR adds fields to Northstar's client model that we can use to limit which grants a client can request. This is important because we don't want to allow users to authorize on behalf of "machine" clients which may have elevated (for example, `admin`) scopes, and because we can use the same field to start to phase out the legacy authentication method client-by-client.

🌎 It also adds the `redirect_uri` field to the client model, which is required by the auth code grant.

🍃 Finally, it removes the old `v1/keys` endpoint, since it's no longer used anywhere and doesn't make sense to continue to maintain.

Tests should continue to pass.

#### Checklist

---
For review: @angaither @weerd


angaither:
🚢

weerd:
¡OLÉ!

PR #414, opened 2016-09-06
Fix issue with UTF-8 encoded last names.

### LINKED ISSUE #413: Poorly encoded UTF-8 strings cause response to explode.
It looks like some records have poorly encoded UTF-8 strings in some of their fields (for example, the user `55bcc177469c64ee15912ce9` in the production database). This causes PHP's `json_encode` to throw up and return "Malformed UTF-8 characters, possibly incorrectly encoded".

Hopefully this isn't the case for many people, but we should try to handle it better anyways.

DFurnes:
#### What's this PR do?
Fixes #413. This patches up a (probably very rare) error with the `last_initial` field. Since PHP's built-in `substr` doesn't support UTF-8 strings, it would return invalid character codes when the last name field started with a UTF-8 character. Switching to Laravel's `Str::substr` fixes things.

I also ran `composer update` just because that's a lovely thing to do.

You can see the test first failing in 02d3bc4, and then fixed in 96b0a2c. 🚥

#### Checklist

---
For review: @angaither @weerd 
/cc @sergii-tkachenko

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -7,6 +7,7 @@ use Illuminate\Foundation\Auth\Access\Authorizable;
 use Illuminate\Auth\Authenticatable;
 use Illuminate\Contracts\Auth\Authenticatable as AuthenticatableContract;
 use Illuminate\Contracts\Auth\Access\Authorizable as AuthorizableContract;
+use Illuminate\Support\Str;
 use Northstar\Auth\Role;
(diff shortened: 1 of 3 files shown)


angaither:
`└(^o^)┘`

weerd:
hehe... wow that's obscure @_@

PR #416, opened 2016-09-08
More cleanup for setting up the web authentication flow.

DFurnes:
#### What's this PR do?

🍃 Removes the remaining calls to the Registrar's `normalize` method, so we can use the new `normalize()` helper function everywhere instead! (2962a18)

💉 Inject the hasher into the Registrar, rather than using the facade. My crusade continues! (d6f72cd)

👋 Remove the Registrar's `login` and `createToken` methods – these were only relevant for legacy authentication so keeping them around could be confusing at first glance, and authentication probably isn't the Registrar's concern anyways! (e22b907)

#### Why are you doing all this?!
I made these changes when adding web sessions & login for the authentication code grant, I promise! The problem I ran into was that I couldn't inject the Registrar into the user provider because it caused a circular dependency (it needs the authentication Guard for that `login` method). 

Like the last PR, it's probably easiest to go commit-by-commit (linked above). 🔢

#### Checklist

---
For review: @angaither @weerd

### DIFF
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -65,18 +65,6 @@ class Registrar
     }
 
     /**
-     * Normalize the given credentials in the array or request (for example, before
-     * validating, or before saving to the database).
-     *
-     * @param \ArrayAccess|array $credentials
-     * @return mixed
-     */
-    public function normalize($credentials)
-    {
-        return normalize('credentials', $credentials);
-    }
-
-    /**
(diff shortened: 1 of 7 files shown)


weerd:
Btw, I ❤️  helper functions 💃

weerd:
> ... rather than using the facade. My crusade continues!

weerd:
Snazzy! 👍  💃

PR #417, opened 2016-09-08
Adds Facebook token verification

deadlybutter:
#### What's this PR do?
Adds a new endpoint at `northstar.app/v2/auth/facebook/validate` which verifies the Facebook token for a given Facebook user. Required for Phoenix to authorize returning users logging in with Facebook.

*Writing the documentation now but code review can go ahead anyway*

1. Get the env vars on our dashboard https://developers.facebook.com/apps/105775762330/dashboard/

2. In Postman (or whatever youre running) hit this URL with a post `northstar.app/v2/auth/facebook/validate`

and supply these params,

```
{
    "facebook_id": "102....",
    "input_token": "EAA....."
}
```

#### Checklist


app/Http/Controllers/FacebookController.php
```diff
@@ -0,0 +1,49 @@
...
+use GuzzleHttp\Client;
+
+class FacebookController extends Controller
+{
+    protected $client;
+
+    public function __construct()
+    {
+        $base_url = config('services.facebook.url');
+        $this->client_secret = config('services.facebook.client_secret');
```
DFurnes: These two properties (`client_secret` and `client_id`) should be declared as protected properties on the class  (rather than implicitly created as public properties when you assign to them). All three need docblocks, like this:

```php
/**
 * The HTTP client.
 * 
 * @var \GuzzleHttp\Client
 */
protected $client;
```

deadlybutter: Gotcha! Not sure what the convention was on using protected vs this

app/Http/Controllers/FacebookController.php
```diff
@@ -0,0 +1,49 @@
...
+        $this->client_id = config('services.facebook.client_id');
+
+        $this->client = new Client([
+            'base_uri' => $base_url,
```
DFurnes: Since it's only used in one place, let's inline this variable (plus it avoids the weird `uri` / `url` shenanigans).

app/Http/Controllers/FacebookController.php
```diff
@@ -0,0 +1,49 @@
...
+     * @return \Illuminate\Http\Response
+     */
+    public function validateToken(Request $request)
+    {
+        $this->validate($request, [
+            'input_token' => 'required',
+            'facebook_id' => 'required',
+        ]);
+
+        $response = $this->client->request('GET', 'debug_token', [
```
DFurnes: I haven't had a chance to look into this very closely, but are you sure this is the endpoint we should be using? It seems like it's meant for debugging (based on the names and the [docs](https://developers.facebook.com/docs/graph-api/reference/v2.7/debug_token)), not as a standard part of the auth flow.

deadlybutter: Yeah this seems to be it,

> The app_id and user_id fields help your app verify that the access token is valid for the person and for your app. 
https://developers.facebook.com/docs/facebook-login/manually-build-a-login-flow#checktoken

app/Http/Controllers/FacebookController.php
```diff
@@ -0,0 +1,67 @@
...
+            'query' => ['access_token' => $this->client_id.'|'.$this->client_secret, 'input_token' => $request->input('input_token')],
+        ]);
+
+        $verification = json_decode($response->getBody()->getContents())->data;
```
DFurnes: Since parsed JSON isn't strictly typed, we tend to use the "associative" option elsewhere to make it an array. At some point, we should probably make a helper for this so we don't have to do it manually everywhere.

```php
$verification = json_decode($response->getBody()->getContents(), true)['data'];

if ($verification['is_valid'] && ...
```

app/Http/routes.php
```diff
@@ -26,6 +26,7 @@
     // Authentication
     $router->post('auth/token', 'OAuthController@createToken');
     $router->delete('auth/token', 'OAuthController@invalidateToken');
+    $router->post('auth/facebook/validate', 'FacebookController@validateToken');
```
DFurnes: Nitpicky, but let's put this under the `v1/` namespace below since it's for legacy Phoenix auth.

app/Http/Controllers/FacebookController.php
deadlybutter: okie dokes

app/Http/Controllers/FacebookController.php
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /**
+     * Verifies if a given Facebook token is valid & corresponds to the Facebook ID
+     *
+     * @param ServerRequestInterface $request
+     * @param ResponseInterface $response
+     * @return \Illuminate\Http\Response
+     */
+    public function validateToken(Request $request)
```
DFurnes: Since this endpoint is making API requests using our org's Facebook credentials, we should require a DS.org API key for this endpoint. Right now, anyone could go to this URL and verify Facebook tokens on our behalf.

deadlybutter: good thinkin!

deadlybutter:
should be good to go now, tested with admin & non admin scope

documentation/endpoints/auth.md
```diff
@@ -214,3 +214,48 @@ curl -X DELETE \
   }
 }
 ```
+
+## Facebook Token Auth
+
+This will check if the given Facebook token is valid & corresponds to the Facebook ID.
+
+```
+POST /v2/auth/facebook/validate
```
DFurnes: This isn't accurate anymore: `POST v1/...`

deadlybutter: Should be good now, also updated the usage below

DFurnes:
We should add (at least) a simple "does this not explode" functional test before merging this.

You can mock the Guzzle client directly for the test, but it also probably also makes sense to extract the Facebook API calls from the controller into a `\Northstar\Services\Facebook` API client with a single `debugToken()` method.

deadlybutter:
okie

deadlybutter:
Updated this to put Facebook into a new service. Going to write the test now & also address the style errors

deadlybutter:
Sweet also added the test, @DFurnes ready for another look!

app/Services/Facebook.php
```diff
@@ -0,0 +1,50 @@
...
+    {
+        $this->client_secret = config('services.facebook.client_secret');
+        $this->client_id = config('services.facebook.client_id');
+
+        $this->client = new Client([
+            'base_uri' => config('services.facebook.url'),
+        ]);
+    }
+
+    public function verifyToken($input_token, $facebook_id)
```
deadlybutter: needs doc block, whoops

app/Http/Controllers/FacebookController.php
```diff
@@ -0,0 +1,45 @@
...
+    {
+        $this->validate($request, [
+            'input_token' => 'required',
+            'facebook_id' => 'required',
+        ]);
+
+        $verified = $this->facebook->verifyToken($request->input('input_token'), $request->input('facebook_id'));
+
+        if ($verified) {
```
DFurnes: This is a nitpick, but can you make this follow the ["happy path"](https://en.wikipedia.org/wiki/Happy_path) so it doesn't end with an indented conditional?

```php

if (! $verified) {
    return $this->respond('Invalid', 401);
}

return $this->respond('Verified', 200);
```

documentation/endpoints/auth.md
```diff
@@ -214,3 +214,48 @@ curl -X DELETE \
   }
 }
 ```
+
+## Facebook Token Auth
```
DFurnes: This should be in [`endpoints/legacy/auth.md`](https://github.com/DoSomething/northstar/blob/d25fb397cec531e3acae8291d05797a96bf495d5/documentation/endpoints/legacy/auth.md) since it's in the `v1/` namespace.

DFurnes:
Two tiny comments, and then looks great to ship! 🚢

deadlybutter:
@DFurnes 👍  ❓

DFurnes:
🚢

PR #418, opened 2016-09-09
Add web sessions, functional login and logout routes.

DFurnes:
#### What's this PR do?
Woo! This implements the fabled web login flow for Northstar:

screen recording 2016-09-09 at 02 39 pm

Step by step:

🎫 Some little tweaks to the Registrar: should ignore any provided passwords when _resolving_ (since that's not an authentication action), and renames `verify` to `validateCredentials` for clarity and because it matches the interface name for the Laravel user provider! (7a17e8a, 5512b65)

🔑 The token guard's `validate` method should check actual credentials, rather than just looking for token deets. This way the `$this->auth->guard(...)->once()` method works as expected. (a756e2b)

👪 Adds a custom user provider so Laravel's built-in SessionGuard works out of the box with our Registrar, and start using the auth factory to allow selecting specific guard rather than always injecting the API-specific one. (7cfc041, 178e5e5)

💻 Fully implements login & logout routes, styled to look like the Phoenix modals. This is how the Auth Code flow will log a user in before redirecting back to the client app. (1b096bd, 1b096bd)

References #419.

Commits for each part of this PR are linked above, and tests are added for the new flows. 🚥

#### Checklist

---
For review: @angaither @weerd

### DIFF
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -9,6 +9,7 @@ use Illuminate\Http\Request;
 use Northstar\Exceptions\NorthstarValidationException;
 use Northstar\Models\User;
 use Northstar\Services\Phoenix;
+use Illuminate\Contracts\Auth\Authenticatable as UserContract;
 
 class Registrar
 {
@@ -138,14 +139,14 @@ class Registrar
     }
 
     /**
-     * Verify the given user and credentials. If the user has a Drupal
+     * Validate a user against the given credentials. If the user has a Drupal
      * password & it matches, re-hash and save to the user document.
      *
-     * @param User $user
+     * @param UserContract|User $user
      * @param array $credentials
      * @return bool
      */
-    public function verify($user, $credentials)
+    public function validateCredentials(UserContract $user, array $credentials)
     {
         if (! $user) {
             return false;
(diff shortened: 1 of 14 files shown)


weerd:
you should show this on login:
image

app/Auth/Registrar.php
```diff
@@ -88,7 +88,9 @@ public function validate(Request $request, User $user = null, array $additionalR
      */
     public function resolve($credentials)
     {
+        // Normalize credentials and remove password if provided.
         $credentials = normalize('credentials', $credentials);
+        $credentials = array_except($credentials, 'password');
```
weerd: Useful Laravel helper 💃

resources/assets/app.scss
```diff
@@ -2,3 +2,52 @@ $forge-path: "~@dosomething/forge/assets/";
...
+// Validation Errors
+.validation-error {
+  color: #ff4747;
+  text-align: left;
+  margin: 12px 0;
+
+  & h4 {
+    color: #ff4747
+  }
+}
```
weerd: CSS in Northstar?! U heard of!

weerd:
¡OLÉ! :sheep: it!

PR #430, opened 2016-09-20
Add registration page.

DFurnes:
#### What's this PR do?
This pull request adds the web registration flow for Northstar! This allows people to register from [northstar.dosomething.org](https://northstar.dosomething.org) (unlikely), and also means that if you use OpenID Connect to authorize with Phoenix, you can register within that flow and be logged in to the newly created account.

👀 

#### Checklist

---
For review: @angaither @weerd

### DIFF
--- a/app/Http/Controllers/WebController.php
+++ b/app/Http/Controllers/WebController.php
@@ -87,7 +87,6 @@ class WebController extends BaseController
         return redirect('login');
     }
 
-
     /**
      * Display the registration form.
      *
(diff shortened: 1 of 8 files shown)


DFurnes:
I'll add tests in the morning! 🚥

app/Http/Controllers/WebController.php
```diff
@@ -86,4 +99,43 @@ public function getLogout()
...
+        // "register" to set a new password on their account.
+        $existing = $this->registrar->resolve($request);
+        if ($existing && $existing->hasPassword()) {
+            throw new NorthstarValidationException(['email' => 'A user with that email or mobile has already been registered.']);
+        }
+
+        $this->registrar->validate($request, $existing, [
+            'first_name' => 'required',
+            'birthdate' => 'required|date',
+            'password' => 'required|confirmed|min:6',
```
angaither: birthday and password required?

how will password required work with voting-app?
Also a ~future problem~ is I don't think we can get birthday with facebook auth.

DFurnes: This is only for the web registration flow! The `POST v1/users/` endpoint will still allow us to programmatically register users without either of those.

DFurnes: For Facebook auth, I believe the idea is to provide a "complete your profile" step where we request that you provide the missing fields.

PR #446, opened 2016-09-28
Remove the redirect_uri specified in the client seeder.

weerd:
#### What's this PR do?
This PR removes the `redirect_uri` from the seeded client.

👀   @DFurnes Can't remember if this was something only I needed to do, or if it's something we decided was needed to be changed.

---
For review: @DFurnes


DFurnes:
\*cough\* #445

weerd:
tee hee! Closing!

PR #447, opened 2016-09-29
Login tweaks

DFurnes:
#### What's this PR do?
This PR includes some improvements to the login flow & general tidiness:

📄 Adds missing `Client` fields to the class's docblock for syntax highlighting goodness.

🔒 Checks authentication for the `/authorize` endpoint via the guard. This allows us to redirect to the login page after we have already figured out the client information for the request.

💻 Adds the human-readable client name to the `ClientEntity`, and adds a `ClientEntity::fromModel` static method that creates a new entity from the corresponding Eloquent model.

🎨 Makes the `.-centered` container fixed-width (just like our modals are), since otherwise the container seemed either a bit too big or a bit too small at either extreme.

🎨 Adds the destination (by default, client name but can be customized with `?destination=...`) to the heading "Log in to continue to ___" when making authorization requests:

screen shot 2016-09-29 at 4 23 40 pm

👀

#### Checklist

---
For review: @weerd @angaither

### DIFF
--- a/app/Auth/Entities/ClientEntity.php
+++ b/app/Auth/Entities/ClientEntity.php
@@ -21,7 +21,9 @@ class ClientEntity implements ClientEntityInterface
      * Make a new OAuth Client entity.
      *
      * @param $client_id
+     * @param $client_name
      * @param $scopes
+     * @param $redirect_uri
      */
     public function __construct($client_id, $client_name, $scopes, $redirect_uri = '')
     {
--- a/app/Auth/Entities/ClientEntity.php
+++ b/app/Auth/Entities/ClientEntity.php
@@ -5,6 +5,7 @@ namespace Northstar\Auth\Entities;
 use League\OAuth2\Server\Entities\ClientEntityInterface;
 use League\OAuth2\Server\Entities\Traits\ClientTrait;
 use League\OAuth2\Server\Entities\Traits\EntityTrait;
+use Northstar\Models\Client;
 
 class ClientEntity implements ClientEntityInterface
 {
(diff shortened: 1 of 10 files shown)


app/Auth/Entities/ClientEntity.php
```diff
@@ -22,15 +23,26 @@ class ClientEntity implements ClientEntityInterface
      * @param $client_id
```
angaither: add newb @param here!

resources/views/login.blade.php
```diff
@@ -6,7 +6,8 @@
     <div class="container -padded">
         <div class="wrapper">
             <div class="container__block -centered">
-                <h1>Log in to get started!</h1>
+                <h1>Let's do this!</h1>
```
angaither: 👀  💯

documentation/endpoints/auth.md
```diff
@@ -24,6 +24,7 @@ Redirect the user to Northstar's "authorize" page with the following query strin
 
 * `response_type` with the value `code`
 * `client_id` with your Client ID
+* `destination` with a destination to display on the login page (optional)
```
angaither: is `client_name` needed here as well?

app/Auth/Repositories/ClientRepository.php
```diff
@@ -37,7 +38,7 @@ public function getClientEntity($clientIdentifier, $grantType, $clientSecret = n
             return null;
         }
 
-        return new ClientEntity($model->client_id, $model->scope, $model->redirect_uri);
+        return ClientEntity::fromModel($model);
```
weerd: Cool 💃

app/Auth/Entities/ClientEntity.php
DFurnes: Good 👀!

documentation/endpoints/auth.md
DFurnes: Nope! It'll load that from what's stored in the Northstar client model (based on the provided `client_id`).

angaither:
🚀 
⭕ 🔏

PR #453, opened 2016-10-13
Updating to jenssegers mongo 3.0

blisteringherb:
#### What's this PR do?
This updates the mongodb driver to v3.0 which is necessary to be compatible with PHP 7. It also appears to be compatible with PHP 5.6 which is running in production now as well. This will enable the ds-homestead environment to be updated to PHP 7 and then production will follow. https://github.com/DoSomething/ds-homestead/pull/15

Run Northstar tests in ds-homestead and eventually use the Ghost inspector tests in QA.

#### Checklist

### DIFF
--- a/app/Console/Commands/CleanDrupalIdsCommand.php
+++ b/app/Console/Commands/CleanDrupalIdsCommand.php
@@ -60,7 +60,7 @@ class CleanDrupalIdsCommand extends Command
 
         $app = $this;
         $blanks->each(function ($this) use (&$app) {
-            $app->info('Found '.$this['count'].' duplicates for '. $this['_id']['drupal_id'] .' ('.config('services.drupal.url').'/user/'. $this['_id']['drupal_id'].'):');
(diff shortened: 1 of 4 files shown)


DFurnes:
@blisteringherb Looks like this is causing our Wercker test suite to fail because the box doesn't have the new Mongo extension installed. Can you update that as a step in the `wercker.yml` file?

blisteringherb:
I'm was seeing the same `CleanDrupalIdsCommandTest::testThatItDeletesTheDupes` test failing with this update. According to the laravel-mongo upgrade info:

"Embedded relations now return an Illuminate\Database\Eloquent\Collection rather than a custom Collection class. If you were using one of the special methods that were available, convert them to Collection operations."

I refactored the code in the `CleanDrupalIdsCommand` to use the new format that's returned. It looks like the `RemoveDuplicateUsersCommand` command also uses the `User::raw()` function and needs to be refactored. I imagine there might be other places where something different is being returned than the application expects. The tests are passing now, but I'm not sure how expansive the test coverage is. @DFurnes

PR #459, opened 2016-10-21
Remove old commands

DFurnes:
#### What's this PR do?
Removes old command from the Artisan kernel's command list, since otherwise it was causing errors when running `php artisan` due to the missing PHP extension class. 💥

This command is not used anymore and relies on the old Mongo extension so would require some refactoring to get back into working shape. (Still keeping it in version control for reference though!)

👀

#### Checklist

---
For review: @weerd @angaither


DFurnes:
Hmm, no Wercker or StyleCI builds. 😕

PR #461, opened 2016-10-21
Fixes errors in Drupal cleanup command.

DFurnes:
#### What's this PR do?
Fixes #461. This prevents an exception from being thrown in our test suite (caused by trying to pass `$this` as a renamed variable pointer). These were failing on my local PHP 7 vagrant box, but weren't caught on Wercker builds since we're still running PHP 5.6 there.

👀

#### Checklist

---
For review: @weerd


DFurnes:
It looks like the webhooks are delayed on this PR too. Come on, Github! ⌛️

PR #463, opened 2016-10-24
Reorganizing and expanding on templating views.

weerd:
#### What's this PR do?

This PR updates the Northstar web views and adds some of the initial pieces for a user to edit their profile, along with some extra fixes/updates. It mostly completes incorporating the new layout design by @lkpttn for the authentication related views (login/register) and also provides a view for a logged in user to view their profile information if necessary.

Login View:
image

Register View:
image

User Profile View:
image
#### Relevant tickets

Refs #469.

Take a thorough 👁 ! And let's chat IRL for any questions.
#### Checklist
---

@DFurnes 

@angaither @jessleenyc


resources/views/layouts/app.blade.php
```diff
@@ -1,9 +1,13 @@
-<!doctype html>
+<!DOCTYPE html>
```
weerd: Capitalized, just how like the browser likes it :trollface:

resources/assets/patterns/_navigation.scss
```diff
@@ -0,0 +1,20 @@
+.navigation {
+  margin: 0 auto;
```
DFurnes: Should these modifications be done under a modifier class, rather than directly to `.navigation`?

resources/assets/patterns/_validation.scss
```diff
@@ -0,0 +1,9 @@
+.validation-error {
```
DFurnes: Nitpicky, let's rename this file to `_validation-error.scss` so it matches pattern classname.

resources/assets/patterns/_navigation.scss
weerd: Given that we'll probably need to expand on this eventually when more of the NS pages get designed, I'd rather just override for now and come up w/ proper modifier classes as needed when it makes more sense.

PR #471, opened 2016-11-01
Cleanup and fixes

weerd:
#### What's this PR do?
This PR does a quick fix to the `.container > .wrapper` to override the default styles which force it to a 75% width and large left margin. However, for the context of the `.-framed` modifier, this is not needed and needs to be overridden to help expand the content to fit the framed container properly. This override was intended for #463 but ended up getting accidentally removed.

It also, does a bit of tweaking and cleanup for the show profile view.

👁 

---
@DFurnes

### DIFF
--- a/resources/assets/patterns/_container.scss
+++ b/resources/assets/patterns/_container.scss
@@ -14,11 +14,11 @@
       box-shadow: $soft-box-shadow;
       max-width: $container-max-width;
       margin: $section-spacing auto;
-    }
 
-    > .wrapper {
-      margin-left: 0;
-      width: 100%;
+      > .wrapper {
+        margin-left: 0;
+        width: 100%;
+      }
     }
 
     .navigation + & {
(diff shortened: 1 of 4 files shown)


resources/assets/patterns/_container.scss
```diff
@@ -1,7 +1,7 @@
 // @TODO: incorporate into the Container pattern in Forge!
 
 .container {
-  // @TODO: patternize in Forge?
+  // @TODO: patternize in Forge
```
DFurnes: ‼️

resources/views/auth/login.blade.php
```diff
@@ -24,7 +24,7 @@
 
             <div class="form-item">
                 <label for="username" class="field-label">Email address or cell number</label>
-                <input name="username" type="text" class="text-field" placeholder="puppet-sloth@example.org" value="{{ old('username') }}">
+                <input name="username" type="text" class="text-field" placeholder="puppet-sloth@example.org" value="{{ old('username') }}" autofocus>
```
DFurnes: Nice! 👍

PR #472, opened 2016-11-01
Redirect to Phoenix after resetting password.

DFurnes:
#### What's this PR do?
To match the current user flow, we want to redirect users to Phoenix's homepage after successfully resetting their password. This is accomplished by initiating Phoenix's authorization flow via the `user/authorize` endpoint over there.

This relies on the change made in DoSomething/phoenix#7168 to prevent a fatal exception if a user was already logged in to Drupal when arriving at the `user/authorize` redirect.

Would appreciate a sanity-check that this is an okay approach. ✌️

#### Checklist

---
For review: @weerd @angaither

### DIFF
--- a/app/Http/Controllers/PasswordController.php
+++ b/app/Http/Controllers/PasswordController.php
@@ -18,13 +18,6 @@ class PasswordController extends BaseController
     protected $guard = 'web';
 
     /**
-     * The path to redirect to after a successful reset.
-     *
-     * @var string
-     */
-    protected $redirectTo = '/';
-
-    /**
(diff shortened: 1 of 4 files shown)


DFurnes:
[As an admin, I want users to be redirected to their phoenix user profile after they've gone through the northstar forgot password flow so that they don't just end up in northstar, confused.](https://trello.com/c/5LOp8SLN/164-as-an-admin-i-want-users-to-be-redirected-to-their-phoenix-user-profile-after-they-ve-gone-through-the-northstar-forgot-password)

PR #475, opened 2016-11-02
Edit profile functionality

### LINKED ISSUE #469: Allow user to view/edit their profile information in Northstar via the web.
This update also needs to include the new layout designs for page views in Northstar.

There are a few pieces that need to be completed to add this feature into Northstar:
- [x] update web login view with new layout/design
- [x] update web register view with new layout/design
- [x] add new web view for a user to view their profile
- [ ] add new web view for a user to edit their profile
- [ ] enabled functionality for user to edit/update their profile information

weerd:
#### What's this PR do?
This PR adds the edit profile functionality for a user's profile information in Northstar along with some tests to make sure it's all functioning correctly.

👁  Try editing your information and then make sure it was updated. Try causing validation errors if you remove a required field like "First name", etc. Try changing your password, logout and then login with new password.

Fixes #469

#### Checklist
- [ ] Post a message in #api if this includes something that causes a rebuild! ]

---
@DFurnes


app/Http/Controllers/Web/UserController.php
```diff
@@ -3,20 +3,31 @@
...
 use Illuminate\Routing\Controller as BaseController;
+use Northstar\Auth\Registrar;
 use Northstar\Models\User;
 
-class UsersController extends BaseController
+class UserController extends BaseController
 {
     /**
-     * Make a new UsersController, inject dependencies and
+     * The authentication factory.
```
DFurnes: This should be updated to describe the registrar! 👾

resources/views/auth/login.blade.php
```diff
@@ -24,12 +24,12 @@
 
             <div class="form-item">
                 <label for="username" class="field-label">Email address or cell number</label>
-                <input name="username" type="text" class="text-field" placeholder="puppet-sloth@example.org" value="{{ old('username') }}" autofocus>
+                <input name="username" type="text" class="text-field" placeholder="puppet-sloth@example.org" value="{{ old('username') }}" autofocus />
```
DFurnes: Retro! 🤓

resources/views/users/edit.blade.php
```diff
@@ -3,75 +3,108 @@
...
+    <form method="POST" action="{{ route('users.update', $user->id) }}">
+        {{ method_field('PATCH') }}
+        {{ csrf_field() }}
+
+        <div class="container__block">
+            <h3 class="heading">Info</h3>
+
+            <div class="form-item">
+                <label for="first-name" class="field-label">First Name</label>
+                <input type="text" id="first-name" class="text-field" name="first_name" value="{{ old('first_name') ?: $user->first_name }}" autofocus />
```
DFurnes: Would Blade's `{{ old('first_name') or $user->first_name}}` read a little nicer?

resources/views/users/show.blade.php
```diff
@@ -36,5 +38,9 @@
             <dt>Country:</dt>
             <dd>{{ $user->country }}</dd>
         </div>
+
+        <div class="form-actions">
+            <a href="{{ url('users/'.$user->id.'/edit') }}" class="button -secondary">Edit Profile</a>
```
DFurnes: It might be nice to use `route('users.edit', $user->id)` here.

resources/views/users/edit.blade.php
weerd: I tried that first, but it doesn't work. It fails because you can't pass a function (like `old(...)`) to `isset()` which is what the `or` expands the logic too. Sadness

resources/views/users/show.blade.php
weerd: Yeah, was debating what would be better. I'll switch it up!

app/Http/Controllers/Web/UserController.php
weerd: oop!

PR #477, opened 2016-11-04
Proof of concept of a Forge presenter class.

DFurnes:
#### What's this PR do?
This is a little proof of concept for a Forge presenter class, which could be pulled out into a standalone library if we're interested in pursuing the idea.

I'm a fan of reducing the amount of noise in some of the longer forms we have, and not having to look up and copy markup from the pattern library. On the other hand, this does add a layer of abstraction from underlying markup that could get in the way.

💡

#### Checklist

---
For thoughts: @weerd @sbsmith86

### DIFF
--- a/app/Http/Presenters/ForgePresenter.php
+++ b/app/Http/Presenters/ForgePresenter.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace Northstar\Http\Presenters;
+
+use Illuminate\Support\HtmlString;
+
+class ForgePresenter
+{
+    public function field($type, $name, $label, $value = '', $placeholder = '')
+    {
+        $value = old($name) ?: $value;
+
+        return new HtmlString('
+          <div class="form-item">
+            <label for="'.e($name).'" class="field-label">'.e($label).'</label>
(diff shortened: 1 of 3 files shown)


weerd:
@DFurnes that seems pretty cool :) I like the idea behind it, but I'm concerned about this -> "this does add a layer of abstraction from underlying markup that could get in the way"... 

It feels a bit like going back to the whole `Form::input()` stuff, which we started to move away from because it caused more headaches when we needed to manipulate the form markup more thoroughly. What about trying to go the route of "publish-able" partials via a _DoSomething Laravel Toolkit_ package or something? Need a login form or registration form, publish it via artisan from package to project and then edit it as needed? Not sure if that actually seems helpful or not.

I do find intriguing some of cool stuff going on for calling the helper functions here. Pretty cool stuff!

DFurnes:
Yeah, packaged partials for each of the Forge patterns is the dream. 👌 I really love the idea of being able to include templates with our npm package and then have this read those in to output markup.

Going to leave this open for now so we can mull over it for a bit.

DFurnes:
This is getting pretty old, so I'll close it until we get a chance to revisit.

PR #479, opened 2016-11-07
Progressively cast birthdates.

### LINKED ISSUE #407: Store birthdate as MongoDate field.
Currently the `birthdate` field is stored as a string. This should be casted as a Date object.

DFurnes:
#### What's this PR do?
This pull request adds support for "progressive casting" for the birthdate field - any new values will be stored as `'Y-m-d H:i:s'`, but existing values stored as plaintext (for example, we've got plenty of values stored `MM/DD/YYYY` in this field on production) will still be read correctly.

Fixes #407.

👀

#### Checklist

---
For review: …

### DIFF
--- a/app/Models/Model.php
+++ b/app/Models/Model.php
@@ -2,6 +2,8 @@
 
 namespace Northstar\Models;
 
+use Carbon\Carbon;
+use InvalidArgumentException;
 use Jenssegers\Mongodb\Eloquent\Model as BaseModel;
 
 /**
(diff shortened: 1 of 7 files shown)


app/Models/User.php
```diff
@@ -136,15 +136,15 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
...
     /**
-     * The attributes which should be stored as MongoDate objects.
-     * @see https://github.com/jenssegers/laravel-mongodb#dates
+     * Indicates if the model should be timestamped.
      *
-     * @var array
+     * @var bool
      */
-    protected $dates = ['created_at', 'updated_at'];
+    public $timestamps = true;
```
weerd: hmmmm, is this overridden somewhere in the mongo base model? It's set to true in the base Eloquent model...

DFurnes: Good catch! Removed altogether in 307a4e0.

app/Models/Model.php
```diff
@@ -32,4 +35,51 @@ public function setAttribute($key, $value)
...
+        try {
+            return parent::asDateTime($value);
+        } catch (InvalidArgumentException $e) {
+            return $this->asDateTimeFallback($value);
+        }
+    }
+
+    /**
+     * Fallback to try to parse poorly formatted date strings, or
+     * return `null` if it's hopeless.
```
weerd: @_@ haha.

weerd:
@DFurnes I updated the edit view birthdate input to match how we require it in the registration form. I also changed the input from `date` back to `text`; you were right and in Safari and Firefox, the date picker doesn't show up and the value isn't straightforward. Plus, it's very unlikely that a member would need a calendar picker to _change_ their birthdate, haha.

everything else looks good though!

DFurnes:
Great, merging! 👍

PR #480, opened 2016-11-09
Allow upserting created_at only if earlier.

### LINKED ISSUE #321: Standardized source values
Currently, a user could use the mobile app API key and tamper with their `auth/register` request to set a nonsense source value. We should consider standardizing this as an enum value with a set of pre-defined "acceptable sources" somewhere.

:runner: **Stretch goal:** It could also be interesting to look into having the source inferred from the API key for the request (e.g. if a request is made with the `LetsDoThis` API key, set that in the "source" field automatically).

### LINKED ISSUE #470: Allow Modification of "Created At" Field if Before Current "Created At "Date, but Not After
After talking to @DFurnes about implications of `created_at` date, he had the brilliant idea of allowing the `created_at` date to be updated to before it may have been created (like if a MC user has a creation date before their Northstar record. 

cc @aaronschachter @jessleenyc 

The lift on this should be pretty minimal, according to @DFurnes.

DFurnes:
#### What's this PR do?
This pull request adds some conditional logic that allows the `created_at` field on a user to be updated, but only to be earlier than it already is. This will allow us to store the user's actual "registration date" when backfilling users who joined by first texting in via SMS, but were already backfilled by their linked Phoenix account.

This PR also refactors how source is set – a manual source can be provided when creating or backfilling a user via the `POST /v1/users` endpoint, but otherwise it is automatically set based on the active OAuth Client ID.

Fixes #470.
Fixes #321.

:eyes: and the included test case/documentation.

#### Checklist

---
For review: @weerd @angaither

### DIFF
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -3,6 +3,7 @@
 namespace Northstar\Providers;
 
 use Illuminate\Support\ServiceProvider;
+use Northstar\Models\User;
 
 class AppServiceProvider extends ServiceProvider
 {
@@ -13,7 +14,10 @@ class AppServiceProvider extends ServiceProvider
      */
     public function boot()
     {
-        //
+        User::creating(function ($user) {
+            // Set source automatically if not provided.
+            $user->source = $user->source ?: client_id();
+        });
     }
 
     /**
--- a/tests/Http/UserTest.php
+++ b/tests/Http/UserTest.php
@@ -243,13 +243,14 @@ class UserTest extends TestCase
      */
     public function testUpsertCreatedAtField()
     {
-        $user = factory(User::class)->create(['first_name' => 'Daisy', 'last_name' => 'Johnson']);
+        $user = factory(User::class)->create(['first_name' => 'Daisy', 'last_name' => 'Johnson', 'source' => 'television']);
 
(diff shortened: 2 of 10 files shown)


app/Providers/AppServiceProvider.php
```diff
@@ -13,7 +14,10 @@ class AppServiceProvider extends ServiceProvider
      */
     public function boot()
     {
-        //
+        User::creating(function ($user) {
+            // Set source automatically if not provided.
+            $user->source = $user->source ?: client_id();
```
angaither: ooo!

angaither:
👍

tests/Http/UserTest.php
```diff
@@ -234,4 +234,56 @@ public function testUTF8Fields()
...
+     */
+    public function testUpsertCreatedAtField()
+    {
+        $user = factory(User::class)->create(['first_name' => 'Daisy', 'last_name' => 'Johnson', 'source' => 'television']);
+
+        // We finally read Secret War #2, and want to update her 'created_at' date to match her first appearance.
+        $this->asAdminUser()->json('POST', 'v1/users', [
+            'email' => $user->email,
+            'first_name' => 'Daisy',
+            'created_at' => '7/1/2004', // first comic book appearance!
```
weerd: LOL! 💃

DFurnes: 🤓

PR #482, opened 2016-11-10
Keep user passwords from unintentionally updating

### LINKED ISSUE #481: Remove empty fields when updating user profile
Bug: when a user updates their profile information, if they don't edit their password info, then the form submits and their password is still changed, but to an empty string. Once they logout, they CANNOT log back in, since submitting an empty field for their password would result in a server side validation error. This basically completely locks out a user from their account. Fix is incoming!

weerd:
#### What's this PR do?
Currently, if a user views their profile and edits the information, even if they don't edit their password and submit the form, their password is still submitted as being updated as an empty string. When they logout, they are basically locked out 🔒  of their account, since their new password is an empty string and the validation on the login form does not allow entering an empty string as a password.

This PR runs the array of fields through a quick function that removes any items where the value is an empty field (implying it wasn't updated). This way, if a user doesn't edit their password, it is not changed.

Fixes #481

👁 

#### Checklist

---
For review: 
@DFurnes @angaither 

CC: @jessleenyc

### DIFF
--- a/app/Http/Controllers/Web/UserController.php
+++ b/app/Http/Controllers/Web/UserController.php
@@ -85,7 +85,10 @@ class UserController extends BaseController
             'password' => 'min:6|confirmed',
         ]);
 
-        $user->fill($request->all())->save();
+        // Remove fields with empty values.
+        $values = array_diff($request->all(), ['']);
+
+        $user->fill($values)->save();
 
         return redirect()->route('users.show', $user->id);
     }


angaither:
:shield:

app/Http/Controllers/Web/UserController.php
```diff
@@ -85,7 +85,10 @@ public function update(Request $request, $id)
             'password' => 'min:6|confirmed',
         ]);
 
-        $user->fill($request->all())->save();
+        // Remove fields with empty values.
+        $values = array_diff($request->all(), ['']);
```
DFurnes: This seems like a good quick fix but it'll prevent people from unsetting other fields as well. For example, if I wanted to remove my home address from my profile I couldn't just make those fields blank.

DFurnes: I'd also love to think about pulling the password field out into it's own form (and talk about adding a "Current Password" confirmation step before letting you change it).

weerd: Indeed. Yeah, I wanted something in that wouldn't break the login. Something better needs to be added in place!

PR #484, opened 2016-11-10
Update to profile access.

weerd:
#### What's this PR do?
This PR ensures that the profile edit form and the process of updating a profile is only accessible to the authenticated user and their own information. If the authenticated user tries to access edit form or perform an update on a profile other than their own, they are sent back to their own profile view while an exception is thrown in the background.

👀 

#### Checklist

---
For review:
@DFurnes @angaither

### DIFF
--- a/app/Http/Controllers/Web/UserController.php
+++ b/app/Http/Controllers/Web/UserController.php
@@ -7,6 +7,7 @@ use Illuminate\Http\Request;
 use Illuminate\Routing\Controller as BaseController;
 use Northstar\Auth\Registrar;
 use Northstar\Models\User;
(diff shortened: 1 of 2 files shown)


weerd:
@DFurnes would there be value in having an `authUser` property on the `UserController`, set upon `__construct()`ing and thus available for when needed to check against access rather than calling `auth()->guard('web')->user()` in a variety of places?

DFurnes:
I think we can set the default guard for the controller in `routes.php` to remove that `->guard('web')` call, which would then leave us with `auth()->user()`. That feels roughly equivalent to a `$this->user`.

weerd:
@DFurnes cool cool. that sounds like a good idea :)

PR #487, opened 2016-11-11
Google analytics powers activate

weerd:
#### What's this PR do?
This PR adds some initial Google Analytics for form submission events (login, registration, edit forms) as well as the "Cancel" link on the edit profile form. There will be additional events to add once the client-side validation library is added to handle client-side validation on these forms.

Need to add a new environment variable in **.env** file `GOOGLE_ANALYTICS_ID=UA-CODE-HERE`

👁 

#### Checklist

---
For review:
@DFurnes 

cc: @jessleenyc

### DIFF
--- a/config/services.php
+++ b/config/services.php
@@ -14,6 +14,11 @@ return [
     |
     */
 
+    'analytics' => [
+        'production' => 'UA-493209-41',
+        'qa' => 'UA-493209-42',
+    ],
+
     'mailgun' => [
         'domain' => env('MAILGUN_DOMAIN'),
         'secret' => env('MAILGUN_KEY'),
--- a/config/services.php
+++ b/config/services.php
@@ -14,11 +14,6 @@ return [
     |
     */
 
-    'analytics' => [
-        'production' => 'UA-493209-41',
-        'qa' => 'UA-493209-42',
-    ],
-
     'mailgun' => [
         'domain' => env('MAILGUN_DOMAIN'),
         'secret' => env('MAILGUN_KEY'),
--- a/config/services.php
+++ b/config/services.php
@@ -14,6 +14,10 @@ return [
     |
     */
 
+    'analytics' => [
+        'id' => env('GOOGLE_ANALYTICS_ID'),
+    ],
+
     'mailgun' => [
         'domain' => env('MAILGUN_DOMAIN'),
         'secret' => env('MAILGUN_KEY'),
--- a/resources/assets/js/app.js
+++ b/resources/assets/js/app.js
@@ -6,4 +6,4 @@ import '../scss/app.scss';
 
 
 // Utilities
-// import Analytics from 'utilities/Analytics';
+import Analytics from 'utilities/Analytics';
--- a/resources/assets/js/app.js
+++ b/resources/assets/js/app.js
@@ -1,9 +1,22 @@
+/**
+ * This is where we load and initialize components of the app.
+ */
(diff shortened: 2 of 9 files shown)


resources/assets/js/app.js
```diff
@@ -1,9 +1,17 @@
+/**
+ * This is where we load and initialize components of the app.
+ */
+
+import $ from 'jquery';
```
DFurnes: It looks like this import isn't used.

weerd: oh ya. I forgot to push up my latest commit that removes it! 😬

config/services.php
```diff
@@ -14,6 +14,11 @@
     |
     */
 
+    'analytics' => [
+        'production' => 'UA-493209-41',
```
DFurnes: Should these be a single open-ended `analytics_id` environment variable, rather than hard-coding a mapping per environment? I could see us wanting to add another separate analytics ID for the Thor environment, for example, and it'd be nice to not need a code change.

resources/assets/js/utilities/Analytics.js
```diff
@@ -0,0 +1,13 @@
+const $ = require('jquery');
+const Analytics = require('@dosomething/analytics');
+
+function init() {
+  Analytics.init();
+
+  // Attach custom form submit events.
+  $(document).delegate('form', 'submit', (event) => {
```
DFurnes: `$.delegate` is deprecated, we should be using `$(document).on('submit', 'form', (event) => ...` here.

resources/views/layouts/google_analytics.blade.php
```diff
@@ -0,0 +1,6 @@
+@if (is_array(config('services.analytics')))
```
DFurnes: Won't this always be an array? It may be more flexible to check against if a single `analytics_id` string config is empty or not and show/hide the snippet based on that.

resources/assets/js/utilities/Analytics.js
weerd: oh. good to know.

config/services.php
weerd: Yeah, that might be better. I thought we had discussed a solution like this, but maybe I misunderstood.

resources/views/layouts/google_analytics.blade.php
```diff
@@ -1,6 +1,7 @@
-@if (is_array(config('services.analytics')))
+@if (env('GOOGLE_ANALYTICS_ID'))
     <script>
         (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)})(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
-        ga('create', '{{ app('env') === 'production' ? config('services.analytics.production') : config('services.analytics.qa') }}', 'auto');ga('send', 'pageview');
+        ga('create', '{{ env('GOOGLE_ANALYTICS_ID') }}', 'auto');ga('send', 'pageview');
```
DFurnes: We should have this as a `services.analytics.id` config setting, since that will get cached by `php artisan config:cache` in production and direct `env()` calls will not.

PR #488, opened 2016-11-11
Client side validation

weerd:
#### What's this PR do?
This PR adds client-side validation to the Northstar user registration form. It makes use of the `dosomething-validation` package, however, we eventually would like to upgrade this to use the newer version of the package once it is completed. This copies the "auth" rules from the Phoenix repository. The newer version of the Validation package comes pre-packaged with a lot of the rules baked in.

👁 

#### Checklist

---
For review:

@angaither

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
   "dependencies": {
     "@dosomething/analytics": "^1.0.2",
     "@dosomething/forge": "^6.7.4",
-    "dosomething-validation": "^0.2.4"
+    "dosomething-validation": "^0.2.4",
+    "mailcheck": "^1.1.1"
   }
 }
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
   },
   "dependencies": {
     "@dosomething/analytics": "^1.0.2",
-    "@dosomething/forge": "^6.7.4"
+    "@dosomething/forge": "^6.7.4",
+    "dosomething-validation": "^0.2.4"
   }
 }
--- a/resources/assets/js/validators/auth.js
+++ b/resources/assets/js/validators/auth.js
@@ -1,0 +1,25 @@
+const Validation = require('dosomething-validation');
+
+// @TODO: This will be removed when Neue's validation gets refactored.
+function validateNotBlank(string, done, success, failure) {
+  if( string !== "" ) {
+    return done({
+      success: true,
+      message: success
+    });
+  } else {
+    return done({
+      success: false,
+      message: failure
+    });
+  }
+}
+
+// ## First Name
+// Greets the user when they enter their first name.
+Validation.registerValidationFunction("first_name", function(string, done) {
+  validateNotBlank(string, done,
+    `Hey, ${string}`,
+    'We need your first name.'
+  );
+});
--- a/resources/assets/js/validators/auth.js
+++ b/resources/assets/js/validators/auth.js
@@ -1,25 +1,231 @@
-const Validation = require('dosomething-validation');
+define(function(require) {
+  "use strict";
 
-// @TODO: This will be removed when Neue's validation gets refactored.
-function validateNotBlank(string, done, success, failure) {
-  if( string !== "" ) {
-    return done({
-      success: true,
-      message: success
-    });
-  } else {
-    return done({
-      success: false,
-      message: failure
-    });
+  const Validation = require('dosomething-validation');
+
+  // We use Mailcheck (https://github.com/mailcheck/mailcheck) to provide
+  // suggestions for typos in email addresses. See the "email" validator below.
+  var Mailcheck = require("mailcheck");
+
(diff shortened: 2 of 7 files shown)


resources/assets/js/validators/auth.js
```diff
@@ -0,0 +1,231 @@
...
+
+  const Validation = require('dosomething-validation');
+
+  // We use Mailcheck (https://github.com/mailcheck/mailcheck) to provide
+  // suggestions for typos in email addresses. See the "email" validator below.
+  var Mailcheck = require("mailcheck");
+
+  // We'll add a few extra domains to the defaults provided.
+  // @see: https://github.com/mailcheck/mailcheck/blob/v1.1.1/src/mailcheck.js#L16-L25
+  Mailcheck.defaultDomains.push("dosomething.org", "sina.com");
```
angaither: what's `sina.com`?

resources/assets/scss/_components/_forms.scss
```diff
@@ -0,0 +1,4 @@
+// Hide asterisks on required form fields.
+.form-required {
+  display: none;
```
angaither: oh, really?

resources/assets/js/validators/auth.js
weerd: :shrug: this was copied over from the Phoenix JS code...

resources/assets/scss/_components/_forms.scss
weerd: We do it on the Phoenix side. I think the whole validation process will be nicer when we upgrade to the newer version that @DFurnes has been working on and I helped out with a bit during hack week.

package.json
```diff
@@ -27,6 +27,8 @@
   },
   "dependencies": {
     "@dosomething/analytics": "^1.0.2",
-    "@dosomething/forge": "^6.7.4"
+    "@dosomething/forge": "^6.7.4",
+    "dosomething-validation": "^0.2.4",
+    "mailcheck": "^1.1.1"
```
sbsmith86: sometimes I get errors when I don't put trailing commas in these files. Has this been working ok for you?

weerd: Ya. I normally _don't_ put trailing commas cause composer.json freaks out. Not sure why composer freaks but npm doesn't...

resources/assets/js/validators/auth.js
```diff
@@ -0,0 +1,231 @@
...
+    );
+  });
+
+  // ## Birthday
+  // Validates correct date input, reasonable birthdate, and says a nice message.
+  Validation.registerValidationFunction("birthday", function(string, done) {
+    var birthday, birthMonth, birthDay, birthYear, format;
+
+    // Date validation format
+    format = 'MM/DD/YYYY';
```
sbsmith86: I'm not sure I understand this. we hard code `format` to be `MM/DD/YYYY` and then check if the format is a particular way in the conditionals below. In what cases would `format` ever be `DD/MM/YYYY`?

weerd: This is mostly all code from the Validation stuff in Phoenix. Nothing new. All I did was removed the `setting()` calls which depended on settings from Phoenix and not needed here. This used to be:

```js
// Use Drupal dateFormat validation format if available
format = setting('dsValidation.dateFormat', 'MM/DD/YYYY');
```
I left it declaring the variable `format` cause these checks happen in a few places and since I'm just repurposing this for now, didn't want to mess with it too much and potentially break something.

weerd: @DFurnes and I have plans to replace all this w/ the new Validation package, but that's not ready yet, so using the old one (which was very much linked to just working in Phoenix)

resources/views/auth/register.blade.php
```diff
@@ -23,33 +23,58 @@
...
             <div class="form-item">
-                <label for="first_name" class="field-label">First Name</label>
-                <input name="first_name" type="text" class="text-field" placeholder="What do we call you?"value="{{ old('first_name') }}" autofocus>
+                <label for="first_name" class="field-label">
+                    <div class="validation">
+                        <div class="validation__label">First Name <span class="form-required" title="This field is required.">*</span></div>
+                        <div class="validation__message"></div>
+                    </div>
+                </label>
+                <input name="first_name" type="text" id="first_name" class="text-field required js-validate" placeholder="What do we call you?"value="{{ old('first_name') }}" autofocus data-validate="first_name" data-validate-required />
```
weerd: lol. I'd call it out too. Good find!

package.json
sbsmith86: cool

resources/assets/js/validators/auth.js
sbsmith86: Ok. Was just wondering how this works in the 🌎 world then. (not needed for merge)

PR #490, opened 2016-11-15
Facebooks bad

### LINKED ISSUE #491: Remove Facebook ID messaging from server side validation
https://trello.com/c/xaMeR5oL/167-4-as-an-admin-i-want-server-side-validations-to-be-the-same-as-phoenix-so-that-users-understand-the-messages

This doesn't pull out all the facebook stuff completely; just the validation related messaging.

weerd:
#### What's this PR do?
This PR updates some of the server side validation language to remove mentions of "Facebook id" which is not a field needed for the registration, and may only end up applying to the API.

It also corrects an earlier bug where I forgot to include the jQuery NPM package in the `package.json` file. Even though I had tested the client-side validation and was working fine locally, it is probably because it was adding a globally stored copy of jQuery. 😥 

Fixes #491

👁 

#### Checklist

---
For review: 
@angaither 

cc: @jessleenyc

### DIFF
--- a/app/Http/Controllers/Web/AuthController.php
+++ b/app/Http/Controllers/Web/AuthController.php
@@ -162,10 +162,10 @@ class AuthController extends BaseController
      */
     public function postRegister(Request $request)
     {
-        $existing = $this->registrar->resolve($request);
-
         // If a user exists but has not set a password yet, allow them to
         // "register" to set a new password on their account.
+        $existing = $this->registrar->resolve($request);
+
         if ($existing && $existing->hasPassword()) {
             throw new NorthstarValidationException(['email' => 'A user with that email or mobile has already been registered.']);
         }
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/Web/AuthController.php
```diff
@@ -162,16 +162,21 @@ public function getRegister()
      */
     public function postRegister(Request $request)
     {
+        $existing = $this->registrar->resolve($request);
```
angaither: this line should still be below the comment, ya?

package.json
```diff
@@ -29,6 +29,7 @@
     "@dosomething/analytics": "^1.0.2",
     "@dosomething/forge": "^6.7.4",
     "dosomething-validation": "^0.2.4",
+    "jquery": "^3.1.1",
```
angaither: tricky

PR #492, opened 2016-11-17
Updating the back link to send back to origin uri if available.

weerd:
#### What's this PR do?
This PR adds a simple script to manipulate the "Back" link on Northstar so that when a user arrives to the login page from a page on Phoenix, they can click the link and head back exactly to that last page on Phoenix.

👁 

#### Additional Info
So there _can_ be wonky behavior if only using Northstar... for example, if you go Northstar Login page, click to "create a new account", and go to the Registration page then click "back", you will go back to the Login page. But if you now click "back" on the login page, you go to the Registration page, and thus can end up going in a bit of a loop. This is the initial step to adding this functionality. I believe the next step is to pass an origin path from Phoenix that Northstar can use to then use when returning from different NS pages to the login page and then informing the login page which Phoenix page to go back to.

#### Checklist

---
For review: 
@sbsmith86 @angaither 

cc: @jessleenyc

### DIFF
--- a/resources/assets/js/app.js
+++ b/resources/assets/js/app.js
@@ -18,9 +18,13 @@ window.Drupal = {
 
 // Utilities
 import Analytics from './utilities/Analytics';
+import DeLorean from './utilities/DeLorean';
 
 // Register validation rules.
 import './validators/auth';
 
 // Initialize analytics.
 Analytics.init();
+
+// Initialize routing back to last page.
+DeLorean.init();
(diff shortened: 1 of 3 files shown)


resources/assets/js/app.js
```diff
@@ -18,9 +18,13 @@ window.Drupal = {
 
 // Utilities
 import Analytics from './utilities/Analytics';
+import DeLorean from './utilities/DeLorean';
```
sbsmith86: 😂 love the name

weerd: 😉

PR #496, opened 2016-11-17
Adding new PHP to JS vars package and making a better back button.

### LINKED ISSUE #495: Improve Back link to send user back to correct Phoenix referrer uri
Mainly for the `/login` route.

weerd:
#### What's this PR do?
This PR snags the `referrer_uri` that was implemented in [Phoenix](https://github.com/DoSomething/phoenix/pull/7199) and uses it to help provide a more logical flow w/ the view flows presented to the user on Northstar. It helps send the user back to the last page they were on in Phoenix after they clicked "sign up" or "log in" (anything that would send them to Northstar for authentication), while also allowing them to use that same "back" link to navigate the current forms on Northstar; it only sends them back to the page they came from on Phoenix if they are on the `/login` or `/` routes in Northstar and click that "back" link.

It also implements the [PHP-Vars-To-Js-Transformer](https://github.com/laracasts/PHP-Vars-To-Js-Transformer) package by the all-knowing code sage, Jeff Way. This package allows us to set variables from PHP and pop them into a namespaced JS object accessible at `window.Northstar`, to help with passing information from the back-end to the front-end!

Fixes #495

👁 

#### Checklist

---
For review:
@angaither @sbsmith86 

cc: @jessleenyc 

@chloealee (pinging you since you seemed interested in the usage of the _PHP-Vars-To-Js-Transformer_ package 😄

### DIFF
--- a/app/Http/Controllers/Web/AuthController.php
+++ b/app/Http/Controllers/Web/AuthController.php
@@ -54,7 +54,7 @@ class AuthController extends BaseController
         $this->oauth = $oauth;
 
         $this->middleware('guest:web', ['only' => ['getLogin', 'postLogin', 'getRegister', 'postRegister']]);
-        $this->middleware('sessionVars');
+        $this->middleware('session_vars');
     }
 
     /**
--- a/app/Http/Controllers/Web/AuthController.php
+++ b/app/Http/Controllers/Web/AuthController.php
@@ -72,6 +72,9 @@ class AuthController extends BaseController
         // Store the Client ID so we can set user source on registrations.
         session(['authorize_client_id' => request()->query('client_id')]);
 
+        // Store the referrer URI so we can redirect back to it if necessary.
+        session(['referrer_uri' => request()->query('referrer_uri')]);
+
         if (! $this->auth->guard('web')->check()) {
             $destination = request()->query('destination', $client->getName());
             session(['destination' => $destination]);
@@ -97,6 +100,10 @@ class AuthController extends BaseController
      */
     public function getLogin()
     {
+        app('JavaScript')->put([
+            'referrerUri' => session('referrer_uri'),
+        ]);
+
         return view('auth.login');
     }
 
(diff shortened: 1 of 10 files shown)


app/Http/Controllers/Web/AuthController.php
```diff
@@ -54,6 +54,7 @@ public function __construct(Auth $auth, Registrar $registrar, AuthorizationServe
         $this->oauth = $oauth;
 
         $this->middleware('guest:web', ['only' => ['getLogin', 'postLogin', 'getRegister', 'postRegister']]);
+        $this->middleware('sessionVars');
```
angaither: should this be 🐫 ?

weerd: I wasn't sure... all the other middleware names are single words... not sure what the standard is here 😬

resources/assets/js/utilities/DeLorean.js
```diff
@@ -1,15 +1,49 @@
...
 
 function init(element = 'back') {
-  let backLink = document.getElementById(element);
-  let originUri = document.referrer;
+  $(document).ready(() => {
+
+    let backLink = document.getElementById(element);
+    let referrerUri = document.referrer;
+
+    if (isSpecifiedRoute('/login') || isSpecifiedRoute('/')) {
```
weerd: bascially, the back link needs to behave a little different if on those pages.

sbsmith86: Ahh got it got it

DFurnes:
Did we consider having the "back" button trigger `window.history.back()` instead of relying on the referrer header? It feels like this is adding a fair amount of extra complexity to keep track of.

/cc @weerd for next week

weerd:
Yep, I looked into this, but it wouldn't fulfill the requirements for what we wanted it to do based on what I saw. I can look into it again if needed; might be something I missed.

weerd:
Actually, I looked into the `history.pushState()` and `history.replaceState()`, but not `back()`.... I can update later to simplify if it ends up working correctly.

DFurnes:
Cool! Definitely not high priority, but would be nice if it does end up working out. 😄

jessleenyc:
[As a dev,  I want to use the "back" button trigger window.history.back() instead of relying on the referrer header.](https://trello.com/c/FQs3np0y/194-as-a-dev-i-want-to-use-the-back-button-trigger-window-history-back-instead-of-relying-on-the-referrer-header)

PR #499, opened 2016-11-21
Use admin API account when registering Phoenix users.

DFurnes:
#### What's this PR do?
This updates our Phoenix API wrapper to use admin API credentials when registering new users, because we're lockin' it down in DoSomething/phoenix#7208. This will prevent rapscallions from sending bogus data to that endpoint!

Must be deployed before merging DoSomething/phoenix#7208.

👀

#### Checklist

---
For review: …

### DIFF
--- a/app/Services/Phoenix.php
+++ b/app/Services/Phoenix.php
@@ -117,7 +117,7 @@ class Phoenix
         $payload = $user->toArray();
 
         // Format user object for consumption by Drupal API.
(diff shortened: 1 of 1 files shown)


DFurnes:
[(4) As a user, I want my Phoenix & NS accounts connected](https://trello.com/c/GaB9z96k/177-4-as-a-user-i-want-my-phoenix-ns-accounts-connected)

PR #503, opened 2016-11-22
Create Modernizr script so it doesn't 404.

DFurnes:
#### What's this PR do?
When testing the Northstar user flow, I noticed that the Modernizr script referenced in our layout file does not exist. This pull request adds a step to build that file, since otherwise it'll always 404.

screen shot 2016-11-22 at 2 27 42 pm

This PR also adds a favicon and touch icon since those were missing. 🎨 

👀

#### Checklist

---
For review: @angaither


angaither:
💯

Issue #406, opened 2016-08-16
Automatically create placeholder Phoenix accounts on registration.


DFurnes:
See update to this issue in [my comment](https://github.com/DoSomething/northstar/issues/406#issuecomment-264937540) below.

~We should remove the `?create_drupal_user` and Northstar's code to create Drupal users – instead, these will be created on-demand the first time a user logs in to Phoenix.~

~Potential gotchas: signup, reportback, and "magic login" flows still expect a Drupal UID. We should probably update these to be possible without (or create the placeholder user on demand).~

The original-original issue:

> The `?create_drupal_user` query parameter has caused some confusion when integrating with other internal services, and doesn't really seem all that useful to have:
> 1. If a user does _not_ have a Drupal ID on their Northstar profile, we _always_ want to create a corresponding Phoenix account. The only case we wouldn't is...
> 2. If the user is coming from Phoenix (the only time we wouldn't want to make an account), they'd have a Drupal ID set (because it's [included in the payload](https://github.com/DoSomething/phoenix/blob/45f8ce9199311d4d560b2f961f503fb6c07848d1/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module#L222))
> 
> 📻

DFurnes:
I guess one "gotcha" scenario is if when we create a mobile user without an email address: currently Northstar doesn't have logic to make `@mobile.import` placeholder Drupal emails when creating the corresponding Phoenix account.

DFurnes:
This may also become more of a non-issue once we complete pulling auth out of Phoenix.

DFurnes:
Going to change this issue – once we move over to using OAuth login in Phoenix, we'll want to completely disable the ability for Northstar to create Phoenix accounts.

jessleenyc:
![](https://github.trello.services/images/mini-trello-icon.png) [As a dev, when oAuth is out of phoenix, I want to disable the ability for Northstar to create Phoenix accounts.](https://trello.com/c/7zCUSYma/145-as-a-dev-when-oauth-is-out-of-phoenix-i-want-to-disable-the-ability-for-northstar-to-create-phoenix-accounts)

DFurnes:
We ran into a problem with this today on Celebs Gone Good – the voting app makes Northstar users without a password, so it doesn't create a corresponding Phoenix account. This means those users also can never signup or reportback on campaigns through Gambit. No good. 🙅‍♂️

[...]

PR #507, opened 2016-12-05
Always make a corresponding Phoenix account.

DFurnes:
#### What's this PR do?
A lot of functionality still relies on a Phoenix account existing for every user. Now that we've shipped Northstar authentication, we don't need every Phoenix account to have a password in Phoenix which simplifies things a lot!

Alongside DoSomething/phoenix#7229, this should ensure all Northstar accounts will have a Drupal ID.

This shouldn't cause any issues, but we should do some thorough testing just in case!

~Relies on DoSomething/phoenix#7229.~ Merged! ✅ 
References #406.

#### Checklist

---
For review: @angaither @weerd

### DIFF
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -208,7 +208,7 @@ class Registrar
             $user->drupal_id = $drupal_id;
         } catch (ClientException $e) {
             // If user already exists (403 Forbidden), try to find the user to get the UID.
-            if ($e->getCode() == 403) {
+            if ($e->getCode() === 403) {
                 $drupal_id = $this->phoenix->getDrupalIdForNorthstarUser($user);
                 $user->drupal_id = $drupal_id;
             }
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -2,6 +2,7 @@
 
 namespace Northstar\Auth;
 
+use GuzzleHttp\Exception\ClientException;
 use Illuminate\Contracts\Hashing\Hasher;
 use Illuminate\Database\Eloquent\ModelNotFoundException;
 use Illuminate\Validation\Factory as Validation;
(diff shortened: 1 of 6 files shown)


DFurnes:
[Always create a Phoenix account when registering users via Northstar API endpoints.](https://trello.com/c/7zCUSYma/145-always-create-a-phoenix-account-when-registering-users-via-northstar-api-endpoints)

app/Auth/Registrar.php
```diff
@@ -185,31 +186,37 @@ public function register($input, $user = null)
...
     {
-        // @TODO: we can't create a Drupal user without an email. Do we just create an @mobile one like we had done previously?
         try {
-            $drupal_id = $this->phoenix->register($user, $password);
+            $drupal_id = $this->phoenix->createDrupalUser($user);
             $user->drupal_id = $drupal_id;
-        } catch (\GuzzleHttp\Exception\ClientException $e) {
+        } catch (ClientException $e) {
             // If user already exists (403 Forbidden), try to find the user to get the UID.
             if ($e->getCode() == 403) {
```
weerd: triple equals?

DFurnes: Ah! Updated in 32a7773609f66e9242e569179471aaaa896cfeeb.

app/Services/Phoenix.php
```diff
@@ -108,18 +109,21 @@ public function campaigns($id = null)
...
      */
-    public function register($user, $password)
+    public function createDrupalUser($user)
     {
         $payload = $user->toArray();
 
         // Format user object for consumption by Drupal API.
         $payload['birthdate'] = format_date($user->birthdate, 'Y-m-d');
         $payload['user_registration_source'] = $user->source;
-        $payload['password'] = $password;
```
angaither: maybe I'm missing this elsewhere, but does it still get included if it is set?

DFurnes: Nope - to simplify things, we're never sending password out of Northstar now. The only case where a password would also be stored within Drupal is if a user registers via the legacy Phoenix registration modal.

DFurnes:
[Unplanned Work: Fix issues caused by Celebs Gone Good users not having Drupal IDs.](https://trello.com/c/kIvMHGsr/202-unplanned-work-fix-issues-caused-by-celebs-gone-good-users-not-having-drupal-ids)

PR #509, opened 2016-12-08
If a user doesn't have a password, ask them to reset it.

DFurnes:
#### What's this PR do?
This pull request adds some better messaging when a passwordless user (for example, someone who registered by voting for Celebs Gone Good) tries to log in to their account.

screen shot 2016-12-08 at 11 09 02 am

#### How should this be reviewed?
👀 and 🚥!

#### Checklist
- [ ] Documentation added for changed endpoints.
- [x] Tests added for new features/bug fixes.
- [ ] Post a message in #api if this includes something that causes a rebuild!

PR #510, opened 2016-12-08
Remove ability to register over an account.

DFurnes:
#### What's this PR do?
We had previously added the ability to "register over" an account if it didn't have a password. This was a little bandaid fix when users who had voted in CGG could not register last year. Now that we're authenticating users in Northstar, we can handle this more elegantly on the login page.

References #509.

👀

#### Checklist


DFurnes:
[Unplanned Work: Fix issues caused by Celebs Gone Good users not having Drupal IDs.](https://trello.com/c/kIvMHGsr/202-unplanned-work-fix-issues-caused-by-celebs-gone-good-users-not-having-drupal-ids)

PR #511, opened 2016-12-08
Expose API for creating password reset links.

DFurnes:
#### What's this PR do?
This pull request adds an endpoint which will allow other DoSomething.org services (for example, the Niche import script) to create password reset links. If the Northstar password reset flow is not enabled, this will proxy to the existing Drupal endpoint.

:eyes:

#### Checklist

### DIFF
--- a/app/Http/routes.php
+++ b/app/Http/routes.php
@@ -57,7 +57,7 @@ $router->group(['prefix' => 'v2', 'middleware' => ['api']], function () use ($ro
     $router->resource('clients', 'ClientController');
 
     // Password Reset
-    $router->resource('resets', 'ResetController');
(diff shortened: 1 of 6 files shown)


app/Http/routes.php
```diff
@@ -56,6 +56,9 @@
     // OAuth Clients
     $router->resource('clients', 'ClientController');
 
+    // Password Reset
+    $router->resource('resets', 'ResetController');
```
weerd: You're only using the `store()` method in the controller, so maybe not define it as a resource, or just limit it to `only` that one? Just a thought, but not vital for merge.

DFurnes: Oh, yeah good catch!

PR #514, opened 2016-12-13
Correctly cast dates when saving to MongoDB.

### LINKED ISSUE #513: Dates are being stored as strings, rather than ISODates in Mongo.
This is causing issues when trying to query by timeframe for data analysis. 🤔 

/cc @sheyd

DFurnes:
#### What's this PR do?
This fixes an issue where date fields (e.g. `created_at`, `updated_at`, `birthdate`) were being stored as ISO-8601 strings rather than `ISODate` objects. This meant that we couldn't query for date ranges with Mongo ("find me records made between these dates!") which is needed for data analysis.

Fixes #513. References #479.

I added a test in acaf5d7 which demonstrates the problem (it will fail since there isn't a record with a casted date). The fix in 0a8d169 fixes the problem (and the failing test!) by correctly casting dates in the base model class.

#### Checklist

### DIFF
--- a/app/Models/Model.php
+++ b/app/Models/Model.php
@@ -6,6 +6,7 @@ use Carbon\Carbon;
 use Exception;
 use InvalidArgumentException;
 use Jenssegers\Mongodb\Eloquent\Model as BaseModel;
+use MongoDB\BSON\UTCDateTime;
 
 /**
  * Base model class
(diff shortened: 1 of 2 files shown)


DFurnes:
Confirming that we just double-checked that IRL and it works as expected! ✅

DFurnes:
[Unplanned Work: Fix data issues to allow data team to query records by time range.](https://trello.com/c/QgCVLGwb/204-unplanned-work-fix-data-issues-to-allow-data-team-to-query-records-by-time-range)

PR #516, opened 2016-12-14
Add allowDiskUse option to query.

DFurnes:
#### What's this PR do?
We need to use the `allowDiskUse` option so we don't run into errors with this command in prod.

😋

#### Checklist


DFurnes:
Made two more fixes (adding a hard cap on total results & prioritizing fixing timestamps before birthdates). Tested the changes manually on Thor and it's working now, though! 😄

PR #519, opened 2016-12-20
Add a command to backfill missing Drupal accounts.

DFurnes:
#### What's this PR do?
This command should (?!) create or link Drupal accounts for any Northstar users who have a `mobile ` but not a `drupal_id` set on their profile. This will prevent fatal Gambit errors for users who were created before #507 was deployed.

Check out the new Artisan command & the included test case. Are there ways this could break things that I haven't thought of? This is so exciting! 🔥

#### Checklist

### DIFF
--- a/app/Console/Commands/BackfillPhoenixAccounts.php
+++ b/app/Console/Commands/BackfillPhoenixAccounts.php
@@ -30,8 +30,7 @@ class BackfillPhoenixAccounts extends Command
      */
     public function fire(Registrar $registrar)
     {
-        // Find all users where the given field is stored as a string type.
-        // @see: https://docs.mongodb.com/manual/reference/operator/query/type/#op._S_type
+        // Find all users with a null `drupal_id` but a non-null `mobile`.
         $users = User::whereNull('drupal_id')
             ->whereNotNull('mobile')
             ->forPage(1, 100000)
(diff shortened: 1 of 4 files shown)


DFurnes:
I also tested this for funsies on my local Northstar/Phoenix and it's all lookin' good:

screen shot 2016-12-20 at 5 20 38 pm

tests/Console/BackfillPhoenixAccountsTest.php
```diff
@@ -0,0 +1,25 @@
...
+
+class BackfillPhoenixAccountsTest extends TestCase
+{
+    /**
+     * Test that it does the thing its supposed to.
+     * GET /users
+     *
+     * @return void
+     */
+    public function testThatItDoesTheThings()
```
angaither: lol

app/Console/Commands/BackfillPhoenixAccounts.php
```diff
@@ -0,0 +1,56 @@
...
+     *
+     * @param Registrar $registrar
+     * @return mixed
+     */
+    public function fire(Registrar $registrar)
+    {
+        // Find all users with a null `drupal_id` but a non-null `mobile`.
+        $users = User::whereNull('drupal_id')
+            ->whereNotNull('mobile')
+            ->forPage(1, 100000)
```
angaither: why the pages?

DFurnes:
Currently re-running this after the fix applied in DoSomething/phoenix#7264.

PR #524, opened 2017-01-09
Fix empty date formatter

DFurnes:
#### What's this PR do?
This pull request fixes an issue where the `format_date()` helper function would print today's date if given `null` rather than an actual time string or object. I noticed this when testing the Mobile Commons syncer, since many of those users do not have birthdates.

👀

#### Checklist

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -37,6 +37,10 @@ function normalize($type = null, $value = null)
  */
 function format_date($date, $format = 'M j, Y')
 {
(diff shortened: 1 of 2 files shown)


DFurnes:
I did too, until noticing every user from MC showed the current date as their birthday! 🙃

It turns out `new Carbon()` is the same as `new Carbon('now')`, which I guess makes more sense than trying to instantiate an object and getting `null` instead. The more you know! 💫

PR #529, opened 2017-01-19
Handle Phoenix server exceptions more gracefully.

DFurnes:
#### What's this PR do?
If Phoenix returns a 500 error, we should log it and continue along. This is consistent with how we treat other errors (like 422s) from Phoenix.

Fixes #529.

👀 

#### Checklist

### DIFF
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -4,6 +4,7 @@ namespace Northstar\Auth;
 
 use Closure;
 use GuzzleHttp\Exception\ClientException;
+use GuzzleHttp\Exception\ServerException;
 use Illuminate\Contracts\Hashing\Hasher;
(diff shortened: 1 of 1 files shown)


app/Auth/Registrar.php
```diff
@@ -224,6 +225,9 @@ public function createDrupalUser($user)
                 logger('Encountered error when creating Drupal user', ['user' => $user, 'error' => $e]);
                 app('stathat')->ezCount('error creating drupal uid for user');
             }
+        } catch (ServerException $e) {
+            logger('Encountered error when creating Drupal user', ['user' => $user, 'error' => $e]);
```
angaither: is there a way to backfill these l8r when phoenix isn't returning terrible errors?

DFurnes: Yeah, it might make sense to queue these up to retry. I'll make an issue for another PR?

PR #531, opened 2017-01-23
Add birthdate to userinfo endpoint.

DFurnes:
#### What's this PR do?
This pull request adds the user's birthdate to the [User Info](https://github.com/DoSomething/northstar/blob/dev/documentation/endpoints/auth.md#get-user-info) endpoint. This allows OpenID Connect clients to pull a user's birthday on login, rather than relying on a manual sync to occur.

This is not normally an issue, since Northstar syncs new accounts to the relevant Phoenix instance upon creation. It does mean that changes to a user's profile after initial registration will not be synced into the Phoenix database which was causing some unexpected behavior when working on DoSomething/phoenix#7283.

👀

#### Checklist

### DIFF
--- a/app/Http/Transformers/UserInfoTransformer.php
+++ b/app/Http/Transformers/UserInfoTransformer.php
@@ -21,6 +21,7 @@ class UserInfoTransformer extends TransformerAbstract
             'family_name' => $user->last_name,
             'email' => $user->email,
             'phone_number' => $user->mobile,
+            'birthdate' => format_date($user->birthdate, 'Y-m-d'),
 
             'address' => [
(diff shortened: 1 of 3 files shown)


DFurnes:
Yowza, this is causing fatal errors when logging into Phoenix after deploying. 😨 Something's not right.

weerd:
Seriously @_@ That seems weird.

DFurnes:
Alrighty, so it looks like Drupal explodes if dates are not provided as UNIX timestamps. Frustratingly, that's [not what the spec shows](http://openid.net/specs/openid-connect-core-1_0.html#rfc.section.5.6.2.1) and it looks like Drupal's OpenID Connect implementation doesn't do anything about that. 😕

weerd:
😱

PR #534, opened 2017-01-30
Removes drupal function from auth validation

deadlybutter:
#### What's this PR do?
Fixes the validations to show the correct messages! Also updated the JS syntax

When you go to the create an account form do you see the proper validation messages?

### DIFF
--- a/resources/assets/js/validators/auth.js
+++ b/resources/assets/js/validators/auth.js
@@ -1,231 +1,226 @@
-define(function(require) {
-  "use strict";
+const Validation = require('dosomething-validation');
 
-  const Validation = require('dosomething-validation');
+// We use Mailcheck (https://github.com/mailcheck/mailcheck) to provide
+// suggestions for typos in email addresses. See the 'email' validator below.
+var Mailcheck = require('mailcheck');
 
-  // We use Mailcheck (https://github.com/mailcheck/mailcheck) to provide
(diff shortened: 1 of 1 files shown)


resources/assets/js/validators/auth.js
```diff
@@ -1,26 +1,26 @@
 define(function(require) {
-  "use strict";
+  'use strict';
```
deadlybutter: this one needs to be double quotes, got caught in my replace all

DFurnes: How come?

DFurnes: I think Babel actually automatically inserts the strict-mode directive when compiling the code, so you could probably remove it altogether (and get rid of the old AMD-style `define` wrapper).

deadlybutter: Sure thing

deadlybutter: And functionality i don't think it makes a difference, I just always saw quotes as the "convention"

jessleenyc:
[(5) As a user, I want to see validation messaging on the Birthday field that makes sense so that I don't see "Cool @age!"](https://trello.com/c/xZPAqIz8/242-5-as-a-user-i-want-to-see-validation-messaging-on-the-birthday-field-that-makes-sense-so-that-i-don-t-see-cool-age)

PR #537, opened 2017-02-01
Adding auth db name to connections config

blisteringherb:
#### What's this PR do?
This fixes this ticket https://github.com/DoSomething/devops/issues/236 where QA Northstar deploys were failing because of the update to the MongoDB package. I updated the format of the database options array to explicitly include authenticating against the db rather than the default admin db.

Check out this deploy https://app.wercker.com/deploystep/589252cb2da796010073b335. This issue here led to a solution. https://github.com/jenssegers/laravel-mongodb/issues/772

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -56,7 +56,7 @@ return [
             'options'  => [
                 env('DB_REPL_SET_NAME') ? ['replicaSet' => env('DB_REPL_SET_NAME', 'rs0')] : [],
                 'database' => env('DB_NAME', 'userapi'),
-            ] ,
+            ],
         ],
 
     ],
--- a/config/database.php
+++ b/config/database.php
@@ -54,7 +54,7 @@ return [
             'password' => env('DB_PASSWORD', ''),
             'database' => env('DB_NAME', 'userapi'),
             'options'  => [
-                env('DB_REPL_SET_NAME') ? ['replicaSet' => env('DB_REPL_SET_NAME', 'rs0')] : [],
+                'replicaSet' => env('DB_REPL_SET_NAME', 'rs0'),
                 'database' => env('DB_NAME', 'userapi'),
             ],
         ],
--- a/config/database.php
+++ b/config/database.php
@@ -54,7 +54,7 @@ return [
             'password' => env('DB_PASSWORD', ''),
             'database' => env('DB_NAME', 'userapi'),
             'options'  => [
-                'replicaSet' => env('DB_REPL_SET_NAME', 'rs0'),
+                env('DB_REPL_SET_NAME') ? ['replicaSet' => env('DB_REPL_SET_NAME', 'rs0')] : [],
                 'database' => env('DB_NAME', 'userapi'),
             ],
         ],
--- a/config/database.php
+++ b/config/database.php
@@ -54,7 +54,7 @@ return [
             'password' => env('DB_PASSWORD', ''),
             'database' => env('DB_NAME', 'userapi'),
             'options'  => [
-                env('DB_REPL_SET_NAME') ? ['replicaSet' => env('DB_REPL_SET_NAME', 'rs0')] : [],
+                'replicaSet' => env('DB_REPL_SET_NAME'),
                 'database' => env('DB_NAME', 'userapi'),
             ],
         ],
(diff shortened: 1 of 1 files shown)


config/database.php
```diff
@@ -53,7 +53,10 @@
             'username' => env('DB_USERNAME', ''),
             'password' => env('DB_PASSWORD', ''),
             'database' => env('DB_NAME', 'userapi'),
-            'options'  => env('DB_REPL_SET_NAME') ? ['replicaSet' => env('DB_REPL_SET_NAME', 'rs0')] : [],
+            'options'  => [
+                env('DB_REPL_SET_NAME') ? ['replicaSet' => env('DB_REPL_SET_NAME', 'rs0')] : [],
```
DFurnes: This will change the options array to look like this if the `DB_REPL_SET_NAME` env variable is set:
```php
'options' => [
   0 => ['replicaSet' => env('DB_REPL_SET_NAME', 'rs0'],
   'database' => env('DB_NAME', 'userapi')
]
```

If we're using that setting anywhere, it should be fixed to output like this:

```php
'options' => [
   'replicaSet' => env('DB_REPL_SET_NAME', 'rs0'),
   'database' => env('DB_NAME', 'userapi')
]
```

config/database.php
```diff
@@ -53,7 +53,10 @@
...
+                'database' => env('DB_NAME', 'userapi'),
```
DFurnes: I'm surprised that this is the fix, since this same release has deployed successfully to both the Thor and the production environments without a code change. Are we sure that there isn't an issue specific to the QA environment?

DFurnes:
It also seems strange that it re-ran a bunch of old migrations on this deploy.

```
  * 2017-02-01 21:28:38 executing `deploy:artisan_migrate'
  * executing "cd /var/www/northstar/releases/20170201212749 && php artisan migrate --force"
    servers: ["northstar-qa"]
    [northstar-qa] executing command
 ** [out :: northstar-qa] Migrated: 2016_07_05_155130_AddIndexToSource
 ** [out :: northstar-qa] 
 ** [out :: northstar-qa] Migrated: 2016_07_12_195635_AddMoreIndexesToRefreshTokens
 ** [out :: northstar-qa] 
 ** [out :: northstar-qa] Migrated: 2016_07_14_182348_AddIndexToRole
 ** [out :: northstar-qa] 
 ** [out :: northstar-qa] Migrated: 2016_08_31_181524_AddIndexToFacebookId
 ** [out :: northstar-qa] 
 ** [out :: northstar-qa] Migrated: 2016_09_12_155104_AddIndexToAuthCodes
 ** [out :: northstar-qa] 
    command finished in 410ms
```

All of these should have already run on QA and been tracked in the migrations collection. For example, the `AddIndexToFacebookId` migration [ran in this QA deploy](https://app.wercker.com/deploystep/57c993594f6f000100281994) from September.

blisteringherb:
@DFurnes The PR itself didn't run a bunch of migrations. That was some testing I was trying to do that reset the migrations.

DFurnes:
@blisteringherb Did you get a chance to look into this any more? I'm happy to merge (pending a fix for the [formatting bug](https://github.com/DoSomething/northstar/pull/537#discussion_r99136419) mentioned above), but it does seem really strange that a code fix is required when the same code works perfectly in Thor and Production.

blisteringherb:
@DFurnes I've checked the versions and the .env files for differences between the environments. The QA db is running on the same mongo cluster with prod, so the only difference I can think is that the QA db auth needs explicit instructions to use the QA db to authenticate instead of using the admin db by default. Deploying to Thor and to QA seem to work thus far.

DFurnes:
Hmm, yeah that's so strange... it sounds like something broken in the MongoDB package. Either way, 👍 for merge from me after fixing that formatting bug!

config/database.php
blisteringherb: Updating the code to remove the check for whether there's a replSet env var makes the Wercker build fail. https://app.wercker.com/buildstep/589cc1bd9a99bd0100753734 Removing it allows the build to pass. It seems to work OK on QA which has a replSet, and on Thor where there is none.

DFurnes: @blisteringherb Can you try it with the `rs0` default value removed, like this?

```php
'options' => [
   'replicaSet' => env('DB_REPL_SET_NAME'),
   'database' => env('DB_NAME', 'userapi')
]
```

blisteringherb: @DFurnes The build works now.

PR #538, opened 2017-02-06
Adds Localization middleware

deadlybutter:
#### What's this PR do?
Adds Localization middleware to accept Fastly headers and convert that into a language.

1. To check it works I did a `dd(App::getLocale());` on the index route, which works as expected (Use ModHeader to change your fastly header).
2. Was attempting to write a test for this, but running into a wall here.

```
    public function testAppLocale()
    {
        $this->call('GET', '/', [], [], [], ['HTTP_X_FASTLY_COUNTRY_CODE' => 'MX']);
        $this->assertEquals(App::getLocale(), 'es-mx');
    }
```

The assertion fails here, getLocale returns `en`. Any ideas on how to properly test this?

#### Checklist

### DIFF
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -27,7 +27,7 @@ class Kernel extends HttpKernel
             \Illuminate\Session\Middleware\StartSession::class,
             \Illuminate\View\Middleware\ShareErrorsFromSession::class,
             \Northstar\Http\Middleware\VerifyCsrfToken::class,
-            \Northstar\Http\Middleware\Localization::class,
+            \Northstar\Http\Middleware\SetLanguageFromHeader::class,
         ],
         'api' => [
             \Northstar\Http\Middleware\ParseOAuthHeader::class,
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -27,6 +27,7 @@ class Kernel extends HttpKernel
             \Illuminate\Session\Middleware\StartSession::class,
             \Illuminate\View\Middleware\ShareErrorsFromSession::class,
             \Northstar\Http\Middleware\VerifyCsrfToken::class,
+            \Northstar\Http\Middleware\localization::class,
         ],
         'api' => [
             \Northstar\Http\Middleware\ParseOAuthHeader::class,
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -27,7 +27,7 @@ class Kernel extends HttpKernel
             \Illuminate\Session\Middleware\StartSession::class,
             \Illuminate\View\Middleware\ShareErrorsFromSession::class,
             \Northstar\Http\Middleware\VerifyCsrfToken::class,
-            \Northstar\Http\Middleware\localization::class,
+            \Northstar\Http\Middleware\Localization::class,
         ],
         'api' => [
             \Northstar\Http\Middleware\ParseOAuthHeader::class,
--- a/app/Http/Middleware/Localization.php
+++ b/app/Http/Middleware/Localization.php
@@ -8,20 +8,6 @@ use Illuminate\Support\Facades\Request;
 
 class Localization
 {
-
-    /**
-     * Map the given two letter ISO country code to
-     * one of our supported languages or the fallback language.
-     *
-     * @param  String $countryCode
-     * @return String
-     */
-    public function mapCountryToLanguage($countryCode) {
-        $countryCode = strtolower($countryCode);
-
-
-    }
-
     /**
      * Handle an incoming request.
      *
--- a/app/Http/Middleware/Localization.php
+++ b/app/Http/Middleware/Localization.php
@@ -1,0 +1,51 @@
+<?php
+
+namespace Northstar\Http\Middleware;
+
+use Closure;
+use App;
+use Illuminate\Support\Facades\Request;
+
+class Localization
+{
+
+    /**
+     * Map the given two letter ISO country code to
+     * one of our supported languages or the fallback language.
+     *
+     * @param  String $countryCode
+     * @return String
+     */
+    public function mapCountryToLanguage($countryCode) {
+        $countryCode = strtolower($countryCode);
+
+
+    }
+
+    /**
+     * Handle an incoming request.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure  $next
+     * @return mixed
+     */
+    public function handle($request, Closure $next)
+    {
+        $countryCode = Request::server('HTTP_X_FASTLY_COUNTRY_CODE');
+        if ($countryCode) {
+            switch ($countryCode) {
+                case 'us':
+                    App::setLocale('en');
+                    break;
+                case 'mx':
+                    App::setLocale('es-mx');
+                    break;
+                default:
+                    App::setLocale('en');
+                    break;
+            }
+        }
+
+        return $next($request);
+    }
+}
--- a/app/Http/Middleware/SetLanguageFromHeader.php
+++ b/app/Http/Middleware/SetLanguageFromHeader.php
@@ -17,7 +17,7 @@ class SetLanguageFromHeader
      */
     public function handle($request, Closure $next)
     {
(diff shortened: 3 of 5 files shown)


app/Http/Kernel.php
```diff
@@ -27,6 +27,7 @@ class Kernel extends HttpKernel
             \Illuminate\Session\Middleware\StartSession::class,
             \Illuminate\View\Middleware\ShareErrorsFromSession::class,
             \Northstar\Http\Middleware\VerifyCsrfToken::class,
+            \Northstar\Http\Middleware\localization::class,
```
DFurnes: This classname is capitalized incorrectly, which is why you're getting [errors on the build](https://app.wercker.com/buildstep/5898a294a4c1760100eed734):

```
ReflectionException: Class Northstar\Http\Middleware\localization does not exist
```

deadlybutter: https://media.giphy.com/media/TwtXMS5EnKDBK/giphy.gif

DFurnes:
@deadlybutter Did fixing the classname on the middleware fix the issue you described? Either way, can you add a test case like the one you wrote in the PR body?

app/Http/Middleware/Localization.php
```diff
@@ -0,0 +1,37 @@
...
+    /**
+     * Handle an incoming request.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure  $next
+     * @return mixed
+     */
+    public function handle($request, Closure $next)
+    {
+        $countryCode = Request::server('HTTP_X_FASTLY_COUNTRY_CODE');
```
DFurnes: It makes sense to use the Request `$request` object being passed as a parameter to the middleware here, and to ask for the header rather than the raw server variable.

```
$countryCode = $request->header('X-Fastly-Country-Code');
```

app/Http/Middleware/Localization.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+namespace Northstar\Http\Middleware;
+
+use Closure;
+use App;
+use Illuminate\Support\Facades\Request;
+
+class Localization
```
DFurnes: This is nitpicky, but can you change this to an "action" like other middleware, e.g. `SetLocaleFromFastlyHeader`?

deadlybutter:
@DFurnes Sorry let me try to re-explain -- The code works fine, even before the capitalization fix. To test that I wrote `dd(App:getLocale());` inside the index route. This returned `es-mx` when I faked my country code as `MX`. So 👍 

The problem is when I try to write a test case (Like the one above), `App::getLocale()` always returns `en`. I'm not familiar with the Localization aspect of Laravel so I'm not sure if there is a better way to check this?

app/Http/Middleware/Localization.php
deadlybutter: Sure!

DFurnes:
@deadlybutter I believe you were running into trouble because you passed `MX` as an uppercase string, but are only checking for lowercase in the middleware. It makes sense to normalize the case in there before comparing it in the `switch` statement.

I'd also switch to the `get` method on the TestCase since it reads a bit better:

```php
class WebLocalizationTest extends TestCase
{
    public function testLocalizationForMexico()
    {
	    $this->get('/', ['X-Fastly-Country-Code' => 'mx']);
	    $this->assertEquals('es-mx', app()->getLocale());
    }
}
```

deadlybutter:
@DFurnes I had a strtolower in there before, must have removed it on accident. good catch!

deadlybutter:
gah yep, i had strtolower in the old version, forgot to copy it https://github.com/DoSomething/northstar/pull/538/commits/1e876e0f9d527dd7daef21c23da925ed5d15db89#diff-822552e705b187a6ed6a4240dd6d0a1dR20

deadlybutter:
annnd that was it. 👍

tests/Http/LocalizationTest.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+use App as App;
```
DFurnes: This file isn't namespaced, so adding a `use` statement for another un-namespaced class won't do anything. Aliasing that class to have it's own name will do even less.

tests/Http/LocalizationTest.php
```diff
@@ -0,0 +1,17 @@
...
+use Northstar\Models\User;
```
DFurnes: This class isn't used here.

tests/Http/LocalizationTest.php
```diff
@@ -0,0 +1,17 @@
...
+
+class LocalizationTest extends TestCase
+{
+
+    /**
+     * Test that the correct Fastly header is applied for the given header.
+     */
+    public function testAppLocale()
+    {
+        $this->call('GET', '/', [], [], [], ['HTTP_X_FASTLY_COUNTRY_CODE' => 'MX']);
```
DFurnes: I'd really like to use the `$this->get` method on the TestCase, as [suggested above](https://github.com/DoSomething/northstar/pull/538#issuecomment-277776042).

DFurnes:
@deadlybutter Remember that there isn't a race to get this PR merged! Take a little time to double-check your work and make sure all the status checks are passing, and I'll take another look. :)

deadlybutter:
@DFurnes Yep, Ill ping you when it's ready! I just push up out of habbit

deadlybutter:
@DFurnes Great, I think it's ready for a look!

deadlybutter:
Actually Dave, I'm gonna knock out one more task in this since its all the same function

deadlybutter:
Added, just a simple function that sets the language in JS, accessible at `Northstar.language`

screen shot 2017-02-06 at 4 00 44 pm

DFurnes:
I'm not convinced about the value of adding the language to a JS global, although that does remind me that we should probably set that for the `<html>` element's [`lang` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang).

deadlybutter:
@DFurnes Adding it for this task: `turn client side validation off for non-US`

deadlybutter:
@DFurnes I used Session to pass the variable to the layout, seemed like the most straightforward way but open to other options if that's not the convention

resources/views/layouts/app.blade.php
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="{{ Session::get('language') }}">
```
deadlybutter: Hm, is it worth adding an 'en' default here as well just in case?

DFurnes: This doesn't need to be tracked separately in the session:

```
<html lang="{{ App::getLocale() }}">
```

DFurnes: I also don't think there's any need to separately place that value in the session's JavaScript, since you can grab the value from the `lang` property pretty easily on the client-side: `document.documentElement.lang`.

DFurnes: The `en` default should come from the default locale on the application.

deadlybutter: Makes sense to me!

deadlybutter:
Re-tested locally, seems to still work! Thanks for pointing out that code we didn't need @DFurnes

app/Http/Middleware/SetLanguageFromHeader.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+namespace Northstar\Http\Middleware;
+
+use Closure;
+use App;
+use Session;
+use Illuminate\Support\Facades\Request;
```
DFurnes: These two imported classes aren't used: `Session` and `Illuminate\Support\Facades\Request`.

deadlybutter:
@DFurnes Darn, you beat me by a second!!

deadlybutter:
[(13) As a user, I want Northstar to support globalization](https://trello.com/c/TGEjNKlt/210-13-as-a-user-i-want-northstar-to-support-globalization)

PR #541, opened 2017-02-07
Fix incorrect label on log.

DFurnes:
#### What's this PR do?
I'm a real dummy and accidentally logged the opposite of `exists` on this line (see the location of the `!` before and after for details). This made the logs real confusing. 💩

👀

#### Checklist

### DIFF
--- a/app/Auth/Repositories/AuthCodeRepository.php
+++ b/app/Auth/Repositories/AuthCodeRepository.php
@@ -62,10 +62,10 @@ class AuthCodeRepository implements AuthCodeRepositoryInterface
      */
     public function isAuthCodeRevoked($codeId)
     {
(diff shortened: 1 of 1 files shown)


deadlybutter:
Took me a second to realize the difference, this is because its the incorrect value in that log statement?

DFurnes:
> Took me a second to realize the difference, this is because its the incorrect value in that log statement?

Yep. The function is asking whether something is revoked, so "not exists" (which is why we returned `! AuthCode::where(...)->exists()` originally). When I added in the logging, I saved the "negated" version of that exists check to `$exists`, even though it probably should have been named `$revoked`.

This PR just shuffles the `!` around so that the `$exists` variable is what it says it is.

PR #542, opened 2017-02-07
Adds EN Localization & translation functions to all auth templates

deadlybutter:
#### What's this PR do?
This PR adds `{{ trans('...') }}` literally everywhere, and my take for the translation file structure (In EN only).

Once this PR is merged I'll add the `es-mx` translations

Questions
- I assume the `validation.php` is what's being used for server errors, looking at most of these I don't think we have 1:1 translations. How should we proceed?
- Off the top of your head, are there strings anywhere in the code I should be searching for, or is everything we need to worry about in the templates?

I've checked all of the templates to make sure the strings are still rendering correctly, (if the translation name is wrong then it displays that instead)

screen shot 2017-02-07 at 3 48 45 pm
screen shot 2017-02-07 at 3 48 53 pm
screen shot 2017-02-07 at 3 49 01 pm
screen shot 2017-02-07 at 3 49 06 pm

### DIFF
--- a/resources/lang/en/auth.php
+++ b/resources/lang/en/auth.php
@@ -54,5 +54,5 @@ return [
         'terms_of_service' => 'Terms of Service',
         'privacy_policy' => 'Privacy Policy',
         'messaging' => 'and to receive our weekly update. Message &amp; data rates may apply. Text STOP to opt-out, HELP for help.',
-    ]
+    ],
 ];
--- a/resources/lang/en/auth.php
+++ b/resources/lang/en/auth.php
@@ -1,0 +1,50 @@
+<?php
+
+return [
+    'greeting' => [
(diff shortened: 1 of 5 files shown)


deadlybutter:
Alrighty, I think this beast is ready for a look...

ngjo:
[(13) As a user, I want Northstar to support globalization](https://trello.com/c/TGEjNKlt/210-13-as-a-user-i-want-northstar-to-support-globalization)

deadlybutter:
@DFurnes Thanks for getting back on that. @ngjo I guess we should coordinate on getting the validation errors sent to our partner? 

As for the string Dave linked me, I can't find any suitable replacement in Poe editor (Searched for a few variants / similar words). Let's just add this to the list of things to send I guess

PR #549, opened 2017-02-15
Fix Mexico login/registration

deadlybutter:
#### What's this PR do?
Previously I was only importing the validation functions if the users lang was English. However the data attributes still existed which broke our ds-validation module

screen shot 2017-02-15 at 2 38 44 pm

This PR instead,
 - checks user lang
 - if not equal to en
   - remove all `data-validate` attributes

Can you register in both ES-MX and EN?

🇲🇽

### DIFF
--- a/resources/assets/js/app.js
+++ b/resources/assets/js/app.js
@@ -21,7 +21,12 @@ import Analytics from './utilities/Analytics';
 import DeLorean from './utilities/DeLorean';
 
 // Register validation rules for en lang only.
-if (document.documentElement.lang === 'en') require('./validators/auth');
+if (document.documentElement.lang !== 'en') {
+  const $ = require('jquery');
+  $('[data-validate]').each(input => $(input).removeAttr('data-validate'));
+}
+
+import Validation from './validators/auth';
 
 
 // Initialize analytics.
--- a/resources/assets/js/app.js
+++ b/resources/assets/js/app.js
@@ -19,15 +19,13 @@ window.Drupal = {
 // Utilities
 import Analytics from './utilities/Analytics';
 import DeLorean from './utilities/DeLorean';
+import Validation from './validators/auth';
 
 // Register validation rules for en lang only.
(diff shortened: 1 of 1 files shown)


resources/assets/js/app.js
```diff
@@ -21,7 +21,12 @@ import Analytics from './utilities/Analytics';
 import DeLorean from './utilities/DeLorean';
 
 // Register validation rules for en lang only.
-if (document.documentElement.lang === 'en') require('./validators/auth');
+if (document.documentElement.lang !== 'en') {
+  const $ = require('jquery');
+  $('[data-validate]').each(input => $(input).removeAttr('data-validate'));
+}
+
+import Validation from './validators/auth';
```
DFurnes: ES module `import` statements get hoisted to the top of the file, so it's a little confusing to have this down below the conditional. It might make more sense to use a conditional `require('./validators/auth')` inside the above if-statement instead of iterating over the DOM nodes and removing the data attribute.

deadlybutter: @DFurnes As I explained in the PR, we have to load the validators otherwise our base module throws errors and prevents the form submission (see screenshot ^). so we unfortunately have to iterate over the DOM

DFurnes: Oh, because we also require the package [within `Analytics.js`](https://github.com/DoSomething/northstar/blob/22f999910df70777318e1e58eb3b4b0895fe98ef/resources/assets/js/utilities/Analytics.js#L2-L2). Bummer.

Either way, let's move this `import` to the top of the file so it's more clear to read.

deadlybutter: Cool with me!

deadlybutter:
[🐛As a user in MX, I want to be able to create an account 🐛](https://trello.com/c/8ojFnWm4/301-as-a-user-in-mx-i-want-to-be-able-to-create-an-account)

deadlybutter:
@DFurnes Ready for another look!

PR #554, opened 2017-02-27
Wercker docker conversion

blisteringherb:
#### What's this PR do?
Removing unnecessary code from wercker.yml and updating the build box to support new-world Docker deploys on Wercker.

Ensure that the builds are running here
https://app.wercker.com/dosomething/northstar/runs

### DIFF
--- a/wercker.yml
+++ b/wercker.yml
@@ -23,7 +23,7 @@ build:
         code: |-
           mkdir -p $WERCKER_CACHE_DIR/wercker/npm
           npm config set cache $WERCKER_CACHE_DIR/wercker/npm
-          sudo npm install
+          npm install
     - script:
         name: build front-end assets
         code: npm run build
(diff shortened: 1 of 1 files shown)


wercker.yml
```diff
@@ -41,8 +19,11 @@ build:
           cp .env.example .env
           vendor/bin/phpunit
     - script:
-        name: install npm dependencies
-        code: npm install
+        name: npm install
+        code: |-
+          mkdir -p $WERCKER_CACHE_DIR/wercker/npm
+          npm config set cache $WERCKER_CACHE_DIR/wercker/npm
+          sudo npm install
```
DFurnes: Is the `sudo` necessary? It's probably not suuuuper dangerous here since the build box is discarded after running tests, but it's generally considered a bad idea to run package install scripts as the root user.

DFurnes:
Oh actually it looks like the webhook broke here too. 😕 I hit the "Repair Webhook" button, but perhaps something to keep an eye on for future Docker migrations?

PR #558, opened 2017-03-15
Experimental: Fast mode.

DFurnes:
#### What's this PR do?
This adds an experimental "fast mode" which _significantly_ speeds up the `v1/users` index __from around 7 seconds to 142ms on Thor__ by using a database cursor rather than "full" pagination. 🚀 

The downside is that we lose some information in the `meta.pagination` portion of the response (the total number of pages & results), but that seems like a fair trade-off for iterating over a huge dataset.

#### Tell me more!
Standard pagination (where it splits the total results into pages) requires an extra query to get the total sum of the query, which is apparently known to be slow with large datasets in MongoDB.

```js
// the (slow) aggregation query (7.09ms on my local)
users.aggregate([{"$group":{"aggregate":{"$sum":1},"_id":null}}],{"typeMap":{"root":"array","document":"array"}});

// the (much faster) result window query (0.68ms on my local)
users.aggregate([{"$limit":20}],{"typeMap":{"root":"array","document":"array"}});
```

The "cursor" pagination _only_ makes that second query for `$limit + 1`.

I deployed this branch on Thor, so feel free to test it out:

```
GET https://northstar-thor.dosomething.org/v1/users?speed=fast
```

#### Checklist

---
For review: @sheyd @weerd @deadlybutter

### DIFF
--- a/app/Http/Controllers/Traits/TransformsResponses.php
+++ b/app/Http/Controllers/Traits/TransformsResponses.php
@@ -90,7 +90,7 @@ trait TransformsResponses
         $pages = (int) $request->query('limit', 20);
 
         // Experimental: Should we go at warp speed?
-        $fastMode = $request->query('speed') === 'fast';
+        $fastMode = $request->query('pagination') === 'cursor';
         if ($fastMode) {
             $paginator = $query->simplePaginate(min($pages, 100));
         } else {
--- a/app/Http/Controllers/Traits/TransformsResponses.php
+++ b/app/Http/Controllers/Traits/TransformsResponses.php
@@ -2,7 +2,9 @@
 
 namespace Northstar\Http\Controllers\Traits;
 
+use Illuminate\Pagination\Paginator;
 use League\Fractal\Manager;
+use League\Fractal\Pagination\Cursor;
 use League\Fractal\Pagination\IlluminatePaginatorAdapter;
 use League\Fractal\Resource\Collection as FractalCollection;
 use League\Fractal\Resource\Item as FractalItem;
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/Traits/TransformsResponses.php
```diff
@@ -86,15 +88,34 @@ public function paginatedCollection($query, $request, $code = 200, $meta = [], $
         }
 
         $pages = (int) $request->query('limit', 20);
-        $paginator = $query->paginate(min($pages, 100));
+
+        // Experimental: Should we go at warp speed?
+        $fastMode = $request->query('speed') === 'fast';
```
weerd: instead of `speed === fast`, maybe `warp === engage`

image

lol, jk

sheyd: I think this is actually a necessary change. Make it so, Number 💯 . :trollface:

DFurnes: Oh man, now I really want to do that. 😢

sheyd: Can we have it as an easter egg? You have to look at the code to find it?

DFurnes:
@sheyd @weerd Great! Updated documentation, and named the parameter `?pagination=cursor`.

DFurnes:
This is now deployed to production! ⚓️

```
GET https://northstar.dosomething.org/v1/users?pagination=cursor
```

deadlybutter:
Late to the party and on my phone, but I also believe paginating in mongo with limit() is actually the least efficient way, I'll try to find a blog post on something when I get in

PR #560, opened 2017-04-20
Adds an experimental merge endpoint.

DFurnes:
#### What's this PR do?
This is totally experimental, but is way nicer than having to try to do this by hand for a bunch of people. If/when we work out some of the kinks, we could expose this as an option in Aurora.

This is going to be used to merge a handful of accounts for the competitions team.

#### Checklist

---
For review: @weerd @deadlybutter

### DIFF
--- a/app/Http/Controllers/MergeController.php
+++ b/app/Http/Controllers/MergeController.php
@@ -80,7 +80,7 @@ class MergeController extends Controller
 
         return $this->item($target, 200, [
             'updated' => array_keys($duplicateFields),
-            'duplicate' => $duplicate
+            'duplicate' => $duplicate,
         ]);
     }
 }
--- a/app/Http/Controllers/MergeController.php
+++ b/app/Http/Controllers/MergeController.php
@@ -1,0 +1,86 @@
+<?php
+
+namespace Northstar\Http\Controllers;
+
+use Northstar\Exceptions\NorthstarValidationException;
+use Northstar\Models\User;
+use Illuminate\Http\Request;
+use Northstar\Http\Transformers\UserTransformer;
+
+class MergeController extends Controller
+{
+    /**
+     * @var UserTransformer
+     */
+    protected $transformer;
+
+    public function __construct()
+    {
+        $this->transformer = new UserTransformer();
+        $this->middleware('role:admin,staff');
+    }
+
+    /**
+     * EXPERIMENTAL: Merge two user accounts into one.
+     * POST /users/:id/merge
+     *
+     * @param string $id - the "destination" account
+     * @param Request $request
+     *
+     * @return \Illuminate\Http\Response
+     * @throws NorthstarValidationException
+     */
+    public function store($id, Request $request)
+    {
+        $this->validate($request, [
+            'id' => ['required', 'exists:users,_id', 'not_in:'.$id],
+        ]);
+
+        /** @var \Northstar\Models\User $target */
+        $target = User::findOrFail($id);
+
+        /** @var \Northstar\Models\User $duplicate */
+        $duplicate = User::findOrFail($request->input('id'));
+
+        // Get all profile fields from the duplicate (except metadata like ID or source).
+        $metadata = ['_id', 'updated_at', 'created_at', 'drupal_id', 'source', 'source_detail', 'role'];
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/MergeController.php
```diff
@@ -0,0 +1,86 @@
...
+        /** @var \Northstar\Models\User $duplicate */
+        $duplicate = User::findOrFail($request->input('id'));
+
+        // Get all profile fields from the duplicate (except metadata like ID or source).
+        $metadata = ['_id', 'updated_at', 'created_at', 'drupal_id', 'source', 'source_detail', 'role'];
+        $duplicateFields = array_except($duplicate->toArray(), $metadata);
+        $duplicateFieldNames = array_keys($duplicateFields);
+
+        // Are there fields we can't automatically merge? Throw an error.
+        if (count(array_intersect_key($target->toArray(), array_flip($duplicateFieldNames)))) {
```
weerd: just curious, why do you flip the array?

app/Http/Controllers/MergeController.php
```diff
@@ -0,0 +1,86 @@
...
+            $errors = array_fill_keys($duplicateFieldNames, 'Cannot merge into non-null field on target.');
+            throw new NorthstarValidationException($errors, ['target' => $target, 'duplicate' => $duplicate]);
+        }
+
+        // Copy the "duplicate" account's fields to the target & unset on the dupe account.
+        $target->fill($duplicateFields);
+        $duplicate->unset($duplicateFieldNames);
+
+        if (empty($duplicate->email) && empty($duplicate->mobile)) {
+            $duplicate->email = 'merged-account-'.$target->id.'@dosomething.invalid';
```
weerd: Any chance this would give you an error for a malformed email `.invalid` not being a 3 char TLD?

app/Http/Controllers/MergeController.php
```diff
@@ -0,0 +1,86 @@
...
+        // Copy over created_at & source information if it's earlier than the target's timestamp.
+        $duplicateUserHasEarlierCreatedTimestamp = $duplicate->created_at->lt($target->created_at);
+        if ($duplicateUserHasEarlierCreatedTimestamp) {
+            $target->created_at = $duplicate->created_at;
+            $target->source = $duplicate->source;
+            $target->source_detail = $duplicate->source_detail;
+        }
+
+        // Are we "pretending" for this request? If so, short-circuit and display the (unsaved) result.
+        if ($request->query('pretend', false)) {
```
weerd: haha, nice.

weerd:
@DFurnes any thoughts on potentially having a separate `DuplicateUsers` model and associated table where we toss the duplicates into so they're officially removed from the main `users` table. Not sure how many duplicates there are but could help with database transactions and future issues if we just remove them and if we do want to save them place them in a separate query-able table?

app/Http/Controllers/MergeController.php
DFurnes: To "intersect" it with the target profile, the field names need to be keys (so `['first_name' => 0, 'last_name' => 1]` rather than `[0 => 'first_name', 1 => 'last_name']`).

app/Http/Controllers/MergeController.php
DFurnes: Nope, that's a [reserved TLD](https://en.wikipedia.org/wiki/.invalid) so it should be safe to use as a placeholder value. It's the same approach we've used with sanitization scripts in Dr00ps.

DFurnes:
Yeah, I wanna revisit this in a few sprints and:
1. Expose it in Aurora so that Hannah can more easily handle these requests.
2. Look into soft-deleting invalid users & providing a useful response for any other services that still reference the deleted/merged ID (maybe just a `301` to the new account?)
3. See if we can do this automatically when users register and a separate SMS account already exists.

weerd:
Ah jes! "Soft-Deletes" would be ideal here :)

DFurnes:
Just added some documentation. Merging!

DFurnes:
[Unplanned Work: Merge Northstar accounts for a competition broadcast.](https://trello.com/c/YkpEDVMu/583-unplanned-work-merge-northstar-accounts-for-a-competition-broadcast)

PR #561, opened 2017-04-20
Unsubscribe page

ScottPaillant:
#### What's this PR do?
Creates a new page in Northstar, to be viewed by users when they want to unsubscribe from a specific competition. References: https://github.com/DoSomething/gladiator/issues/358 , as a step towards competition of it. 

Is this the proper/right way to add this page to Northstar? Is the code clean?

#### Checklist

---
For review: @DFurnes @sbsmith86

### DIFF
--- a/app/Http/routes.php
+++ b/app/Http/routes.php
@@ -24,7 +24,7 @@ $router->group(['namespace' => 'Web', 'guard' => 'web', 'middleware' => ['web']]
     $router->get('logout', 'AuthController@getLogout');
 
     //Unsubscribe
-    $router->get('unsubscribe', 'AuthController@unsubscribe');
+    $router->get('unsubscribe', 'UnsubscribeController@unsubscribe');
 
     // Registration
     $router->get('register', 'AuthController@getRegister');
--- a/app/Http/routes.php
+++ b/app/Http/routes.php
@@ -10,6 +10,16 @@
 // Web Experience for https://northstar.dosomething.org/
 
 $router->group(['namespace' => 'Web', 'guard' => 'web', 'middleware' => ['web']], function () use ($router) {
+    // @TODO: Temporarily added for testing. Remove this.
+    $router->get('test', function (\Illuminate\Http\Request $request) {
+        return [
+            'ip' => $request->ip(),
+            'remote_addr' => $request->server->get('REMOTE_ADDR'),
+            'forwarded' => $request->header('forwarded'),
+            'x-forwarded-for' => $request->header('x-forwarded-for'),
+        ];
+    });
+
     $router->get('/', 'UserController@home');
 
     // Users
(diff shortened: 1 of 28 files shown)


composer.json
```diff
@@ -6,7 +6,7 @@
   "require": {
     "laravel/framework": "5.2.*",
     "guzzlehttp/guzzle": "~6.2.1",
-    "jenssegers/mongodb": "^3.0",
+    "jenssegers/mongodb": "^3.2",
```
DFurnes: We should double-check that this doesn't cause any unexpected issues since it looks like this version of the package only officially support Laravel 5.3 and higher. (Although we should also change this from a `^` to a `~` since that update may have already happened in the background anyways!)

resources/views/auth/subscriptions.blade.php
```diff
@@ -0,0 +1,15 @@
+@extends('layouts.app')
+
+@section('title', 'Unsubscribe | DoSomething.org')
+
+@section('content')
+
+	<div class="container__block -centered ">
+		To unsubscribe from this competition , click 'Unsubscribe' !
```
DFurnes: Wrap this guy in a paragraph tag!

```html
<p>To unsubscribe from this competition , click 'Unsubscribe'!</p>
```

resources/views/auth/subscriptions.blade.php
```diff
@@ -0,0 +1,15 @@
...
+
+		<form role="form" method="POST" action="  ">
+			<a class="button :hover"> Unsubscribe </a>
```
DFurnes: This `:hover` class looks like it might have been added by mistake. In CSS, [`:hover`](https://developer.mozilla.org/en-US/docs/Web/CSS/:hover), [`:active`](https://developer.mozilla.org/en-US/docs/Web/CSS/:active) and so on are "pseudo-classes" that are automatically applied by the browser when certain actions occur (so in this case, the `:hover` rules get applied when a user's mouse hovers over the button).

app/Http/routes.php
```diff
@@ -23,6 +23,9 @@
     $router->post('login', 'AuthController@postLogin');
     $router->get('logout', 'AuthController@getLogout');
 
+    //Unsubscribe
+    $router->get('unsubscribe', 'AuthController@unsubscribe');
```
DFurnes: Let's make a new controller for this, since it's more related to a "subscription" resource than a user itself. That would nicely map to an `index` (list of a user's current subscriptions at `GET /subscriptions`) and `delete` (remove a subscription at `DELETE /subscriptions/:id`) resourceful controller methods.

composer.json
ScottPaillant: Sure thing, I tested with = 3.0 and nothing immediately crashed, I could do some more testing if necessary. After testing set to ~3.0 !

resources/views/auth/subscriptions.blade.php
ScottPaillant: Sure thing!

resources/views/auth/subscriptions.blade.php
ScottPaillant: Yeah, I think initially 'button' didnt show me generate the forge style, so I put hover as well. Works good now though with just 'button'

app/Http/routes.php
ScottPaillant: I agree, especially since in the future we might want a page thats similar where users can manage all their subscriptions (not just including competitions).

PR #563, opened 2017-04-25
Ensure users provide valid country codes.

DFurnes:
#### What's this PR do?
This pull request makes some updates related to the lax formatting we had for `country` field.

2️⃣ Ensure we only return 2-letter country codes in the transformer. This way we won't continue to send bad data to Drupal when users with unsanitized values try to log in.

🔀 Fixes the user profile form to redirect you to the homepage after changing your profile, rather than to the unimplemented `show` route (a blank page)!

🗺 Adds the League's [`iso3166 `](https://github.com/thephpleague/iso3166) package and uses it to display [a dropdown "country" field](https://cloud.githubusercontent.com/assets/583202/25401086/b395401a-29c2-11e7-9f0e-bb5bd7aabd65.png) for users rather than an open-ended text field.

👁 Adds a mutator (to standardize uppercase values) and a `country` validator so that API clients can't accidentally set bad data on the `country` field.

References DoSomething/phoenix#7304.

Should be good to peek commit-by-commit. I also added a lil' functional test at the end.

Here's the [Wercker build](https://app.wercker.com/dosomething/northstar/runs/build/58ff97b6cc05940001b2b6a5) since it no longer links automatically. 😕

#### Checklist

### DIFF
--- a/app/Http/Controllers/Web/UserController.php
+++ b/app/Http/Controllers/Web/UserController.php
@@ -2,7 +2,6 @@
 
 namespace Northstar\Http\Controllers\Web;
 
-use Illuminate\Contracts\Auth\Factory as Auth;
 use Illuminate\Http\Request;
 use Illuminate\Routing\Controller as BaseController;
 use Northstar\Auth\Registrar;
@@ -51,9 +50,8 @@ class UserController extends BaseController
      */
     public function show($id)
     {
-        $user = User::findOrFail($id);
-
-        $authUser = auth()->guard('web')->user()->toArray();
+        // @TODO: Implement this route.
+        return redirect('/');
     }
 
     /**
@@ -100,6 +98,6 @@ class UserController extends BaseController
 
         $user->fill($values)->save();
 
-        return redirect()->route('users.show', $user->id);
+        return redirect('/');
     }
 }
--- a/app/Http/Transformers/UserTransformer.php
+++ b/app/Http/Transformers/UserTransformer.php
@@ -55,7 +55,9 @@ class UserTransformer extends TransformerAbstract
         }
 
         $response['language'] = $user->language;
-        $response['country'] = $user->country;
+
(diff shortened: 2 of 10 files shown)


weerd:
> the League™'s

app/Http/Transformers/UserTransformer.php
```diff
@@ -55,7 +55,9 @@ public function transform(User $user)
         }
 
         $response['language'] = $user->language;
-        $response['country'] = $user->country;
+
+        // Ensure we only return 2-letter country codes.
+        $response['country'] = strlen($user->country) <= 2 ? $user->country : null;
```
weerd: Can the country code be 1 char?

app/Http/Controllers/Web/UserController.php
```diff
@@ -51,9 +50,8 @@ public function home()
      */
     public function show($id)
     {
-        $user = User::findOrFail($id);
-
-        $authUser = auth()->guard('web')->user()->toArray();
+        // @TODO: Implement this route.
```
weerd: 😬

app/helpers.php
```diff
@@ -93,3 +93,13 @@ function client_id()
     // If not an API request, use Client ID from `/authorize` call or just 'northstar'.
     return session('authorize_client_id', 'northstar');
 }
+
+/**
+ * Get a list of countries keyed by ISO country code.
+ */
+function get_countries()
+{
+    $iso = (new League\ISO3166\ISO3166)->getAll();
```
weerd: coooool!

app/Http/Transformers/UserTransformer.php
DFurnes: Ideally not! Yeah, lemme change this to a `==`.

DFurnes: Updated in ef4b379!

PR #566, opened 2017-04-26
Update to use Webpack 2.

DFurnes:
#### What's this PR do?
This pull request updates Northstar to use Webpack 2 and our [updated webpack config](https://github.com/DoSomething/webpack-config/releases/tag/v2.0.0).

👀

#### Checklist

### DIFF
--- a/resources/views/layouts/app.blade.php
+++ b/resources/views/layouts/app.blade.php
@@ -13,7 +13,7 @@
     <link rel="apple-touch-icon-precomposed" href="{{ asset('apple-touch-icon-precomposed.png') }}">
 
     <link rel="stylesheet" href="{{ elixir('app.css', 'dist') }}">
-    <script src="{{ asset('build/modernizr.js') }}"></script>
+    <script src="{{ asset('dist/modernizr.js') }}"></script>
     <meta name="csrf-token" content="{{ csrf_token() }}">
 </head>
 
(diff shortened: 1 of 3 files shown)


resources/views/layouts/app.blade.php
```diff
@@ -12,8 +12,8 @@
     <link rel="shortcut icon" href="{{ asset('favicon.ico') }}">
     <link rel="apple-touch-icon-precomposed" href="{{ asset('apple-touch-icon-precomposed.png') }}">
 
-    <link rel="stylesheet" href="{{ asset('dist/app.css') }}">
-    <script src="{{ asset('dist/modernizr.js') }}"></script>
+    <link rel="stylesheet" href="{{ elixir('app.css', 'dist') }}">
+    <script src="{{ asset('build/modernizr.js') }}"></script>
```
deadlybutter: where does the /build folder come from? is that a ds webpack thing?

DFurnes: Oh I'm a real dummy... that's a leftover from a different approach. Laravel defaults to that `build` directory for `elixir()` so I'd played around with switching over the build path before realizing it was gonna make a nasty diff.

DFurnes: Fixed in 11034e1.

PR #575, opened 2017-05-09
Send new users to Blink.

DFurnes:
#### What's this PR do?
This pull request adds Blink integration, starting by sending new user profiles to Blink whenever we create an account, so that we can keep accurate records in [Customer.io](https://customer.io).

Check things out commit-by-commit and see the included test case.

#### Checklist

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -323,6 +323,34 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
     }
 
     /**
+     * Transform the user model for Blink.
+     *
+     * @return array
+     */
+    public function toBlinkPayload()
+    {
+        return [
+            'id' => $this->id,
+            'email' => $this->email,
+            'mobile' => null, // <-- send once verified.
+            'first_name' => $this->first_name,
+            'last_name' => $this->last_name,
+            'addr_street1' => $this->addr_street1,
+            'addr_street2' => $this->addr_street2,
+            'addr_city' => $this->addr_city,
+            'addr_state' => $this->addr_state,
+            'addr_zip' => $this->addr_zip,
+            'language' => $this->language,
+            'country' => $this->country,
+            'source' => $this->source,
+            'source_detail' => $this->source_detail,
+            'last_authenticated_at' => $this->last_authenticated_at ? $this->last_authenticated_at->toIso8601String() : null,
+            'updated_at' => $this->created_at->toIso8601String(),
+            'created_at' => $this->created_at->toIso8601String(),
+        ];
+    }
+
+    /**
      * Scope a query to get all of the users in a group.
      *
      * @return \Illuminate\Database\Eloquent\Builder
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -336,8 +336,6 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
             'mobile_status' => $this->mobilecommons_status,
             'first_name' => $this->first_name,
             'last_name' => $this->last_name,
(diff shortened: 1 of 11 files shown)


DFurnes:
One open question – what do we do if Blink API requests fail? Do we catch the exception?

app/Providers/AppServiceProvider.php
```diff
@@ -14,10 +15,17 @@ class AppServiceProvider extends ServiceProvider
      */
     public function boot()
     {
-        User::creating(function ($user) {
+        User::creating(function (User $user) {
             // Set source automatically if not provided.
             $user->source = $user->source ?: client_id();
         });
+
+        User::created(function (User $user) {
```
deadlybutter: is it worth keeping a log somewhere if this import fails?

weerd: 🤔

DFurnes: Yeah, I'm not sure! I'm tempted to say we queue this on our own end so it's non-blocking and something we can retry if Blink is down, but that seems to defeat the purpose of Blink.

DFurnes: @sergii-tkachenko I'd love your thoughts here!

app/Models/User.php
```diff
@@ -323,6 +323,34 @@ public static function drupalIDForNorthstarId($northstar_id)
...
+            'addr_city' => $this->addr_city,
+            'addr_state' => $this->addr_state,
+            'addr_zip' => $this->addr_zip,
+            'language' => $this->language,
+            'country' => $this->country,
+            'source' => $this->source,
+            'source_detail' => $this->source_detail,
+            'last_authenticated_at' => $this->last_authenticated_at ? $this->last_authenticated_at->toIso8601String() : null,
+            'updated_at' => $this->created_at->toIso8601String(),
+            'created_at' => $this->created_at->toIso8601String(),
```
sergii-tkachenko: Could you also send `mobile_status`?

app/Providers/AppServiceProvider.php
sergii-tkachenko: Yes, you should just assume blink always returns a positive response. It's still possible that validation fails, or there are network connectivity problems, but that would be edge cases. 

In the first example, we'll be able to track malformed messages in blink logs.
Second, we should treat network connectivity issues as force majeure.

app/Models/User.php
```diff
@@ -323,6 +323,34 @@ public static function drupalIDForNorthstarId($northstar_id)
...
+     */
+    public function toBlinkPayload()
+    {
+        return [
+            'id' => $this->id,
+            'email' => $this->email,
+            'mobile' => null, // <-- send once verified.
+            'first_name' => $this->first_name,
+            'last_name' => $this->last_name,
+            'addr_street1' => $this->addr_street1,
```
sergii-tkachenko: We don't need `addr_street1` and `addr_street2`, you can exclude them.

app/Providers/AppServiceProvider.php
DFurnes: Okay, good to know!

app/Models/User.php
DFurnes: Done in 40f31dd!

app/Models/User.php
DFurnes: Done in 2237f38!

DFurnes:
[(3) Northstar send a request to Blink to send a transactional message](https://trello.com/c/bpERY4m0/569-3-northstar-send-a-request-to-blink-to-send-a-transactional-message)

PR #584, opened 2017-05-11
Save country and language from Fastly in Northstar.

DFurnes:
#### What's this PR do?
This pull request updates the web registration flow to save a user's country code & language to their account. This was originally only saving in Phoenix as part of the OpenID Connect flow, but that meant the "created" hook would not include the full profile for Customer.io.

I made some updates to the base `TestCase` so we can add headers per request, and extracted a `country_code` helper from the original middleware. Currently `country` will be set to the upper-cased code provided by Fastly (or `null`), and `language` will be set to the app locale (either `en` or `es-mx`).

#### Checklist

### DIFF
--- a/app/Http/Controllers/Web/AuthController.php
+++ b/app/Http/Controllers/Web/AuthController.php
@@ -185,7 +185,15 @@ class AuthController extends BaseController
         ]);
 
         // Register and login the user.
-        $user = $this->registrar->register($request->except(User::$internal));
+        $editableFields = $request->except(User::$internal);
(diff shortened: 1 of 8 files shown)


DFurnes:
[(3) Unplanned Work: As Northstar, I want users' country and language to be set on their profile via their Fastly header.](https://trello.com/c/TfVMHCyg/642-3-unplanned-work-as-northstar-i-want-users-country-and-language-to-be-set-on-their-profile-via-their-fastly-header)

PR #590, opened 2017-06-06
Don't upsert source when `created_at` is the same.

DFurnes:
### What's this PR do?
__Background:__ We had added the ability to upsert users' `source` field only if also providing an earlier `created_at` date in the same request (so, for example, we could update a user's "true" source to be from SMS if they'd texted us first, even if we'd first "created" the account via Phoenix afterwards).

__Problem:__ We create a user's MobileCommons account at (effectively) the exact same time as their Northstar account when doing Niche imports. The comparison check allows us to upsert `created_at` if it's the exact same time (rather than "exclusively" earlier), so all those users would end up with `sms` as a source as soon as their MobileCommons ID was processed by [northstar-mobilecommonsd](https://github.com/DoSomething/northstar-mobilecommonsd).

__Solution:__ Only allow clients to update `source` if it's _actually_ an earlier source, not just the same.

References DoSomething/quasar#356.

See the included test case for a demo of the change.

### Checklist

### DIFF
--- a/app/Http/Controllers/UserController.php
+++ b/app/Http/Controllers/UserController.php
@@ -110,7 +110,7 @@ class UserController extends Controller
             // Optionally, allow setting a custom "created_at" (useful for back-filling from other services).
             // We'll only update this value on existing records if it's earlier than the existing timestamp.
             $created_at = $request->input('created_at');
-            $existingUserHasEarlierCreatedTimestamp = $existingUser && $existingUser->created_at->lt(Carbon::createFromTimestamp($created_at));
+            $existingUserHasEarlierCreatedTimestamp = $existingUser && $existingUser->created_at->lte(Carbon::createFromTimestamp($created_at));
             if ($created_at && ! $existingUserHasEarlierCreatedTimestamp) {
                 $user->created_at = $created_at;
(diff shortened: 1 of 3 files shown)


DFurnes:
We've also talked about just removing the ability to "upsert" `source` and `created_at` altogether now that we're done running big user backfills, but holding off on that until we can discuss in a future story grooming.

jamjensen:
@DFurnes saw you merged this pr - does it mean that the issue is fixed?  I'm still seeing an unusual number of new "sms" users and a much less than expected number of new niche users. Here's a look that outlines new users by northstar source: https://data.dosomething.org/looks/577

DFurnes:
Yep, this is deployed to production. That's a bummer that it's still happening... can you send me some example users that are mistakenly filed as `sms` in the dashboard? We've added some more logging to profile changes so we can probably better debug what's going on. 🕵️ 

Let's also move discussion back to the original issue (or Pivotal now I guess?).

jamjensen:
@DFurnes taking this to the pivotal ticket - https://www.pivotaltracker.com/story/show/149379309 

I'll add profiles to the ticket of users we believe should have been tagged with niche as registration source.

PR #593, opened 2017-06-07
Unsubscribe competitors

sbsmith86:
#### What's this PR do?
The PR hooks up the unsubscribe button on the `/unsubscribe` page so that when it is clicked, a request is made to the gladiator API endpoint that unsubscribes users from competitions. 

unsubscribe

 
You should be able to review commit by commit.

### DIFF
--- a/app/Http/Controllers/Web/UnsubscribeController.php
+++ b/app/Http/Controllers/Web/UnsubscribeController.php
@@ -34,7 +34,9 @@ class UnsubscribeController extends BaseController
     }
 
     /**
+     * Post unsubscribe requests to gladiator.
      *
+     * @param \Illuminate\Http\Request $request
      */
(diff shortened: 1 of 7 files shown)


app/Http/Controllers/Web/UnsubscribeController.php
```diff
@@ -2,15 +2,55 @@
...
+     * @param \Illuminate\Http\Request $request
+     */
+    public function postSubscriptions(Request $request)
+    {
+        $user = $request->input('user');
+        $competition = $request->input('competition');
+
+        if ($user && $competition) {
+            try {
+                $response = $this->gladiator->unsubscribeUser($user, $competition);
```
DFurnes: Looks like `$response` is unused here.

PR #594, opened 2017-06-07
Fix issue where we could add `null` values to Mongo documents on updates.

DFurnes:
#### What's this PR do?
This fixes an issue where editing a users profile on Aurora would insert `null` values for any fields that were already empty on that database record.

Unfortunately, this meant as soon as you edited an account without a mobile you'd insert `null` on that field and the Mongo uniqueness constraint would explode (since Mongo bizarrely treats `null` as a distinct value when checking uniqueness). 💥

I added a test case demonstrating the problem & the fix. Originally, the `last_name` field in that example would have been left in the database & now it gets properly cleaned up.

#### Checklist

### DIFF
--- a/app/Models/Model.php
+++ b/app/Models/Model.php
@@ -49,12 +49,25 @@ class Model extends BaseModel
         $clearable = [];
 
         foreach ($this->original as $key => $value) {
+            // If a field was unset from the model's attributes or assigned null, mark it as clearable.
             if (! array_key_exists($key, $this->attributes) || is_null($this->attributes[$key])) {
                 $clearable[$key] = null;
             }
         }
 
         return $clearable;
+    }
+
+    /**
+     * Remove any null values from the model's attributes.
+     *
+     * @return void
+     */
+    public function removeNullAttributes()
+    {
+        $this->attributes = array_filter($this->attributes, function ($value) {
+            return ! is_null($value);
+        });
     }
 
     /**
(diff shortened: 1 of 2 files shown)


DFurnes:
[Unplanned Work: Fix issue where records could become unwritable on Northstar. 🐛](https://trello.com/c/dvb7XHFa/687-unplanned-work-fix-issue-where-records-could-become-unwritable-on-northstar-%F0%9F%90%9B)

deadlybutter:
Yeah I've had Mongo do that to me before on side projects not realizing its a unique value

app/Models/Model.php
```diff
@@ -84,8 +95,12 @@ protected function performInsert(Builder $query, array $options = [])
      */
     protected function performUpdate(Builder $query, array $options = [])
     {
-        $success = parent::performUpdate($query, $options);
+        // Mark existing attributes that can be cleared, and remove any null values that
+        // may have been added in this "update" operation.
         $clearable = $this->getClearable();
+        $this->removeNullAttributes();
```
weerd: So we never set `null` attributes? Mongo doesn't like them for any field?

DFurnes: It only specifically creates problems for us with unique indexes, since two records that both had `null` would conflict. It seems easiest to just have everything work the same, though, and we save some trivial amount of space by removing the fields.

weerd: gotcha. fair enough. oh silly Mongo

weerd:
post-merge 👍

PR #595, opened 2017-06-08
Fix reporting rate limited requests to StatHat.

DFurnes:
#### What's this PR do?
I had set up the event listener for rate limited requests incorrectly before, so it wasn't properly reporting to StatHat. This fixes that so we get reliable metrics here.

I also renamed the other two listeners for consistency, but otherwise it's pretty straightforward! I just added a `Throttled` event and registered it to the `ReportThrottledRequest` listener in the event service provider.

#### Checklist

### DIFF
--- a/app/Events/Throttled.php
+++ b/app/Events/Throttled.php
@@ -1,0 +1,8 @@
+<?php
+
+namespace Northstar\Events;
+
+class Throttled extends Event
+{
+    // ...
+}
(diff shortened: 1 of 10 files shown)


DFurnes:
[Unplanned Work: Fix rate limited requests reporting to StatHat](https://trello.com/c/AvvO0Nlz/690-unplanned-work-fix-rate-limited-requests-reporting-to-stathat)

PR #596, opened 2017-06-12
Experimental: Improved SSO and extracted authentication.

DFurnes:
#### What's this PR do?
This pull request is a little experimental. ☢️ I've been tossing around the idea of using JWTs for stateless web authentication so that each individual app doesn't need to manage it's own separate sessions & users can transparently move between apps without any extra steps.

It shouldn't! 🙅‍♂️ I'd love to chat big picture approach, but code is very uncooked.

#### What's the idea?
Right now, each app individually completes the OAuth flow to get an access & refresh token. This is a battle-tested approach to authentication (used by Google, Blizzard, and like _so_ many other people). It starts to fall apart, though, when the lines between applications blur (like they do with our new campaign template).

One idea I've been tossing around is using the same ingredients (a short-lived stateless JWT and long-lived stateful refresh token) to share sessions between any applications on the same base domain:

 - A user visits some DoSomething.org site and hits the "Log In" button.
 - They get redirected to Northstar and see the login page. Upon successful login, they get a secure cookie on `*.dosomething.org` with a short-lived JWT. They also get a long-lived secure & HTTP-only cookie on `profile.dosomething.org` that can be used to refresh the JWT later.
 - The user is redirected back to the original application, which can now verify the JWT cookie using Northstar's public key. When it expires, they can either redirect or make an AJAX request to Northstar to refresh that cookie.
 - When the user wants to log out, Northstar can just remove the two cookies and they'll be logged out of all the DoSomething.org applications they've visited at the same time.

This gives us most of the security & performance guarantees of our current approach with the user experience benefits we'd get from all applications just reading & writing to a shared session store.

#### Checklist

### DIFF
--- a/app/Auth/NorthstarJwtGuard.php
+++ b/app/Auth/NorthstarJwtGuard.php
@@ -1,0 +1,302 @@
+<?php
+
+namespace Northstar\Auth;
+
+use Illuminate\Auth\GuardHelpers;
+use InvalidArgumentException;
+use Lcobucci\JWT\Parser;
+use Lcobucci\JWT\Signer\Rsa\Sha256;
+use Lcobucci\JWT\ValidationData;
+use League\OAuth2\Server\CryptKey;
+use League\OAuth2\Server\Exception\OAuthServerException;
+use Illuminate\Http\Request;
+use Illuminate\Contracts\Auth\UserProvider;
+use Illuminate\Contracts\Auth\Guard;
+
+class NorthstarJwtGuard implements Guard
+{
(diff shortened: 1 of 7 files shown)


DFurnes:
[As a user, I want more seamless single sign-on when switching between multiple apps on same domain.](https://trello.com/c/3HDSfIc9/692-as-a-user-i-want-more-seamless-single-sign-on-when-switching-between-multiple-apps-on-same-domain)

DFurnes:
Closing this for now.

PR #598, opened 2017-06-29
Allow filtering users by `created_at` or `updated_at` timestamp.

DFurnes:
#### What's this PR do?
This pull request adds support for querying users by timeframes. For example:

```sh
# users updated before 1/1/2010
GET v1/users?before[updated_at]=1/1/2010

# users updated after 1/1/2010
GET v1/users?after[updated_at]=1/1/2010

# users created between 1/1/2010 and 10/25/2011
GET v1/users?after[created_at]=1/1/2010&before[created_at]=10/25/2011
```

It also includes various little touch-ups, like improved docblocks & test helpers.

I'd recommend going commit-by-commit.

#### Checklist

### DIFF
--- a/app/Auth/Normalizer.php
+++ b/app/Auth/Normalizer.php
@@ -2,6 +2,8 @@
 
 namespace Northstar\Auth;
 
+use Carbon\Carbon;
+
 class Normalizer
 {
     /**
@@ -69,6 +71,21 @@ class Normalizer
         }
 
         return $sanitizedValue;
+    }
+
+    /**
+     * Parse a string into a Carbon date.
+     *
+     * @param string[] $strings
+     * @return string
+     */
+    public function dates($strings)
+    {
+        $dates = collect($strings)->map(function ($string) {
+            return new Carbon($string);
+        });
+
+        return $dates->toArray();
     }
 
     /**
--- a/app/Auth/Normalizer.php
+++ b/app/Auth/Normalizer.php
@@ -77,7 +77,7 @@ class Normalizer
      * Parse a string into a Carbon date.
      *
      * @param string[] $strings
-     * @return string
+     * @return Carbon[]
      */
     public function dates($strings)
     {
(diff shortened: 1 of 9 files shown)


DFurnes:
[(5) As a data analyst, I want to be able to query Northstar for users by timeframe.](https://trello.com/c/w6aMPWG1/710-5-as-a-data-analyst-i-want-to-be-able-to-query-northstar-for-users-by-timeframe)

sheyd:
@DFurnes Are date ranges queryable by hours, e.g. `GET v1/users?after[updated_at]=1/1/2010T12:00`?

DFurnes:
@sheyd Yes! There's a fair number of ranges [supported](http://php.net/manual/en/datetime.formats.php), including ISO-8601 datetimes.

app/Auth/Normalizer.php
```diff
@@ -72,6 +74,21 @@ public function mobile($mobile)
     }
 
     /**
+     * Parse a string into a Carbon date.
+     *
+     * @param string[] $strings
```
weerd: stray `[]`?

app/Auth/Normalizer.php
```diff
@@ -72,6 +74,21 @@ public function mobile($mobile)
...
+     * @return string
```
weerd: Also, isn't the return an array and not a string?

app/Auth/Normalizer.php
DFurnes: An [array of strings](https://phpdoc.org/docs/latest/guides/types.html#arrays)!

app/Auth/Normalizer.php
DFurnes: Oops, fixed!

app/Auth/Normalizer.php
weerd: Ah cool!

PR #604, opened 2017-07-17
Installing socialite and setting the library up for usage

deadlybutter:
#### What's this PR do?
- Adds the Socialite package at version [2.0.20](https://github.com/laravel/socialite/tree/2.0) so it works with our Laravel setup.
- Configures provider / alias. It didn't feel like this was a third party service or a framework service so I started a new comment section in the config, but feel free to suggest a better idea!
- Updated the env for the values required for Socialite

For now just to make sure the app still turns on & nothing breaks!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
     "league/iso3166": "^1.0",
     "fideloper/proxy": "^3.3",
     "dosomething/gateway": "^1.2",
-    "laravel/socialite": "2.0.20"
+    "laravel/socialite": "~2.0.20"
   },
   "require-dev": {
     "phpunit/phpunit": "~4.0",
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,8 @@
     "zendframework/zend-diactoros": "^1.3",
     "league/iso3166": "^1.0",
     "fideloper/proxy": "^3.3",
-    "dosomething/gateway": "^1.2"
+    "dosomething/gateway": "^1.2",
+    "laravel/socialite": "2.0.20"
   },
   "require-dev": {
     "phpunit/phpunit": "~4.0",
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -16,7 +16,8 @@
     "zendframework/zend-diactoros": "^1.3",
     "league/iso3166": "^1.0",
     "fideloper/proxy": "^3.3",
-    "dosomething/gateway": "^1.2"
+    "dosomething/gateway": "^1.2",
+    "laravel/socialite": "2.0.20"
```
DFurnes: Can we use `~2.0.20` here, or are we specifically trying to avoid the `2.0.21` release?

config/app.php
```diff
@@ -161,6 +161,11 @@
         Illuminate\View\ViewServiceProvider::class,
 
         /*
+         * Laravel Service Providers
+         */
+        Laravel\Socialite\SocialiteServiceProvider::class,
```
DFurnes: I think it's fine to include this in "third party service providers" since it's not included in the framework.

composer.json
deadlybutter: Nope! I'll update that

config/app.php
deadlybutter: 👍

config/app.php
```diff
@@ -161,6 +161,11 @@
         Illuminate\View\ViewServiceProvider::class,
 
         /*
+         * Laravel Service Providers
```
weerd: Maybe name to `First Party Service Providers...`?

DFurnes: Personally, I don't think it's a super important distinction.

deadlybutter:
Decided to just merge with the third-party list!

PR #605, opened 2017-07-17
Socialite controller

deadlybutter:
#### What's this PR do?
- Removes the old prototype Facebook code
- Adds the new Socialite controller
- Adds some web routes that handle the redirecting & response accepting (Currently just prints the user out)
- Updates env value
- Simple test to make sure the route maps to a redirect

Visit `northstar.dev/facebook/continue`, you should be redirected to Facebook. You should see your profile printed out like so

screen shot 2017-07-17 at 3 08 44 pm

#### Checklist


deadlybutter:
Simple test added to confirm user is redirected

app/Http/Controllers/Web/FacebookController.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+namespace Northstar\Http\Controllers\Web;
+
+use Socialite;
+
+class FacebookController extends Controller
+{
+    /**
+     * Redirect the user to the GitHub authentication page.
```
deadlybutter: I should update these comments.

tests/Http/Web/FacebookTest.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+class FacebookTest extends TestCase
+{
+
+    /**
+     * Test that a user is redirected to Facebook
+     * @expectedException \Illuminate\Foundation\Testing\HttpException
```
DFurnes: Are these two annotations necessary?

deadlybutter: @DFurnes Yeah, to stop this from happening

```
1) FacebookTest::testFacebookRedirect
A request to [https://www.facebook.com/v2.8/dialog/oauth?client_id=....] failed. Received status code [404].
```

DFurnes: Not high priority, but try making the HTTP call with `$this->get` instead of `$this->visit`. It shouldn't automatically try to follow the redirect so you can then just make a plain ol' assertion against the response.

deadlybutter: Hmm interesting, I'll give that a try!

PR #606, opened 2017-07-18
Create a user when logging in via Facebook

deadlybutter:
#### What's this PR do?
- Takes some of the public profile fields and converts them into a Northstar user
- Leaves several TODO's for finishing this feature out (at least enough to call it an MVP, still need to think about Facebook scopes for more profile fields)
- Adds some testing

**SECURITY NOTE**: This has a security hole I pointed out in the comments, but it only applies to account merging, which is not handled (yet), so it's safe to merge this PR.

When you hit the facebook continue endpoint you should be logged into a new account

#### Checklist

### DIFF
--- a/app/Http/Controllers/Web/FacebookController.php
+++ b/app/Http/Controllers/Web/FacebookController.php
@@ -54,6 +54,14 @@ class FacebookController extends Controller
      */
     public function handleProviderCallback()
     {
+        // TODO: I think we should verify the Facebook token here to make sure
+        // someone can't send a fake FB response & take over an account.
+        // This will require re-introducing the Facebook service I just removed
+        // in a prior commit. :facepalm:
+        //
+        // https://stackoverflow.com/a/16822904/2129670
+        // https://github.com/DoSomething/northstar/pull/605/files#diff-84b65dc95c66e295b3b94d21c873ea18L46
+
         $facebookUser = Socialite::driver('facebook')->user();
         $northstarUser = User::where('email', '=', $facebookUser->email)->first();
 
@@ -70,7 +78,6 @@ class FacebookController extends Controller
         } else {
             // TODO: Sync the existing user with Facebook fields.
         }
-
 
         $this->auth->guard('web')->login($northstarUser, true);
 
--- a/app/Http/Controllers/Web/FacebookController.php
+++ b/app/Http/Controllers/Web/FacebookController.php
@@ -79,7 +79,6 @@ class FacebookController extends Controller
             // TODO: Sync the existing user with Facebook fields.
         }
 
-
         $this->auth->guard('web')->login($northstarUser, true);
 
         // TODO: Implement a feature flag for Phoenix requests, then uncomment this email request.
--- a/app/Http/Controllers/Web/FacebookController.php
+++ b/app/Http/Controllers/Web/FacebookController.php
@@ -69,7 +69,7 @@ class FacebookController extends Controller
             $fields = [
                 'email' => $facebookUser->email,
                 'facebook_id' => $facebookUser->id,
-                'first_name' => $facebookUser->name, // QUESTION: Should we attempt to split the string here or leave it?
+                'first_name' => explode(" ", $facebookUser->name)[0],
                 'country' => country_code(),
                 'language' => app()->getLocale(),
             ];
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/Web/FacebookController.php
```diff
@@ -23,7 +54,36 @@ public function redirectToProvider()
      */
     public function handleProviderCallback()
     {
-        $user = Socialite::driver('facebook')->user();
-        dd($user);
+        // TODO: I think we should verify the Facebook token here to make sure
+        // someone can't send a fake FB response & take over an account.
+        // This will require re-introducing the Facebook service I just removed
+        // in a prior commit. :facepalm:
```
weerd: Have there been any forum posts or whatever where Socialite and this issue come up?

app/Http/Controllers/Web/FacebookController.php
```diff
@@ -23,7 +54,36 @@ public function redirectToProvider()
...
+        // https://github.com/DoSomething/northstar/pull/605/files#diff-84b65dc95c66e295b3b94d21c873ea18L46
+
+        $facebookUser = Socialite::driver('facebook')->user();
+        $northstarUser = User::where('email', '=', $facebookUser->email)->first();
+
+        if (! $northstarUser) {
+            $fields = [
+                'email' => $facebookUser->email,
+                'facebook_id' => $facebookUser->id,
+                'first_name' => $facebookUser->name, // QUESTION: Should we attempt to split the string here or leave it?
```
weerd: `$facebookUser->name` is the full name? If so, I'd say we could have a method we toss it into and do some quick guessing on how to best split it.

app/Http/Controllers/Web/FacebookController.php
deadlybutter: No idea as I didn't search around. Also, not seeing anything in Socialite for this either, which makes sense given they don't ask for our app token which is needed to verify the user token came from our specific app.

deadlybutter: Did some digging and found this PR actually https://github.com/laravel/socialite/pull/126

deadlybutter: This technically doesn't do exactly what I had in mind, but I think it could accomplish the task for now

app/Http/Controllers/Web/FacebookController.php
deadlybutter: Yep, my full name is first+last, so the first name field for me was "Joe Kent". I was debating how to split it, since I have a feeling, not every name will be able to split on the first space in the string. But probably not worth a ton of effort either, people can always change it.

weerd: Yeah, we can try and make our best educated guess on trying to split it.

deadlybutter: Just did the name split locally. I'm cutting at the first space and if people have a more complex name they'll have to fix that post signup.

deadlybutter:
@weerd Yeah im gonna add a test so for the name splitting. As for the security issue, no shouldn't be an issue to merge this PR, see the PR description ^

deadlybutter:
Added name splitting + styling

app/Http/Controllers/Web/FacebookController.php
```diff
@@ -23,7 +54,36 @@ public function redirectToProvider()
...
+        // https://github.com/DoSomething/northstar/pull/605/files#diff-84b65dc95c66e295b3b94d21c873ea18L46
+
+        $facebookUser = Socialite::driver('facebook')->user();
+        $northstarUser = User::where('email', '=', $facebookUser->email)->first();
+
+        if (! $northstarUser) {
+            $fields = [
+                'email' => $facebookUser->email,
+                'facebook_id' => $facebookUser->id,
+                'first_name' => explode(' ', $facebookUser->name)[0],
```
weerd: Why not make a helper method? It could be something we'll need down the road for other social logins. The method could return an array and then we could assign both first and last name to the NS profile.

deadlybutter: 🤔 👍

app/Http/Controllers/Web/FacebookController.php
deadlybutter: Also, just a clarification note in case it wasn't clear: The token auth will happen for both new accounts & merged accounts, but the security hole is really only a concern on merging accounts.

The other thing that pops into mind is rate limiting like we do on the other auth endpoints, and an invalid FB token would trigger the throttle.

deadlybutter:
🚂 🚋 🚋

PR #607, opened 2017-07-19
Verify Facebook token before auth

deadlybutter:
#### What's this PR do?
- This PR uses the socialite `getUserFromToken` function in combination with the token supplied by the callback response. Thus, if a user faked a random string, Facebook graph will not be able to use that token & return the proper user profile.
- Adds some more testing around the token response

- Run the test suite to check for bad token handling, or manually add this change`$requestUser->token = 'blahblah'`
- If you hit the `facebook/continue` endpoint as a regular non-hacker human being nothing should change for you

#### Checklist

### DIFF
--- a/tests/Http/Web/FacebookTest.php
+++ b/tests/Http/Web/FacebookTest.php
@@ -9,7 +9,8 @@ class FacebookTest extends TestCase
      * @param  array  $fields
      * @param  string $method
      */
-    private function mockSocialiteFacade($fields, $method) {
+    private function mockSocialiteFacade($fields, $method)
+    {
         $abstractUser = Mockery::mock('Laravel\Socialite\Two\User');
 
         $user = new Laravel\Socialite\Two\User();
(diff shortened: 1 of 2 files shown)


tests/Http/Web/FacebookTest.php
```diff
@@ -7,19 +7,38 @@ class FacebookTest extends TestCase
...
+     * @param  string  $token token
+     */
+    private function mockSocialiteFromUserToken($email, $name, $id, $token)
+    {
+        $abstractUser = Mockery::mock('Laravel\Socialite\Two\User');
+
+        $user = new Laravel\Socialite\Two\User();
+        $user->map(compact('id', 'name', 'email', 'token'));
+
+        Socialite::shouldReceive('driver->userFromToken')->andReturn($user);
```
DFurnes: Do these need to be two separate methods? Looks like they're always called together with the same parameters.

deadlybutter: The only part thats different across both is `driver->userFromToken` and `driver->user` so I could potentially dry some of this code up. one sec

PR #608, opened 2017-07-19
Merge a Facebook account and a Northstar account on login

deadlybutter:
#### What's this PR do?
- This PR handles cases where a user logs in & the email on Facebook matches a Northstar email.
- In this scenario, we make sure the Facebook Id is set, along with the first+last name. If any of these are not set, we fill them in & save the user. Eventually, this will expand to include birthday, mobile, address, etc

Try logging in via Facebook, then delete either your Facebook ID or a last+first name value. I've been doing this in Mongo with,

```
db.users.update({email: "thedeadlybutter@gmail.com"}, {$set: {"last_name": ""}})
```

Then, log out of Northstar and log back in via Facebook. The missing value should be filled in.

_You can also just run the test!_

#### Checklist

### DIFF
--- a/app/Http/Controllers/Web/FacebookController.php
+++ b/app/Http/Controllers/Web/FacebookController.php
@@ -66,9 +66,9 @@ class FacebookController extends Controller
         }
 
         $northstarUser = User::where('email', '=', $facebookUser->email)->first();
+        $name = get_first_and_last($facebookUser->name);
 
         if (! $northstarUser) {
-            $name = get_first_and_last($facebookUser->name);
             $fields = [
                 'email' => $facebookUser->email,
                 'facebook_id' => $facebookUser->id,
@@ -80,7 +80,11 @@ class FacebookController extends Controller
 
             $northstarUser = $this->registrar->register($fields, null);
         } else {
-            // TODO: Sync the existing user with Facebook fields.
+            // Fill in any blank fields with Facebook values.
+            $northstarUser->facebook_id ? null : $northstarUser->facebook_id = $facebookUser->id;
+            $northstarUser->first_name ? null : $northstarUser->fist_name = $name['first_name'];
+            $northstarUser->last_name ? null : $northstarUser->last_name = $name['last_name'];
+            $northstarUser->save();
         }
 
         $this->auth->guard('web')->login($northstarUser, true);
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/Web/FacebookController.php
```diff
@@ -66,9 +66,9 @@ public function handleProviderCallback()
         }
 
         $northstarUser = User::where('email', '=', $facebookUser->email)->first();
+        $name = get_first_and_last($facebookUser->name);
```
deadlybutter: Interesting find, I can play around with that in a bit!

app/Http/Controllers/Web/FacebookController.php
```diff
@@ -80,7 +80,11 @@ public function handleProviderCallback()
 
             $northstarUser = $this->registrar->register($fields, null);
         } else {
-            // TODO: Sync the existing user with Facebook fields.
+            // Fill in any blank fields with Facebook values.
+            $northstarUser->facebook_id ? null : $northstarUser->facebook_id = $facebookUser->id;
```
deadlybutter: fancy, works for me!

deadlybutter: @DFurnes Actually whats the array_filter for?

DFurnes: Removes null (specifically, non-truthy) values from the array.

DFurnes: To be clearer, we could also make a `fillUnlessNull` method on the base model.

deadlybutter: I think that could be interesting & useful, I'll go ahead & do that!

deadlybutter:
@DFurnes Great, updated with the new method on the user model, let me know if that reads better

app/Models/User.php
```diff
@@ -375,4 +375,16 @@ public function scopeGroup($query, $id)
...
+    /**
+     * Fill & save the user with the given array of fields.
+     * Filter out any fields that have a null value.
+     *
+     * @param  array $fields
+     */
+    public function fillUnlessNull($fields)
+    {
+        $this->fill(array_filter($fields));
+        $this->save();
```
DFurnes: This method shouldn't call `save`.

deadlybutter: Okie dokes, swapped it back to the controller

PR #610, opened 2017-07-21
Happy birthday

deadlybutter:
#### What's this PR do?
- Adds user_birthday scope and retrieves user birthday
- Refactors the FacebookController a bit
- Refactors the Facebook tests to be a bit more friendly to work with
- Adds some new tests specific for birthday 

1. run all tests (duh)
2. try signing up + logging in again to just make sure it all still works
3. revoke FB perms ([handy link](https://www.facebook.com/settings?tab=applications)) and try again. this time try preventing the DS app from getting the birthday scope.

#### Checklist

### DIFF
--- a/app/Http/Controllers/Web/FacebookController.php
+++ b/app/Http/Controllers/Web/FacebookController.php
@@ -59,7 +59,7 @@ class FacebookController extends Controller
     {
         $requestUser = Socialite::driver('facebook')->user();
 
-        // Grab the user profile using there oauth token.
+        // Grab the user profile using their oauth token.
         try {
             $facebookUser = Socialite::driver('facebook')
                 ->fields(['email', 'first_name', 'last_name', 'birthday'])
--- a/app/Http/Controllers/Web/FacebookController.php
+++ b/app/Http/Controllers/Web/FacebookController.php
@@ -45,7 +45,9 @@ class FacebookController extends Controller
      */
     public function redirectToProvider()
     {
-        return Socialite::driver('facebook')->redirect();
+        return Socialite::driver('facebook')
+            ->scopes(['user_birthday'])
+            ->redirect();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/Web/FacebookController.php
```diff
@@ -57,35 +59,38 @@ public function handleProviderCallback()
...
-                'first_name' => $name['first_name'],
-                'last_name' => $name['last_name'],
-                'country' => country_code(),
-                'language' => app()->getLocale(),
-            ];
+        if ($northstarUser) {
+            $northstarUser->fillUnlessNull($fields);
+            $northstarUser->save();
+        } else {
+            $fields['email'] = $facebookUser->email;
```
DFurnes: This isn't necessary for this pull request, but we should see if we can pull this out into the Registrar so we can share the logic for setting these fields between Facebook auth and the "standard" registration form.

app/helpers.php
```diff
@@ -138,17 +138,22 @@ function array_replace_keys($array, $keys, $value)
...
- * @return array
+ * @param  string $birthday
+ * @return date|null
  */
-function get_first_and_last($full_name)
+function format_birthdate($birthdate)
 {
-    $parts = explode(' ', $full_name);
+    if (is_null($birthdate) || empty($birthdate)) {
+        return null;
```
DFurnes: Are these cases significantly different than what's already handled in `format_date`? I think they might already be handled by the `is_null` and Carbon `InvalidArgumentException` checks in there.

deadlybutter: The important exception I wanted to cover is the next `if` block where it checks if it has 3 slashes in the string. I figured might as well do a quick check to see if its empty or not as well!

app/Http/Controllers/Web/FacebookController.php
deadlybutter: Sure! I'll think on it

app/helpers.php
deadlybutter: im gonna quickly try removing this & just using format_date to see what happens

DFurnes: Hmm, got it. That makes sense.

DFurnes: Tested it out on my local, and you're right on the two-slashes scenario. Bummer that not everyone would have their year on their Facebook profile, but that makes sense as a response.

```php
new Carbon\Carbon('1/5');
=> Carbon\Carbon {#833
     +"date": "2017-01-05 00:00:00.000000",
     +"timezone_type": 3,
     +"timezone": "UTC",
   }
```

deadlybutter: Yep just did it as well and got the same dealio

deadlybutter:
```

1) FacebookTest::testFacebookPartialBirthday
Failed asserting that Carbon\Carbon Object &000000007592486f0000000028c63bc5 (
    'date' => '2017-01-01 00:00:00.000000'
    'timezone_type' => 3
    'timezone' => 'UTC'
) is null.

/home/vagrant/Sites/northstar/tests/Http/Web/FacebookTest.php:175
```

deadlybutter:
@DFurnes Good to merge then?

DFurnes:
Yep!

PR #611, opened 2017-07-26
New experience

deadlybutter:
#### What's this PR do?
- Adds a new extended layout
- Updates the register page to have the dual layout
- Makes the default page for unauthenticated users the register instead of login 

NOTE: In order to maintain **Small PR's** ™️ this is not the final Luke design, but I talked with Ashley and we're ok with this being its own PR at its current state. The next PR will make everything look closer to the mock, and we'll need another one at some point for the header.

screen shot 2017-07-26 at 2 11 45 pm
screen shot 2017-07-26 at 2 11 35 pm

NOTE 2:
The cover image is a div since I intentionally wanted to use `background-cover`, given the images being thrown at us by clients can be of any dimension and we want a simple way to handle anything. If there is a better way of doing this via an Image tag, I'm all ears!


weerd:
> The cover image is a div since I intentionally wanted to use background-cover, given the images being thrown at us by clients can be of any dimension and we want a simple way to handle anything. If there is a better way of doing this via an Image tag, I'm all ears!

In this scenario using the image as a `background-image` is the most ideal. You just need to think of it as "is this image part of the content or is it just a complementary image for enhancing the design"? In this case it's the latter; you could remove the background image and it's _not_ crucial to the login form. When the image _is_ part of the content, then you want it as an ``. ✨

resources/assets/scss/_regions/_navigation.scss
```diff
@@ -7,6 +7,10 @@
   @include media($tablet) {
     max-width: $container-max-width;
     padding: $section-spacing ($base-spacing / 2) ($base-spacing / 2);
+
+    &.-extended {
```
DFurnes: It's a bummer that we need to do this in more than one place. Might be nice to pull the container up a level, and then have full-width `.navigation` and `.card` (as a visual replacement for the existing `.container.-framed`.

resources/assets/scss/_components/_cover-photo.scss
```diff
@@ -0,0 +1,14 @@
+.cover-photo {
+  height: 300px;
+  margin-top: $base-spacing;
+  background-size: cover;
+  background-position: center center;
```
DFurnes: Silly little nitpick, but you could replace this with `background-position: center` shorthand! ✋

app/Exceptions/Handler.php
```diff
@@ -125,7 +125,7 @@ protected function unauthenticated($request, AuthenticationException $e)
             return $this->buildJsonResponse(new HttpException(401, 'Unauthorized.'));
         }
 
-        return redirect()->guest('login');
+        return redirect()->guest('register');
```
DFurnes: Is this part something we'd want to consider A/B testing? (Just a question, not for this pull request.)

resources/assets/scss/_regions/_navigation.scss
deadlybutter: Hm, not sure I'm following. Can we chat more about this?

resources/assets/scss/_components/_cover-photo.scss
deadlybutter: 👍

app/Exceptions/Handler.php
deadlybutter: Couldn't hurt!

resources/assets/scss/_regions/_navigation.scss
DFurnes: Yes! Also not something that needs to block merging.

PR #613, opened 2017-07-28
More styling for the register page, pulling in new variables

deadlybutter:
#### What's this PR do?
- More styling to the registration page so it nearly matches the mock
- Pulling in the new query params 

screen shot 2017-07-28 at 10 42 42 am
screen shot 2017-07-28 at 10 42 31 am

#### Checklist
**- [ ] Documentation added for changed endpoints.** ([ref](https://github.com/DoSomething/northstar/blob/02795b9f2f7a3ad0ca9fff016e457e29a580a410/documentation/endpoints/auth.md#L27))
**- [ ] Tests added for new features/bug fixes.**

@DFurnes What's the best way to test the session stuff locally?

### DIFF
--- a/app/Http/Controllers/Web/AuthController.php
+++ b/app/Http/Controllers/Web/AuthController.php
@@ -186,7 +186,7 @@ class AuthController extends BaseController
             'birthdate' => 'required|date',
             'email' => 'required|email|unique:users',
             'mobile' => 'mobile|unique:users',
-            'password' => 'required|confirmed|min:6|max:512',
+            'password' => 'required|min:6|max:512',
         ]);
 
         // Register and login the user.
--- a/app/Http/Controllers/Web/AuthController.php
+++ b/app/Http/Controllers/Web/AuthController.php
@@ -77,7 +77,6 @@ class AuthController extends BaseController
         session(['referrer_uri' => request()->query('referrer_uri')]);
 
         if (! $this->auth->guard('web')->check()) {
-            // TODO: Wat is this
             $authorizationRoute = request()->query('mode') === 'login' ? 'login' : 'register';
 
             session([
--- a/app/Http/Controllers/Web/AuthController.php
+++ b/app/Http/Controllers/Web/AuthController.php
@@ -78,8 +78,7 @@ class AuthController extends BaseController
 
         if (! $this->auth->guard('web')->check()) {
             $authorizationRoute = request()->query('mode') === 'register' ? 'register' : 'login';
-            $destination = request()->query('destination', $client->getName());
-            session(['destination' => $destination]);
+            set_auth_experience_session();
 
             return redirect()->guest($authorizationRoute);
         }
--- a/app/Http/Controllers/Web/AuthController.php
+++ b/app/Http/Controllers/Web/AuthController.php
@@ -77,7 +77,8 @@ class AuthController extends BaseController
         session(['referrer_uri' => request()->query('referrer_uri')]);
 
         if (! $this->auth->guard('web')->check()) {
-            $authorizationRoute = request()->query('mode') === 'register' ? 'register' : 'login';
+            // TODO: Wat is this
+            $authorizationRoute = request()->query('mode') === 'login' ? 'login' : 'register';
 
             session([
                 'destination' => request()->query('destination', $client->getName()),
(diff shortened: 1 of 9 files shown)


deadlybutter:
Ah, should probably remove the code related to the password confirmation as well, one sec

app/Http/Controllers/Web/AuthController.php
```diff
@@ -78,8 +78,7 @@ public function authorize(ServerRequestInterface $request, ResponseInterface $re
 
         if (! $this->auth->guard('web')->check()) {
             $authorizationRoute = request()->query('mode') === 'register' ? 'register' : 'login';
-            $destination = request()->query('destination', $client->getName());
-            session(['destination' => $destination]);
+            set_auth_experience_session();
```
DFurnes: I don't think this needs to be a helper since it's not used in more than one place.

app/helpers.php
```diff
@@ -157,3 +157,16 @@ function format_birthdate($birthdate)
...
     return format_date($birthdate, 'Y-m-d');
 }
+
+/**
+ * Set the session variables from the auth request.
+ */
+function set_auth_experience_session()
+{
+    session([
+        'destination' => request()->query('destination'),
```
DFurnes: Was the fallback to the client name removed on purpose?

DFurnes:
> @DFurnes What's the best way to test the session stuff locally?

Make a test case! If you visit the `/authorize` endpoint with your additional query parameters, you should be able to follow the redirect and then assert the login page looks as expected.

app/Http/Controllers/Web/AuthController.php
deadlybutter: Alrighty!

app/helpers.php
deadlybutter: When I moved it to the helper file I didn't have access to `$client` so I just did a string default, but since we're moving that back I'll just re-add it!

app/Http/Controllers/Web/AuthController.php
```diff
@@ -77,7 +77,8 @@ public function authorize(ServerRequestInterface $request, ResponseInterface $re
         session(['referrer_uri' => request()->query('referrer_uri')]);
 
         if (! $this->auth->guard('web')->check()) {
-            $authorizationRoute = request()->query('mode') === 'register' ? 'register' : 'login';
+            // TODO: Wat is this
```
deadlybutter: (Removed this is the following commit)

deadlybutter:
@DFurnes Updated this PR so everything should be working & have test coverage, see the latest commit:

https://github.com/DoSomething/northstar/pull/613/commits/7c492cb96834cf8c6991d3001629504af5cf0583

resources/views/auth/register.blade.php
```diff
@@ -22,24 +23,26 @@
...
-                    <div class="validation">
-                        <div class="validation__label">{{ trans('auth.fields.first_name') }} <span class="form-required" title="This field is required.">*</span></div>
-                        <div class="validation__message"></div>
-                    </div>
-                </label>
-                <input name="first_name" type="text" id="first_name" class="text-field required js-validate" placeholder="{{ trans('auth.validation.placeholder.call_you') }}" value="{{ old('first_name') }}" autofocus data-validate="first_name" data-validate-required />
-            </div>
+            <div>
+                <div class="form-item -reduced">
+                    <label for="first_name" class="field-label">
```
DFurnes: I know this isn't new to this pull request, but you don't need a lot of this markup since it's generated automatically by the validation module. See the [example markup](https://github.com/DoSomething/validation/blob/22195ef48e5a6e0f669c219d7999b58247c40d25/example/index.html#L35-L38) included in the package.

```diff
-                <label for="birthdate" class="field-label">
-                    <div class="validation">
-                        <div class="validation__label">{{ trans('auth.fields.birthday') }} <span class="form-required" title="This field is required.">*</span></div>
-                        <div class="validation__message"></div>
-                    </div>
-                </label>
+                <label  for="birthdate" class="field-label">{{ trans('auth.fields.first_name') }}</label>
```

DFurnes:
@deadlybutter Can you update the documentation for the authorize endpoint with the new `title`, `callToAction`, and `coverPhoto` query parameters as well?

deadlybutter:
👍 👍

deadlybutter:
@DFurnes Hows that?

PR #615, opened 2017-08-01
Adds default session values

deadlybutter:
#### What's this PR do?
If an app does not pass along `title`, `callToAction`, etc, there will be a default value to take its place. Unfortunately, I didn't experience this bug until I tried logging in from a separate application and was presented with lots of white space!

### DIFF
--- a/app/Http/Controllers/Web/AuthController.php
+++ b/app/Http/Controllers/Web/AuthController.php
@@ -81,9 +81,9 @@ class AuthController extends BaseController
 
             session([
                 'destination' => request()->query('destination', $client->getName()),
-                'title' => request()->query('title'),
-                'callToAction' => request()->query('callToAction'),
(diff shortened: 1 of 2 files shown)


deadlybutter:
The long version:

Locally I've always seen this

screen shot 2017-08-01 at 1 13 00 pm

The defaults pull in because 

1) on the template itself we specify what those defaults are,
2) `session` was never called, and thus it always went to those defaults

When I tried it logging in from a campaign on Phoenix Ashes, I was given a pretty blank page. This is because, now that we hit the authorize route, the session function was called. And because I didnt merge the phoenix ashes PR for adding `title, cta, etc`, it had nothing to fill for these session values.

I was able to further prove this by
1) Refreshing the page, it's still blank (Blank values are now cached)
2) going to the same page in incognito, default values show (No cached blank values)

PR #616, opened 2017-08-01
Sending facebook id in the blink payload

deadlybutter:
#### What's this PR do?
After some debugging with Sergii we realized Blink was never getting the Facebook ID from Northstar! Whoops.

Just added it to the user function and updated the blink test. (Facebook ID was already in the model generator, just had to ensure it was in the blink payload)


sergii-tkachenko:
Looks right to me.

PR #618, opened 2017-08-03
Facebook detail

deadlybutter:
#### What's this PR do?
- Ensures Facebook users have a source detail specifying they came from Facebook
- Adds a quick addition to the test case for this

Run the tests!

#### Checklist

### DIFF
--- a/app/Http/Controllers/UserController.php
+++ b/app/Http/Controllers/UserController.php
@@ -118,7 +118,7 @@ class UserController extends Controller
             $existingUserHasEarlierCreatedTimestamp = $existingUser && $existingUser->created_at->lte(Carbon::createFromTimestamp($created_at));
             if ($created_at && ! $existingUserHasEarlierCreatedTimestamp) {
                 $user->created_at = $created_at;
-                $user->setSource(client_id(), $request->input('source_detail'));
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/UserController.php
```diff
@@ -118,14 +118,12 @@ public function store(Request $request)
             $existingUserHasEarlierCreatedTimestamp = $existingUser && $existingUser->created_at->lte(Carbon::createFromTimestamp($created_at));
             if ($created_at && ! $existingUserHasEarlierCreatedTimestamp) {
                 $user->created_at = $created_at;
-                $user->source = $request->input('source') ?: client_id();
-                $user->source_detail = $request->input('source_detail');
+                $user->setSource($request->input('source') ?: client_id(), $request->input('source_detail'));
```
DFurnes: You can set a default fallback with the second parameter to `input`:

```php
$user->setSource($request->input('source', client_id()), $request->input('source_detail'));
```

deadlybutter: ✨  thanks for the suggestion!

deadlybutter:
@DFurnes Should be good to go!

PR #620, opened 2017-08-07
Password visibility

deadlybutter:
#### What's this PR do?
- This PR adds a little 👁 to the password input on the registration page so people can toggle the password visibility while typing. We removed the password confirmation input, so we're adding this for the user to still be able to confirm.

password

- Can you still make a new account?
- Does the toggle work as expected?


deadlybutter:
@DFurnes Ready for review!

resources/assets/js/utilities/PasswordVisibility.js
```diff
@@ -0,0 +1,34 @@
...
+  }
+}
+
+function init() {
+  $(document).ready(() => {
+    const passwordToggles = document.getElementsByClassName('password-visibility__toggle');
+    if (! passwordToggles) return;
+
+    for (const toggle of passwordToggles) {
+      if (toggle) toggle.addEventListener('click', clickHandler);
```
DFurnes: When is `toggle` falsy here? Could we remove this check?

resources/views/auth/register.blade.php
```diff
@@ -52,9 +52,10 @@
                 </div>
             @endif
 
-            <div class="form-item">
+            <div class="form-item password-visibility">
                 <label for="password" class="field-label">{{ trans('auth.fields.password') }}</label>
                 <input name="password" type="password" id="password" class="text-field required js-validate" placeholder="{{ trans('auth.validation.placeholder.password') }}" data-validate="password" data-validate-required data-validate-trigger="#password_confirmation" />
+                <span class="password-visibility__toggle -hide">&#128065;</span>
```
DFurnes: How's this work on platforms that don't support emoji?

resources/assets/js/utilities/PasswordVisibility.js
deadlybutter: Hm true, the one above would likely catch anything wrong. 👍

resources/views/auth/register.blade.php
deadlybutter: 🤔 I guess we could use an image? I didn't want to do anything super over kill for this

DFurnes: An image is probably the most robust choice. I'm not necessarily opposed to using an emoji, but let's make sure to take a peek at how it works in BrowserStack (specifically Android, Windows, Edge, IE11, etc.)

deadlybutter:
@DFurnes Updated to use an svg instead!

screen shot 2017-08-07 at 3 01 58 pm
screen shot 2017-08-07 at 3 01 54 pm

resources/assets/scss/_components/_password-visibility.scss
```diff
@@ -0,0 +1,36 @@
...
+  .password-visibility__toggle {
+    width: 1em;
+    height: 1em;
+    position: absolute;
+    bottom: 0;
+    right: 0;
+    margin-bottom: .8em;
+    margin-right: $base-spacing / 2;
+    user-select: none;
+    background-image: url('/eye.svg');
```
DFurnes: Can you move the SVG into this directory and change this to `url('./eye.svg')` so Webpack can optimize it?

deadlybutter:
✅

PR #621, opened 2017-08-07
Remove ability to backfill 'source' or 'created_at'.

DFurnes:
#### What's this PR do?
__This pull request removes the ability to backfill `source` or `created_at` via Northstar's general-purpose user endpoint.__ This prevents an issue where clock drift between servers was causing it to appear that MobileCommons users were always "older" (by 6 seconds! ⏱) than the one's we'd just created in Northstar, and therefore unexpectedly overwrite sources as `sms`. 

Since we're no longer backfilling old user data into Northstar (and likely won't ever need to again), I think this is safe to remove so we can reduce some complexity.

Closes [#149379309](https://www.pivotaltracker.com/story/show/149379309).

I updated the test suite for the new expected behavior.

#### Checklist

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -397,7 +397,11 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
      */
     public function setSource($source, $detail = null)
     {
-        $this->source = $source;
+        if ($this->source) {
+            return;
+        }
+
+        $this->source = $source ?: client_id();
         $this->source_detail = $detail;
     }
 }
--- a/app/Http/Controllers/UserController.php
+++ b/app/Http/Controllers/UserController.php
@@ -2,7 +2,6 @@
 
 namespace Northstar\Http\Controllers;
 
-use Carbon\Carbon;
 use Illuminate\Http\Request;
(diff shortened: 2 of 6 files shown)


weerd:
> This pull request removes the ability to backfill source or created_at via Northstar's general-purpose user endpoint. This prevents an issue where clock drift between servers was causing it to appear that MobileCommons users were always "older" (by 6 seconds! ⏱) than the one's we'd just created in Northstar, and therefore unexpectedly overwrite sources as sms.

that's so sneaky!

app/Models/User.php
```diff
@@ -397,7 +397,11 @@ public function fillUnlessNull($fields)
      */
     public function setSource($source, $detail = null)
     {
-        $this->source = $source;
+        if ($this->source) {
+            return;
+        }
```
weerd: So we never set `source_detail` if `$this->source` is already set?

DFurnes: That's the idea (since it should always be extended info related to that specific source).

weerd: Ok. So only really set when setting the source the first time! Coolio :)

PR #629, opened 2017-08-31
(Discussion) Adding redirect uri to Facebook

itsjoekent:
#### What's this PR do?
In order to allow for one click auth, we're adding redirect uri support to the Facebook login flow.

At the moment, I did this by placing the `session()->referer` code in the `redirectToProvider()` function. However, I'm now thinking this would break the normal client based redirect, because the `referer_uri` is going to override the intended redirect for the client (Basically, if i click signup, and click login with facebook, the referer uri will be northstar/register, not phoenix/campaign)

So I'm now thinking, maybe we need one more Facebook route specifically for 1 click login that puts the referrer in the session and then sends you to `redirectToProvider`

…

#### Checklist

---
For review: …

### DIFF
--- a/app/Http/Controllers/Web/FacebookController.php
+++ b/app/Http/Controllers/Web/FacebookController.php
@@ -54,6 +54,9 @@ class FacebookController extends Controller
      */
     public function redirectToProvider()
     {
+        // Store the referrer URI so we can redirect back to it if necessary.
+        session()->put('referrer_uri', request()->headers->get('referer'));
+
         return Socialite::driver('facebook')
             ->scopes(['user_birthday'])
             ->redirect();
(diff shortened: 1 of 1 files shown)


DFurnes:
Let's handle this the same way we allow clients to specify [whether to show login or registration page](https://github.com/DoSomething/northstar/blob/84ae526158894cf8aab338308d054d68f8e2d346/app/Http/Controllers/Web/AuthController.php#L80-L89) (so `/authorize?mode=facebook&...` would automatically redirect unauthenticated users to Facebook rather than showing them a login or registration page).

DFurnes:
Going to close this since it's been open for a while.

itsjoekent:
rip

itsjoekent:
press f to pay respects

Issue #570, opened 2017-05-02
Rename `mobilecommons_status` to `mobile_status`


sergii-tkachenko:
In order to generalize mobile accessibility status, I'd like to recommend renaming `mobilecommons_status` to `mobile_status`.

Possible values should stay the same, as declared in [northstar-mobilecommonsd](https://github.com/DoSomething/northstar-mobilecommonsd/blob/dev/app/Jobs/SendUserToNorthstar.php#L78-L87):

- `active` - For reachable users
- `undeliverable` - Invalid mobile, can't receive texts or texted STOP (note: think about re-opting-in)
- `unknown` - For everything else
- `null` for members with no mobile number. It's ok if it's not empty now for them, we can gradually change that.

As we gain access to Twilio message-level data, Blink will be able to update it on Northstar as following:

- `active` - For reachable users
- `undeliverable` - [Twilio errors](https://www.twilio.com/docs/api/rest/message#error-values) `30004`, `30005`, `30006`, or texted STOP
- `unknown` -  [Twilio errors](https://www.twilio.com/docs/api/rest/message#error-values) `30007 `, `30008`
- `null` for members with no mobile number

cc @justkika @sheyd

DFurnes:
> In order to generalize mobile accessibility status, I'd like to recommend renaming `mobilecommons_status` to `mobile_status`.

Yeah, this makes sense to me! 👍

aaronschachter:
Linking to https://github.com/DoSomething/gambit-conversations/issues/63 to document additional `'less'` value to indicate User's broadcast frequency preferences.

If a User texts STOP, we'll set the value to `'undeliverable'`. Per the note above about opting-in again: If user then texts us again after STOP, we will set the value to `'active'` (unless it was a STOP or LESS)

DFurnes:
@aaronschachter Implemented this as `sms_status` in #631, including the additional `less` value. I'm leaving this as a top-level field on the user document for now, since the Laravel MongoDB package we're using makes nested fields a little complicated.

DFurnes:
This is done, closing.

PR #631, opened 2017-09-15
Rename SMS status field & add "paused" field.

DFurnes:
#### What's this PR do?
This pull request includes field updates for the upcoming SMS work:

☎️ It renames the existing `mobilecommons_status` field to `sms_status` so that it's applicable to more messaging platforms in the future. It maintains support for [applications](https://github.com/DoSomething/northstar-mobilecommonsd) that either read or write to the old `mobilecommons_status` field. (#570)

📴 It adds a `sms_paused` field which Gambit can use to track whether a user is in a bot conversation or talking to a support agent (DoSomething/gambit-conversations#65).

📆 It also adds a `last_messaged_at` field for tracking the last time a user messaged us via Gambit (similar to how we track `last_authenticated_at` or `last_accessed_at` for web traffic).

I've added a test for the backwards-compatible `mobilecommons_status` field support, and updated documentation to reference the new field name. It'll be interesting to see how long the migration takes when deploying this to Thor... hopefully not too long!

#### Checklist

### DIFF
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -72,6 +72,7 @@ class Registrar
             'mobilecommons_status' => 'in:active,undeliverable,unknown', // for backwards compatibility.
             'sms_status' => 'in:active,less,undeliverable,unknown',
             'sms_paused' => 'boolean',
+            'last_messaged_at' => 'date',
         ];
 
         // If existing user is provided, merge indexes into the request so
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -69,6 +69,8 @@ class Registrar
             'birthdate' => 'date',
             'country' => 'country',
             'password' => 'min:6|max:512',
+            'mobilecommons_status' => 'in:active,undeliverable,unknown', // for backwards compatibility.
+            'sms_status' => 'in:active,less,undeliverable,unknown',
         ];
 
         // If existing user is provided, merge indexes into the request so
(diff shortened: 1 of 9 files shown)


app/Auth/Registrar.php
```diff
@@ -69,6 +69,10 @@ public function validate(Request $request, User $existingUser = null, array $add
             'birthdate' => 'date',
             'country' => 'country',
             'password' => 'min:6|max:512',
+            'mobilecommons_status' => 'in:active,undeliverable,unknown', // for backwards compatibility.
+            'sms_status' => 'in:active,less,undeliverable,unknown',
```
weerd: what's `less`?

DFurnes: It'll be [used by Gambit](https://github.com/DoSomething/gambit-conversations/issues/63) to track users who want to stay subscribed, but get less messaging.

database/migrations/2017_09_15_152015_RenameMobileStatusField.php
```diff
@@ -0,0 +1,44 @@
...
+     *
+     * @return void
+     */
+    public function down()
+    {
+        $this->renameField('users', 'sms_status', 'mobilecommons_status');
+    }
+
+    /**
+     * Rename the given field on any documents in the collection.
```
weerd: @_@

app/Auth/Registrar.php
weerd: ah! cool. just curious :)

PR #632, opened 2017-09-18
Convert existing mobile numbers to E.164 format.

DFurnes:
#### What's this PR do?
This pull request gets us ready to convert all numbers to [E.164 format](https://www.twilio.com/docs/glossary/what-e164) (which is needed to be able to send broadcasts via [Customer.io](https://customer.io/) in the future, and also is just a nicer way to format this data).

📝 I've added a `northstar:e164` script which iterates over all existing phone numbers and saves an E.164-formatted version in an `e164` field on the user document. We can later swap this to be the "real" mobile field (kinda like what we did [with `mobilecommons_status` and `sms_status`](https://github.com/DoSomething/northstar/pull/631)).

🔀 Adds a mutator so any new values for `mobile` also set a corresponding `e164` attribute on the document. This means any new or edited users will automatically have a correct E.164 formatted number, so we don't miss any records between running the script and swapping the fields.

I've added a test for the script, and the existing tests still pass with the new mutator. 🚥

One thing I'm worried about is performance when running this script against millions of records in production... I'm hoping that `User::chunk` helps with this, but I have a feeling it'll still be sluggish. I'll give it a test run on Thor and we'll see. 🤷‍♂️

#### Checklist

### DIFF
--- a/app/Console/Commands/ConvertMobilesCommand.php
+++ b/app/Console/Commands/ConvertMobilesCommand.php
@@ -1,0 +1,65 @@
+<?php
+
+namespace Northstar\Console\Commands;
+
+use Northstar\Models\User;
+use Illuminate\Console\Command;
+use libphonenumber\PhoneNumberUtil;
+use libphonenumber\PhoneNumberFormat;
+
+class ConvertMobilesCommand extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'northstar:e164';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Convert mobile numbers to E.164 format.';
+
+    /**
+     * Execute the console command.
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        $counter = 1;
+
+        // Iterate over users where the `mobile` field is not null.
+        User::whereNotNull('mobile')->chunk(200, function ($users) use (&$counter) {
+            $parser = PhoneNumberUtil::getInstance();
+
+            /** @var User $user */
+            foreach ($users as $user) {
+                $this->line('['.$counter++.'] '.$user->id.' - '.$user->mobile);
+
+                try {
+                    // Parse & format as E.164.
+                    $number = $parser->parse($user->mobile, 'US');
+                    $formattedNumber = $parser->format($number, PhoneNumberFormat::E164);
+
+                    // Save to the `e164` field. We'll swap this with `mobile` later.
+                    $user->e164 = $formattedNumber;
+                    $user->save();
+
+                    $this->info('['.$counter.'] '.$user->id.' - Formatted as '.$formattedNumber);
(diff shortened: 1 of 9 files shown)


app/Console/Commands/ConvertMobilesCommand.php
```diff
@@ -0,0 +1,65 @@
...
+        User::whereNotNull('mobile')->chunk(200, function ($users) use (&$counter) {
+            $parser = PhoneNumberUtil::getInstance();
+
+            /** @var User $user */
+            foreach ($users as $user) {
+                $this->line('['.$counter++.'] '.$user->id.' - '.$user->mobile);
+
+                try {
+                    // Parse & format as E.164.
+                    $number = $parser->parse($user->mobile, 'US');
```
sergii-tkachenko: But what if this is not a US number?
Would it make more sense to use a lib that can actually make an educated guess if it's not a US number?
I know one in JS: https://github.com/aftership/phone.
See https://github.com/AfterShip/phone/blob/master/lib/iso3166_data.js for examples.

app/Providers/ValidationServiceProvider.php
```diff
@@ -26,10 +27,21 @@ public function boot()
...
+            $parser = PhoneNumberUtil::getInstance();
 
-            return ! empty($valid) && empty($repeat);
+            try {
+                // Make sure that libphonenumber can parse this phone.
+                // @TODO: Consider testing stricter validity here.
+                $parser->parse($value, 'US');
+
+                // And sanity-check the format is okay:
+                preg_match('#^(?:\+?([0-9]{1,3})([\-\s\.]{1})?)?\(?([0-9]{3})\)?(?:[\-\s\.]{1})?([0-9]{3})(?:[\-\s\.]{1})?([0-9]{4})#', preg_replace('#[\-\s\.]#', '', $value), $valid);
```
itsjoekent: giphy 17

app/Console/Commands/ConvertMobilesCommand.php
DFurnes: This sets the default format if a country code is not included in the number. And seeing as we only store US phone numbers, I think this is a safe assumption for now.

sergii-tkachenko: My concern here is that we could inadvertently convert phones that are non US to US.
For example, +38 (347) 555-555-55 would become `+1 347 555-5555` that could be a real number that belongs to another person.

weerd: @sergii-tkachenko if we only store US numbers why would this be an issue? Not sure if I understand.

DFurnes: Again, the `US` is only used as a default if a country code is not included in the number. So if a user were to provide a valid number with a country code, it would be parsed correctly:

```
$parser->parse('+44 7427 747204', 'US');
// => "Country Code: 44 National Number: 7427747204"
```

PR #634, opened 2017-09-21
Swap temporary 'e164' to 'mobile'.

DFurnes:
#### What's this PR do?
This pull request finishes up work for storing mobiles in E.164 format.

🚚 Adds a migration which swaps the [pre-generated](https://github.com/DoSomething/northstar/pull/632) `e164` field with the existing `mobile`. The old field is kept in `_old_mobile` in case we need to quickly reverse this.*

📱 Always normalize the `mobile` field in E.164 format (instead of a separate `e164` field).

🍂 For backwards compatibility, we'll continue to return the "old" format from the existing API endpoints. I'm adding a ticket to remember to make a `v2/users` endpoint sometime soon that will return E.164 formatted numbers & removes old fields (like `_id`, `mobilecommons_status`, etc.)

I updated the test suite where it made specific assertions against the database contents, otherwise everything should continue to function exactly the same as it did before!

*__*Important:__* This cannot be deployed until the `artisan northstar:e164` script is run!! 🚨

#### Checklist

### DIFF
--- a/app/Auth/Normalizer.php
+++ b/app/Auth/Normalizer.php
@@ -68,6 +68,12 @@ class Normalizer
             return null;
         }
 
+        // Normalize "1 (555) 555-5555" format without leading "+".
+        $digits = preg_replace('/[^0-9]/', '', $mobile);
+        if (strlen($digits) === 11 && $digits[0] === '1') {
+            $mobile = '+'.$mobile;
+        }
+
         try {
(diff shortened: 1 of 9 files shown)


DFurnes:
Finished running through the `northstar:e164` script on QA & Thor on Friday, and everything looks good. Going to deploy this to those environments to test out the "swap" migration, and then send it on to production once the script finishes running on production as well.

PR #636, opened 2017-09-22
Switch to chunking/offsetting by ID.

DFurnes:
#### What's this PR do?
Alright, hopefully this should alleviate our performance issues – instead of offsetting by a number of records, the script now chunks by Object ID (and if we need to restart it, you specify the ID to start from).

Unfortunately Laravel's built-in `chunkById` method assumes integer IDs, and needed to be overridden to work with Mongo's string ObjectIds (and to also provide a "start" offset). It's not super elegant, but it gets the job done. 😅

#### Checklist

### DIFF
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -41,24 +41,32 @@ class AppServiceProvider extends ServiceProvider
 
         // Add 'chunkWithLimit' method to the query builder.
         // @see: https://github.com/laravel/internals/issues/103
-        Builder::macro('chunkWithOffset', function ($count, $from, callable $callback) {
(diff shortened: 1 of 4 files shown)


app/Providers/AppServiceProvider.php
```diff
@@ -41,24 +41,32 @@ public function boot()
 
         // Add 'chunkWithLimit' method to the query builder.
         // @see: https://github.com/laravel/internals/issues/103
-        Builder::macro('chunkWithOffset', function ($count, $from, callable $callback) {
+        Builder::macro('chunkFromId', function ($count, $startId, callable $callback, $column) {
             /** @var Builder $this */
-            $page = 1;
+
+            // Literally copy-pasting `chunkById` so we can override this value... :'(
```
weerd: 😱

Issue #637, opened 2017-09-26
Troubleshooting: E.164 conversion woes.


DFurnes:
It turns out there were some unexpected issues when converting numbers to E.164 format:

### What Happened
1. Capistrano deploys timed out when running the migration on production. This happened when testing the deploy on Thor as well, but it seemed to complete successfully on the following run so I assumed it was a fluke (as we've had other Capistrano deploys randomly time out in the past).
2. This caused the [final step](https://github.com/DoSomething/northstar/blob/e75920c4ca7be9d65c5b69ca178ae0781a7a6cab/database/migrations/2017_09_19_204233_SwapE164ForMobileField.php#L23-L25) of re-adding the unique index on the newly transformed `mobile` field to not occur (although the rename finished as expected, since Mongo continued running the query even after Cap died).
3. Because the index was not re-added, anything that needed to look up a mobile number (like logging in via a mobile, or searching for a user on Rogue) would time out. It seems like these timeouts caused the server to get confused and give `502 Bad Gateway` errors until restarted.
4. Adding a sparse (non-unique) index via the Mongo CLI stopped these issues from occurring.

### Where We Are
It turns out in both Thor & production, we ended up with duplicates after having transformed numbers to E.164 format (since numbers that had previously co-existed as `122233355555` and `22233355555` both got transformed into the same E.164 format). (#410, #415)

I didn't catch this since the `e164` temporary field we were using didn't get a unique index, and I only checked that the total counts were consistent between the two (and a few data spot-checks here and there).

### What's Next
Everything seems to have stabilized after adding that sparse index, but we'll now have weird behavior when trying to look up any of those "dupe" users. I'm going to make a quick script to fix those records by discarding the records with the leading `1##########` on their `_old_mobile` & we can then re-add the unique index on the `mobile` field.

DFurnes:
Merging #638 and going to test it out on QA/Thor. If all seems good, will push to prod as well.

DFurnes:
I tried running the de-dupe script on Thor, and am getting a MongoDB fatal error:

```sh
$ php artisan northstar:e164-dupes --pretend


  [MongoDB\Driver\Exception\RuntimeException]
  BSONObj size: 63913327 (0x3CF3D6F) is invalid. Size must be between 0 and 16793600(16MB)
```

[...]

PR #638, opened 2017-09-26
Fix index creation & add script to remove dupes from E.164 conversion.

DFurnes:
#### What's this PR do?
This pull request fixes two issues:

🍂 The method signature for `index` [changed](https://git.io/vdt4y) in the [v3.1 release](https://github.com/jenssegers/laravel-mongodb/releases/tag/v3.1.0) of the Laravel MongoDB adapter that we use. As far as I can tell this was undocumented, which is frustrating. I've updated the calls to this function in our migrations so tests & new instances get the proper settings on indexes.

📴 Adds a `northstar:e164-dupes` command to remove any duplicate accounts that were [created by the E.164 conversion](https://github.com/DoSomething/northstar/issues/637). Since we would always resolve accounts without the leading `1`, these accounts have been effectively deleted for a long time.

If you haven't, give a read to #637 for full context. I wrote a test case for the new script added in this pull request. Because the database uniqueness constraint on the `mobile` column is working again, it uses a `new_mobile` column for simulating instances where the constraint doesn't exist.

#### Checklist

### DIFF
--- a/app/Console/Commands/FixE164DuplicatesCommand.php
+++ b/app/Console/Commands/FixE164DuplicatesCommand.php
@@ -1,0 +1,87 @@
+<?php
+
+namespace Northstar\Console\Commands;
+
+use Northstar\Models\User;
+use Illuminate\Console\Command;
+
+class FixE164DuplicatesCommand extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'northstar:e164-dupes {column=mobile}
+                            {--pretend : List the duplicates that would be deleted.}';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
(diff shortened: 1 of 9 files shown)


weerd:
> As far as I can tell this was undocumented, which is frustrating. 😞

app/Console/Commands/FixE164DuplicatesCommand.php
```diff
@@ -0,0 +1,87 @@
...
+            $trash = User::find($ids)->filter(function ($user) use ($originalColumn) {
+                return ! preg_match('#^[0-9]{10}$#', $user->{$originalColumn});
+            });
+
+            /** @var User $user */
+            foreach ($trash as $user) {
+                $verb = $pretending ? 'Will remove' : 'Removing';
+                $this->line($verb.' duplicate user '.$user->id);
+
+                if (! $pretending) {
```
weerd: hehe :)

PR #647, opened 2017-10-05
Logout security

itsjoekent:
#### What's this PR do?
This pr checks if the logout redirect is a `*.dosomething.org` domain so a third party cant' hijack a users experience.

Try going to `/logout?redirect="http://heyheyhey.com"` and `/logout?redirect="http://dosomething.org"`

*note: the HTTP part matters. if we have apps not including http in the redirect url they need to be updated as well.

**Fixes**: https://www.pivotaltracker.com/story/show/150597261

#### Checklist

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -173,6 +173,12 @@ function format_birthdate($birthdate)
     return format_date($birthdate, 'Y-m-d');
 }
 
+/**
+ * Format a legacy phone number to a proper number format.
+ *
+ * @param  string $mobile
+ * @return string
+ */
 function format_legacy_mobile($mobile)
 {
     try {
@@ -184,4 +190,21 @@ function format_legacy_mobile($mobile)
     } catch (\libphonenumber\NumberParseException $e) {
         return null;
     }
+}
+
+/**
+ * Check if the given url is a *.dosomething.org domain.
+ *
+ * @param  string  $url
+ * @return boolean
+ */
+function is_dosomething_domain($url)
+{
+    $parsed = parse_url($url);
+
+    if (! array_key_exists('host', $parsed)) {
+        return false;
+    }
+
+    return strpos($parsed['host'], 'dosomething.org') !== FALSE;
 }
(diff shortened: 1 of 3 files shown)


app/helpers.php
```diff
@@ -185,3 +191,20 @@ function format_legacy_mobile($mobile)
...
+ */
+function is_dosomething_domain($url)
+{
+    $parsed = parse_url($url);
+
+    if (! array_key_exists('host', $parsed)) {
+        return false;
+    }
+
+    return strpos($parsed['host'], 'dosomething.org') !== false;
```
DFurnes: Let's use [`ends_with`](https://laravel.com/docs/5.3/helpers#method-ends-with) here to make sure this wouldn't allow someone to pass a domain like `dosomething.org.evil.herokuapp.com`.

app/helpers.php
```diff
@@ -173,6 +173,12 @@ function format_birthdate($birthdate)
     return format_date($birthdate, 'Y-m-d');
 }
 
+/**
+ * Format a legacy phone number to a proper number format.
```
DFurnes: 👏 🙇

app/helpers.php
itsjoekent: @DFurnes Updated & also updated the test to mimic this

tests/Http/Web/WebAuthenticationTest.php
```diff
@@ -154,6 +154,25 @@ public function testLogoutRedirect()
...
+     * Test that we can't be redirected to a third party domain
+     * in the custom post-logout redirect.
+     */
+    public function testLogoutRedirectThirdPartyDomain()
+    {
+        $user = factory(User::class)->create();
+
+        $this->be($user, 'web');
+
+        $this->get('logout?redirect=http://puppet.sloth.com');
```
DFurnes: Can you update the query param here to match the new assertion you're making below?

itsjoekent: Oh yes, good catch 🌮

PR #651, opened 2017-10-10
Always return redirect_uri as an array.

DFurnes:
#### What's this PR do?
We have some `redirect_uri` fields stored as strings. To make our logic easier on Aurora, let's always return these as an array from Northstar (since that's how we'll be setting them from now on).

👀

#### Checklist


itsjoekent:
Hm this is why I mentioned the thing in Slack earlier, can we just migrate all of the existing strings to an array?

DFurnes:
We can, but I think this is probably a little quicker. Any future edits will save as an array.

itsjoekent:
alrighty!

PR #654, opened 2017-10-13
Add ability to rate-limit the Customer.io backfill script.

DFurnes:
#### What's this PR do?
This pull request adds a `--throughput` option to the `artisan northstar:cio` script. It can be used to set a cap on the number of requests that may be processed per minute.

References [#151680875](https://www.pivotaltracker.com/story/show/151680875).

👀

#### Checklist

### DIFF
--- a/app/Console/Commands/BackfillCustomerIoProfiles.php
+++ b/app/Console/Commands/BackfillCustomerIoProfiles.php
@@ -29,10 +29,9 @@ class BackfillCustomerIoProfiles extends Command
     /**
      * Execute the console command.
      *
-     * @param Blink $blink
      * @return mixed
      */
-    public function handle(Blink $blink)
+    public function handle()
     {
         $start = new Carbon($this->argument('start'));
 
@@ -42,7 +41,7 @@ class BackfillCustomerIoProfiles extends Command
             $query->whereNotNull('mobile')->orWhere('updated_at', '>', $start);
         })->where('cio_backfilled', '!=', true);
 
-        $query->chunkById(200, function (Collection $users) use ($blink) {
+        $query->chunkById(200, function (Collection $users) {
             // Send each of the loaded users to Blink's user queue.
             $users->each(function (User $user) {
                 try {
--- a/app/Console/Commands/BackfillCustomerIoProfiles.php
+++ b/app/Console/Commands/BackfillCustomerIoProfiles.php
@@ -31,7 +31,7 @@ class BackfillCustomerIoProfiles extends Command
     /**
      * Execute the console command.
      *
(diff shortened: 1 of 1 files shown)


app/Console/Commands/BackfillCustomerIoProfiles.php
```diff
@@ -29,22 +31,22 @@ class BackfillCustomerIoProfiles extends Command
...
         // profile was updated after the given date, skipping ones we have already backfilled.
         $query = User::where(function (Builder $query) use ($start) {
             $query->whereNotNull('mobile')->orWhere('updated_at', '>', $start);
         })->where('cio_backfilled', '!=', true);
 
-        $query->chunkById(200, function (Collection $users) use ($blink) {
+        $query->chunkById(200, function (Collection $users) use ($throughput) {
             // Send each of the loaded users to Blink's user queue.
-            $users->each(function (User $user) {
+            $users->each(function (User $user) use ($throughput) {
```
itsjoekent: dumb question, how does $throughput know that you made an API call?

itsjoekent: OH i see its used in the sleep below, that makes sense now

app/Console/Commands/BackfillCustomerIoProfiles.php
```diff
@@ -56,6 +58,12 @@ public function handle(Blink $blink)
                 } catch (Exception $e) {
                     $this->error('Failed to backfill user '.$user->id);
                 }
+
+                // If the `--throughput #` parameter is set, make sure we can't
+                // process more than # users per minute by taking a little nap.
+                if ($throughput) {
+                    sleep(60 / $throughput);
```
sergii-tkachenko: Wait. If `$throughput` is null, won't you get a `Division by zero` Fatal error?

app/Console/Commands/BackfillCustomerIoProfiles.php
sergii-tkachenko: I had the same question and got the same answer 😄 .

app/Console/Commands/BackfillCustomerIoProfiles.php
itsjoekent: This code only runs if `$throughput` is greater than 0?

DFurnes: Yep, `$throughput` is optional and either `null` or `0` would not trigger this conditional.

sergii-tkachenko: Oops, I see it now. Sorry.

PR #655, opened 2017-10-13
Add Discovery & Key endpoints.

DFurnes:
#### What's this PR do?
This pull request adds [OpenID Connect Discovery](http://openid.net/specs/openid-connect-discovery-1_0.html) and [JSON Web Key](https://tools.ietf.org/html/draft-ietf-jose-json-web-key-41) endpoints to Northstar. This will allow resource servers to automatically load Northstar's public key [using a command in Gateway](https://github.com/DoSomething/gateway/pull/85).

The "discovery endpoint" is a standardized way to communicate information about how an OpenID Connect server works, and is specifically useful here so we don't have to hardcode the JWKS endpoint into Gateway. Here's [Google's discovery endpoint](https://accounts.google.com/.well-known/openid-configuration) for comparison.

The JWKS endpoint is a standardized way to share public keys with the world. This is useful for setting up other servers with Northstar's public key (rather than using `scp` to bop a file around), and especially handy for quickly setting up local environments! Here's [Google's](https://www.googleapis.com/oauth2/v3/certs) and [Twitch's](https://api.twitch.tv/api/oidc/keys) for comparison.

References [#151957763](https://www.pivotaltracker.com/story/show/151957763).

👀

#### Checklist

### DIFF
--- a/app/Http/Controllers/KeyController.php
+++ b/app/Http/Controllers/KeyController.php
@@ -2,6 +2,9 @@
 
 namespace Northstar\Http\Controllers;
 
+use JOSE_JWK;
+use phpseclib\Crypt\RSA;
+
 class KeyController extends Controller
 {
     /**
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/KeyController.php
```diff
@@ -2,6 +2,9 @@
 
 namespace Northstar\Http\Controllers;
 
+use JOSE_JWK;
```
weerd: `JOSÉ_JWK`? 💃

DFurnes: ¡Olé!

PR #659, opened 2017-10-17
Use usleep for accurate sub-second delays.

DFurnes:
#### What's this PR do?
It turns out that PHP's `sleep` [only accepts integers](http://php.net/manual/en/function.sleep.php#54587), and so `sleep(0.5)` would just not sleep at all! Instead, let's use [`usleep`](http://php.net/manual/en/function.usleep.php) which takes an argument in microseconds.

I'm a dummy & only tested this with a really low throughput.

#### Checklist

---
For review: …

### DIFF
--- a/app/Console/Commands/FixSourcesCommand.php
+++ b/app/Console/Commands/FixSourcesCommand.php
@@ -67,7 +67,8 @@ class FixSourcesCommand extends Command
             // If the `--throughput #` parameter is set, make sure we can't
             // process more than # users per minute by taking a little nap.
             if ($throughput) {
-                sleep(60 / $throughput);
+                $seconds = 60 / $throughput;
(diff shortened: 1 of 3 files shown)


app/Console/Commands/FixSourcesCommand.php
```diff
@@ -67,7 +67,8 @@ public function handle()
             // If the `--throughput #` parameter is set, make sure we can't
             // process more than # users per minute by taking a little nap.
             if ($throughput) {
-                sleep(60 / $throughput);
+                $seconds = 60 / $throughput;
```
itsjoekent: `//TODO: make this a helper function` ?

DFurnes: Mmm, that's a good call!

PR #660, opened 2017-10-18
Add tests for nullable fields.

DFurnes:
#### What's this PR do?
~A fun bug I noticed when upgrading Aurora – when we [updated Northstar to Laravel 5.3](https://github.com/DoSomething/northstar/pull/600), we forgot to add the `nullable` validation rule. This means that explicitly setting `null` for a field would fail. Luckily we don't usually do that (we either sent empty strings from Aurora, or partial fields from other clients).~

~However, now that [Aurora is on Laravel 5.5](https://github.com/DoSomething/aurora/pull/163), the default `TrimStrings` middleware turns empty strings into nulls, which causes this to be a big deal! This pull request adds a test case to demonstrate the problem and fixes the unexpected behavior by marking fields as nullable~ 

...kidding, this pull request just adds a few tests (since the actual issue was fixed in #677).

👀

#### Checklist

### DIFF
--- a/tests/Http/UserTest.php
+++ b/tests/Http/UserTest.php
@@ -169,6 +169,38 @@ class UserTest extends BrowserKitTestCase
         $this->assertNotEquals('Hamilton', $user->last_name);
     }
 
+    /** @test */
+    public function testUnsetFieldWithEmptyString()
+    {
+        $user = factory(User::class)->create();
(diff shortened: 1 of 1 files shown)


DFurnes:
There's some [unexpected behavior with `nullable` and `required_without`](https://git.io/vdQEO). Putting this on pause until we update Northstar to Laravel 5.4.

PR #666, opened 2017-10-26
New `toCustomerIoPayload` function

itsjoekent:
#### What's this PR do?
This PR makes a new function for sending customer io specific data that abides by the rules created by Blink. It's similar to the existing `toBlinkPayload` but adds a few new variables and has a slightly different data structure. 

I additionally made updates to the Customer IO service so we were calling this new function and also ensured a user is never sent to customer io that doesn't have an email or phone.

#### How should this be reviewed
Generic existing admin user
screen shot 2017-10-26 at 3 59 44 pm

New user
screen shot 2017-10-26 at 4 05 35 pm

New user with no email
screen shot 2017-10-26 at 4 07 01 pm

User with no email or mobile
screen shot 2017-10-26 at 4 08 11 pm

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -388,6 +388,57 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
     }
 
     /**
+     * Transform the user model for Customer.io
+     *
+     * @return array
+     */
+    public function toCustomerIoPayload()
+    {
+        $requiredCustomerIoFields = collect(['id', 'updated_at', 'created_at']);
+
+        // If this user was just created and has an email, mark them as subscribed.
+        $isNewUser = $this->updated_at === $this->created_at;
+        $unsubscribed = ! ($isNewUser && $this->email);
+
+        $data = collect([
+            'id' => $this->id,
+            'email' => $this->email,
+            'phone' => $this->mobile,
+            'mobile_status' => $this->sms_status, // @TODO: Remove!
+            'sms_status' => $this->sms_status,
+            'facebook_id' => $this->facebook_id,
+            'first_name' => $this->first_name,
+            'last_name' => $this->last_name,
+            'birthdate' => format_date($this->birthdate, 'Y-m-d'),
+            'role' => $this->role,
+            'addr_city' => $this->addr_city,
+            'addr_state' => $this->addr_state,
+            'addr_zip' => $this->addr_zip,
+            'language' => $this->language,
+            'country' => $this->country,
+            'source' => $this->source,
+            'source_detail' => $this->source_detail,
+            'last_messaged_at' => iso8601($this->last_messaged_at),
+            'last_authenticated_at' => iso8601($this->last_authenticated_at),
+            'updated_at' => iso8601($this->updated_at),
+            'created_at' => iso8601($this->created_at),
+            'unsubscribed' => $unsubscribed,
+        ])->filter(function($value, $key) {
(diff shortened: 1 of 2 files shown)


sergii-tkachenko:
@itsjoekent Looks good so far. Could you also show one with `sms_status` field?

app/Models/User.php
```diff
@@ -388,6 +388,57 @@ public function toBlinkPayload()
...
+
+        // If this user was just created and has an email, mark them as subscribed.
+        // Otherwise set unsubscribed to null so it's removed in the filter later.
+        $isNewUser = $this->updated_at->timestamp === $this->created_at->timestamp;
+        $unsubscribed = ($this->email && $isNewUser) ? false : null;
+
+        $data = collect([
+            'email' => $this->email,
+            'phone' => $this->mobile,
+            'mobile_status' => $this->sms_status, // @TODO: Remove!
```
sergii-tkachenko: I think, this one should be `sms_status`.

app/Services/CustomerIo.php
```diff
@@ -47,8 +47,13 @@ private function getAuthParams()
...
     public function updateProfile(User $user)
     {
+        // If the user doesn't have an email or phone number, don't send them.
+        if (! $user->email || ! $user->phone) {
+            return false;
+        }
+
         $response = $this->client->put('customers/'.$user->id, [
-            'json' => $user->toBlinkPayload(),
+            'json' => $user->toCustomerIoPayload(),
```
sergii-tkachenko: Hm, will `$this->client` be customer.io client?

itsjoekent: Yes see above in the constructor

app/Models/User.php
itsjoekent: Check the row below it, I think Dave was sending both to Blink. Should I remove this one @sergii-tkachenko ?

app/Services/CustomerIo.php
sergii-tkachenko: Ah, perfect. Thanks

app/Models/User.php
sergii-tkachenko: Yes, we don't need `mobile_status`, only `sms_status`. We renamed the field and, as I understand, it has been sent to blink, just in case.

itsjoekent: :poof:

app/Models/User.php
```diff
@@ -388,6 +388,56 @@ public function toBlinkPayload()
...
+        $data = collect([
+            'email' => $this->email,
+            'phone' => $this->mobile,
+            'sms_status' => $this->sms_status,
+            'facebook_id' => $this->facebook_id,
+            'first_name' => $this->first_name,
+            'last_name' => $this->last_name,
+            'birthdate' => format_date($this->birthdate, 'Y-m-d'),
+            'role' => $this->role,
+            'addr_city' => $this->addr_city,
```
sergii-tkachenko: Could you also please convert all `addr_*` fields to strings? Their value is denormalized sometimes, see https://github.com/DoSomething/blink/issues/142#issuecomment-334179997

itsjoekent: @sergii-tkachenko I think in order to quickly get something out the door, I'm going to just discard anything that isn't a string.

itsjoekent:
New logic to handle addresses

screen shot 2017-10-26 at 5 14 15 pm

PR #667, opened 2017-10-27
Cio mobile import script

weerd:
#### What's this PR do?
This PR adds an Artisan™ command to import mobile users to Customer.io. There's an additional flag to only consider mobile users that _have not been backfilled when running the script.

There's also an additional flag `--dry` to do a "dry run" of the script and quickly see if it seems to be working correctly and outputting expected IDs, etc.

👁 

#### Checklist

---
For review: @itsjoekent

### DIFF
--- a/app/Console/Commands/CustomerIoMobileUserImport.php
+++ b/app/Console/Commands/CustomerIoMobileUserImport.php
@@ -88,16 +88,16 @@ class CustomerIoMobileUserImport extends Command
                 if ($dryRun) {
                     $this->line('Dry run to backfill user '.$user->id.' #'.$itemNumber);
                 } else {
-                    try {
-                        // Send user data to CustomerIo.
-                        // $this->customerIo->updateProfile($user);
+                    // Send user data to CustomerIo.
+                    $response = $this->customerIo->updateProfile($user);
 
+                    if ($response) {
                         // Mark this user as processed.
-                        // $user->cio_backfilled = true;
-                        // $user->save(['touch' => false]);
+                        $user->cio_backfilled = true;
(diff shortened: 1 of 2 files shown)


app/Console/Commands/CustomerIoMobileUserImport.php
```diff
@@ -0,0 +1,114 @@
...
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        $this->info('::: Script started on '.Carbon::now()->format('m/d/Y \a\t H:i:s'));
+
+        $dryRun = $this->option('dry');
+        $throughput = (int) $this->option('throughput');
+        $notBackfilled = $this->option('not_backfilled');
```
itsjoekent: Smart 👏

weerd:
@itsjoekent btw, my dream of building in a `--skip` flag that skips a certain number of records ultimately did not work. The complications came about because the `skip()` modifies the query _before_ it gets run and chunked, but the chunking process uses the same query every time. So say you want to skip the first 10 items in a collection and the query is chunked into a batch 100 records grabbed, _each_ run would skip 10, not just the first run. Eventually it turned out more complicated than I hoped; I had some further ideas but ended up not wanting to go down a 🐰 🕳  😬 .

itsjoekent:
Yeah I figured thats why you went with the skip already processed records, which seems like it would basically achieve the same thing

weerd:
Yeah, it defs would work great if we could add a skip to the query first and then _after_ grabbing all the collection of records using the `chunk` method on the `Collection` class, but that would mean grabbing **all** the records in one shot which for the 3million+ records would likely kill the instance @_@

PR #668, opened 2017-10-27
Adding the clean customer io command

itsjoekent:
#### What's this PR do?
This PR adds a new command to mark all previously backfilled users as false.

original_billy_mays_oxiclean_ad_from_october_2000

Run it!

screen shot 2017-10-27 at 11 42 57 am

### DIFF
--- a/app/Console/Commands/CleanCustomerIoImport.php
+++ b/app/Console/Commands/CleanCustomerIoImport.php
@@ -1,0 +1,45 @@
+<?php
+
+namespace Northstar\Console\Commands;
+
+use Illuminate\Console\Command;
+use Northstar\Models\User;
+
+class CleanCustomerIoImport extends Command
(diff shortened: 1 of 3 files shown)


app/Console/Commands/CleanCustomerIoImport.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+namespace Northstar\Console\Commands;
+
+use Illuminate\Console\Command;
+use Northstar\Models\User;
+
+class CleanCustomerIoImport extends Command
```
weerd: lol. while a great Billy May's gif, can we change to `RevertCustomerIoImport`?

itsjoekent: UGH BUT BRANDING

Issue #662, opened 2017-10-24
Task: Running Customer.io backfill.


DFurnes:
### Background
When we originally enabled Customer.io integration, we only did so for _new_ accounts, and excluded the `mobile` field & mobile-only profiles. We did this because Blink wasn't able to accept profile updates at launch & Customer.io requires E.164-formatted numbers for messaging (and we wanted to convert this in the database rather than on-the-fly in the Blink payload).

Since then, we've [started sending profile updates](https://github.com/DoSomething/northstar/pull/642) to Blink, converted existing mobile numbers to E.164, and started including the `mobile` field in Customer.io payloads.

### Task
We're running a new backfill to get any profile updates that occurred since the original backfill on 5/12/2017 (but before we started sending updates on 10/03/2017), and _any_ accounts that have a `mobile` field. This is being done with the [`northstar:cio` Artisan command](https://github.com/DoSomething/northstar/blob/dev/app/Console/Commands/BackfillCustomerIoProfiles.php).

We also want to limit our throughput so we don't overload Blink (maximum of 20 requests/second), and pipe our output into a log file for later auditing.

### Steps

We can run the following command in a `tmux` session to start the backfill:

```
php artisan northstar:cio 5/10/2017 10/05/2017 --throughput=1200 | tee -a ~/cio-2017-backfill.log
```

Accounts that have been sent to Customer.io will have the `_cio_backfill` field marked on their account so we can safely stop and restart this command at any time.

DFurnes:
We want to be careful that we don't overload Blink & clog up the queue, so we want to be sure to ping @sergii-tkachenko whenever starting or stopping the backfill so he can keep an eye on Blink's health.

DFurnes:
This looks good (with some good scrambling in #667, #668, #670, #671). Just checked and after running the `northstar:cio_mobile_import` script we've got 6.33m users with `mobile` numbers in Customer.io, and the same in Northstar. 👌

DFurnes:
We'll still want to run the original `northstar:cio` script (again, bypassing Blink) so that we can get any profile updates that occurred in that time window as well (since we were only sending _new_ accounts through Blink during that time period).

DFurnes:
There are a remaining 736,260 profiles without mobiles that were updated in that timeframe, let's start up the `northstar:cio` script tomorrow morning to get them in there. ⛸

DFurnes:
[...]

PR #673, opened 2017-10-30
Add command for importing address data.

DFurnes:
#### What's this PR do?
This pull request adds a simple command to import the (already processed) address information from Mobile Commons from a script created by Sohaib. Since Blink has been having some trouble, we'll update these records directly on [Customer.io](https://customer.io).

References [#152249269](https://www.pivotaltracker.com/n/projects/2019313/stories/152249269).

👀

#### Checklist

### DIFF
--- a/app/Console/Commands/AddAddressesCommand.php
+++ b/app/Console/Commands/AddAddressesCommand.php
@@ -1,0 +1,61 @@
+<?php
+
+namespace Northstar\Console\Commands;
+
+use League\Csv\Reader;
+use Illuminate\Console\Command;
+use Northstar\Models\User;
+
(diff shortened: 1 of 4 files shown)


DFurnes:
This should be safe to run on production once we've finished with #662. I'll triple check tomorrow, but the CSV provided by Sohaib should have exact `addr_*` values we want for each ID (based on the logic described in the Pivotal ticket), and we'll be providing that separately to the [Customer.io](https://customer.io) folks so we don't need to make any API requests here.

One change I _did_ have to make – I renamed `addr_country` to `country` in the CSV so it would match the Northstar field name. We should make this change in the original CSV we're handing off as well.

PR #679, opened 2017-11-02
Remove deprecated endpoints & commands.

DFurnes:
#### What's this PR do?
~Spring~ Fall cleaning! This pull request removes Artisan commands that are no longer relevant & API endpoints that are no longer used by any other apps (mainly things we had built for the mobile app).

Tests should continue to pass!

#### Checklist


DFurnes:
This also removes the Phoenix Ashes API requests to trigger Quicksilver registration emails, since those are now sent via Customer.io & the request was just a no-op.

DFurnes:
I've also deleted the old `Token` model (which was used for the mobile app's "sessions"). I'll make a note to create a migration to drop this collection later on as well.

app/Console/Commands/CustomerIoMobileUserImport.php
```diff
@@ -1,114 +0,0 @@
...
-
-namespace Northstar\Console\Commands;
-
-use Carbon\Carbon;
-use Northstar\Models\User;
-use Illuminate\Console\Command;
-use Illuminate\Support\Collection;
-use Northstar\Services\CustomerIo;
-
-class CustomerIoMobileUserImport extends Command
```
weerd: You did well my friend. Goodbye for now.... ⚰️

app/Http/Controllers/AvatarController.php
```diff
@@ -1,76 +0,0 @@
...
-use Northstar\Http\Transformers\UserTransformer;
-use Northstar\Auth\Scope;
-use Northstar\Services\AWS;
-use Northstar\Models\User;
-use Illuminate\Http\Request;
-use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
-use Symfony\Component\HttpKernel\Exception\UnauthorizedHttpException;
-use Gate;
-
-class AvatarController extends Controller
```
weerd: Why's the `AvatarController` out?

DFurnes: This was only used by the mobile app, and I think we'd want to handle this differently if we re-add this feature again in the future (for example, supporting web uploads).

weerd: Ok gotcha. Yeah, I think we'd eventually want something similar when supporting web uploads and customizing profiles 🎉

DFurnes: Agreed!

PR #681, opened 2017-11-02
Accept any string format for 'last_messaged_at'.

DFurnes:
#### What's this PR do?
It turns out `toIsoString()` in Javascript provides something different than in PHP, fab! For now, we'll use `strtodate` to parse any ol' time & revisit this later.

👀

#### Checklist


DFurnes:
Pushed this up to unblock @aaronschachter and @rapala61 working during the event. Will see if there's a more elegant solution for this in the future.

cc: @weerd @itsjoekent @mendelB

PR #687, opened 2017-11-06
Adding Puck to Northstar

itsjoekent:
#### What's this PR do?
- Adds Puck configuration & package 
- Adds two custom events to be tracked by Puck (if validation errors exist, FB auth is clicked)
- Adds `scriptify` function we use elsewhere to stick the Northstar ID on the page

You can actually check for Puck events firing in the console even if you're not sending them to the puck server. So I've just been testing that those are firing as intended.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -9,5 +9,10 @@ module.exports = configure({
   output: {
     // Override output path for Laravel's "public" directory.
     path: path.join(__dirname, '/public/dist'),
-  }
+  },
+  plugins: [
+    new webpack.DefinePlugin({
+      PUCK_URL: JSON.stringify(process.env.PUCK_URL) || null,
+    }),
+  ],
 });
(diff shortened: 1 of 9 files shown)


itsjoekent:
@DFurnes tagged you on this but no rush

webpack.config.js
```diff
@@ -9,5 +9,10 @@ module.exports = configure({
   output: {
     // Override output path for Laravel's "public" directory.
     path: path.join(__dirname, '/public/dist'),
-  }
+  },
+  plugins: [
+    new webpack.DefinePlugin({
```
DFurnes: I'd like to handle this with a `window.ENV` like we do in [Phoenix Next](https://github.com/DoSomething/phoenix-next/blob/4c6138caf16509732073f7f63cd4d09924410945/resources/views/layouts/master.blade.php#L31)? I'd rather not have to rebuild the JS if we want to ever change this value, since I remember that being a headache.

itsjoekent:
@DFurnes 💥 

screen shot 2017-11-06 at 5 05 29 pm

PR #691, opened 2017-11-08
Improve puck integration

itsjoekent:
#### What's this PR do?
- This PR appends validation errors to the window object so they can be parsed by Puck. 
- Puck takes these values and stashes them to the data property of the `has validation errors` event.
- Finally, I wrapped this validation code in the same document.ready as the Facebook tracker to ensure consistency when binding to the dom.

QUESTION: In the `scriptify` function I passed `$errors->default`. I can't find any case where the errors are under a different subkey besides default, but curious if anyone feels the need to flag this.

Check out `window.ERRORS` and look at the Puck event.

screen shot 2017-11-08 at 3 06 20 pm
screen shot 2017-11-08 at 2 55 19 pm

### DIFF
--- a/resources/assets/js/utilities/Puck.js
+++ b/resources/assets/js/utilities/Puck.js
@@ -12,12 +12,12 @@ function init() {
     $('.facebook-login').on('click', () => (
       puck.trackEvent('clicked facebook auth')
     ));
-  });
 
-  const $validationErrors = $('.validation-error');
-  if ($validationErrors && $validationErrors.length) {
-    puck.trackEvent('has validation errors');
-  }
+    const $validationErrors = $('.validation-error');
+    if ($validationErrors && $validationErrors.length) {
+      puck.trackEvent('has validation errors');
+    }
+  });
 }
 
 export default { init };
--- a/resources/assets/js/utilities/Puck.js
+++ b/resources/assets/js/utilities/Puck.js
@@ -15,7 +15,14 @@ function init() {
 
     const $validationErrors = $('.validation-error');
     if ($validationErrors && $validationErrors.length) {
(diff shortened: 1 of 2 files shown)


resources/views/layouts/app.blade.php
```diff
@@ -36,6 +36,7 @@
     @include('layouts.variables')
     {{ scriptify(auth()->user() ? auth()->user()->id : null, 'NORTHSTAR_ID') }}
     {{ scriptify(get_client_environment_vars(), 'ENV') }}
+    {{ scriptify(count($errors) > 0 ? $errors->default : null, 'ERRORS') }}
```
DFurnes: Can this be replaced with `scriptify($errors->messages(), 'ERRORS')`?

resources/assets/js/utilities/Puck.js
```diff
@@ -12,12 +12,22 @@ function init() {
...
-  const $validationErrors = $('.validation-error');
-  if ($validationErrors && $validationErrors.length) {
-    puck.trackEvent('has validation errors');
-  }
+    const $validationErrors = $('.validation-error');
+    if ($validationErrors && $validationErrors.length) {
+      const errors = window.ERRORS || {};
+      const invalidFields = Object.keys(errors);
+
+      const validationMessages = invalidFields.reduce((list, key) => (
```
DFurnes: Absolutely optional, but I think you could use Lodash's [`flattenDeep`](https://lodash.com/docs/#flattenDeep) here:

```js
import { flattenDeep } from 'lodash';

const validationMessages = flattenDeep(Object.values(errors));
```

itsjoekent:
⚡️

PR #694, opened 2017-11-27
Some fixes for OpenID Connect.

DFurnes:
#### What's this PR do?
This pull request includes two fixes for parts of the OpenID Connect spec that I hadn't implemented faithfully (it may make sense to polish some of this up and contribute to the community as a package):

🆔 Adds a separate `id_token` field to the token response (for now, the same as the `access_token`, but can optionally include some of the same claims included in `userinfo` endpoint).

ℹ️ Adds a new `v2/userinfo` endpoint that doesn't have the nested `data` wrapper like `v2/auth/info`. This isn't in the spec, and so causes some clients (specifically [node-openid-client](https://github.com/panva/node-openid-client)) to fail.

Tests should still pass!

#### Checklist

### DIFF
--- a/app/Auth/Responses/BearerTokenResponse.php
+++ b/app/Auth/Responses/BearerTokenResponse.php
@@ -1,0 +1,23 @@
+<?php
+
+namespace Northstar\Auth\Responses;
+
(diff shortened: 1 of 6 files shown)


weerd:
Post-merge 👍

PR #695, opened 2017-11-28
Adding addr source field to user

### LINKED ISSUE #693: New address source field
We'll be saving User address information based on inbound twilio messages if we don't have any values saved. The address source will be saved as `'sms'`.

Refs https://github.com/DoSomething/gambit-conversations/issues/226

mendelB:
#### What's this PR do?
Adds an `addr_source` field for `users`, in the User models `$fillable` fields as well as the `UserTransformer`.

closes #693 
https://www.pivotaltracker.com/story/show/153096331

### DIFF
--- a/app/Http/Transformers/UserTransformer.php
+++ b/app/Http/Transformers/UserTransformer.php
@@ -43,6 +43,9 @@ class UserTransformer extends TransformerAbstract
             $response['addr_state'] = $user->addr_state;
             $response['addr_zip'] = $user->addr_zip;
 
+            // Source for the address fields (e.g. 'sms')
+            $response['addr_source'] = $user->addr_source;
+
             // Signup source (e.g. drupal, cgg, mobile...)
             $response['source'] = $user->source;
             $response['source_detail'] = $user->source_detail;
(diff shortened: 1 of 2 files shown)


aaronschachter:
Do we plan to update this field when a User adds or edits their address via web?

We'll begin to save `'sms'` to this field when we deploy https://github.com/DoSomething/gambit-conversations/pull/263 next week, storing the address info we get from a Twilio request. 

There's an edge case: a user is created via SMS and we store their info. User logs into web and updates their info (e.g.  my address will show up as NJ since I still have a NJ phone number, even though I live in CA). User updates their address on web but the `addr_source` would still remain `'sms'`.

mendelB:
@aaronschachter We have this card open https://www.pivotaltracker.com/story/show/153181447 which I believe is meant to address this scenario. @DFurnes @weerd can confirm

DFurnes:
Yes! We're all set to accept Gambit's `addr_source` fields for now (and will _not_ change that when a user edits their profile via web). Once we tackle [#153181447](https://www.pivotaltracker.com/story/show/153181447) (hopefully this sprint), Northstar will keep track of this itself (so we'll always have a record of where the field was last updated from without having to manually maintain this).

PR #696, opened 2017-12-11
Upgrade to Laravel 5.5.

DFurnes:
#### What's this PR do?
This pull request upgrades Northstar to use [Laravel 5.5](https://laravel.com/docs/5.5/upgrade)! I'd wanted to do this a while back, but it got pushed aside since [laravel-mongodb](https://github.com/jenssegers/laravel-mongodb) wasn't updated with 5.5 support. Now that it is, let's get everything on the same page! ✨ 

There's relatively few changes for this update – notably:

🌱 Console commands are auto-loaded from the `app/Console/Commands` directory, so we don't need to manually register them in the `$commands` array anymore.

🔑 We still need a [patched version](https://github.com/DoSomething/laravel-mongodb/tree/3.3.1-patched-reset) of the MongoDB package, since it has a bug with reading password reset tokens from the database... I'll try submitting this as another PR. :v:

🚰 The `ThrottleRequests` had some internal changes, so we had to move our custom overrides (to properly display throttled web requests) into the Handler class.

I'd like to stress-test this on Thor just in case, but hopefully should be an uneventful upgrade!

#### Checklist

### DIFF
--- a/app/Console/Kernel.php
+++ b/app/Console/Kernel.php
@@ -8,17 +8,6 @@ use Illuminate\Foundation\Console\Kernel as ConsoleKernel;
 class Kernel extends ConsoleKernel
 {
     /**
-     * The Artisan commands provided by your application.
-     *
-     * @var array
-     */
-    protected $commands = [
-        \Northstar\Console\Commands\AddAddressesCommand::class,
-        \Northstar\Console\Commands\KeysCommand::class,
-        \Northstar\Console\Commands\SetupCommand::class,
-    ];
-
-    /**
      * Define the application's command schedule.
      *
      * @param  \Illuminate\Console\Scheduling\Schedule $schedule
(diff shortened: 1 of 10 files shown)


app/Console/Kernel.php
```diff
@@ -28,4 +17,14 @@ protected function schedule(Schedule $schedule)
...
+
+    /**
+     * Register the commands for the application.
+     *
+     * @return void
+     */
+    protected function commands()
+    {
+        $this->load(__DIR__.'/Commands');
+    }
```
weerd: Noice!

app/Exceptions/Handler.php
```diff
@@ -68,7 +69,9 @@ public function render($request, Exception $e)
         }
 
         // Re-cast specific exceptions or uniquely render them:
-        if ($e instanceof AuthenticationException) {
+        if ($e instanceof HttpException && $e->getStatusCode() == 429) {
```
weerd: triple `===`?

DFurnes: Oh yah.

PR #699, opened 2018-01-11
Audit information

mendelB:
#### What's this PR do?
Adds optional audit information to models so that upon attribute edits, source and time information will be stored per specific attribute.

Example audit info for updating `first_name` from northstar web:

```
audit => [
 "first_name" => [
   "source" => "northstar",
   "updated_at" => [
     "date" => "2018-01-11 18:07:05.000000",
     "timezone_type" => 3,
     "timezone" => "UTC",
   ],
 ],
...
]
```

This is specifically relevant to [#153181447](https://www.pivotaltracker.com/story/show/153181447) and #693 and would retroactively have us tracking the source for any updated field, thus implicitly tracking address field source information.

#### Questions and Comments

Do we want a whitelist of attributes to not track audit info for (`updated_at` field for example)? 
Also, Upon User creation (manually, say from within `artisan tinker`), audit fields are created for `_id`, `updated_at` `created_at` and `source`, and when creating from northstar web, tons more fields are updated.

Perhaps adding the `audit` logic as an `else` to the empty string conditional, to ensure that only updated fields get audit info?

#### Checklist

### DIFF
--- a/app/Http/Controllers/MergeController.php
+++ b/app/Http/Controllers/MergeController.php
@@ -49,7 +49,7 @@ class MergeController extends Controller
         $duplicate = User::findOrFail($request->input('id'));
 
         // Get all profile fields from the duplicate (except metadata like ID or source).
-        $metadata = ['_id', 'updated_at', 'created_at', 'drupal_id', 'source', 'source_detail', 'role'];
+        $metadata = ['_id', 'updated_at', 'created_at', 'drupal_id', 'source', 'source_detail', 'role', 'audit'];
         $duplicateFields = array_except($duplicate->toArray(), $metadata);
         $duplicateFieldNames = array_keys($duplicateFields);
 
--- a/app/Models/Model.php
+++ b/app/Models/Model.php
@@ -2,6 +2,7 @@
 
 namespace Northstar\Models;
 
+use Carbon\Carbon;
 use Illuminate\Database\Eloquent\Builder;
 use Jenssegers\Mongodb\Eloquent\Model as BaseModel;
 
@@ -26,6 +27,13 @@ class Model extends BaseModel
         // Empty strings should be saved as `null`.
         if (empty($this->attributes[$key])) {
             $this->attributes[$key] = null;
+        }
+
+        if ($this->audited) {
+            $this->attributes['audit'][$key] = [
+                'source' => client_id(),
+                'updated_at' => Carbon::now(),
+            ];
         }
 
         return $this;
--- a/app/Models/Model.php
+++ b/app/Models/Model.php
@@ -29,7 +29,7 @@ class Model extends BaseModel
             $this->attributes[$key] = null;
         }
 
-        $isDirty = ! array_key_exists($key, $this->original) || $this->original[$key] != $this->attributes[$key];
+        $isDirty = $this->isDirty($key);
         $shouldAudit = ! in_array($key, ['updated_at', 'created_at']);
 
         if ($this->audited && $isDirty && $shouldAudit) {
--- a/app/Models/Model.php
+++ b/app/Models/Model.php
@@ -22,6 +22,8 @@ class Model extends BaseModel
      */
     public function setAttribute($key, $value)
     {
+        $currentValue = $this->$key;
+
         parent::setAttribute($key, $value);
 
         // Empty strings should be saved as `null`.
@@ -29,7 +31,10 @@ class Model extends BaseModel
             $this->attributes[$key] = null;
         }
 
-        if ($this->audited) {
+        if ($this->audited &&
+            $currentValue != $this->$key &&
+            ! in_array($key, ['updated_at', 'created_at'])
+        ) {
             $this->attributes['audit'][$key] = [
                 'source' => client_id(),
                 'updated_at' => Carbon::now(),
--- a/app/Models/Model.php
+++ b/app/Models/Model.php
@@ -22,8 +22,6 @@ class Model extends BaseModel
      */
     public function setAttribute($key, $value)
     {
-        $currentValue = $this->$key;
-
         parent::setAttribute($key, $value);
 
         // Empty strings should be saved as `null`.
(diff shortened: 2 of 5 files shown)


app/Models/User.php
```diff
@@ -79,6 +79,9 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
 {
     use Authenticatable, Authorizable, Notifiable, CanResetPassword;
 
+    // Set class as 'audited', so audit information will be added on set attributes.
```
DFurnes: Let's use a docblock for this & make it a protected field:

```php
/**
 * Should changes to this model's attributes be stored
 * in an audit property on the database record?
 *
 * @var bool
 */
protected $audited = true;
```

DFurnes:
> Do we want a whitelist of attributes to not track audit info for (`updated_at` field for example)? 

Ahh, good catch. Let's exclude `updated_at` and `created_at`.

> Also, Upon User creation (manually, say from within `artisan tinker`), audit fields are created for `_id`, `updated_at` `created_at` and `source`, and when creating from northstar web, tons more fields are updated.

That's fine in my opinion! Let's just make sure that future changes to that user _only_ update the audit field for the particular things that have changed (so even if I submit the whole profile form, if I only change my first name, that's the only field that should have it's audit record updated.)

app/Models/Model.php
```diff
@@ -21,13 +22,25 @@ class Model extends BaseModel
      */
     public function setAttribute($key, $value)
     {
+        $currentValue = $this->$key;
```
mendelB: This somewhat weird way of determining 'dirtyness' of the new value is to address the issue of fields that have their own mutators on them that change what the stored field will be (such as [`birthdate`](https://github.com/DoSomething/northstar/blob/dev/app/Models/User.php#L235)) which would always return a difference to the user given value even if the stored value would be the same (`12/12/1990` vs some stored MongoDB `UTCDateTime` object for example).

app/Models/Model.php
```diff
@@ -21,13 +22,25 @@ class Model extends BaseModel
...
         parent::setAttribute($key, $value);
 
         // Empty strings should be saved as `null`.
         if (empty($this->attributes[$key])) {
             $this->attributes[$key] = null;
         }
 
+        if ($this->audited &&
+            $currentValue != $this->$key &&
+            ! in_array($key, ['updated_at', 'created_at'])
```
mendelB: I wasn't sure if this should be a list in the User model (or whichever specific model) itself (like an additional `$shouldNotAudit` prop or something) or this general list here.

app/Models/Model.php
```diff
@@ -21,13 +22,25 @@ class Model extends BaseModel
...
+
         parent::setAttribute($key, $value);
 
         // Empty strings should be saved as `null`.
         if (empty($this->attributes[$key])) {
             $this->attributes[$key] = null;
         }
 
+        if ($this->audited &&
+            $currentValue != $this->$key &&
```
DFurnes: For the "dirty" check, let's try comparing against the original attribute:

```php
$isDirty = $this->original[$key] !== $this->attributes[$key];
$shouldAudit = ! in_array($key, ['updated_at', 'created_at']);
if ($this->audited && $isDirty && $shouldAudit) {
  // ...
}
```

app/Models/Model.php
DFurnes: Left a comment below!

app/Http/Controllers/MergeController.php
```diff
@@ -49,7 +49,7 @@ public function store($id, Request $request)
         $duplicate = User::findOrFail($request->input('id'));
 
         // Get all profile fields from the duplicate (except metadata like ID or source).
-        $metadata = ['_id', 'updated_at', 'created_at', 'drupal_id', 'source', 'source_detail', 'role'];
+        $metadata = ['_id', 'updated_at', 'created_at', 'drupal_id', 'source', 'source_detail', 'role', 'audit'];
```
mendelB: the MergeTest was failing because merging wouldn't go through since both profiles had `audit` information which resulted in the `Cannot merge "audit" into non-null field on target.` error.

I figured we should ignore the audit property anyways when merging. (The profiles will receive necessary audit info as their attributes are updated).

app/Models/Model.php
```diff
@@ -28,6 +29,16 @@ public function setAttribute($key, $value)
             $this->attributes[$key] = null;
         }
 
+        $isDirty = ! array_key_exists($key, $this->original) || $this->original[$key] != $this->attributes[$key];
```
mendelB: (Copying comment from above that got squashed)
Is it OK if the comparison operator is !=?
For birthdate (Or I imagine any sort of object being stored) it will always be dirty otherwise.
(I'm assuming so, I did test this for birthdate and every time you update your profile, birthdate audit is also updated)

app/Http/Controllers/MergeController.php
DFurnes: Good catch!

app/Models/Model.php
DFurnes: Mmm, tricky!

Let's try using the `Model`'s [`isDirty`](https://github.com/laravel/framework/blob/fcd2b9eea3d7a7b610d792deed82d1e82437eec0/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L948-L959) instead. Looks like it is able to better check equality w/ dates [in here](https://github.com/laravel/framework/blob/fcd2b9eea3d7a7b610d792deed82d1e82437eec0/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L1041-L1070).

mendelB: Yeesh, sorry, I should have been using that from the start!

DFurnes: Ahh not at at all! I only found it when I started poking around now. 😄

mendelB:
🎊 Tests Are Passing!!!

PR #704, opened 2018-01-25
Last setup steps for Customer.io queued backfill.

DFurnes:
#### What's this PR do?
Two last bits before we kick this off:

1. Going to continue routing things through Blink for now. I don't want to accidentally introduce any issues with mis-formatted fields while running this, and this should also set us up for queueing _all_ Customer.io API calls in a future update.

2. Restarts the queue worker whenever we make deploys! References DoSomething/devops#365.

👀

#### Checklist

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -42,15 +42,19 @@ namespace :deploy do
     run "cd #{release_path} && php artisan cache:clear"
   end
 
+  task :artisan_queue_restart do
+    run "cd #{release_path} && php artisan queue:restart"
+  end
+
   task :restart_php do
     run "sudo /usr/sbin/service php7.0-fpm restart"
   end
-
 end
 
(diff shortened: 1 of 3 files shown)


config/deploy.rb
```diff
@@ -42,15 +42,19 @@
     run "cd #{release_path} && php artisan cache:clear"
   end
 
+  task :artisan_queue_restart do
+    run "cd #{release_path} && php artisan queue:restart"
```
weerd: Any fear in losing things in the queue if restarted on deploy? I'm assuming we'd just do a deploy once queue is cleared, but just wanted to ask anyhoo.

DFurnes: Nope! This should instruct workers to finish their current job & then let Supervisor restart them with the new code, so no fear of losing anything.

PR #705, opened 2018-01-29
Change some fields we send to Blink.

DFurnes:
#### What's this PR do?
🔥 Stop sending along a `mobile_status` field since Blink [no longer uses it](https://github.com/DoSomething/blink/pull/133). ab5d439

⏲ Send `last_messaged_at` as UNIX timestamp, rather than sending as ISO-8601 and forcing Blink to convert it to a timestamp for us. 166ab07

🇹🇬 Add TODOs for some future field changes we may want to do (mainly for context). aaee064

👀

#### Checklist

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -416,7 +416,7 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
         return [
             'id' => $this->id,
             'email' => $this->email,
-            'mobile' => $this->mobile,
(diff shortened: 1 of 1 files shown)


DFurnes:
I also want to give a shoutout to the new [`optional()` helper](https://laravel.com/docs/5.5/helpers#method-optional) in Laravel 5.5! So handy!

weerd:
> I also want to give a shoutout to the new optional() helper in Laravel 5.5! So handy!

OMG! This is so handy for form inputs too (given the example too)!

PR #709, opened 2018-02-16
Update Handler and NorthstarValidationException for 5.5.

DFurnes:
#### What's this PR do?
I noticed that Northstar was not rendering validation exceptions for traditional `Illuminate\Validation\ValidationException` errors, and we were instead just getting blank 200s. No good! This didn't affect most traffic since we use a custom validation class for the user resource, but caused issues when I was trying to create new clients on Aurora.

👀

#### Checklist

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -126,17 +126,24 @@ class Handler extends ExceptionHandler
      */
     protected function invalidated($request, $e)
     {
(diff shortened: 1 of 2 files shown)


DFurnes:
I'd like to take some time to refactor this further in a later sprint, since we should be able to drastically simplify our `Handler` class with some of the changes made in Laravel 5.5.

PR #711, opened 2018-02-26
Removes alpha validation from last name

chloealee:
#### What's this PR do?
- Changes the last name validation to allow all characters so a user can submit their last name if they have `'` or `-`. 

👀 

#### Checklist

---
For review: …

### DIFF
--- a/app/Http/Controllers/Web/UserController.php
+++ b/app/Http/Controllers/Web/UserController.php
@@ -87,7 +87,7 @@ class UserController extends BaseController
 
         $this->registrar->validate($request, $user, [
             'first_name' => 'required|max:50',
-            'last_name' => 'nullable',
+            'last_name' => 'nullable|max:50',
             'birthdate' => 'nullable|required|date',
             'password' => 'nullable|min:6|max:512|confirmed', // @TODO: Split into separate form.
         ]);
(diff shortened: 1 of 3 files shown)


DFurnes:
Before we merge this, can we [either update the existing test](https://github.com/DoSomething/northstar/blob/a41af99f9917f3a131f312354cbee696747b584c/tests/Http/Web/WebUserTest.php#L63) or add a new one to prove this is fixed by the change & catch any accidental regressions in the future?

chloealee:
I think @DoSomething/team-bleed was added automatically as a Reviewer - is it ok to remove @DoSomething/team-bleed as an automatic reviewer?

DFurnes:
Yes! Let's swap that out to @DoSomething/team-bleed 👍

chloealee:
@DFurnes great idea! Just updated the test :)

katiecrane:
YAY Northstar!! This is so exciting!!!

app/Http/Controllers/Web/UserController.php
```diff
@@ -87,7 +87,7 @@ public function update(Request $request, $id)
 
         $this->registrar->validate($request, $user, [
             'first_name' => 'required|max:50',
-            'last_name' => 'nullable|alpha',
+            'last_name' => 'nullable',
```
DFurnes: OH WAIT. Can we add a `max:50` on this field too, to prevent abuse?

chloealee: added!

PR #712, opened 2018-02-27
Simplified "search" parameter

### LINKED ISSUE #354: Simplified "search" parameter
References DoSomething/gladiator#318.

> I've also been considering changing this endpoint to just be ?search=key and have it just automatically search by all indexes (rather than having to specify each index you want to use like search[id]=key&search[email]=key&search[mobile]=key).

This would also let us customize search logic a bit. Maybe searching by `username` (inferred email/mobile by formatting) rather than "normalizing" into each respective key, etc.

chloealee:
#### What's this PR do?
- Simplifies "search" parameter so you no longer have to specify the index you want to search by. E.g. instead of searching like `/v1/users?search[email]=admin@dosomething.org, you can now just write `/v1/users?search=admin@dosomething.org`(but both ways still work!)
- Updates documentation
- Adds test to make sure both ways of searching are supported 

- All tests should pass!
- Test out search params with new and old way and you should get same and accurate results.

#### Relevant Tickets/Cards
Fixes #354 and [this](https://www.pivotaltracker.com/n/projects/2019429/stories/148973259) pivotal card.

#### Checklist

---
For review: …


app/helpers.php
```diff
@@ -255,3 +255,64 @@ function get_client_environment_vars()
...
+    }
+}
+
+/**
+ * Normalizes email and mobile values.
+ *
+ * @param string $value
+ * @return string normalized value
+ */
+function normalizeEmailOrMobile($value)
```
DFurnes: Instead of a global helper method, let's make a `username` method on the [`Normalizer`](https://github.com/DoSomething/northstar/blob/master/app/Auth/Normalizer.php) to do this:

```php
public function normalizeUsername($value) {
    $type = $this->isEmail($credentials['username']) ? 'Email' : 'Mobile';

    return $this->normalize{$type}($value);
}
```

We can then use it as `normalize('username', $value)`.

app/helpers.php
```diff
@@ -255,3 +255,63 @@ function get_client_environment_vars()
...
+{
+    if (is_numeric($value) && strlen($value) === 10) {
```
chloealee: I actually don't think we'll have to normalize the mobile number as long as we make clear that when you use this search, you should strip the mobile number of any `+` or `1` and only have the 10 digit number? This doesn't really do anything since it doesn't look for a value that is longer than 10 digits and won't pass if someone searches by `search=+`+16178888888`

app/Http/Controllers/Traits/FiltersRequests.php
```diff
@@ -64,15 +64,18 @@ public function search($query, $searches, $indexes)
...
-        $searches = array_intersect_key($searches, array_flip($indexes));
-
-        // For the first `where` query, we want to limit results... from then on,
-        // we want to append (e.g. `SELECT * WHERE _ OR WHERE _ OR WHERE _`)
-        $firstWhere = true;
-        foreach ($searches as $term => $value) {
-            $query->where($term, '=', $value, ($firstWhere ? 'and' : 'or'));
-            $firstWhere = false;
+        if (is_string($searches)) {
+            multipleValueQuery($query, $searches, $indexes);
```
DFurnes: Let's not worry about supporting comma-separated search terms for this simplified `?search=string`:

```php
// If given a string, apply that to every indexed field on the model.
if (is_string($searches)) {
    // 'searchTerm' → ['email' => 'searchTerm', 'mobile' => 'searchTerm', ...]
    $searches = array_fill_keys($indexes, $searches);
}

// Searches may only be performed on indexed fields.
$searches = array_intersect_key($searches, array_flip($indexes));

// For the first `where` query, we want to limit results... from then on,
// we want to append (e.g. `SELECT * WHERE _ OR WHERE _ OR WHERE _`)
// ...
```

app/helpers.php
DFurnes: As long as we use the [`Normalizer`](https://github.com/DoSomething/northstar/blob/master/app/Auth/Normalizer.php)'s `mobile` normalizer for anything that looks like a phone number, this will be handled automatically!

app/helpers.php
chloealee: oh cool @DFurnes ! Where does it get the `$credentials['username']` from though?

app/Http/Controllers/Traits/FiltersRequests.php
chloealee: @DFurnes ok! Is there any reason we don't want to support this?

app/helpers.php
DFurnes: OH DUH. That should be `$value`. Updated my original comment.

app/Http/Controllers/Traits/FiltersRequests.php
DFurnes: I'm having a hard time thinking of when we'd use it, so my initial thought is to prefer the simpler change and keep the codebase cleaner. If we want to add support for comma-separated searches, let's add that capability to both the simplified `?search=string1,string2` and complex `?search[field]=string1,string2` queries in this method:

```php
// If given a string, apply that to every indexed field on the model.
if (is_string($searches)) {
    // 'searchTerm' → ['email' => 'searchTerm', 'mobile' => 'searchTerm', ...]
    $searches = array_fill_keys($indexes, $searches);
}

// Searches may only be performed on indexed fields.
$searches = array_intersect_key($searches, array_flip($indexes));

// For the first `where` query, we want to limit results... from then on,
// we want to append (e.g. `SELECT * WHERE _ OR WHERE _ OR WHERE _`)
$firstWhere = true;
foreach ($searches as $term => $values) {
    $values = explode(',', $queryString);
    foreach ($values as $value) {
        $query->where($term, '=', $value, ($firstWhere ? 'and' : 'or'));
        $firstWhere = false;
    }
}
```

chloealee: ah ok got it! Yeah I think you're right it might make more sense to not support until we need it and to also keep bother new and old versions the same. I'll update like you suggested in your first comment!

app/helpers.php
chloealee: cool! Do we only want this to work with email and mobile? I tried to do something like `role` and `source` but with this change, these don't work anymore but not sure if that matters?

DFurnes: I think we'd only want to search by unique identifiers, so just any of the fields in [`User::$uniqueIndexes`](https://github.com/DoSomething/northstar/blob/429a2a8aace98e35acf2f22e7a6597e1e97af1ab/app/Models/User.php#L127-L137). If we were looking for people with a certain `source` or `field`, we'd use `?filter[source]=value`.

chloealee: ah ok that makes sense - I'll update the documentation!

chloealee: sorry @DFurnes just to double check! Right now we allow any fields in (`User::$indexes`)[https://github.com/DoSomething/northstar/blob/master/app/Http/Controllers/UserController.php#L70] but happy to change! Just want to make sure if I change this to `User::$uniqueIndexes` that this won't break anything ?

documentation/endpoints/users.md
```diff
@@ -15,7 +15,7 @@ GET /v1/users
...
 - `filter`: Filter the collection to include _only_ users matching the following comma-separated values. For example, `/v1/users?filter[drupal_id]=10123,10124,10125` would return users whose Drupal ID is either 10123, 10124, or 10125. You can filter by one or more indexed fields.
-- `search`: Search the collection for users with fields whose value match the query. For example, `/v1/users?search[id]=test@example.com&search[email]=test@example.org` would return all users with either an ID or email address matching `test@example.org`. You can search by one or more indexed fields.
+- `search`: Search the collection for users with fields whose value match the query. You can search by `id`, `email`, `mobile`,`source`, and `role`.For example, `/v1/users?search[id]=test@example.com&search[email]=test@example.org` would return all users with either an ID or email address matching `test@example.org`. You can search by one or more indexed fields. You can also search without providing the fields. E.g. `v1/users?search=clee@dosomething.org,12345689` to search all indexes by any value. Both ways are now supported.
```
sbsmith86: If we aren't supporting comma-separated searches, does this documentation need to be updated?

app/Http/Controllers/UserController.php
```diff
@@ -65,9 +65,10 @@ public function index(Request $request)
         $afters = normalize('dates', $request->query('after'));
         $query = $this->filter($query, $afters, [User::CREATED_AT, User::UPDATED_AT], '>');
 
-        // Use `?search[column]=value` to find users matching one or more criteria.
+        // Use `?search[column]=value` or `?search=key` tto find users matching one or more criteria.
         $searches = $request->query('search');
-        $query = $this->search($query, normalize('credentials', $searches), User::$indexes);
+
+        $query = $this->search($query, normalize('credentials', $searches), User::$uniqueIndexes);
```
chloealee: will using `$uniqueIndexes` break anything for the old way?

app/Auth/Normalizer.php
```diff
@@ -110,4 +110,17 @@ protected function isEmail($value)
...
+     * Normalizes email or mobile value.
+     *
+     * @param string $value
+     * @return string
+     */
+    public function username($value)
+    {
+        $type = $this->isEmail($value) ? 'Email' : 'Mobile';
+
+        return $this->{$type}($value);
```
DFurnes: This should be `$this->normalize{$type}($value);` so it calls `$this->normalizeEmail` or `$this->normalizeMobile`, instead of calling `$this->Email` or `$this->Mobile`.

app/Http/Controllers/Traits/FiltersRequests.php
```diff
@@ -63,13 +63,25 @@ public function search($query, $searches, $indexes)
         if (! $searches) {
             return $query;
         }
+        if (is_string($searches)) {
+            // 'searchTerm' → ['email' => 'searchTerm', 'mobile' => 'searchTerm', ...]
```
DFurnes: Let's attach this comment to the line it's describing:

```php
// 'searchTerm' → ['email' => 'searchTerm', 'mobile' => 'searchTerm', ...]
$searches = array_fill_keys($indexes, $value);
```

app/Http/Controllers/UserController.php
```diff
@@ -65,9 +65,10 @@ public function index(Request $request)
         $afters = normalize('dates', $request->query('after'));
         $query = $this->filter($query, $afters, [User::CREATED_AT, User::UPDATED_AT], '>');
 
-        // Use `?search[column]=value` to find users matching one or more criteria.
+        // Use `?search[column]=value` or `?search=key` tto find users matching one or more criteria.
```
DFurnes: Typo! `tto`

documentation/endpoints/users.md
```diff
@@ -15,7 +15,7 @@ GET /v1/users
...
 - `after`: Filter the collection to include _only_ users with timestamps after the given date or datetime. For example, `/v1/users?after[created_at]=1/1/2015`. 
 - `filter`: Filter the collection to include _only_ users matching the following comma-separated values. For example, `/v1/users?filter[drupal_id]=10123,10124,10125` would return users whose Drupal ID is either 10123, 10124, or 10125. You can filter by one or more indexed fields.
-- `search`: Search the collection for users with fields whose value match the query. For example, `/v1/users?search[id]=test@example.com&search[email]=test@example.org` would return all users with either an ID or email address matching `test@example.org`. You can search by one or more indexed fields.
+- `search`: Search the collection for users with fields whose value match the query. You can search by `id`, `drupal_id`, `email`, and `mobile`. For example, you can search with or without an index: `/v1/users?search[email]=test@example.org` or `v1/users?search=clee@dosomething.org`. Both ways are now supported.
```
DFurnes: Can we re-word "with or without an index" to "with or without specifying an indexed column"?

app/Http/Controllers/UserController.php
DFurnes: I don't believe so. We've used this functionality for searching by ID, email, mobile, and (historically) Drupal ID.

chloealee: ok cool! Also, looks like searching by `facebook_id` is not working for both ways... I thought we talked about ignoring this yesterday but can't find the convo in Slack! Did we have this convo or am I 🙃 ?

DFurnes: Yeah, I wouldn't worry about Facebook ID!

app/Http/Controllers/UserController.php
chloealee: oops thanks!

app/Http/Controllers/UserController.php
chloealee: ok cool, thanks for confirming !!

app/Auth/Normalizer.php
chloealee: isn't that what we want since we're already in the `Normalizer` class? When I swap out to `$this->normalize{$type}($value);`, I get this error:
```
{
    "error": {
        "code": 500,
        "message": "Undefined property: Northstar\\Auth\\Normalizer::$normalize"
    },
    "debug": {
        "file": "/home/vagrant/DoSomething.org/northstar/app/Auth/Normalizer.php",
        "line": 124
    }
}
```

DFurnes: OMG I'm a real dummy! I thought we'd named the methods in this class differently.

Ignore that comment, although let's lower-case the methods `'email' : 'mobile'` so they match case.

chloealee: no worries!! Ah yes that makes more sense! Updated :)

PR #713, opened 2018-02-28
Fixes bug that makes you choose a country if it is blank when editing your profile

chloealee:
#### What's this PR do?
- If the user does not input a value, set country code to default value so the below error doesn't pop up and the user can edit their profile. 
screen shot 2018-02-13 at 1 04 41 pm

👀 

#### Relevant Tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/155194216) pivotal card.

#### Checklist

---
For review: …

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -4,6 +4,9 @@
 #### How should this be reviewed?
 …
 
+#### Relevant Tickets
+…
+
 #### Checklist
 - [ ] Documentation added for changed endpoints.
 - [ ] Tests added for new features/bug fixes.
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -82,7 +82,7 @@ class Registrar
         }
 
         // If there is no country, default to Fastly headers
-        if ($fields['country'] === '---') {
+        if (isset($fields['country']) && $fields['country'] === '---') {
             $fields['country'] = country_code();
         }
 
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -75,15 +75,11 @@ class Registrar
             'sms_paused' => 'boolean',
             'last_messaged_at' => 'date',
         ];
+
         // If existing user is provided, merge indexes into the request so
         // that we can validate that they exist on the "updated" document.
         if ($existingUser) {
             $fields = array_merge($existingUser->indexes(), $fields);
-        }
-
-        // If there is no country, default to Fastly headers
-        if (isset($fields['country']) && $fields['country'] === '---') {
-            $fields['country'] = country_code();
         }
 
         $validator = $this->validation->make($fields, array_merge($rules, $additionalRules));
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -75,11 +75,15 @@ class Registrar
             'sms_paused' => 'boolean',
             'last_messaged_at' => 'date',
         ];
-
         // If existing user is provided, merge indexes into the request so
         // that we can validate that they exist on the "updated" document.
         if ($existingUser) {
             $fields = array_merge($existingUser->indexes(), $fields);
+        }
+
+        // If there is no country, default to Fastly headers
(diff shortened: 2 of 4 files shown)


chloealee:
hmmm I've changed the `CODEOWNERS` file to `@dosomething/team-bleed` but when I type in `team-bleed` into the "Reviewers" section, nothing pops up! Anything else I need to do?

DFurnes:
> hmmm I've changed the `CODEOWNERS` file to `@dosomething/team-bleed` but when I type in `team-bleed` into the "Reviewers" section, nothing pops up! Anything else I need to do?

Oh!! Yes! I just added it to [Team Bleed's repositories](https://github.com/orgs/DoSomething/teams/team-bleed/repositories), so it should be good now.

chloealee:
@DFurnes ah that is it! Thank you!

app/Auth/Registrar.php
```diff
@@ -75,13 +75,17 @@ public function validate(Request $request, User $existingUser = null, array $add
...
             'last_messaged_at' => 'date',
         ];
-
         // If existing user is provided, merge indexes into the request so
         // that we can validate that they exist on the "updated" document.
         if ($existingUser) {
             $fields = array_merge($existingUser->indexes(), $fields);
         }
 
+        // If there is no country, default to Fastly headers
```
DFurnes: Let's move this into the [`UserController`](https://github.com/DoSomething/northstar/blob/2e3b2a0942162005eb06adf77d72c576d49c9fd7/app/Http/Controllers/Web/UserController.php#L73-L101). I'm a little uncomfortable with modifying the input in the `validate` method, since that doesn't really imply something that'd have a side-effect that changes the provided input. I'm also not sure we want this to run for API requests, since our server there would be in the US even if the user isn't.

chloealee: oh yes, great call! Updated :)

PULL_REQUEST_TEMPLATE.md
```diff
@@ -4,6 +4,9 @@
 #### How should this be reviewed?
 …
 
+#### Relevant Tickets
+…
+
```
katiecrane: YAY!!!

app/Http/Controllers/Web/UserController.php
```diff
@@ -85,6 +85,11 @@ public function update(Request $request, $id)
             throw new AccessDeniedHttpException;
         }
 
+        // If there is no country, default to Fastly headers
+        if ($request['country'] === '---') {
```
DFurnes: Actually, I think we can do one better on this! Comparing against `'---'` felt a little off to me, so I did a quick look for where that was actually getting set. It's in [our form template](https://github.com/DoSomething/northstar/blob/d847658f45e9eda61f55e48dd12fa0eb992618e5/resources/views/users/edit.blade.php#L84)!

Let's see if we can update that to instead be a default of the current location!

```blade
<option value="{{ $defaultCountry }}" {{ $user->country === $defaultCountry ? 'selected' : null }}>{{ get_countries()->get($defaultCountry) }}</option>
@foreach (get_countries()->except($defaultCountry) as $code => $country)
    <option value="{{ $code }}" {{ $user->country === $code ? 'selected' : null }}>{{ $country }}</option>
@endforeach
```

And then just pass in that Fastly country code (or default `US`) from [the controller's `edit` method](https://github.com/DoSomething/northstar/blob/2e3b2a0942162005eb06adf77d72c576d49c9fd7/app/Http/Controllers/Web/UserController.php#L56-L71):

```php
$defaultCountry = get_country() ?: 'US';

return view('users.edit', compact('user', 'defaultCountry'));
```

chloealee: ah @DFurnes yes this feels a lot better and safer! Thanks for digging into this! Just updated :)

DFurnes: Yay, glad that worked!! 🎉 If so we should be safe to remove this code block in the `update` method too.

chloealee: oh yes oops! Just removed :)

PR #715, opened 2018-03-01
adds v2/user endpoints

chloealee:
#### What's this PR do?
- Adds `v2/users` [resource controller & routes](https://laravel.com/docs/5.5/controllers#resource-controllers), using existing v1 controller and transformers as a template.
- Removes deprecated fields from the newly created v2 `UserTransformer` (like `_id`, `mobilecommons_id`, `mobilecommons_status`, `parse_installation_ids`, `drupal_id`, and `addr_source`)
- The `POST /v2/users` method should _not_ upsert by default (but allow `?upsert=true` to enable that behavior). (https://www.pivotaltracker.com/story/show/148973253)
- Moves existing tests in [UserTest](https://github.com/DoSomething/northstar/blob/master/tests/Http/UserTest.php) into [LegacyUserTest](https://github.com/DoSomething/northstar/blob/master/tests/LegacyHttp/UserTest.php), and creates a baseline test suite for the v2 endpoints in `UserTest`.
- Updates documentation.

- All endpoints should work correctly and all tests should pass! 
- I'd love an extra hard look at the `POST /v2/users` method to make sure I got the logic correct there to not upsert by default! 

#### Relevant Tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/155201607) Pivotal Card.

#### Checklist

---
For review: …


tests/Http/UserTest.php
```diff
@@ -297,41 +294,21 @@ public function testFieldsAreNormalized()
         ]);
     }
 
-    /**
-     * Test that we can still set old `mobilecommons_status` field.
-     * POST /users
-     *
-     * @return void
-     */
-    public function testMobileCommonsStatusFieldTransform()
```
chloealee: This is relevant anymore so just deleted!

sbsmith86:
I'm not sure why the diffs are not showing for the `UserTests` stuff but should we also have a set of tests that are namespaced to a `/two` directory?

app/Http/Controllers/Two/UserController.php
```diff
@@ -0,0 +1,190 @@
...
+     * @param Request $request
+     * @return \Illuminate\Http\Response
+     * @throws NorthstarValidationException
+     */
+    public function store(Request $request)
+    {
+        $existingUser = $this->registrar->resolve($request->only('id', 'email', 'mobile', 'drupal_id', 'facebook_id'));
+
+        // If `?upsert=true` and a record already exists, update a user with the $request fields.
+        if ($request->query('upsert') && $existingUser) {
```
DFurnes: I think there are some unintended behavior changes here. For example, there's no `$this->registrar->validate` call in the second part of the conditional that runs when `?upsert=false`.

We can re-arrange and combine these two conditionals to be a bit less confusing by using a guard. Once we check for the case where `?upsert=true` and there's an existing user, we can continue on the same for both.

```php
$existingUser = ...;

if ($existingUser && ! $request->query('upsert')) {
    throw new NorthstarValidationException(['id' => ['A record matching one of the given indexes already exists.']], $existingUser);
}

// Normalize input and validate the request
$request = normalize('credentials', $request);
$this->registrar->validate($request, $existingUser);
// and so on...
```

app/Http/Controllers/Two/UserController.php
```diff
@@ -0,0 +1,190 @@
...
+     * Display the specified resource.
+     * GET /users/:term/:id
+     *
+     * @param string $term - term to search by (eg. mobile, drupal_id, id, email, etc)
+     * @param string $id - the actual value to search for
+     *
+     * @return \Illuminate\Http\Response
+     * @throws NotFoundHttpException
+     */
+    public function show($term, $id)
```
DFurnes: For the v2 endpoints, these should not have `term`. We just want to support `/v2/users/{id}`.

app/Http/Controllers/Two/UserController.php
```diff
@@ -0,0 +1,190 @@
...
+     * Update the specified resource in storage.
+     * PUT /users/:term/:id
+     *
+     * @param string $term - term to search by (eg. mobile, drupal_id, id, email, etc)
+     * @param string $id - the actual value to search for
+     * @param Request $request
+     *
+     * @return \Illuminate\Http\Response
+     */
+    public function update($term, $id, Request $request)
```
DFurnes: Same here.

routes/api.php
```diff
@@ -18,7 +18,11 @@
     $this->get('userinfo', 'UserInfoController@show');
 
     // Users
-    // ...
+    $this->resource('users', 'Two\UserController', ['except' => ['show', 'update']]);
```
DFurnes: Once we remove `{term}` from the show & update routes, those can be merged in to this resource route!

chloealee:
@sbsmith86 great q! The pivotal card outlined to [move these to the Legacy Tests](https://www.pivotaltracker.com/n/projects/2019429/stories/155201607) and just replace here.

tests/Http/UserTest.php
```diff
@@ -297,86 +293,47 @@ public function testFieldsAreNormalized()
...
-    public function testValidatesCountryCode()
-    {
-        $this->asAdminUser()->json('POST', 'v1/users', [
-            'email' => 'american@example.com',
-            'country' => 'united states',
-        ]);
+        $this->assertResponseStatus(201);
 
-        $this->assertResponseStatus(422);
+        // The user should not have their invalid country removed.
```
DFurnes: I think this comment is backwards. Instead: `// The user should have their invalid country removed.`

tests/LegacyHttp/UserTest.php
```diff
@@ -135,30 +136,30 @@ public function testRetrieveUser()
...
-            ],
-        ]);
-
-        // And test that private profile fields are hidden for 'user' scope.
-        $data = $this->decodeResponseJson()['data'];
-        $this->assertArrayNotHasKey('email', $data);
-        $this->assertArrayNotHasKey('mobile', $data);
-        $this->assertArrayNotHasKey('last_name', $data);
-    }
+    // public function testGetPublicDataFromUser()
```
DFurnes: It looks like we're using the one below, right? If so, let's remove these commented lines!

tests/LegacyHttp/UserTest.php
```diff
@@ -797,4 +798,493 @@ public function testDeleteNoResource()
...
+        $this->assertResponseStatus(401);
+    }
+
+    /**
+     * Test that an admin can create a new user.
+     * GET /users/:term/:id
+     *
+     * @return void
+     */
+    // public function testCreateUser()
```
DFurnes: It looks like this is also already in this file, should we delete the commented lines?

sbsmith86:
Cool cool. Looks good to me!

tests/LegacyHttp/UserTest.php
chloealee: oh right sorry I forgot to put these back in!

app/Http/Controllers/Two/UserController.php
chloealee: ah yes makes so much more sense! updated :)

app/Http/Controllers/Two/UserController.php
chloealee: oooh i didn't know that! So we should not support anything like `/v2/users/email/test@dosomething.org`? If so, i'll update that along with tests and documentation.

DFurnes: Correct! The idea is for this to be a 100% standard resourceful route. There's a follow-up card for creating a `v2/email` and `v2/mobile` resource with "show" routes for those.

chloealee:
@DoSomething/team-bleed this is ready for another look!

app/Http/Controllers/Two/UserController.php
```diff
@@ -0,0 +1,183 @@
...
+                        'existing' => $existingUser->{$index},
+                    ]);
+
+                    throw new NorthstarValidationException([$index => ['Cannot upsert an existing index.']], $user);
+                }
+
+                $user = $this->registrar->register($request->except('role'), $existingUser);
+            }
+        } else {
+            $user = $this->registrar->register($request->except('role'), null);
```
DFurnes: This line doesn't need to be in both clauses of the conditional, and I think it actually accidentally snuck into the `foreach` loop in one of those cases! 😨

Let's pull it out so it's clear it's always run.

```diff
-          $user = $this->registrar->register($request->except('role'), $existingUser);
      }
- } else {
-    $user = $this->registrar->register($request->except('role'), null);
  }
+
+ $user = $this->registrar->register($request->except('role'), $existingUser);
  $code = ! is_null($existingUser) ? 200 : 201;

  return $this->item($user, $code);
```

routes/api.php
```diff
@@ -18,7 +18,9 @@
     $this->get('userinfo', 'UserInfoController@show');
 
     // Users
-    // ...
+    $this->resource('users', 'Two\UserController');
+    $this->post('users/{id}/avatar', 'Two\AvatarController@store');
```
DFurnes: This controller doesn't exist! `Two\AvatarController`

routes/api.php
```diff
@@ -18,7 +18,9 @@
...
+    $this->post('users/{id}/merge', 'Two\MergeController@store');
```
DFurnes: And likewise with this!

routes/api.php
chloealee: oh right! do we want that to work for this PR ?

routes/api.php
chloealee: and this? Or will this be handled with the merging users card?

routes/api.php
DFurnes: Just double-checked and we actually removed this altogether, since it was only used for the mobile app! Safe to remove here and the leftover reference for `v1/users/{id}/avatar`.

routes/api.php
DFurnes: Let's consider doing that when we tackle more merge work.

routes/api.php
chloealee: awesome, thanks for checking @DFurnes !

PR #716, opened 2018-03-01
Makes some little developer experience updates.

DFurnes:
#### What's this PR do?
This pull request updates bundled dependencies & makes some little developer UX improvements:

💂 Uses Gateway's `SetGuard` middleware so that calls to `Auth::check()` or `Auth::user()` always work how you'd expect, whether you're in an API route or "standard" web session.

🤙 Adds a `/callback` page that we can use when making requests on Postman or Paw. This offers a simple "Logout" button to be able to switch users without any complicated steps to clear cookies:

screen shot 2018-03-01 at 4 30 21 pm

👀

#### Checklist

### DIFF
--- a/resources/views/auth/callback.blade.php
+++ b/resources/views/auth/callback.blade.php
@@ -1,0 +1,37 @@
+@extends('layouts.app')
+
+@section('title', 'Login Status | DoSomething.org')
+
+@section('content')
+    @guest
+        <div class="container__block -centered">
+            <figure class="figure -medium">
+                <div class="figure__body footnote">
+                    You are do not have a single sign-on session.
+                </div>
+            </figure>
+        </div>
+
+        <div class="form-actions">
+            <a href="{{ url('/login') }}" class="button -secondary">Log In</a>
+        </div>
+    @endguest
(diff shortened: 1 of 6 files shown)


resources/views/auth/callback.blade.php
```diff
@@ -0,0 +1,37 @@
...
+
+
+    @auth
+        <div class="container__block -centered">
+            <figure class="figure -medium">
+                <div class="figure__media">
+                    <img class="avatar" alt="avatar" src="{{ $user->avatar or asset('avatar-placeholder.png') }}" />
+                </div>
+                <div class="figure__body">
+                    You are logged in as <strong>{{ $user->displayName() }}</strong>.
```
katiecrane: @DFurnes I'm thinking it might be helpful to have email too here

DFurnes: Oh yeah, that's a good call! ✉️

DFurnes: Boom! d237d38

screen shot 2018-03-01 at 4 55 56 pm

PR #719, opened 2018-03-06
Set up merge strategy scaffolding and logic for 2 fields

katiecrane:
#### What's this PR do?
- I did a lot of research about the strategy pattern, and full disclosure this is a toned down version of it. I didn't think we needed a new class for each type of field we can merge, so I took the general idea of the pattern and implemented it as you see here 👀 Seems to me like it will be easy enough to add new merge logic, all you gotta do is add the case to the `merge` function and then add the function with the particular logic that you want. Yay!
- Added fresh hot 🍞 merge logic for `last_authenticated_at` and `last_messaged_at`, as well as tests for those
- Update `MergeController` to:
    - Use the fresh hot 🍞 merging logic
    - Don't use `fill` since some fields can't be `filled` so we might as well manually set them all because that is cleaner

Is this what folks had in mind? Based on the card this is where it sounded like the `Merger` class should live, but if anyone has strong opinions let's hear 'em!

#### Relevant Tickets
[Card](https://www.pivotaltracker.com/story/show/155149254)

#### Checklist

### DIFF
--- a/app/Http/Controllers/MergeController.php
+++ b/app/Http/Controllers/MergeController.php
@@ -20,7 +20,6 @@ class MergeController extends Controller
      */
     protected $merger;
 
-
     /**
      * Make a new MergeController, inject dependencies,
      * and set middleware for this controller's methods.
@@ -66,7 +65,6 @@ class MergeController extends Controller
 
         // Fields that we can automatically merge
         $fieldsToMerge = array_except($duplicateFields, array_keys($intersectedFields));
-
         // Are there fields we can't automatically merge? Throw an error.
         if (count(array_intersect_key($target->toArray(), array_flip($duplicateFieldNames)))) {
             $unmergedFields = [];
--- a/app/Http/Controllers/MergeController.php
+++ b/app/Http/Controllers/MergeController.php
@@ -68,24 +68,9 @@ class MergeController extends Controller
         $fieldsToMerge = array_except($duplicateFields, array_keys($intersectedFields));
         // Are there fields we can't automatically merge? Throw an error.
         if ($intersectedFields) {
-            $unmergedFields = [];
-
             // Call merge on intersecting fields
             foreach ($intersectedFields as $field => $value) {
-                $merged = $this->merger->merge($field, $target, $duplicate);
-                if (! $merged) {
-                    array_push($unmergedFields, $field);
-                } else {
-                    $fieldsToMerge[$field] = $merged;
-                }
-            }
-
-            if ($unmergedFields) {
-                $errors = collect($unmergedFields)->map(function ($fieldName) {
-                    return 'Cannot merge "'.$fieldName.'" into non-null field on target.';
-                });
-
-                throw new NorthstarValidationException($errors, ['target' => $target, 'duplicate' => $duplicate]);
+                $fieldsToMerge[$field] = $this->merger->merge($field, $target, $duplicate);
             }
         }
 
--- a/app/Http/Controllers/MergeController.php
+++ b/app/Http/Controllers/MergeController.php
@@ -72,7 +72,7 @@ class MergeController extends Controller
 
             // Call merge on intersecting fields
             foreach ($intersectedFields as $field => $value) {
-                $merged = $this->merger->merge($field, $target->$field, $duplicate->$field);
+                $merged = $this->merger->merge($field, $target, $duplicate);
                 if (! $merged) {
                     array_push($unmergedFields, $field);
                 } else {
--- a/app/Http/Controllers/MergeController.php
+++ b/app/Http/Controllers/MergeController.php
@@ -29,8 +29,9 @@ class MergeController extends Controller
     public function __construct(UserTransformer $transformer, Merger $merger)
     {
         $this->transformer = $transformer;
-        $this->middleware('role:admin,staff');
         $this->merger = $merger;
+
+        $this->middleware('role:admin,staff');
     }
 
     /**
@@ -66,7 +67,7 @@ class MergeController extends Controller
         // Fields that we can automatically merge
         $fieldsToMerge = array_except($duplicateFields, array_keys($intersectedFields));
         // Are there fields we can't automatically merge? Throw an error.
-        if (count(array_intersect_key($target->toArray(), array_flip($duplicateFieldNames)))) {
+        if ($intersectedFields) {
             $unmergedFields = [];
 
             // Call merge on intersecting fields
--- a/app/Http/Controllers/MergeController.php
+++ b/app/Http/Controllers/MergeController.php
@@ -6,6 +6,7 @@ use Northstar\Exceptions\NorthstarValidationException;
 use Northstar\Models\User;
 use Illuminate\Http\Request;
 use Northstar\Http\Transformers\UserTransformer;
+use Northstar\Merge\Merger;
 
 class MergeController extends Controller
 {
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/MergeController.php
```diff
@@ -14,16 +15,22 @@ class MergeController extends Controller
...
      * and set middleware for this controller's methods.
      *
      * @param UserTransformer $transformer
      */
-    public function __construct(UserTransformer $transformer)
+    public function __construct(UserTransformer $transformer, Merger $merger)
     {
         $this->transformer = $transformer;
         $this->middleware('role:admin,staff');
+        $this->merger = $merger;
```
DFurnes: Let's keep injected dependencies grouped together!

```php
$this->transformer = $transformer;
$this->merger = $merger;

$this->middleware('role:admin,staff');
```

app/Http/Controllers/MergeController.php
```diff
@@ -53,17 +60,39 @@ public function store($id, Request $request)
         $duplicateFields = array_except($duplicate->toArray(), $metadata);
         $duplicateFieldNames = array_keys($duplicateFields);
 
+        // Find out which fields we need to handle merging
+        $intersectedFields = array_intersect_key($target->toArray(), array_flip($duplicateFieldNames));
+
+        // Fields that we can automatically merge
+        $fieldsToMerge = array_except($duplicateFields, array_keys($intersectedFields));
         // Are there fields we can't automatically merge? Throw an error.
         if (count(array_intersect_key($target->toArray(), array_flip($duplicateFieldNames)))) {
```
DFurnes: We can use the `$intersectedFields` variable you created above to make this a bit more readable.

app/Http/Controllers/MergeController.php
```diff
@@ -53,17 +60,39 @@ public function store($id, Request $request)
...
-            $errors = collect($duplicateFieldNames)->map(function ($fieldName) {
-                return 'Cannot merge "'.$fieldName.'" into non-null field on target.';
-            });
-
-            throw new NorthstarValidationException($errors, ['target' => $target, 'duplicate' => $duplicate]);
+            $unmergedFields = [];
+
+            // Call merge on intersecting fields
+            foreach ($intersectedFields as $field => $value) {
+                $merged = $this->merger->merge($field, $target->$field, $duplicate->$field);
```
DFurnes: I've went in circles thinking about this, so here's some unstructured thoughts:

I'm wary about using `$merged` to track both the _new_ value and an "unmergeable field" is what we do when we have merged values that are false (e.g. `sms_paused`). Right now if the Merger's mergeSmsPaused` method returned false, it'd end up in the "unmerged" array. Instead, let's throw an exception when something goes wrong.

We might want to let the Merger handle assigning the new values. For example, we'll probably want to "bundle" `sms_status` and `sms_paused` with whichever account gets the `mobile` field... and actually, speaking of, should we be resetting those to `null` if a user's mobile number changes?

This might be a good thing to huddle on IRL.

app/Merge/Merger.php
```diff
@@ -0,0 +1,32 @@
...
+
+class Merger
+{
+    public function __construct()
+    {
+    }
+
+    public function merge($field, $targetValue, $duplicateValue)
+    {
+        switch ($field) {
```
DFurnes: Instead of making a `case` for each method here, let's load these dynamically if the method exists:

```php
$mergeMethod = 'merge' . Str::studly($field);

if (! method_exists($this, $mergeMethod)) {
    return false;
}

return $this->{$mergeMethod}($targetValue, $duplicateValue);
```

app/Http/Controllers/MergeController.php
katiecrane: @DFurnes I had that at first, but that only includes the fields that we need to shove into the target that both users have. I wanted this to include all fields that we need to shove into the target, even those that only exist on the duplicate (so like duplicate might have `mobile` but target does not)

app/Merge/Merger.php
katiecrane: didn't know you could do this! way cool and way less code!

app/Http/Controllers/MergeController.php
DFurnes: Aren't they the same? Sorry, my eyes might be playing tricks on me. Specifically, thinking this:

```diff
 $intersectedFields = array_intersect_key($target->toArray(), array_flip($duplicateFieldNames));

- if (count(array_intersect_key($target->toArray(), array_flip($duplicateFieldNames)))) {
+ if (count($intersectedFields)) {
```

katiecrane: OH I was looking at the wrong thing. Yes you are right and I meant to do this!

sbsmith86:
This is very neat. Awesome, work!

katiecrane:
@DFurnes updated! After some research I decided that there wasn't much reason to make a custom exception so went with `NorthstarValidationException` because Aurora can deal with that, and it can't deal with `InvalidArgumentException`. I would like to point out though that previously we collected all the fields that we couldn't merge and presented an error for each one. This will now just throw an error on the first field that it is unable to merge.

app/Http/Controllers/MergeController.php
```diff
@@ -14,15 +14,22 @@ class MergeController extends Controller
      */
     protected $transformer;
 
+    /**
+     * @var UserTransformer
```
DFurnes: `@var Merger`

app/Http/Controllers/MergeController.php
```diff
@@ -53,17 +59,25 @@ public function store($id, Request $request)
...
-            });
+        // Find out which fields we need to handle merging
+        $intersectedFields = array_intersect_key($target->toArray(), array_flip($duplicateFieldNames));
+
+        // Fields that we can automatically merge
+        $fieldsToMerge = array_except($duplicateFields, array_keys($intersectedFields));
 
-            throw new NorthstarValidationException($errors, ['target' => $target, 'duplicate' => $duplicate]);
+        // Are there fields we can't automatically merge? Throw an error.
+        if ($intersectedFields) {
```
DFurnes: Try removing this `if`! This `$intersectedFields` variable should always be an array (it'll just be `[]` if there's no intersected fields), so it should be safe to just loop over it.

app/Http/Controllers/MergeController.php
```diff
@@ -53,17 +59,25 @@ public function store($id, Request $request)
...
+            // Call merge on intersecting fields
+            foreach ($intersectedFields as $field => $value) {
+                $fieldsToMerge[$field] = $this->merger->merge($field, $target, $duplicate);
+            }
         }
 
         // Copy the "duplicate" account's fields to the target & unset on the dupe account.
-        $target->fill($duplicateFields);
+        foreach ($fieldsToMerge as $field => $value) {
```
DFurnes: Can we combine these two loops? Let's also wrap the dynamic fields in `{}` so it's clearer what's going on.

```php
foreach ($fieldsToMerge as $field => $value) {
    $target->{$field} = $fieldsToMerge[$field];
    $duplicate->{$field} = null;
}

```

app/Merge/Merger.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+namespace Northstar\Merge;
+
+use Northstar\Exceptions\NorthstarValidationException;
+
+class Merger
+{
+    public function __construct()
```
DFurnes: This empty `__construct` can be omitted!

app/Http/Controllers/MergeController.php
katiecrane: Ahhh didn't realize you could wrap them like that yeah I like that! And yes! this is a nice catch!!!

PR #720, opened 2018-03-09
Adds `GET /v2/mobile/{mobile}` and `GET /v2/email/{email}` endpoints 

chloealee:
#### What's this PR do?
- Adds `GET /v2/mobile/{mobile}` and `GET /v2/email/{email}` endpoints 
- Updates documentation
- Adds tests for these endpoints

👀 
- All tests should pass! 

#### Relevant Tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/155562902) pivotal card. 

#### Checklist

---
For review: …

### DIFF
--- a/app/Http/Controllers/Two/EmailController.php
+++ b/app/Http/Controllers/Two/EmailController.php
@@ -2,20 +2,12 @@
 
 namespace Northstar\Http\Controllers\Two;
 
-use Northstar\Auth\Registrar;
+use Northstar\Models\User;
 use Northstar\Http\Transformers\Two\UserTransformer;
 use Northstar\Http\Controllers\Controller;
 
 class EmailController extends Controller
 {
-    /**
-     * The registrar handles creating, updating, and
-     * validating user accounts.
-     *
-     * @var Registrar
-     */
-    protected $registrar;
-
     /**
      * @var UserTransformer
      */
@@ -25,12 +17,10 @@ class EmailController extends Controller
      * Make a new EmailController, inject dependencies,
      * and set middleware for this controller's methods.
      *
-     * @param Registrar $registrar
      * @param UserTransformer $transformer
      */
-    public function __construct(Registrar $registrar, UserTransformer $transformer)
+    public function __construct(UserTransformer $transformer)
     {
-        $this->registrar = $registrar;
         $this->transformer = $transformer;
 
         $this->middleware('role:admin,staff');
@@ -40,15 +30,13 @@ class EmailController extends Controller
      * Display the specified resource.
      * GET /email/:id
      *
-     * @param string $email - the actual value to search for
+     * @param object $user
      *
      * @return \Illuminate\Http\Response
      * @throws NotFoundHttpException
      */
-    public function show($email)
+    public function show(User $user)
     {
-        $user = $this->registrar->resolveOrFail(['email' => $email]);
-
         return $this->item($user);
     }
 }
--- a/app/Http/Controllers/Two/EmailController.php
+++ b/app/Http/Controllers/Two/EmailController.php
@@ -32,20 +32,22 @@ class EmailController extends Controller
     {
         $this->registrar = $registrar;
         $this->transformer = $transformer;
+
+        $this->middleware('role:admin,staff');
     }
 
     /**
      * Display the specified resource.
      * GET /email/:id
      *
-     * @param string $id - the actual value to search for
+     * @param string $email - the actual value to search for
      *
      * @return \Illuminate\Http\Response
      * @throws NotFoundHttpException
      */
-    public function show($id)
+    public function show($email)
     {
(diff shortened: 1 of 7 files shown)


app/Http/Controllers/Two/EmailController.php
```diff
@@ -0,0 +1,52 @@
...
+     * and set middleware for this controller's methods.
+     *
+     * @param Registrar $registrar
+     * @param UserTransformer $transformer
+     */
+    public function __construct(Registrar $registrar, UserTransformer $transformer)
+    {
+        $this->registrar = $registrar;
+        $this->transformer = $transformer;
+    }
```
chloealee: do we want to restrict these endpoints to just admins and staff members (add that middleware here)? Right now, the transformer still takes roles into consideration and doesn't return sensitive info if user is not admin or staff but not sure if we wanted to restrict these endpoints to just admin/staff like is done in the v1 endpoint.

DFurnes: Yes! These should definitely be locked down to admins & staff, since otherwise a malicious user could use these endpoints to find out details about someone with a particular email or phone from our records.

app/Http/Controllers/Two/MobileController.php
```diff
@@ -0,0 +1,52 @@
...
+    /**
+     * Display the specified resource.
+     * GET /mobile/:id
+     *
+     * @param string $id - the actual value to search for
+     *
+     * @return \Illuminate\Http\Response
+     * @throws NotFoundHttpException
+     */
+    public function show($id)
```
DFurnes: Can we rename this to `$mobile`?

app/Http/Controllers/Two/EmailController.php
```diff
@@ -0,0 +1,52 @@
...
+    /**
+     * Display the specified resource.
+     * GET /email/:id
+     *
+     * @param string $id - the actual value to search for
+     *
+     * @return \Illuminate\Http\Response
+     * @throws NotFoundHttpException
+     */
+    public function show($id)
```
DFurnes: Can we rename this to `$email`?

DFurnes:
Not needed before merging, but I'd love to see if we can set up [explicit route-model-bindings](https://laravel.com/docs/5.6/routing#route-model-binding) for `{user}`, `{email}` and `{mobile}`! For example, if we set that up for email:

```php
// AppServiceProvider:
Route::bind('email', function ($value) {
    return $this->registrar->resolveOrFail(['email' => $value]);
});

// routes/api.php:
$this->get('email/{email}', 'Two\EmailController@show');

// EmailController:
public function show(User $user)
{
    return $this->item($user);
}
```

chloealee:
@DFurnes cool! Should this be in the `AppServiceProvider` or the `RouteServiceProvider`?

DFurnes:
Oh, let's drop it in `RouteServiceProvider`! 👍

app/Providers/RouteServiceProvider.php
```diff
@@ -16,16 +17,26 @@ class RouteServiceProvider extends ServiceProvider
...
      */
     public function boot()
     {
-        //
-
         parent::boot();
+
+        Route::bind('email', function ($value) {
+            return $this->registrar->resolveOrFail(['email' => $value]);
+        });
```
chloealee: thanks @DFurnes ! I'm having a little trouble using `$registrar` here - I'm getting this error: 
```
{"error":{"code":500,"message":"Call to a member function resolveOrFail() on null"},"debug":{"file":"\/home\/vagrant\/DoSomething.org\/northstar\/app\/Providers\/RouteServiceProvider.php","line":40}}
``` 
I tried to pass `Registrar $registrar` to the `boot method but wasn't able to do that :-/

DFurnes: Hmm, according to the docs that should work. Could you push up the code w/ that injected here?

chloealee: @DFurnes it's pushed up now! 

When passing in `Registrar $registrar` to the `boot()` function and trying to hit the endpoint, I get this error:

```
"Declaration of Northstar\Providers\RouteServiceProvider::boot(Northstar\Auth\Registrar $registrar) should be compatible with Illuminate\Foundation\Support\Providers\RouteServiceProvider::boot()"
```

DFurnes: Weird, maybe you can't inject dependencies into the RouteServiceProvider...? 🤔 Can you try using the `app()` helper to ask for the dependency instead?

```php
Route::bind('email', function ($value) {
    return app(Registrar::class)->resolveOrFail(['email' => $value]);
});
```

chloealee: oh @DFurnes that worked!! thank you!

PR #721, opened 2018-03-09
155798781 Add merge logic for `language` and `last_accessed_at`

katiecrane:
#### What's this PR do?
- Adds merge logic for `language`
    - Takes language from account that has most recent `last_accessed_at`
- Adds merge logic for `last_accessed_at`
    - Takes most recent one
- Adds tests for both of these mergers

Do the two merge logics make sense? Is `last_accessed_at` the field that should be used when choosing language, or should it be `last_authenticated_at` or another field?

#### Relevant Tickets
[Card](https://www.pivotaltracker.com/n/projects/2019429/stories/155798781)

#### Checklist

### DIFF
--- a/app/Merge/Merger.php
+++ b/app/Merge/Merger.php
@@ -32,4 +32,22 @@ class Merger
 
         return $targetValue > $duplicateValue ? $targetValue : $duplicateValue;
     }
+
+    public function mergeLastAccessedAt($target, $duplicate)
+    {
+        $targetValue = $target->last_accessed_at;
+        $duplicateValue = $duplicate->last_accessed_at;
+
+        return $targetValue > $duplicateValue ? $targetValue : $duplicateValue;
+    }
+
+    public function mergeLanguage($target, $duplicate)
+    {
+        // should this be last accessed at or last authenticated at?
+        if ($target->last_accessed_at > $duplicate->last_accessed_at) {
+            return $target->language;
+        }
+
+        return $duplicate->language;
+    }
 }
--- a/app/Merge/Merger.php
+++ b/app/Merge/Merger.php
@@ -43,7 +43,6 @@ class Merger
 
     public function mergeLanguage($target, $duplicate)
     {
-        // should this be last accessed at or last authenticated at?
         if ($target->last_accessed_at > $duplicate->last_accessed_at) {
             return $target->language;
         }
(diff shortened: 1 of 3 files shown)


app/Merge/Merger.php
```diff
@@ -32,4 +32,21 @@ public function mergeLastMessagedAt($target, $duplicate)
 
         return $targetValue > $duplicateValue ? $targetValue : $duplicateValue;
     }
+
+    public function mergeLastAccessedAt($target, $duplicate)
+    {
+        $targetValue = $target->last_accessed_at;
```
DFurnes: It might be handy to extract these into a shared `chooseMostRecentDate` method:

```php
public function mergeLastAuthenticatedAt($target, $duplicate)
{
    return $this->keepMostRecentDate('last_authenticated_at', $target, $duplicate);
}

public function mergeLastMessagedAt($target, $duplicate)
{
    return $this->keepMostRecentDate('last_authenticated_at', $target, $duplicate);
}

public function mergeLastAccessedAt($target, $duplicate)
{
    return $this->keepMostRecentDate('last_authenticated_at', $target, $duplicate);
}

public function chooseMostRecentDate($field, $target, $duplicate)
{
    $targetValue = $target->{$field};
    $duplicateValue = $duplicate->{$field};

    return $targetValue > $duplicateValue ? $targetValue : $duplicateValue;
}
```

app/Merge/Merger.php
```diff
@@ -32,4 +32,21 @@ public function mergeLastMessagedAt($target, $duplicate)
...
+        $duplicateValue = $duplicate->last_accessed_at;
+
+        return $targetValue > $duplicateValue ? $targetValue : $duplicateValue;
+    }
+
+    public function mergeLanguage($target, $duplicate)
+    {
+        if ($target->last_accessed_at > $duplicate->last_accessed_at) {
```
DFurnes: Wondering the same thing here with a `chooseByLastAccessedAt` helper?

DFurnes: Sidenote: another fun tool we have at our disposal now is the (hidden) `audits` object, where we track when each field was updated! Just mentioning in case it becomes handy on future merge methods.

app/Merge/Merger.php
katiecrane: yes yes!!! great call

app/Merge/Merger.php
katiecrane: OOOOOO okay, so that could make sense to use for some of the others where the field itself isn't some sort of `updated_at` timestamp situation

katiecrane:
@DFurnes updated! I did not make the `chooseByLastAccessedAt` helper because I'm not sure there is another use case for it yet

PR #727, opened 2018-03-30
Anonymous `v2/user/:id` requests should return public profile.

DFurnes:
#### What's this PR do?
We want to lock down "privileged" access to resources by using scopes (so that, for example, if an app doesn't _need_ full user profiles we just don't give it access to them). This is helpful because internal apps that use the Client Credentials grants are always given admin privileges, and so a bug or leaked key could get up to a lot of mischief!

HOWEVER, we also have scenarios where we want anonymous access to those same resources – for example, to grab first names for a set of Northstar IDs in a photo gallery. In these cases, the request won't have a scope because it doesn't have an access token at all!

This pull request updates Northstar's `RequireScope` middleware to ignore scopes on anonymous requests, just [like we did in Gateway](https://github.com/DoSomething/gateway/pull/114).

I added a failing test in f4b9bdb, and the fix in 4e7e409.

#### Relevant Tickets
I ran into this issue while working on GraphQL galleries for [#155944555](https://www.pivotaltracker.com/story/show/155944555).

#### Checklist

### DIFF
--- a/app/Auth/Scope.php
+++ b/app/Auth/Scope.php
@@ -110,8 +110,9 @@ class Scope
      */
     public static function gate($scope)
     {
-        if (has_middleware('web')) {
-            return true;
+        // Only check scopes if request is made with an OAuth token.
+        if (! request()->attributes->has('oauth_access_token_id')) {
+            return;
         }
 
         if (! static::allows($scope)) {
--- a/tests/Http/UserTest.php
+++ b/tests/Http/UserTest.php
@@ -83,7 +83,30 @@ class UserTest extends BrowserKitTestCase
      *
      * @return void
      */
-    public function testV2GetPublicDataFromUser()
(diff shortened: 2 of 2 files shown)


katiecrane:
@DFurnes this would only be a problem if I was anonymously hitting that Northstar endpoint directly? Wondering if when requesting stuff from Rogue with `include=user` you would have expected me to see this issue.

DFurnes:
Rogue is a bit weird since it has the `Registrar` make Northstar requests using its own client credentials (so they all go out as "admin" requests), and then it caches them locally. That makes for better performance but we lose some of the benefits of encoding the viewer's identity in the request.

I ran into this particular issue when my local Phoenix Next made a GraphQL request to https://graphql-qa.dosomething.org, which forwards along the token (if any) to ask Rogue for signups/posts, and Northstar for users.

PR #729, opened 2018-04-13
Adding email subscription status to Northstar

katiecrane:
#### What's this PR do?
🐻 Bear with me!
- Updates `it_should_backfill_users()` test to expect 10 calls to Blink's `userCreate`
     - Previously this was only called 5 times because the model event did not trigger because `$this->app->runningInConsole()` was `true`
    - Now that we have removed that conditional, each user gets sent TWICE, once when we manually make the call to send it in the job and once when we save the user and it gets triggered in the model event. This doesn't seem ideal, but still preserves the functionality of the backfill command.
- Removes some code in `/Jobs/SendUserToCustomerIo.php` that was outdated
- Removes the original `toCustomerIoPayload()` and renames `toBlinkPayload()` as the new `toCustomerIoPayload()`
- Adds `email_status` as a fillable field on users
- Includes `email_status` and `unsubscribed` (`true` if status is `stop`, `false` if status is anything else) in payload to Customer.io
- Removes some old conditionals from the model events
 
- Is the logic for when to return `unsubscribed` as `true` or `false` correct?
- Is there anywhere else that we should return email subscription status?

#### Relevant Tickets
[Card](https://www.pivotaltracker.com/story/show/152081036)

#### Checklist


app/Providers/AppServiceProvider.php
```diff
@@ -20,42 +20,34 @@ public function boot()
...
+            // Send payload to Blink for Customer.io profile.
 
-                $blinkPayload = $user->toBlinkPayload();
-                info('blink: user.create', $blinkPayload);
-                if (config('features.blink')) {
-                    gateway('blink')->userCreate($blinkPayload);
-                }
+            $blinkPayload = $user->toCustomerIoPayload();
+            info('blink: user.create', $blinkPayload);
+            if (config('features.blink')) {
```
DFurnes: We don't have to do it it in this PR, but I'd love to have this trigger the [SendUserToCustomerIo](https://github.com/DoSomething/northstar/blob/master/app/Jobs/SendUserToCustomerIo.php) queued job (so that this is not blocking the main thread & Blink failures wouldn't take down Northstar as well).

I'd _also_ love to pull in [that beautiful logging](https://github.com/DoSomething/rogue/blob/2baabfae167ab8c4308f7cb01588106095a5efde/app/Jobs/SendPostToQuasar.php#L59-L60) that we do in Rogue too!

tests/Console/BackfillCustomerIoTest.php
```diff
@@ -10,8 +10,8 @@ public function it_should_backfill_users()
     {
         factory(User::class, 5)->create();
 
-        // Mock Blink client & set expectation that it'll be called 5 times.
-        $this->mock(Blink::class)->shouldReceive('userCreate')->times(5);
+        // Mock Blink client & set expectation that it'll be called 10 times.
```
DFurnes: Let's keep the `->times(5)` and mark this test as incomplete until we fix that in the command.

DFurnes:
> Is the logic for when to return `unsubscribed` as `true` or `false` correct?

This seems good to me!

> Is there anywhere else that we should return email subscription status?

We should start returning it from our v2 user routes, but it might make sense to hold off until we've run our backfill and actually have real data to return.

DFurnes:
Let's also remove the [`CustomerIo`](https://github.com/DoSomething/northstar/blob/master/app/Services/CustomerIo.php) service class.

app/Providers/AppServiceProvider.php
katiecrane: @DFurnes That's a great idea! Right now I think that would create a loop though since in that job it sets `cio_full_backfill` and saves the user, which is an update and would thus trigger the job again. Is that field still something that we use?

DFurnes: We can kill that field!

katiecrane: @DFurnes Awesome! I'll make a card to kill that field and then start using the job instead!

PR #730, opened 2018-04-17
Use SendUserToCustomerIo in model events

katiecrane:
#### What's this PR do?
- Updates the model events to call `SendUserToCustomerIo` when a user is created or updated
- Update the `SendUserToCustomerIo` job to check if we should actually send to Customer.io and update logging so we get a message regardless
- Update `BackfillCustomerIoTest` - I have updated it to expect 10 calls to the Blink `userCreate` method for several reasons:
    - I tried using `assertDispatched` to check that the job was dispatched and we can do that and we can verify that it got dispatched for each Northstar ID that we want. We cannot (as far as I can tell) however, make sure that it got sent TWICE for each user. It should get triggered twice for each user because it gets triggered when we create the user and then when we run the command.
    - Checking that Blink `userCreate` was hit 5 times really only tells us that it got hit when we created the users, and what we really want to test is that it gets hit when the command gets called
- Adds `Bus::fake()` to `StandardizeBirthdatesTest` because we make a User with an improperly formatted birthday on purpose, and we don't want to trigger an event to Blink on that because it will error out (alternatively, should Blink be off for tests?)

- Question: What does the `DS_ENABLE_BLINK_UPDATES` environment variable do? It seems that it is only used in conjunction with `DS_ENABLE_BLINK` before sending an updated user to Customer.io. In this update, the `DS_ENABLE_BLINK_UPDATES` check happens in the same place, and the `DS_ENABLE_BLINK` check happens in the `SendUserToCustomerIo` job, but I'm wondering if the first check can be remove.

#### Relevant Tickets
[Card](https://www.pivotaltracker.com/story/show/156774352)

#### Checklist

### DIFF
--- a/tests/Console/StandardizeBirthdatesTest.php
+++ b/tests/Console/StandardizeBirthdatesTest.php
@@ -7,6 +7,9 @@ class StandardizeBirthdatesTest extends TestCase
     /** @test */
     public function it_should_fix_birthdates()
     {
+        // We are creating some bad data on purpose, so don't try to send it to Customer.io
+        Bus::fake();
+
         // Create some regular and borked birthday users
         factory(User::class, 5)->create();
         $this->createMongoDocument('users', ['birthdate' => '2018-03-15 00:00:00.000']);
--- a/.env.example
+++ b/.env.example
@@ -5,7 +5,6 @@ APP_KEY=
 
 # Feature Flags
 DS_ENABLE_BLINK=false
-DS_ENABLE_BLINK_UPDATES=false
 DS_ENABLE_RATE_LIMITING=false
 
 # DB_HOST=localhost
--- a/app/Jobs/SendUserToCustomerIo.php
+++ b/app/Jobs/SendUserToCustomerIo.php
@@ -41,13 +41,18 @@ class SendUserToCustomerIo implements ShouldQueue
         // Rate limit Blink/Customer.io API requests to 10/s.
         $throttler = Redis::throttle('customerio')->allow(10)->every(1);
         $throttler->then(function () {
-            if (config('features.blink')) {
-                // @TODO: update logging to be like in Rogue
+            // Send to Customer.io
+            $shouldSendToCustomerIo = config('features.blink');
+            if ($shouldSendToCustomerIo) {
                 $blinkPayload = $this->user->toCustomerIoPayload();
-                info('blink: user', $blinkPayload);
                 gateway('blink')->userCreate($blinkPayload);
             }
 
+            // Log
+            $verb = $shouldSendToCustomerIo ? 'sent' : 'would have been sent';
+            info('User ' . $this->user->id . ' ' . $verb . ' to Customer.io');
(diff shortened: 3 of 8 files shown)


DFurnes:
> Question: What does the `DS_ENABLE_BLINK_UPDATES` environment variable do? It seems that it is only used in conjunction with `DS_ENABLE_BLINK` before sending an updated user to Customer.io. In this update, the `DS_ENABLE_BLINK_UPDATES` check happens in the same place, and the `DS_ENABLE_BLINK` check happens in the `SendUserToCustomerIo` job, but I'm wondering if the first check can be remove.

This was used for a while when Blink wasn't able to accept "updates" for users. It can be removed (and just rely on `DS_ENABLE_BLINK` in both places)!

tests/Console/StandardizeBirthdatesTest.php
```diff
@@ -7,6 +7,9 @@ class StandardizeBirthdatesTest extends TestCase
     /** @test */
     public function it_should_fix_birthdates()
     {
+        // We are creating some bad data on purpose, so don't try to send it to Customer.io
+        Bus::fake();
```
DFurnes: Is this necessary? Blink calls should be faked in tests by default, I think.

DFurnes:
Let's double-check that the AWS queue runner is still happily functioning when we deploy this!

tests/Console/StandardizeBirthdatesTest.php
katiecrane: @DFurnes without this I get things like `User 5ad762f09a89200a1b261ae5 sent to Customer.io` in the logs so it seems like it's not being faked. Maybe because Blink gets called in a job, so kind of indirectly? The other option would be to turn off Blink in `phpunit.xml`.

DFurnes: Oh, gotcha. The actual API client is faked, but the job isn't. We could maybe look into whether we can disable writing to the actual log file when running tests.

katiecrane: @DFurnes The logging isn't the problem, it's the proof! The problem is that when in the job we try to get the payload for Customer.io on the user we've made with a bad birthday on purpose, we error out doing this `'birthdate' => format_date($this->birthdate, 'Y-m-d'),`

katiecrane: So we do want to make sure that we're not actually dispatching the job during this test in particular, because we know we will break it

DFurnes: OH OH OH! I see now, sorry! That makes total sense to skip then. 👍

katiecrane:
@DFurnes is the best way to test this on Thor to see that the job is getting fired?

DFurnes:
> @DFurnes is the best way to test this on Thor to see that the job is getting fired?

Yeah, let's double-check on both environments! I can swing by and look with ya.

PR #732, opened 2018-04-19
Add merges for first and last name, birthdate

katiecrane:
#### What's this PR do?
- Adds merge logic for `first_name`, `last_name`, and `birthdate` as outlined in the card
- Adds tests for merging those fields
- Adds new helper `chooseMostRecentFromAudit` which for a given field and the users to merge, chooses the field to keep by selecting the one which was updated the most recently.
    - If neither account has and `audit` field, we fall back on `last_accessed_at` and choose the field from the account most recently accessed

Is the logic right?

#### Relevant Tickets
[Card](https://www.pivotaltracker.com/story/show/155853358)

#### Checklist

### DIFF
--- a/app/Merge/Merger.php
+++ b/app/Merge/Merger.php
@@ -41,11 +41,34 @@ class Merger
         return $duplicate->language;
     }
 
+    public function mergeFirstName($target, $duplicate)
+    {
+        return $this->chooseMostRecentFromAudit('first_name', $target, $duplicate);
+    }
+
+    public function mergeLastName($target, $duplicate)
+    {
+        return $this->chooseMostRecentFromAudit('last_name', $target, $duplicate);
+    }
+
+     public function mergeBirthdate($target, $duplicate)
+    {
+        return $this->chooseMostRecentFromAudit('birthdate', $target, $duplicate);
+    }
+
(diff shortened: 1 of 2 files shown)


app/Merge/Merger.php
```diff
@@ -41,11 +41,34 @@ public function mergeLanguage($target, $duplicate)
...
     {
         $targetValue = $target->{$field};
         $duplicateValue = $duplicate->{$field};
 
         return $targetValue > $duplicateValue ? $targetValue : $duplicateValue;
     }
+
+    public function chooseMostRecentFromAudit($field, $target, $duplicate)
+    {
+        $targetUpdatedTimestamp = $target->audit[$field]['updated_at']['date'];
```
DFurnes: For older accounts, we might not have an audit record for all fields (we only started tracking these changes [earlier this year](https://github.com/DoSomething/northstar/pull/699)!) Let's choose a default to return if the audit field doesn't exist for this user.

DFurnes: Yep!

katiecrane: @DFurnes good news! If there is no audit record then `$duplicate->audit[$field]['updated_at']['date']` returns `null` rather than erroring out. Just gotta add business logic for if both are `null` 👍

PR #736, opened 2018-04-26
sends birthdate to customer.io as timestamp instead of formatted date

chloealee:
#### What's this PR do?
- Customer.io wants the user's birthdate sent as a timestamp, not formatted date. This PR updates the logic to do that! 

👀 

#### Relevant Tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/156696372) Pivotal Card.

#### Checklist

### DIFF
--- a/app/Console/Commands/BackfillCustomerIo.php
+++ b/app/Console/Commands/BackfillCustomerIo.php
@@ -31,6 +31,7 @@ class BackfillCustomerIo extends Command
     public function handle()
     {
         // @TODO: Running this command will cause users to be sent to customer.io twice
+        $query = $this->newQuery(User::class);
         $progress = $this->output->createProgressBar($query->count());
 
         $query->chunkById(200, function (Collection $users) use ($progress) {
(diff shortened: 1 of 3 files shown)


chloealee:
@katiecrane in the [Pivotal Card](https://www.pivotaltracker.com/n/projects/2019429/stories/156696372), the third task says that we need to update the [BackfillCustomerIo](https://github.com/DoSomething/northstar/blob/master/app/Console/Commands/BackfillCustomerIo.php) job but I don't think we need to since it sends through [`toCustomerIoPayload`](https://github.com/DoSomething/northstar/pull/736/files#diff-4614b2d052a2cb7486640775e2bdbb55L415) updated in this PR anyway?

katiecrane:
@chloealee I think we want to update it to remove [`$query = User::where('cio_full_backfill', '!=', true);
`](https://github.com/DoSomething/northstar/blob/master/app/Console/Commands/BackfillCustomerIo.php#L34) because we want to make sure to re-send everyone

app/Console/Commands/BackfillCustomerIo.php
```diff
@@ -31,7 +31,7 @@ class BackfillCustomerIo extends Command
     public function handle()
     {
         // @TODO: Running this command will cause users to be sent to customer.io twice
```
DFurnes: This `TODO` can be removed!

chloealee: oh right!

PR #739, opened 2018-04-30
Update birthdate un-borker to update directly in the DB

katiecrane:
#### What's this PR do?
Updates the birthdate un-borker to update date-like strings to actual dates directly in the database (we have to do it this way due to a bug in the mongo package we use, PR on that front coming soon).

Question about exception catching: @DFurnes and I talked about only catching the specific exception thrown when calling `Carbon::parse($value)` on a `$value` that can't be converted to a date. We only want to catch the specific exception thrown because if we catch one, we will unset `birthdate` on that user. However, the exception thrown in that case is of class `Exception`, so it seems like we _have_ to catch that?

#### Relevant Tickets
[Card](https://www.pivotaltracker.com/story/show/156050050)

#### Checklist

### DIFF
--- a/app/Console/Commands/StandardizeBirthdates.php
+++ b/app/Console/Commands/StandardizeBirthdates.php
@@ -57,7 +57,7 @@ class StandardizeBirthdates extends Command
                     $user->save();
                 }
 
-                if (! $date) {
+                if (! $user->birthdate) {
                     info('northstar:bday - removed invalid birthdate from '.$user->id.' - '.$value);
                 } else {
                     info('northstar:bday - updated user '.$user->id.' birthdate from '.$value.' to '.$date);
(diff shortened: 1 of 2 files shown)


DFurnes:
> We only want to catch the specific exception thrown because if we catch one, we will unset `birthdate` on that user. However, the exception thrown in that case is of class `Exception`, so it seems like we _have_ to catch that?

Darn it, PHP! 😞

app/Console/Commands/StandardizeBirthdates.php
```diff
@@ -49,14 +50,14 @@ public function handle()
                 // If possible, switch the birthday to a date type, otherwise, wipe it!
                 try {
                     $date = Carbon::parse($value);
+                    app('db')->collection('users')->where(['_id' => $user->id])
+                ->update(['birthdate' => new UTCDateTime($date->getTimestamp() * 1000)]);
```
DFurnes: Can you indent this a wee bit?

PR #743, opened 2018-05-09
updates .env.example file

chloealee:
#### What's this PR do?
Updates `.env.example` file to include `SQS_DEFAULT_QUEUE` (thanks @mirie for noting it was missing!) 

👀 

#### Relevant Tickets
Forgot to add this in #741 

#### Checklist

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -28,6 +28,10 @@ QUEUE_DRIVER=sync
 SQS_DEFAULT_QUEUE=northstar-dev-customerio
 SQS_BACKFILL_QUEUE=northstar-dev-backfill
 
+#SQS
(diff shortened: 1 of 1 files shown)


DFurnes:
Do we want to include other SQS environment variables here too? I'd left them out since we shouldn't be connecting to SQS on our local machines, but could be nice to see in a table of contents (rather than having to peek at [config/queue.php](https://github.com/DoSomething/northstar/blob/e0e5c9b420199d055f9af0ece2695684d52c6ccb/config/queue.php#L65-L71).

chloealee:
that's a great call @DFurnes ! Just added :)

PR #744, opened 2018-05-11
Remove unnecessary data and make it unfillable

katiecrane:
#### What's this PR do?
- Removes the attributes specified [here](https://www.pivotaltracker.com/n/projects/2019429/stories/157162537) as fillable fields on a User
- Removes `setInterestsAttribute` because that field was removed
- Removes `cgg_id` from `$casts`
- Creates `northstar:unset` (instead of a migration because the equivalent of "dropping columns" in Mongo requires going through every user)
    - Takes as arguments all the fields that you want to 🔥 and shoves them into an array
    - Asks the user to confirm that they really want to remove that field (since we are deleting a bunch of stuff this felt nice and safe)

Example run of `northstar:unset`:
```
vagrant@homestead:~/Sites/northstar$ php artisan northstar:unset pig dog cat

 Have you made sure that there is NOT a broadcast in progress or starting soon? (yes/no) [no]:
 > yes

 Are you sure you want to remove this field from ALL USERS? `pig` (yes/no) [no]:
 > yes

Field removed from all users: pig

 Are you sure you want to remove this field from ALL USERS? `dog` (yes/no) [no]:
 > no

Did NOT remove dog

 Are you sure you want to remove this field from ALL USERS? `cat` (yes/no) [no]:
 > yes

Field removed from all users: cat
```

Can we use this to unset all the fields that we want to? Did I fail to remove any fields? 

#### Relevant Tickets
[Card](https://www.pivotaltracker.com/story/show/157162537)

#### Checklist

### DIFF
--- a/app/Console/Commands/UnsetFieldCommand.php
+++ b/app/Console/Commands/UnsetFieldCommand.php
@@ -13,7 +13,7 @@ class UnsetFieldCommand extends Command
      *
      * @var string
      */
-    protected $signature = 'northstar:unset {field*}';
+    protected $signature = 'northstar:unset {field*} {--force}';
 
     /**
      * The console command description.
@@ -40,11 +40,16 @@ class UnsetFieldCommand extends Command
     public function handle()
     {
         $fieldsToRemove = $this->argument('field');
+        $force = $this->option('force');
 
         foreach ($fieldsToRemove as $field) {
-            $burnItDown = $this->confirm('Are you sure you want to remove this field from ALL USERS? `'.$field.'`');
+            $burnItDown = false;
 
-            if ($burnItDown) {
+            if (! $force) {
+                $burnItDown = $this->confirm('Are you sure you want to remove this field from ALL USERS? `'.$field.'`');
+            }
+
+            if ($burnItDown || $this->option('force')) {
                 info('Removing field from all users: '.$field);
 
                 $usersToUnset = (new User)->newQuery();
@@ -62,6 +67,8 @@ class UnsetFieldCommand extends Command
 
                 $progressBar->finish();
                 info('Field removed: '.$field);
+            } else {
+                $this->line('Did NOT remove '.$field);
             }
         }
     }
--- a/app/Console/Commands/UnsetFieldCommand.php
+++ b/app/Console/Commands/UnsetFieldCommand.php
@@ -1,0 +1,58 @@
+<?php
+
+namespace Northstar\Console\Commands;
+
+use Illuminate\Console\Command;
+use Illuminate\Support\Collection;
+
+class UnsetFieldCommand extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'northstar:unset {field*}';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Unset the given field from all users.';
+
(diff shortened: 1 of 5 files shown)


app/Console/Commands/UnsetFieldCommand.php
```diff
@@ -0,0 +1,75 @@
...
+
+            if (! $force) {
+                $burnItDown = $this->confirm('Are you sure you want to remove this field from ALL USERS? `'.$field.'`');
+            }
+
+            if ($burnItDown || $this->option('force')) {
+                info('Removing field from all users: '.$field);
+
+                $usersToUnset = (new User)->newQuery();
+                $usersToUnset->whereRaw([$field => ['$exists' => true]]);
```
DFurnes: Can this be combined to `$usersToUnset = User::whereRaw([...])`?

app/Console/Commands/UnsetFieldCommand.php
```diff
@@ -0,0 +1,75 @@
...
+
+                $this->line('Progess removing '.$field.':');
+                $progressBar = $this->output->createProgressBar($usersToUnset->count());
+
+                $usersToUnset->chunkById(200, function (Collection $users) use ($progressBar, $field) {
```
DFurnes: I'm not sure how the performance differs, but you should be able to use an unset query to do these at once:

```php
User::whereRaw([$field => ['$exists' => true]])->unset($field);
```

katiecrane: @DFurnes Yeah that's what I had originally but then I thought it might not be the besttttt way to go. I wasn't sure how many users we'd be dealing with for any given field, but it will probably be most of them for `cio_full_backfill`

katiecrane: @DFurnes based on the thread here: https://dosomething.slack.com/archives/C02BBRN5A/p1526073769000241 seems like we do need to chunk because fields that are on more users time out when trying to use `unset` like that? Or do you think that was just the Tinker session?

DFurnes: Yeah, we may be in trouble if the request is timing out. I'm still pretty convinced a unset query is going to be orders of magnitude faster than iterating over each user (although who knows without testing it out!).

Can you try removing the timeout from this query [with `->timeout(-1)`](https://github.com/jenssegers/laravel-mongodb/blob/17b6864f2e27cb9a274840f1b1364dea1760be40/README.md#mongodb-specific-operations)?

katiecrane: @DFurnes should I still do that in an `artisan tinker` or should this just get written into the command now?

DFurnes: Let's try it out in an Artisan tinker first!

app/Console/Commands/UnsetFieldCommand.php
```diff
@@ -0,0 +1,74 @@
...
+     */
+    public function handle()
+    {
+        $force = $this->option('force');
+
+        if ($force || $this->confirm('Have you made sure that there is NOT a broadcast in progress or starting soon?')) {
+            $fieldsToRemove = $this->argument('field');
+
+            foreach ($fieldsToRemove as $field) {
+                $burnItDown = false;
```
DFurnes: 🔥🔥🔥

PR #745, opened 2018-05-18
Run composer update

katiecrane:
#### What's this PR do?
Just ran `composer update` and included the updated `composer.lock` file.

👀 

#### Relevant Tickets
[Card](https://www.pivotaltracker.com/story/show/157288532)

#### Checklist

### DIFF
--- a/app/Auth/Repositories/ClientRepository.php
+++ b/app/Auth/Repositories/ClientRepository.php
@@ -19,7 +19,7 @@ class ClientRepository implements ClientRepositoryInterface
      *
      * @return \League\OAuth2\Server\Entities\ClientEntityInterface
      */
-    public function getClientEntity($clientIdentifier, $grantType, $clientSecret = null, $mustValidateSecret = true)
+    public function getClientEntity($clientIdentifier, $grantType = null, $clientSecret = null, $mustValidateSecret = true)
     {
         /** @var \Northstar\Models\Client $model */
         $model = Client::where('client_id', $clientIdentifier)->first();
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
     "league/flysystem-aws-s3-v3": "~1.0",
     "league/fractal": "0.13.*",
     "league/iso3166": "^1.0",
-    "league/oauth2-server": "~6.0.0",
+    "league/oauth2-server": "^7.0.0",
     "predis/predis": "^1.1",
     "symfony/psr-http-message-bridge": "^1.0.0",
     "zendframework/zend-diactoros": "^1.3",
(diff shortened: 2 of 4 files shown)


DFurnes:
We can do this in a separate pull request, but we should also upgrade `oauth2-server` to the [latest release](https://github.com/thephpleague/oauth2-server/blob/master/CHANGELOG.md#710---released-2018-04-22). We're still on 6.x, but luckily looks like the only breaking change is dropping old PHP support.

katiecrane:
@DFurnes would that be to update to `~7.0.0`? And would that require any additional testing?

DFurnes:
I think it'd be a good idea to set our constraint to `^7.0.0` so we get minor updates as well when we run `composer update` (like `7.1.0`, which is the current latest release).

DFurnes:
As long as the test suite passes, I think we should be safe to update.

katiecrane:
@DFurnes I tried updating locally and tests are failing 😩 
```
vagrant@homestead:~/Sites/northstar$ vendor/bin/phpunit
PHPUnit 5.7.27 by Sebastian Bergmann and contributors.

........................PHP Fatal error:  Declaration of Northstar\Auth\Repositories\ClientRepository::getClientEntity($clientIdentifier, $grantType, $clientSecret = NULL, $mustValidateSecret = true) must be compatible with League\OAuth2\Server\Repositories\ClientRepositoryInterface::getClientEntity($clientIdentifier, $grantType = NULL, $clientSecret = NULL, $mustValidateSecret = true) in /home/vagrant/Sites/northstar/app/Auth/Repositories/ClientRepository.php on line 9
PHP Fatal error:  Uncaught Illuminate\Contracts\Container\BindingResolutionException: Target [Illuminate\Contracts\Debug\ExceptionHandler] is not instantiable. in /home/vagrant/Sites/northstar/vendor/laravel/framework/src/Illuminate/Container/Container.php:918
```

DFurnes:
If you update that `getClientEntity` to have the default value `= NULL` for `$clientSecret`, does that clear up the test failures?

katiecrane:
@DFurnes ah I think it was `$grantType = null` that was missing (we had no default before). But one test is still failing:
```
1) OAuthTest::testAuthorizationCodeGrant
Expected status code 200, got 400.
Failed asserting that 400 matches expected 200.

/home/vagrant/Sites/northstar/vendor/laravel/browser-kit-testing/src/Concerns/MakesHttpRequests.php:744
/home/vagrant/Sites/northstar/tests/Http/OAuthTest.php:17
/home/vagrant/Sites/northstar/tests/Http/OAuthTest.php:78
```

DFurnes:
Oh oof yeah, that's the one I meant! I can swing by if you want to dig into that `OAuthTest` failure.

PR #750, opened 2018-05-25
Voter registration status on registration form a/b test

katiecrane:
#### What's this PR do?
Creates an a/b called `voter-status-reg-form` test to include a question about voter registration status as part of the registration form. Users who see the new form will see this:
image
The new question is optional and no `voter_registration_status` will be saved to their profile if they do not answer. If they do answer, we save it!

QUESTIONS:
1. What do we want to call the "no" option? For yes I have it saving as "confirmed" and for not sure I have it saving as "uncertain" to follow the conventions that we currently use with TurboVote.

2. Right now folks only convert if they submit the registration form. Meaning, folks who hit this page and click “login” will count as unconverted participants. However, traffic is split equally among the two options (new form and original form), so I’d expect there to be an equal number of folks already registered and trying to login hitting each alternative of the experiment. Therefore, while this will deflate conversion rates on both tests, I think it will do so approximately equally and we will still get all the information we need to know. I’ve also never done a/b testing before and there could be some incorrect assumptions in there, so this is in the questions section to make sure this is okay.

Does the form look okay? Are we converting the right folks? Saving the right information?

#### Relevant Tickets
[Card](https://www.pivotaltracker.com/story/show/157196236)

#### Checklist

### DIFF
--- a/app/Http/Controllers/Web/AuthController.php
+++ b/app/Http/Controllers/Web/AuthController.php
@@ -2,17 +2,18 @@
 
 namespace Northstar\Http\Controllers\Web;
 
-use Illuminate\Contracts\Auth\Factory as Auth;
-use Illuminate\Foundation\Validation\ValidatesRequests;
-use Illuminate\Http\Request;
-use Illuminate\Routing\Controller as BaseController;
-use League\OAuth2\Server\AuthorizationServer;
-use Northstar\Auth\Entities\UserEntity;
-use Northstar\Auth\Registrar;
-use Northstar\Exceptions\NorthstarValidationException;
 use Northstar\Models\User;
+use Illuminate\Http\Request;
+use Northstar\Auth\Registrar;
 use Psr\Http\Message\ResponseInterface;
+use Northstar\Auth\Entities\UserEntity;
 use Psr\Http\Message\ServerRequestInterface;
+use League\OAuth2\Server\AuthorizationServer;
+use SeatGeek\Sixpack\Session\Base as Sixpack;
+use Illuminate\Contracts\Auth\Factory as Auth;
+use Illuminate\Routing\Controller as BaseController;
+use Northstar\Exceptions\NorthstarValidationException;
+use Illuminate\Foundation\Validation\ValidatesRequests;
 
 class AuthController extends BaseController
 {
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/Web/AuthController.php
```diff
@@ -173,7 +174,17 @@ public function getLogout(Request $request)
      */
     public function getRegister()
     {
-        return view('auth.register');
+        $experiment = false;
```
DFurnes: Let's rename this `$showVoterStatusForm` for clarity.

app/Http/Controllers/Web/AuthController.php
```diff
@@ -173,7 +174,17 @@ public function getLogout(Request $request)
...
+
+        if (config('services.sixpack.enabled')) {
+            $sixpack = app(Sixpack::class);
```
DFurnes: I think we could simplify this a bit with a helper functions that handle feature-flagging and returning the instance:

```php
function participate($experiment, $alternatives) {
    if (! config('services.sixpack.enabled')) {
        return $alternatives[0];
    }

    return app(Sixpack:class)
        ->participate($experiment, $alternatives)
        ->getAlternative();
}

function convert($experiment) {
    if (! config('services.sixpack.enabled')) {
        return;
    }

    return app(Sixpack::class)->convert($experiment);
}

```

Then we can just call those anywhere in our controllers:

```php
public function getRegister()
{
    $showVoterStatusForm = participate('voter-status-reg-form', ['normal_form', 'voter_form']);
    
    return view('auth.register', ['voter_reg_status_form' => showVoterStatusForm]);
}

{
    convert('voter-status-reg-form');

    $this->registrar->validate($request, null, [
        // ...
}
```

resources/views/auth/register.blade.php
```diff
@@ -28,43 +28,109 @@
             </div>
         @endif
 
-        <form id="profile-registration-form" method="POST" action="{{ url('register') }}">
-            <input name="_token" type="hidden" value="{{ csrf_token() }}">
+        @if ($voter_reg_status_form)
```
DFurnes: Do we need to have the entire form in this conditional, or could we just wrap the voter reg question?

PR #751, opened 2018-05-31
Updates to allow keys to be stored locally or elsewhere

katiecrane:
#### What's this PR do?
- Adds new `KeyRepository` for accessing all our keys. If we don't have it cached locally, we grab it from our storage driver (which for now is "local" but could be "S3" in the future) and cache it. 
- Updates everything else to use this new `KeyRepository` when reading and writing keys
- Moves keys to be stored locally into `app/keys` from plain `keys` and updates symlink to reflect this

Is the symlink correct? Is everything access keys in the new way correctly?

#### Relevant Tickets
[Card](https://www.pivotaltracker.com/story/show/157289052)

#### Checklist


sheyd:
@katiecrane For the path directory permissions here: https://github.com/DoSomething/northstar/pull/751/files#diff-80af33ee3b258b7cc62886fd82c20343R40, can we set it to `mkdir(storage_path('cache/keys'), 0770, true);` instead? `0777` dir perms allow the directory to world-writeable, and a potential security vulnerability. The web-user/group should be able to write/read with `0770` perms.

katiecrane:
@DFurnes those environment variables are now set on staging and Thor, so I'm merging!

PR #755, opened 2018-06-05
Command to update given fields on users from csv

katiecrane:
#### What's this PR do?
- Adds `UpdateUserFieldsCommand` (which I foresee being used mostly for data clean-up) that takes in a csv and which fields will be updated. It then iterates through each user given in the csv, and updates each field based on the new value in the csv.
- Adds tests for this command too! (Which of course took longer than writing the command itself 😋)

Is this substantially reusable? Any errors get written into the test?

#### Relevant Tickets
[Card](https://www.pivotaltracker.com/story/show/157479967)

#### Checklist


app/Console/Commands/UpdateUserFieldsCommand.php
```diff
@@ -0,0 +1,71 @@
...
+    public function handle()
+    {
+        // Make a local copy of the CSV
+        $path = $this->argument('path');
+        $this->line('Loading in csv from '.$path);
+
+        $temp = tempnam('temp', 'command_csv');
+        file_put_contents($temp, fopen($this->argument('path'), 'r'));
+
+        // Load the missing signups from the CSV
```
DFurnes: 👾 `missing signups` → `users`

app/Console/Commands/UpdateUserFieldsCommand.php
```diff
@@ -0,0 +1,71 @@
...
+        $users_csv = Reader::createFromPath($temp, 'r');
```
DFurnes: Super unimportant, but we tend to write variables in `$camelCase` and only use `snake_case` for database column fields, like `$user->first_name`.

PR #761, opened 2018-06-13
Adding Procfile

blisteringherb:
#### What's this PR do?
This adds a Procfile so that Northstar can be successfully deployed on Heroku. 

This branch has been deployed on dosomething-northstar-dev and dosomething-northstar-qa successfully.

#### Relevant Tickets
https://github.com/DoSomething/devops/issues/383

#### Checklist

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,0 +1,3 @@
+web: composer warmup && vendor/bin/heroku-php-nginx -C nginx.conf public/
+
+release: php artisan migrate --force
--- a/Procfile
+++ b/Procfile
@@ -1,3 +1,5 @@
 web: composer warmup && vendor/bin/heroku-php-nginx -C nginx.conf public/
 
 release: php artisan migrate --force
+
+queue: php artisan queue:work --tries=3 --sleep=5
--- a/composer.json
+++ b/composer.json
@@ -77,7 +77,6 @@
       "@php artisan package:discover"
     ],
     "warmup": [
-      "@php artisan route:cache",
       "@php artisan config:cache"
     ]
   },
--- a/composer.json
+++ b/composer.json
@@ -75,6 +75,11 @@
     "post-autoload-dump": [
       "Illuminate\\Foundation\\ComposerScripts::postAutoloadDump",
       "@php artisan package:discover"
+    ],
+    "warmup": [
+      "@php artisan gateway:key",
+      "@php artisan route:cache",
+      "@php artisan config:cache"
     ]
   },
   "config": {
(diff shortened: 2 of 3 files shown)


Procfile
```diff
@@ -0,0 +1,3 @@
+web: composer warmup && vendor/bin/heroku-php-nginx -C nginx.conf public/
```
DFurnes: The file being pointed at here `-C nginx.conf` doesn't exist. You can use [Rogue's `nginx.conf` as a template](https://github.com/DoSomething/rogue/blob/master/nginx.conf).

Procfile
```diff
@@ -0,0 +1,3 @@
...
+
+release: php artisan migrate --force
```
DFurnes: We need to define a queue worker for this application too. You can use [Rogue's Procfile](https://github.com/DoSomething/rogue/blob/cda20f052d4a8fbcbeccb918881dd93e7a02e1fd/Procfile#L5) as a template.

Procfile
DFurnes: Oh, the `composer warmup` script doesn't exist here either. Again, Rogue is a [good template](https://github.com/DoSomething/rogue/blob/cda20f052d4a8fbcbeccb918881dd93e7a02e1fd/composer.json#L59-L63). In this case, we'll want to only run `@php artisan route:cache` and `@php artisan config:cache` since Northstar doesn't need to ask itself for a public key.

blisteringherb: @DFurnes I've added that nginx conf file.

blisteringherb: I put that warmup code into the composer.json file.

DFurnes: Let's remove `@php artisan gateway:key` since Northstar doesn't need to ask itself for a public key.

blisteringherb:
The deploy finishes without errors, but there's still a `Unable to prepare route [status] for serialization. Uses Closure.` Any hints on that @DFurnes ?

DFurnes:
> The deploy finishes without errors, but there's still a `Unable to prepare route [status] for serialization. Uses Closure.` Any hints on that @DFurnes ?

Ah, we need to remove closure routes, like [this one](https://github.com/DoSomething/northstar/blob/3f69f4f10c947b3d4a1b1526380f3c93cedf32ec/routes/api.php#L61-L64), before we can enable route caching. For now, we can just remove that from the warmup script.

blisteringherb:
This looks good. There are errors, but they're now related to parsing the Redis URL. I'll open a new PR for that.

PR #777, opened 2018-07-18
Prevent password overlay overlap in Safari.

DFurnes:
#### What's this PR do?
This had been bugging me a bit. In Safari, before:

no good

And after:

good

👀

#### Relevant Tickets
N/A

#### Checklist


katiecrane:
Does this also fix this funkiness in Chrome? 
image

DFurnes:
It might! I don't have that on my Chrome, but I wouldn't be surprised if it worked the same way.

PR #781, opened 2018-07-19
Add debug middleware & update dependencies.

DFurnes:
#### What's this PR do?
This pull request adds a `LogMemoryUsage` middleware that prints out how much memory was used by the request to see if we can [drop our memory limit](https://github.com/DoSomething/northstar/commit/4cec55896d2eac9f00e9bc3b8ebbdc48daf44eb8) even more & get more concurrent requests per dyno. I've added this as a "debug" log message so we can toggle it on and off with `APP_LOG_LEVEL`.

Here's an example of what these log messages look like from my local:

```
[2018-07-19 15:13:14] local.DEBUG: memory_usage {"mb":3.274952,"path":"v2/users/5b50a5fb9a892007c52df162"} {"user_id":"5b50a5fc9a892007c52df163","client_id":"oauth-test-client","request_id":null}
[2018-07-19 15:13:16] local.DEBUG: memory_usage {"mb":3.731272,"path":"v2/users"} {"user_id":"5b50a5fc9a892007c52df163","client_id":"oauth-test-client","request_id":null}
```

It also updates dependencies (including Webpack 2 → 4) since builds were failing on my machine.

👀

#### Relevant Tickets
DoSomething/devops#383

#### Checklist

### DIFF
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -15,6 +15,7 @@ class Kernel extends HttpKernel
         \Illuminate\Foundation\Http\Middleware\CheckForMaintenanceMode::class,
         \Illuminate\Foundation\Http\Middleware\ConvertEmptyStringsToNull::class,
         \Illuminate\Foundation\Http\Middleware\ValidatePostSize::class,
+        \Northstar\Http\Middleware\LogMemoryUsage::class,
         \Northstar\Http\Middleware\TrimStrings::class,
         \Fideloper\Proxy\TrustProxies::class,
     ];
--- a/app/Http/Middleware/LogMemoryUsage.php
+++ b/app/Http/Middleware/LogMemoryUsage.php
@@ -1,0 +1,30 @@
+<?php
+
+namespace Northstar\Http\Middleware;
+
+use Closure;
+use Illuminate\Support\Facades\Log;
+
+class LogMemoryUsage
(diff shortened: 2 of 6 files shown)


katiecrane:
@DFurnes is that memory limit per request? So is what we're doing here getting a feel for how big a request could be to see if we can safely lower that limit?

DFurnes:
Yep, exactly! Right now we [allocate a maximum of 32MB per-request](https://github.com/DoSomething/northstar/commit/4cec55896d2eac9f00e9bc3b8ebbdc48daf44eb8), which Heroku uses to set our limit to 80 concurrent requests per performance dyno (w/ 2.5GB RAM). Dropping that further (say, to `16M` or `8M`) might let us handle more requests per dyno (and save some moolah).

chloealee:
so cool @DFurnes ! One q - if we log every request, will that fill up the log limit?

DFurnes:
It probably would! The plan is to only set `APP_LOG_LEVEL` to `debug` for a little bit so we can get a good sample of memory usage, and then toggle off this log message by swapping it to `info`.

chloealee:
ah I see! Love it, @DFurnes !

katiecrane:
@DFurnes this is really cool I can't wait to see the results!

PR #787, opened 2018-08-20
Reduce 'max_execution_time' to match Fastly timeout.

DFurnes:
#### What's this PR do?
I've noticed some 503s when this app is under load that aren't ending up in our logs, and I think that's because Fastly is cutting things off while the request continues to process behind the scenes. This pull request reduces the max amount of time we'll let PHP process a request on this application so we can see how much this is actually happening.

Here's [the Heroku default](https://github.com/heroku/heroku-buildpack-php/blob/b5c4a0b77a725c4ab9dc1736e5a8217824790093/support/build/_conf/php/php.ini#L437-L440) and [PHP documentation](https://github.com/heroku/heroku-buildpack-php/blob/b5c4a0b77a725c4ab9dc1736e5a8217824790093/support/build/_conf/php/php.ini#L437-L440).

#### Relevant Tickets
References DoSomething/devops#435.

#### Checklist

### DIFF
--- a/public/.user.ini
+++ b/public/.user.ini
@@ -1,3 +1,8 @@
 # Reduce memory limit (from default 128MB) so
 # we can process more concurrent requests.
 memory_limit = 32M
+
+# Reduce max execution time (from default 30s)
+# so we don't continue processing requests that
+# Fastly has already returned a 503 for.
+max_execution_time = 15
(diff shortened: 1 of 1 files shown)


DFurnes:
Thanks for the review! Going to hold off on merging this until after tomorrow's big broadcast so we can get some additional logging from the Fastly side of things (see linked issue above for deets).

PR #789, opened 2018-08-28
Update to use key object instead of string password

katiecrane:
#### What's this PR do?
- Adds new environment variable `APP_AUTH_KEY` to be generated by running `vendor/bin/generate-defuse-key`
    - Adds this variable to `config/auth.php` as `key`
- In a couple of oauth related places where we used to use `APP_KEY` we now use `Key::loadFromAsciiSafeString(config('auth.key'))`
- Update docs to match what we have in the tests

Did I update the docs correctly? I was trying to update the docs corresponding to the final request in the `testRevokeRefreshToken` test, but it's kind of confusing because there are multiple requests you can make to that same `token` endpoint.

Also, before deploying to an environment we will have to add this new environment variable.

#### Relevant Tickets
https://github.com/DoSomething/northstar/issues/779
[Card](https://www.pivotaltracker.com/story/show/159171074)

#### Checklist


app/Providers/OAuthServiceProvider.php
```diff
@@ -3,31 +3,32 @@
...
+use Northstar\Auth\Repositories\AuthCodeRepository;
+use League\OAuth2\Server\Grant\ClientCredentialsGrant;
+use Northstar\Auth\Repositories\AccessTokenRepository;
+use Northstar\Auth\Repositories\RefreshTokenRepository;
+use League\OAuth2\Server\Repositories\UserRepositoryInterface;
+use League\OAuth2\Server\Repositories\ScopeRepositoryInterface;
+use League\OAuth2\Server\Repositories\ClientRepositoryInterface;
+use League\OAuth2\Server\Repositories\AuthCodeRepositoryInterface;
+use League\OAuth2\Server\Repositories\AccessTokenRepositoryInterface;
+use League\OAuth2\Server\Repositories\RefreshTokenRepositoryInterface;
```
DFurnes: Beautiful! 🔠

config/auth.php
```diff
@@ -71,6 +71,8 @@
         ],
     ],
 
+    'key' => env('APP_AUTH_KEY'),
```
DFurnes: Let's add some documentation for this, following the format used by the other config properties:

```php
    /*
     |--------------------------------------------------------------------------
     | CUSTOM: OAuth Encryption Key
     |--------------------------------------------------------------------------
     |
     | This is the plain-text encryption key used when creating refresh tokens.
     | It may be generated by running `vendor/bin/generate-defuse-key`.
     |
     */
```

DFurnes: And actually, let's move this to the top of the file too so it's more noticeable.

.env.example
```diff
@@ -2,6 +2,7 @@ APP_URL=http://northstar.test
 APP_ENV=local
 APP_DEBUG=true
 APP_KEY=
+APP_AUTH_KEY=
```
DFurnes: Let's add a comment above this line also describing where it comes from:

```sh
# Token encryption key, generated by `php artisan northstar:setup`
# or `vendor/bin/generate-defuse-key` scripts:
```

DFurnes:
> Did I update the docs correctly? I was trying to update the docs corresponding to the final request in the testRevokeRefreshToken test, but it's kind of confusing because there are multiple requests you can make to that same token endpoint.

I think that reads clearly! 👌

DFurnes:
Let's also make sure we update the `northstar:setup` command to create an encryption key and automatically place it in `APP_AUTH_KEY` (kinda like we do for `APP_KEY`) so we don't have to remember to do this step when setting up a new local environment!

katiecrane:
@DFurnes yup! I'm working on updating `northstar:setup` currently!

.env.example
```diff
@@ -2,6 +2,9 @@ APP_URL=http://northstar.test
 APP_ENV=local
 APP_DEBUG=true
 APP_KEY=
+# Token encryption key, generated by `php artisan northstar:setup`
```
DFurnes: Nitpicky, could you add a new-line separating this from the section above it?

DFurnes:
Let's also do some testing before we push this to production – will this invalidate all existing refresh tokens? If so, we should probably just clear out that collection. (I'd also like to consider adding an expiration to those, but that's another ticket.)

katiecrane:
@DFurnes yes this is way to scary to go to prod without some good testing!

PR #797, opened 2018-09-14
Enable CORS for Northstar's API.

DFurnes:
#### What's this PR do?
This pull request enables [cross-origin resource sharing](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing) for Northstar (for example, so we can load a user's profile from https://vote.dosomething.org). We'd previously overridden this in Fastly in DoSomething/devops#416, but I'd like to clean this up and move it into the application now.

Here's an working example from my local (simulating an AJAX request from [example.com](https://example.com)):

screen shot 2018-09-14 at 2 24 07 pm

One note – I had to change our PHP version constraint to accept any `7.2.*` release since my Homestead [is running PHP 7.2.7 and so failing with the `~7.2.9` constraint](https://user-images.githubusercontent.com/583202/45568047-7c194180-b829-11e8-94bc-dbd32df7716a.png) we'd added in #794.

#### Relevant Tickets
DoSomething/devops#439

#### Checklist

### DIFF
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -39,6 +39,7 @@ class Kernel extends HttpKernel
         'api' => [
             \Northstar\Http\Middleware\ParseOAuthHeader::class,
             \Illuminate\Routing\Middleware\SubstituteBindings::class,
+            \Barryvdh\Cors\HandleCors::class,
             'guard:api',
         ],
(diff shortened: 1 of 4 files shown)


DFurnes:
And with the updated Wercker box we're back in business!

PR #801, opened 2018-09-17
exposes email_frequency in user transformer

chloealee:
#### What's this PR do?
exposes email_frequency in user transformer

👀 

#### Relevant Tickets
[This](https://www.pivotaltracker.com/n/projects/2019429/stories/160204153) Pivotal Card.

#### Checklist

### DIFF
--- a/app/Http/Transformers/UserTransformer.php
+++ b/app/Http/Transformers/UserTransformer.php
@@ -59,6 +59,7 @@ class UserTransformer extends TransformerAbstract
             // Subscription status
             $response['sms_status'] = $user->sms_status;
             $response['sms_paused'] = (bool) $user->sms_paused;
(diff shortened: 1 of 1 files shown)


app/Http/Transformers/UserTransformer.php
```diff
@@ -59,6 +59,7 @@ public function transform(User $user)
             // Subscription status
             $response['sms_status'] = $user->sms_status;
             $response['sms_paused'] = (bool) $user->sms_paused;
+            $response['email_frequency'] = isset($user->email_frequency) ? $user->email_frequency : null;
```
DFurnes: I don't believe this needs to be wrapped in an `isset`. If `email_frequency` is not set, it'll just be `null`.

chloealee: ah good point! 💡

PR #802, opened 2018-09-18
northstar:update - Declare variable outside foreach, add unset

katiecrane:
#### What's this PR do?
- Do what someone did [here](https://stackoverflow.com/questions/9985030/how-does-the-garbage-collector-work-in-php), which is declaring the variable that we create in the `foreach` loop outside of it instead
- Unset `$userToUpdate` each time we loop through

Does this make sense to test? Locally I've been seeing different results than @DFurnes or on QA. This didn't make anything better for me, but neither did @DFurnes changes yesterday that helped in QA and for him.

#### Relevant Tickets
…

#### Checklist

### DIFF
--- a/app/Console/Commands/UpdateUserFieldsCommand.php
+++ b/app/Console/Commands/UpdateUserFieldsCommand.php
@@ -73,6 +73,8 @@ class UpdateUserFieldsCommand extends Command
         $this->totalCount = count($usersCsv);
         $currentCount = 0;
 
+        $user = null;
+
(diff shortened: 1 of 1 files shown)


katiecrane:
@chloealee @DFurnes any thoughts on if I should deploy this to QA and test there first or if we should just merge it in since it can't hurt (I don't think?)

DFurnes:
> @chloealee @DFurnes any thoughts on if I should deploy this to QA and test there first or if we should just merge it in since it can't hurt (I don't think?)

I'm all good with just merging this in.

PR #804, opened 2018-09-27
Voter registration status mutator

katiecrane:
#### What's this PR do?
- Adds a setter for `voter_registration_status` to compare the old status and the new status, and only update to the new status if it is higher up on the voter registration status hierarchy
- Adds 2 tests:
    - One to make sure a user can be updated to a higher status
    - One to make sure that a user cannot be updated to a lower status

These are the values we want to use?

#### Relevant Tickets
[Card](https://www.pivotaltracker.com/story/show/160807904)

#### Checklist

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -464,4 +464,28 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
     {
         $this->notify(new ResetPasswordNotification($token));
     }
+
+    /**
+     * Set the voter_registration_status based on the hierarchy.
+     *
+     * @param string $status
+     */
+    public function setVoterRegistrationStatusAttribute($status)
+    {
+        $statusHierarchy = [
+            'uncertain',
+            'ineligible',
+            'confirmed',
+            'registration_complete',
+        ];
+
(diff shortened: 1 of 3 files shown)


app/Models/User.php
```diff
@@ -464,4 +464,28 @@ public function sendPasswordResetNotification($token)
...
     }
+
+    /**
+     * Set the voter_registration_status based on the hierarchy.
+     *
+     * @param string $status
+     */
+    public function setVoterRegistrationStatusAttribute($status)
+    {
+        $statusHierarchy = [
```
DFurnes: Let's add these to [a validation rule in the Registrar](https://github.com/DoSomething/northstar/blob/62df9013aea09f44aed46fc62ac7b29ef7a3b66f/app/Auth/Registrar.php#L66-L78) as well.

DFurnes:
One suggestion, but otherwise looks perfect. Thanks for adding tests for this! 🙌

app/Models/User.php
katiecrane: @DFurnes ah yeah, I was thinking we already did this but never actually checked!

PR #805, opened 2018-09-27
Adds fields to User for Voting Make a Plan 

chloealee:
#### What's this PR do?
- Adds fields to User so we can save someone's make a voting plan info:
  - Adds `voting_plan_status`
  - Adds `voting_plan_method_of_transport`
  - Adds `voting_plan_time_of_day`
  - Adds `voting_plan_attending_with`
- Fields are fillable and sent to C.io / Quasar 
- Fields added to transformers 

- Can you set these on a user? 
- Are they sent to C.io / Quasar?
- Do you see these in the transformer?

#### Relevant Tickets
[This](https://www.pivotaltracker.com/n/projects/2019429/stories/160652546) Pivotal Card.

cc @rapala61 @aaronschachter @sheyd @blisteringherb 

#### Checklist


chloealee:
Just to double check - [`toCustomerIoPayload`](https://github.com/DoSomething/northstar/pull/805/files#diff-4614b2d052a2cb7486640775e2bdbb55R389) also sends to Quasar, correct?

katiecrane:
@chloealee I believe Quasar still scrapes Northstar data from the API

app/Models/User.php
```diff
@@ -410,6 +419,23 @@ public function toCustomerIoPayload()
...
+        }
+
+        if ($this->voting_plan_time_of_day) {
+            $payload['voting_plan_time_of_day'] = $this->voting_plan_time_of_day;
+        }
+
+        if ($this->voting_plan_attending_with) {
+            $payload['voting_plan_attending_with'] = $this->voting_plan_attending_with;
+        }
+
```
katiecrane: What your thoughts are on this alternative (doesn't feel like there is really an elegant way to do this):
```
$votingPlanAttributes = ['voting_plan_status', 'voting_plan_method_of_transport', 'voting_plan_time_of_day', 'voting_plan_attending_with'];

foreach ($votingPlanAttributes as $votingPlanAttribute) {
     if ($this->{$votingPlanAttribute}) {
         $payload[$votingPlanAttribute] = $this->{$votingPlanAttribute};
    }
}
```

Do you think this makes it too hard to follow?

database/migrations/2018_09_23_222315_add_voting_make_a_plan_fields_to_user_table.php
```diff
@@ -0,0 +1,37 @@
...
+    public function down()
+    {
+        Schema::table('users', function ($table) {
+            $table->dropColumn('voting_plan_status');
+            $table->dropColumn('voting_plan_method_of_transport');
+            $table->dropColumn('voting_plan_time_of_day');
+            $table->dropColumn('voting_plan_attending_with');
+        });
+    }
+}
```
katiecrane: We don't need a migration to add new attributes in Mongo! :parrot:

katiecrane:
@chloealee Do we need any validation for the voting_plan fields?

chloealee:
@katiecrane great q! No validation is needed yet because O'Doyle is still working on what will go in the fields.

database/migrations/2018_09_23_222315_add_voting_make_a_plan_fields_to_user_table.php
chloealee: oh, right! So they just automatically get added if sent in?

app/Models/User.php
chloealee: I think this is a lot cleaner - I'll update! Thanks, Katie!

aaronschachter:
Question about our `voting_plan` fields -- will they be exposed in a public `GET /users` request? We'll need them available anonymously for the work we're doing ito speed up broadcast sends over in https://www.pivotaltracker.com/story/show/160585940.

chloealee:
@aaronschachter currently in this PR, these will only be available for [admins and if the user is able to view the full profile](https://github.com/DoSomething/northstar/pull/805/files#diff-bc53d4c1a0c9ecf56419002440b32f78R30). I can move this out of this though if we're ok with exposing this information to everyone!

aaronschachter:
We don't have a need for `voting_plan_status` to be public, but we do want the other fields exposed so we can render them in reminder messaging - e.g. `"Don't forget to {{user.voting_plan_method_of_transport}} to the {{user.voting_plan_time_of_day}} tomorrow!"`

chloealee:
cool @aaronschachter updated and moved those out of the admin and full profile checks!

database/migrations/2018_09_23_222315_add_voting_make_a_plan_fields_to_user_table.php
katiecrane: @chloealee yeah it just has to be a fillable field on the user, no db updates necessary!

chloealee: cool, thanks @katiecrane !

app/Models/User.php
DFurnes: Sorry for the belated comment, but I don't think we need to exclude these at all! We have some logic around `email_frequency` because we have to worry about the order of operations as we migrate this field from Customer.io into Northstar (so we don't overwrite existing data). For these, just sending null is totally fine.

chloealee: ah makes sense! @DFurnes updated this [this PR](https://github.com/DoSomething/northstar/compare/voting-plan-update?expand=1)!

PR #807, opened 2018-10-11
Ready for review but not merge - Add Fastly caching for users requested anonymously

katiecrane:
#### What's this PR do?
- Adds `.env` variables for Fastly (will have to be updated manually in Heroku)
- Adds Fastly to `config/services.php`
- Adds a Fastly service class to purge keys
- Uses the Fastly service class in the UserController (`v1`) for purging keys when a user is updated
- Includes the `Surrogate-Key` header in a response when a user is requested anonymously

- Do we want to use the same feature flag for adding the Surrogate Key header? I figure that doesn't really break anything if it is sent unnecessarily

#### Relevant Tickets
[Card](https://www.pivotaltracker.com/story/show/160290066)

#### Checklist


app/Http/Controllers/UserController.php
```diff
@@ -144,7 +152,13 @@ public function show($term, $id)
 
         $user = $this->registrar->resolveOrFail([$term => $id]);
 
-        return $this->item($user);
+        $response = $this->item($user);
+
+        if (! Auth::user()) {
+            $response->headers->set('Surrogate-Key', 'user-'.$user->id);
```
katiecrane: @DFurnes Hmm locally I see this header show up, but not sure why it's not showing up on `dev`? Other changes are showing up so I know the branch is really deployed

DFurnes: Oh, I think Fastly might remove that in responses.

katiecrane: hmm [this is how we do it in Rogue](https://github.com/DoSomething/rogue/blob/ec8f8fdfffb10eb3adaa3a15f6837392b78c60ba/app/Http/Controllers/Legacy/Web/ImagesController.php#L69) - I just went off of that. How can we include it in the headers if Fastly removes it?

DFurnes: Oh sorry, to clarify: Fastly _does_ use that header in order to mark cached pages & remove them via the API, it just doesn't seem to forward it along in the response to users. For example, check the headers on this cached image in Rogue: https://activity.dosomething.org/images/56292

katiecrane: @DFurnes oooo okay I see that that one has a `x-cache` `HIT` so much have had the `Surrogate-Key` header. 

On `dev` re-loading the same user still comes up as a `MISS` and my only thought was that the `surrogate-key` header wasn't getting sent - any ideas on how to debug this?

DFurnes: Huh, that's definitely strange. Let's look into this together tomorrow AM!

katiecrane: @DFurnes I wonder if it has to do with [this](https://github.com/DoSomething/devops/issues/321#issuecomment-344313051) where api responses aren't cached?

DFurnes: It shooooouldn't, since we've since rebuilt those Fastly properties. Maybe something's off between the new "Backends (Development)" and "Backends (QA)" configs...

app/Http/Controllers/Two/UserController.php
```diff
@@ -135,7 +143,13 @@ public function show($id)
     {
         $user = User::findOrFail($id);
 
-        return $this->item($user);
+        $response = $this->item($user);
+
+        if (! Auth::user()) {
```
DFurnes: Let's use `Auth::check` here instead of returning the full user model:

```suggestion
        if (! Auth::check()) {
```

katiecrane: @DFurnes omg I didn't know that existed!!!

DFurnes: Oh wait, or better yet `Auth::guest()`!

app/Http/Controllers/Two/UserController.php
```diff
@@ -135,7 +143,13 @@ public function show($id)
...
+            $response->headers->set('Surrogate-Key', 'user-'.$user->id);
```
DFurnes: Leaving a note here for future reference. Katie and I dug into this a bit using [Fastly's `Fastly-Debug` header](https://docs.fastly.com/guides/debugging/checking-cache#using-a-fastly-debug-header-with-curl) to verify that the `Surrogate-Key` was being set correctly, and see what `Cache-Control` was provided for this response:

```
$ curl -Is "https://identity-dev.dosomething.org/v1/users/id/5639066ba59dbfe6598b4567" -H "Fastly-Debug:1"
...
cache-control: no-cache, private
surrogate-key: user-5639066ba59dbfe6598b4567
x-cache: MISS
x-cache-hits: 0
```

By default, Laravel says that requests aren't cacheable (with `no-cache, private`), so we need to use `$response->setPublic()` to set the correct `Cache-Control` header so that Fastly knows it can cache the response!

app/Services/Fastly.php
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    /**
+     * Purge object from Fastly cache based on give cache key
+     *
+     * @param $cacheKey String
+     */
+    public function purgeKey($cacheKey)
+    {
+        $fastlyConfigured = config('features.glide') &&
```
DFurnes: Oh, we gotta remove `config('features.glide')` here!

app/Services/Fastly.php
```diff
@@ -0,0 +1,53 @@
...
+            ! is_null(config('services.fastly.url')) &&
+            ! is_null(config('services.fastly.key')) &&
+            ! is_null(config('services.fastly.service_id')) &&
+            isset($cacheKey);
+
+        if (! $fastlyConfigured) {
```
DFurnes: We can also remove the check for these conditionals in the controllers since it checks here!

app/Services/Fastly.php
```diff
@@ -0,0 +1,53 @@
...
+            info('image_cache_purge_failed', ['response' => 'Fastly not configured correctly on this environment.']);
```
DFurnes: Let's rename this log message to `cache_purge_failed`, and remove the success one below.

app/Services/Fastly.php
katiecrane: @DFurnes a-ha! so we can! we can do that in Rogue too then!

app/Http/Controllers/UserController.php
```diff
@@ -172,6 +186,11 @@ public function update($term, $id, Request $request)
 
         $this->registrar->register($request->all(), $user);
 
+        // Purge Fastly cache of user
+        if (! is_null(config('services.fastly.url')) && ! is_null(config('services.fastly.key')) && ! is_null(config('services.fastly.service_id'))) {
+            $this->fastly->purgeKey('user-'.$user->id);
```
DFurnes: Let's do this [with a model event](https://github.com/DoSomething/northstar/blob/85cc3a4e9d3ceafb60d0d54a881a180990a3d832/app/Providers/AppServiceProvider.php#L41-L44) instead, in case we edit a user elsewhere (like the Northstar web UI or a script)

katiecrane: @DFurnes brilliant!!!

config/services.php
```diff
@@ -79,4 +79,10 @@
         'prefix' => env('SIXPACK_COOKIE_PREFIX'),
         'timeout' => env('SIXPACK_TIMEOUT'),
     ],
+
+    'fastly' => [
+        'url' => env('FASTLY_URL'),
```
DFurnes: This is nitpicky, but let's just inline `'https://api.fastly.com/'` here [like we do in Aurora](https://github.com/DoSomething/aurora/blob/ecb10ff32389b26e2f4b1adf8ab3d7dcf02cc15f/config/services.php#L37). There's not much use in making an environment variable for something we never swap out between environments.

katiecrane:
Update: Deployed this branch to `dev` and successfully did this a couple times!
1. Grab a user for the first time and see a cache `MISS`
2. Grab that user a couple more times and see cache `HIT`s
3. Update that user in Aurora
4. Grab that user again and see a `MISS`
5. Grab that user a couple more times and see cache `HIT`s

So this PR is ready for final review @DoSomething/team-bleed

PR #809, opened 2018-10-19
Mandatory voter reg ab test

katiecrane:
#### What's this PR do?
- Adds new a/b test `voter-status-mandatory`
- If the person gets opted in to the test and submit the form without picking a voter registration status option they get one of these:
image

- Otherwise, they get the usual form.

- They convert after the new user is created.

Are the alternatives correct? Also a note that testing the normal form with `http://northstar.test/register?sixpack-force-voter-status-mandatory=normal_form` does not always work because in `postRegister` we call the `participate` function again to determine how we should validate, and I guess that overwrites the parameter passed in the url. If your session was opted into the other alternative, you'd have the reverse problem.

Also, I thought I was going to have to use `auth.js` but turns out I don't but when I saved the file a bunch of formatting stuff changed and I'm not sure if that's good and is cleaning it up or is bad and is messing it up.

#### Relevant Tickets
[Card](https://www.pivotaltracker.com/story/show/158556196)

#### Checklist

### DIFF
--- a/app/Http/Controllers/Web/AuthController.php
+++ b/app/Http/Controllers/Web/AuthController.php
@@ -173,7 +173,9 @@ class AuthController extends BaseController
      */
     public function getRegister()
     {
-        return view('auth.register');
+        $mandatoryVoterStatus = participate('voter-status-mandatory', ['normal_form', 'mandatory_voter_form']);
+
+        return view('auth.register', ['voter_status_mandatory' => $mandatoryVoterStatus]);
     }
 
     /**
--- a/app/Http/Controllers/Web/AuthController.php
+++ b/app/Http/Controllers/Web/AuthController.php
@@ -190,21 +190,16 @@ class AuthController extends BaseController
     public function postRegister(Request $request)
     {
         // Grab alternative for this user again (it will be the same)
-        $mandatoryVoterStatus = session('ab_voter_status_mandatory');
+        $mandatoryVoterStatus = session('ab_voter_status_mandatory') === 'mandatory_voter_form' ? true : false;
 
-        $validationRules = [
+        $this->registrar->validate($request, null, [
             'first_name' => 'required|max:50',
             'birthdate' => 'required|date|before:now',
             'email' => 'required|email|unique:users',
             'mobile' => 'mobile|nullable|unique:users',
             'password' => 'required|min:6|max:512',
-        ];
-
-        if ($mandatoryVoterStatus === 'mandatory_voter_form') {
-            $validationRules['voter_registration_status'] = 'required';
-        }
-
-        $this->registrar->validate($request, null, $validationRules);
+            'voter_registration_status' => $mandatoryVoterStatus ? 'required' : '',
+        ]);
 
         // Register and login the user.
         $editableFields = $request->except(User::$internal);
--- a/app/Http/Controllers/Web/AuthController.php
+++ b/app/Http/Controllers/Web/AuthController.php
@@ -210,6 +210,8 @@ class AuthController extends BaseController
             }
         });
 
+        convert('voter-status-mandatory');
+
         $this->auth->guard('web')->login($user, true);
 
         return redirect()->intended('/');
--- a/app/Http/Controllers/Web/AuthController.php
+++ b/app/Http/Controllers/Web/AuthController.php
@@ -175,8 +175,9 @@ class AuthController extends BaseController
     {
         // Mandatory voter reg question experiment
         $mandatoryVoterStatus = participate('voter-status-mandatory', ['normal_form', 'mandatory_voter_form']);
+        session(['ab_voter_status_mandatory' => $mandatoryVoterStatus]);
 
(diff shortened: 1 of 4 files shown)


resources/views/auth/register.blade.php
```diff
@@ -66,30 +66,57 @@
...
+                    <div class="form-item -reduced">
+                        <label class="option -radio">
+                          <input type="radio" name="voter_registration_status" value="uncertain">
+                          <span class="option__indicator"></span>
+                          <span>I'm not sure</span>
+                        </label>
+                    </div>
                 </div>
-            </div>
+            @endif
```
chloealee: @katiecrane I could totally be overlooking something but these look like the same code to me between the (lines 70-92 and lines 95-118?)?

chloealee:
> Also a note that testing the normal form with http://northstar.test/register?sixpack-force-voter-status-mandatory=normal_form does not always work because in postRegister we call the participate function again to determine how we should validate, and I guess that overwrites the parameter passed in the url. If your session was opted into the other alternative, you'd have the reverse problem.

Interesting! Would this give accurate results if it doesn't always work and overrides then?

app/Http/Controllers/Web/AuthController.php
```diff
@@ -187,13 +188,22 @@ public function getRegister()
      */
     public function postRegister(Request $request)
     {
-        $this->registrar->validate($request, null, [
+        // Grab alternative for this user again (it will be the same)
+        $mandatoryVoterStatus = participate('voter-status-mandatory', ['normal_form', 'mandatory_voter_form']);
```
DFurnes: Another idea that just occurred to me – what if we store the participated A/B test in the session? So:

```php
public function getRegister()
{
    $mandatoryVoterStatus = participate('voter-status-mandatory', ['normal_form', 'mandatory_voter_form']);
    session(['ab_voter_status_mandatory' => $mandatoryVoterStatus]);
    // ...
}

{
    $mandatoryVoterStatus = session('ab_voter_status_mandatory');
    // ...
```

This would avoid making two hits to Sixpack, and should fix the issue you flagged w/ manually testing this.

resources/assets/js/validators/auth.js
```diff
@@ -1,23 +1,23 @@
-const Validation = require('dosomething-validation');
+const Validation = require("dosomething-validation");
```
DFurnes: Hmm, is this from a Prettier editor integration without a config in this project? Might be nice to set up here with our settings (notably, single quotes & trailing commas), or if not just remove this file from the commit.

resources/views/auth/register.blade.php
DFurnes: Same! Were these left over from your stab into using the front-end validation package?

katiecrane: Ah! @DFurnes @chloealee good eye!!! Yes this is leftover from when I thought there would be something different on the frontend - totally forgot that nothing actually ended up being different there!

app/Http/Controllers/Web/AuthController.php
katiecrane: @DFurnes I LOVE THIS IDEA!

resources/assets/js/validators/auth.js
katiecrane: Opted to remove the file since no other changes were made to it

katiecrane:
@chloealee It would have given inaccurate results when testing using the `sixpack-force-voter-status-mandatory=normal_form` parameter in the URL, but with @DFurnes suggestion to add the chosen alternative to the session, that works too now!

app/Http/Controllers/Web/AuthController.php
```diff
@@ -185,13 +189,22 @@ public function getRegister()
...
             'mobile' => 'mobile|nullable|unique:users',
             'password' => 'required|min:6|max:512',
-        ]);
+        ];
+
+        if ($mandatoryVoterStatus === 'mandatory_voter_form') {
+            $validationRules['voter_registration_status'] = 'required';
+        }
+
+        $this->registrar->validate($request, null, $validationRules);
```
DFurnes: Can we inline this conditional rule to the array, or would Laravel pout?

```php
$this->registrar->validate($request, null, [
    'first_name' => 'required|max:50',
    'birthdate' => 'required|date|before:now',
    'email' => 'required|email|unique:users',
    'mobile' => 'mobile|nullable|unique:users',
    'password' => 'required|min:6|max:512',
    'voter_registration_status' => $mandatoryVoterStatus ? 'required' : null,
]);
```

katiecrane: @DFurnes It doesn't seem to like that, it throws `array_map(): Argument #2 should be an array` when I try to do that and `$mandatoryVoterStatus` is `false`

DFurnes: Can you try again with `: ''` or `: []` instead of `: null`?

katiecrane: a-ha @DFurnes ! worked with `''`

PR #811, opened 2018-11-02
Set max-age on cached user profiles.

DFurnes:
#### What's this PR do?
This pull request sets a `max-age` of a year when caching user profiles in Fastly (this seems to be [considered a reasonable maximum](https://stackoverflow.com/a/25201898/811624), and is what Glide uses for image responses in Rogue).

Here's an example of the headers set for `v2/users/5bc4bfc5fdce270a29006472` on my local:

```http
Cache-Control: max-age=31536000, public
Surrogate-Key: user-5bc4bfc5fdce270a29006472
```

#### Relevant Tickets
DoSomething/infrastructure#31

#### Checklist

### DIFF
--- a/app/Http/Controllers/Two/UserController.php
+++ b/app/Http/Controllers/Two/UserController.php
@@ -139,8 +139,9 @@ class UserController extends Controller
         $response = $this->item($user);
 
         if (Auth::guest()) {
+            // If this is an anonymous request, cache in Fastly until it changes.
             $response->headers->set('Surrogate-Key', 'user-'.$user->id);
-            $response->setPublic();
+            $response->setPublic()->setMaxAge(60 * 60 * 24 * 365);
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/Two/UserController.php
```diff
@@ -139,8 +139,9 @@ public function show($id)
         $response = $this->item($user);
 
         if (Auth::guest()) {
+            // If this is an anonymous request, cache in Fastly until it changes.
             $response->headers->set('Surrogate-Key', 'user-'.$user->id);
-            $response->setPublic();
+            $response->setPublic()->setMaxAge(60 * 60 * 24 * 365);
```
katiecrane: @DFurnes thank you for writing the math out this way!!!!

PR #816, opened 2018-11-27
changes oauth2 to version 7.3.1

chloealee:
#### What's this PR do?
changes oauth2 to version 7.3.1 to see if this clears up issues with login

👀 

#### Relevant Tickets
#814 

#### Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
     "league/flysystem-aws-s3-v3": "~1.0",
     "league/fractal": "0.13.*",
     "league/iso3166": "^1.0",
-    "league/oauth2-server": "^7.0.0",
+    "league/oauth2-server": "7.3.1",
     "mongodb/mongodb": "~1.2.0",
     "predis/predis": "^1.1",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -29,7 +29,7 @@
     "league/flysystem-aws-s3-v3": "~1.0",
     "league/fractal": "0.13.*",
     "league/iso3166": "^1.0",
-    "league/oauth2-server": "^7.0.0",
```
katiecrane: was `7.3.1` what we ended up using before since we had it set to `^7.0.0`?

chloealee: @katiecrane so @DFurnes flagged that he thought [this](https://github.com/thephpleague/oauth2-server/compare/7.3.0...7.3.2#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR12) might be the culprit! So we are going to try to set it before this release to see if that fixes things!

katiecrane: @chloealee a-ha! clever!!

PR #820, opened 2018-12-13
Import email subscription status from cio

katiecrane:
#### What's this PR do?
- Script: The script grabs all of our users who have email addresses, and kicks off a `GetEmailSubStatusFromCustomerIo` job for each user. 
- Job: The `GetEmailSubStatusFromCustomerIo` job is passed one user. The job then grabs that user from [the Customer.io Beta API](https://customer.io/docs/api/#apibeta-apicustomerscustomers_get_attributes) and updates their `email_frequency` to `none` or `active` depending on if they are currently subscribed or not. These jobs will run on the low priority queue so nothing gets clogged up. The jobs are also limited to 10 per second as per the Customer.io Beta API requirements.

We will be able to kick of the script and just sit back as the queue processes through all the `GetEmailSubStatusFromCustomerIo` jobs. No further steps will need to be taken, yay!!! Huge shoutout to @DFurnes for his idea to use queue jobs in this way!!

Is this properly rate-limited? Will we end up having current email subscription status for all users?

#### Relevant Tickets
[Card 1](https://www.pivotaltracker.com/story/show/161428034)
[Card 2](https://www.pivotaltracker.com/story/show/160204038)

#### Checklist

### DIFF
--- a/app/Console/Commands/ImportSubStatusFromCio.php
+++ b/app/Console/Commands/ImportSubStatusFromCio.php
@@ -34,8 +34,8 @@ class ImportSubStatusFromCio extends Command
         $query = (new User)->newQuery();
         $query = $query->where('email', 'exists', true);
 
-        $query->chunkById(200, function (Collection $users){
-            $users->each(function (User $user){
+        $query->chunkById(200, function (Collection $users) {
+            $users->each(function (User $user) {
                 $queue = config('queue.names.low');
 
                 dispatch(new GetEmailSubStatusFromCustomerIo($user))->onQueue($queue);
--- a/app/Console/Commands/ImportSubStatusFromCio.php
+++ b/app/Console/Commands/ImportSubStatusFromCio.php
@@ -59,7 +59,7 @@ class ImportSubStatusFromCio extends Command
 
                     // Do not update this user if they were last updated after the specified date
                     if ($lastUpdated > $date) {
-                        continue;
+                        return false;
                     }
                 }
 
--- a/app/Console/Commands/ImportSubStatusFromCio.php
+++ b/app/Console/Commands/ImportSubStatusFromCio.php
@@ -2,7 +2,6 @@
 
 namespace Northstar\Console\Commands;
 
-use Carbon\Carbon;
 use Northstar\Models\User;
 use Illuminate\Console\Command;
 use Illuminate\Support\Collection;
@@ -14,15 +13,14 @@ class ImportSubStatusFromCio extends Command
      *
      * @var string
      */
-    protected $signature = 'northstar:importsub
-                            {--beforeDate= : Only pull updates for users whose status was last updated before this date.}';
+    protected $signature = 'northstar:importsub';
 
     /**
      * The console command description.
      *
      * @var string
      */
-    protected $description = 'Grab email subscription status from Customer.io for users on which it is not already set.';
+    protected $description = 'For each user, kick off a job to grab email subscription status from Customer.io.';
 
     /**
      * Execute the console command.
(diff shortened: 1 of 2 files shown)


DFurnes:
> Includes `--beforeDate=` option for re-runability. Say you ran the script on December 10th and come in on December 11th to find that Heroku has brought down your one-off dyno. Never fear! Re-start the script with `--beforeDate=2018-12-10` to only update users who were last updated _before_ December 10th, so you're not just replaying the work that the script did yesterday.

I think it'd be a good idea to create a queue job for these (and run it on the low-priority `backfill` queue), so the script just has to queue up the jobs & then they can continue processing without human intervention until they're done. (Mirroring the approach we took for this C.io backfill in #741).

app/Console/Commands/ImportSubStatusFromCio.php
```diff
@@ -0,0 +1,89 @@
...
+                }
+                if (! $unsubscribed) {
+                    $user->email_frequency = 'active';
+                }
+                $user->save();
+
+                $this->line('Updated user: '.$user->id);
+
+                // Make sure we don't go over 10 requests per second
+                sleep(.1);
```
DFurnes: This is tricky! [`sleep`](http://php.net/manual/en/function.sleep.php) actually accepts an integer and so this will be rounded down to `0`. We created a [`throttle`](https://github.com/DoSomething/northstar/blob/fa873dc92081a8468997d592490dbc24eb831754/app/helpers.php#L214-L234) helper for this that uses `usleep` behind-the-scenes and accepts a rate-limit per minute:

```suggestion
                throttle(600);
```

app/Console/Commands/ImportSubStatusFromCio.php
```diff
@@ -0,0 +1,89 @@
...
+                // Make request to c.io to get that user's subscription status
+                $response = $client->get('/v1/api/customers/'.$user->id.'/attributes', ['auth' => $auth]);
+                $body = json_decode($response->getBody());
+                $unsubscribed = $body->customer->unsubscribed;
+
+                // Update subscription status on user
+                if ($unsubscribed) {
+                    $user->email_frequency = 'none';
+                }
+                if (! $unsubscribed) {
```
DFurnes: These two `if` statements could be combined to a ternary for readability:

```php
$user->email_frequency = $unsubscribed ? 'none' : 'active';
```

app/Console/Commands/ImportSubStatusFromCio.php
```diff
@@ -0,0 +1,89 @@
...
+        // See if the --beforeDate option was used and convert it from a string to a date
+        $date = $this->option('beforeDate');
+        $date = new Carbon($date);
+
+        // Iterate through all the users and do the things!
+        $query->chunkById(200, function (Collection $users) use ($client, $auth, $date) {
+            $users->each(function (User $user) use ($client, $auth, $date) {
+                if ($date) {
+                    // Grab when the user's subscription status was last updated and convert from a string to a date
+                    $userUpdated = $user->audit['email_frequency']['updated_at']['date'];
```
DFurnes: This seems rough, since we'd still have to iterate over every single user to see if they were updated or not (and we don't have any indexes on the `audit` fields). We should be able to avoid this by extracting all this work into per-user queue jobs.

app/Console/Commands/ImportSubStatusFromCio.php
```diff
@@ -0,0 +1,89 @@
...
+                    $lastUpdated = new Carbon($userUpdated);
+
+                    // Do not update this user if they were last updated after the specified date
+                    if ($lastUpdated > $date) {
+                        return false;
+                    }
+                }
+
+                // Make request to c.io to get that user's subscription status
+                $response = $client->get('/v1/api/customers/'.$user->id.'/attributes', ['auth' => $auth]);
```
DFurnes: You can provide `['auth' => $auth]` as part of the options array in the `\GuzzleHttp\Client` constructor above so its used as the default for each request (and therefore you don't need to pass it down into this closure).

katiecrane: @DFurnes updated in new Job!

app/Console/Commands/ImportSubStatusFromCio.php
katiecrane: @DFurnes the Job has arrived!

app/Console/Commands/ImportSubStatusFromCio.php
katiecrane: @DFurnes updated in new Job!

app/Jobs/GetEmailSubStatusFromCustomerIo.php
```diff
@@ -0,0 +1,63 @@
...
+        $unsubscribed = $body->customer->unsubscribed;
+
+        // Update subscription status on user
+        $this->user->email_frequency = $unsubscribed ? 'none' : 'active';
+        $this->user->save();
+
+        info('User '.$this->user->id.' email subscription status grabbed from Customer.io');
+
+        // Make sure we don't go over 10 requests/second
+        throttle(600);
```
DFurnes: Ahh whoops, I tricked you with my earlier comment!

Since this is a queue job & could be processed by more than one concurrent worker (at least theoretically, we've always only had one on this app), we should handle throttling using Redis as [described in Laravel's queue docs](https://laravel.com/docs/5.5/queues#rate-limiting).

app/Jobs/GetEmailSubStatusFromCustomerIo.php
```diff
@@ -0,0 +1,63 @@
...
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        // Customer.io authentication
+        $auth = [config('services.customerio.username'), config('services.customerio.password')];
+
+        // Create a Guzzle Client to use with the Customer.io Beta API
+        $client = new \GuzzleHttp\Client([
```
DFurnes: Totally up to you, but it might be nice to extract this into a `CustomerIoBeta` API class using Gateway's [`AuthorizesWithBasicAuth`](https://github.com/DoSomething/gateway/blob/master/docs/traits/AuthorizesWithBasicAuth.md) trait. That'd also help by automatically JSON decoding the response down below.

app/Jobs/GetEmailSubStatusFromCustomerIo.php
```diff
@@ -0,0 +1,68 @@
...
+    /**
+     * Execute the job.
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        $key = 'GetEmailSubStatusFromCustomerIo-'.$this->user->id;
+
+        Redis::throttle($key)->allow(10)->every(1)->then(function () {
```
katiecrane: @DFurnes did I use `$key` correctly? The docs say:

> the key may be any string that uniquely identifies the type of job you would like to rate limit. For example, you may wish to construct the key based on the class name of the job and the IDs of the Eloquent models it operates on.

I wasn't sure if it meant the key should be unique per specific instance of one job or just per each different type of job.

app/Jobs/GetEmailSubStatusFromCustomerIo.php
katiecrane: @DFurnes going to leave it as it is for now since I don't think we have any other plans currently to use the Customer.io Beta API in the future

app/Jobs/GetEmailSubStatusFromCustomerIo.php
DFurnes: Oh yeah, those docs are totally confusing. The key is used to differentiate what "group" of things should be limited to that rate, so for our purposes we want it to be unique for this type of job.

Here's, for example, how we do it [in the `SendUserToCustomerIo` job](https://github.com/DoSomething/northstar/blob/b316c865283fd5e0ecdc6de9e78634c0e8a7e9a5/app/Jobs/SendUserToCustomerIo.php#L41-L43).

PR #823, opened 2019-01-04
Use standardized environment variables.

DFurnes:
#### What's this PR do?
This pull request updates Northstar to use standardized environment variables set by our S3, SQS, and IAM User Terraform modules (so, for example, instead of separate `S3_KEY` and `SQS_PUBLIC_KEY`, they both just use `AWS_ACCESS_KEY`).

I've also tidied up our `.env.example` to be a bit more approachable, and removed settings we shouldn't be using on local machines (for example, we don't need or want to connect to S3 buckets or Fastly's API from local machines).

Finally, I set the default `MAIL_DRIVER` to `log` so that we can see mail in the logs on local machines rather than having to hook it up to our real Mandrill account (yikes).

👀

#### Relevant Tickets
References DoSomething/infrastructure#99.

#### Checklist


.env.example
```diff
@@ -11,79 +14,43 @@ APP_AUTH_KEY=
...
 
-GLADIATOR_URL=http://gladiator.dev/api/
-GLADIATOR_API_KEY=GladiatorAPIKey
-
-MAIL_DRIVER=mandrill
-MANDRILL_SECRET=xxxxxxxxx
+# Mail:
 MAIL_ADDRESS=noreply@dosomething.org
 MAIL_NAME=DoSomething.org
+MANDRILL_SECRET=
```
katiecrane: Should this be removed since we have `MAIL_DRIVER=log`?

DFurnes: Oh whoops, it sure should!

DFurnes: Removed in 9dc0e68.

.env.example
```diff
@@ -1,6 +1,9 @@
 APP_URL=http://northstar.test
 APP_ENV=local
 APP_DEBUG=true
+
+# Generated by running `php artisan northstar:setup`
+# or `php artisan key:generate` scripts:
```
chloealee: great to have this context!

config/filesystems.php
```diff
@@ -50,22 +50,10 @@
 
         's3' => [
             'driver' => 's3',
-            'key' => env('S3_KEY'),
-            'secret' => env('S3_SECRET'),
+            'key' => env('AWS_ACCESS_KEY'),
+            'secret' => env('AWS_SECRET_KEY'),
```
chloealee: I feel like I am totally missing something but where do these come from? Not seeing them in the `.env.example` file!

config/queue.php
```diff
@@ -27,8 +27,8 @@
     |
     */
     'names' => [
-        'customerio' => env('SQS_DEFAULT_QUEUE'),
-        'backfill' => env('SQS_BACKFILL_QUEUE'),
+        'high' => env('SQS_DEFAULT_QUEUE'),
+        'low' => env('SQS_LOW_PRIORITY_QUEUE'),
```
chloealee: and same with these! Not seeing these in the `.env.example` file?

config/filesystems.php
DFurnes: They come from environment variables, but not ones we should be setting on our local machines. These ones come from [here](https://github.com/DoSomething/infrastructure/blob/80729502492f9175661056cdeb1c69b89cdb498d/shared/iam_app_user/main.tf#L25-L30).

config/queue.php
DFurnes: As above, these aren't in `.env.example` because they shouldn't be used on local machines. They're set [here](https://github.com/DoSomething/infrastructure/blob/80729502492f9175661056cdeb1c69b89cdb498d/shared/sqs_queue/main.tf#L33-L38) and [here](https://github.com/DoSomething/infrastructure/blob/80729502492f9175661056cdeb1c69b89cdb498d/applications/northstar/main.tf#L87-L89) for dev/QA/production apps.

chloealee: ooo interesting - thanks for explaining, @DFurnes !

PR #742, opened 2018-05-09
Rename email_status to email_frequency

katiecrane:
#### What's this PR do?
- Renames `email_status` to `email_frequency` in the database and updates all the references to that field. 
- In the migration, I re-used the [`renameField`](https://github.com/DoSomething/northstar/blob/master/database/migrations/2017_09_15_152015_RenameMobileStatusField.php#L27-L43) function from the `RenameMobileStatusField` migration.

#### How should this be reviewed?
Are all the old `email_status`s now `email_frequency`s? 

#### Relevant Tickets
[Card](https://www.pivotaltracker.com/story/show/157162083)

#### Checklist
- [ ] Documentation added for changed endpoints.
- [ ] Tests added for new features/bug fixes.
- [ ] Post a message in #api if this includes something that causes a rebuild!

PR #824, opened 2019-01-11
Rename `email_frequency` to `email_subscription_status`

katiecrane:
#### What's this PR do?
It's deja vu all over again: https://github.com/DoSomething/northstar/pull/742

✍️ Changes all references to `email_frequency` to `email_subscription_status` instead. We also now expect `email_subscription_status` to be a boolean so there is a validation update and a logic update to reflect that.

🐦 Includes a migration file to update the name of this attribute in the database which uses our existing `renameField` function.

Did I miss any references? Does this migration work as expected?

#### Relevant Tickets
Related to this work: https://www.pivotaltracker.com/story/show/161428034
[Slack discussion](https://dosomething.slack.com/archives/C02BBRN5A/p1547055755030600)

#### Checklist

### DIFF
--- a/app/Jobs/GetEmailSubStatusFromCustomerIo.php
+++ b/app/Jobs/GetEmailSubStatusFromCustomerIo.php
@@ -54,7 +54,7 @@ class GetEmailSubStatusFromCustomerIo implements ShouldQueue
             $unsubscribed = $body->customer->unsubscribed;
 
             // Update subscription status on user
-            $this->user->email_subscription_status = $unsubscribed ? 'none' : 'active';
+            $this->user->email_subscription_status = $unsubscribed ? false : true;
             $this->user->save();
 
             info('User '.$this->user->id.' email subscription status grabbed from Customer.io');
--- a/app/Jobs/GetEmailSubStatusFromCustomerIo.php
+++ b/app/Jobs/GetEmailSubStatusFromCustomerIo.php
@@ -54,7 +54,7 @@ class GetEmailSubStatusFromCustomerIo implements ShouldQueue
             $unsubscribed = $body->customer->unsubscribed;
 
             // Update subscription status on user
-            $this->user->email_frequency = $unsubscribed ? 'none' : 'active';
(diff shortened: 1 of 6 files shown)


app/Jobs/GetEmailSubStatusFromCustomerIo.php
```diff
@@ -54,7 +54,7 @@ public function handle()
             $unsubscribed = $body->customer->unsubscribed;
 
             // Update subscription status on user
-            $this->user->email_frequency = $unsubscribed ? 'none' : 'active';
+            $this->user->email_subscription_status = $unsubscribed ? 'none' : 'active';
             $this->user->save();
```
chloealee: should `none` be updated to `0` and `active` be updated to `1`?

app/Models/User.php
```diff
@@ -418,9 +418,9 @@ public function toCustomerIoPayload()
         ];
 
         // Only include email subscription status if we have that information
-        if ($this->email_frequency) {
-            $payload['email_frequency'] = $this->email_frequency;
-            $payload['unsubscribed'] = ($this->email_frequency === 'stop');
+        if ($this->email_subscription_status) {
+            $payload['email_subscription_status'] = $this->email_subscription_status;
+            $payload['unsubscribed'] = (! $this->email_subscription_status);
         }
```
chloealee: does this expect `$payload['unsubscribed'] to be either `true` or `false`?

app/Jobs/GetEmailSubStatusFromCustomerIo.php
katiecrane: @chloealee ah good catch, these should be updated to `true` and `false`!

app/Models/User.php
katiecrane: @chloealee yes! this should be `true` if the user IS unsubscribed

chloealee: @katiecrane ah got it thanks for explaining!

PR #836, opened 2019-02-12
Adds more logging to job that pulls email subscription status from Customer.io

katiecrane:
#### What's this PR do?

🍭 Adds a log of what we got when we pulled `unsubscribed` from c.io, except we log the opposite value so it is taken out of double negative land. Also adds what we set as `email_subscription_status`. These two values should be the same.

Should we log `unsubscribed` as it is and not flip it?

#### Relevant Tickets
[Card](https://www.pivotaltracker.com/story/show/163562051)

#### Checklist

### DIFF
--- a/app/Jobs/GetEmailSubStatusFromCustomerIo.php
+++ b/app/Jobs/GetEmailSubStatusFromCustomerIo.php
@@ -52,7 +52,7 @@ class GetEmailSubStatusFromCustomerIo implements ShouldQueue
             $response = $client->get('/v1/api/customers/'.$this->user->id.'/attributes');
             $body = json_decode($response->getBody());
             $unsubscribed = $body->customer->unsubscribed;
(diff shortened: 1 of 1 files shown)


app/Jobs/GetEmailSubStatusFromCustomerIo.php
```diff
@@ -52,12 +52,12 @@ public function handle()
             $response = $client->get('/v1/api/customers/'.$this->user->id.'/attributes');
             $body = json_decode($response->getBody());
             $unsubscribed = $body->customer->unsubscribed;
+            info('[GetEmailSubStatusFromCustomerIo] For user '.$this->user->id.' got subscribed='.! $unsubscribed);
```
DFurnes: It might be handier to actually print the raw `$unsubscribed` value here (just in case flipping that boolean does anything weird, like if C.io sent strings there or `null` or something).

katiecrane: @DFurnes yeah that feels safer! I went back and forth (as you can see in the commit history 😝)

PR #837, opened 2019-02-20
Add email subscription type attributes to users

katiecrane:
#### What's this PR do?

🔍 Added four new subscription type fields: 
- `news_email_subscription_status`
- `lifestyle_email_subscription_status`
- `action_email_subscription_status`
- `scholarship_email_subscription_status`

🍹 Validates that these fields are `boolean`.

📨 Included these fields in `v1` and `v2` user transformers, with the values cast as booleans.

🗄 Added these fields to the `fillable` array on users.

🎣 Cast these fields as `boolean`.

How are the names? Should I include any tests?

#### Relevant Tickets
[Card](https://www.pivotaltracker.com/story/show/163779556)

#### Checklist

### DIFF
--- a/app/Http/Transformers/Two/UserTransformer.php
+++ b/app/Http/Transformers/Two/UserTransformer.php
@@ -53,7 +53,7 @@ class UserTransformer extends TransformerAbstract
             $response['news_email_subscription_status'] = (bool) $user->news_email_subscription_status;
             $response['lifestyle_email_subscription_status'] = (bool) $user->lifestyle_email_subscription_status;
             $response['action_email_subscription_status'] = (bool) $user->action_email_subscription_status;
-            $response['scholarship_email_subscription_status']  = (bool) $user->scholarship_email_subscription_status;
+            $response['scholarship_email_subscription_status'] = (bool) $user->scholarship_email_subscription_status;
 
             // Voter registration status
             $response['voter_registration_status'] = $user->voter_registration_status;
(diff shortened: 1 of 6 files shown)


app/Http/Transformers/Two/UserTransformer.php
```diff
@@ -48,8 +48,12 @@ public function transform(User $user)
             // Internal & third-party service IDs:
             $response['slack_id'] = null;
 
-            // Email subscription status
+            // Email subscription statuses
             $response['email_subscription_status'] = (bool) $user->email_subscription_status;
+            $response['news_email_subscription_status'] = (bool) $user->news_email_subscription_status;
```
DFurnes: I wonder if this might make more sense as an array that we can then add/remove "topics" from over time?

```json
{
  "email_subscription_topics": ["news", "scholarship"]
}
```

katiecrane: @DFurnes I was thinking about this too, but I thought that might be harder to work with. I remember doing something like that in Longshot that was kind of a pain. Do you know if that'd be any easier or harder to use on the c.io side?

DFurnes: Ah, looks like Customer.io [only supports strings, booleans, and numbers](https://customer.io/docs/documentation/attributes.html#attribute-data-types). I'd still be tempted to say it's better to store this as one "logical" field in Northstar and expand it out to booleans in the C.io transformer though.

DFurnes: > I remember doing something like that in Longshot that was kind of a pain.

My guess is that was because we were dealing with MySQL there. It shouldn't be a problem to store this as an array in Mongo.

katiecrane: a-ha! I'll fiddle with it and learn just how easy it can be! thanks @DFurnes !

PR #841, opened 2019-02-27
Refactors ResetPassword Notification as Mail

aaronschachter:
#### What's this PR do?

This PR refactors Forgot Password emails as a `Mailable`, so we can eventually call [`render`](https://laravel.com/api/5.5/Illuminate/Mail/Mailable.html#method_render) to send rendered HTML to Blink's new `/events/user-password-reset` endpoint (https://github.com/DoSomething/blink/pull/234).

* Following this [tutorial on converting a MailMessage to Mailable](https://medium.com/@woganmay/convert-mailmessage-to-mailable-in-laravel-5-3-d26e74b55c51), I ran `php artisan vendor:publish --tag=laravel-notifications`, which added `resources/views/vendor/notifications/email.blade.php`.

* I then ran `php artisan make:mail ResetPassword` to add `app/Mail/ResetPassword.php`, and copied content from `app/Auth/Notifications/ResetPassword.php`.

Next up: Add Gateway support to post to the password-reset to Blink, and refactor the User `sendPasswordResetNotification` function to send the email via Gateway instead of Mandrill mail.

#### How should this be reviewed?

Verify that the Forgot Password form works as expected (email is sent from Mandrill)

#### Relevant Tickets
Begins work on https://www.pivotaltracker.com/story/show/163779567. 

#### Checklist
- [x] Tests added for new features/bug fixes.

PR #842, opened 2019-03-01
Adds PasswordResetType

aaronschachter:
#### What's this PR do?

This PR adds a `PasswordResetType` class to begin defining various types of password reset emails we'll send, such as the existing Forgot Password type and the upcoming Rock The Vote Activate Account type.

* It also renames the `ResetPassword` mail from #841 as `PasswordReset`, because  `PasswordResetType` felt more readable than `ResetPasswordType` ... and also because the end goal is to post to the `events/user-password-reset` Blink endpoint. I kept going back and forth on this, as Laravel uses both the `PasswordReset` and `ResetPassword` naming within existing classes/functions.

* Refactors the currently deprecated `POST /resets` endpoint:
    * Requires a `type` parameter, expecting a  valid `PasswordResetType` 
    * Send a PasswordReset email. This endpoint will be used by Chompy imports to send users "Activate Account" types of resets per [spec](https://docs.google.com/document/d/1RbWAAJA-zTfQYv6dBVIcpJqegKJUM0dH4DkVpikRpE0/edit?ts=5c6f0bd1#heading=h.f68k8ccmho5)

* Verify the Forgot Password webform (e.g. `/password/reset`) sends a Forgot Password email upon submitting a valid email

* Verify executing a `POST /resets` request with type `rock-the-vote-activate-account` sends an Activate Account email, prompting user to "Set Password"

#### Relevant Tickets

https://www.pivotaltracker.com/story/show/163779579

#### Checklist


aaronschachter:
> I kept going back and forth on this, as Laravel uses both the `PasswordReset` and `ResetPassword` naming within existing classes/functions.

I also kept going back and forth on whether to add said `PasswordReset`as a `Notification`, as it was before I refactored it as a `Mailable` in #841 (as Notifications can return a Mailable in the [`toMail`](url)) -- and we could potentially create a [custom notification channel](https://laravel.com/docs/5.7/notifications#custom-channels) to send the notification from Blink... but this seemed potentially like overcomplicating.

app/Http/Controllers/ResetController.php
```diff
@@ -29,18 +29,18 @@ public function store(Request $request)
...
+            'type' => 'required',
         ]);
 
         /** @var \Northstar\Models\User $user */
         $user = User::findOrFail($request['id']);
 
         $tokenRepository = $this->createTokenRepository();
         $token = $tokenRepository->create($user);
-        $email = $user->getEmailForPasswordReset();
+        // TODO: Throw error if invalid type is passed.
```
DFurnes: This can be handled in the validation rules above, rather than explicitly throwing an error:

```php
use Illuminate\Validation\Rule;

$this->validate($request, [
    'id' => 'required',
    'type' => ['required', Rule::in(PasswordResetType::all())]
```

app/Http/Controllers/ResetController.php
```diff
@@ -29,18 +29,18 @@ public function store(Request $request)
...
+        $message = $user->sendPasswordReset($token, $request['type']);
 
-        return [
-            'url' => config('app.url').'/password/reset/'.$token.'?email='.urlencode($email),
-        ];
+        return ['url' => $message->url];
```
DFurnes: Since this endpoint is now actually sending the email, I wonder if we even need to return the `url`? Would it make more sense to just return a simple `success` response?

tests/Http/ResetTest.php
```diff
@@ -33,7 +34,10 @@ public function testCreatePasswordResetLink()
     {
         $user = factory(User::class)->create();
 
-        $this->asAdminUser()->post('v2/resets', ['id' => $user->id]);
+        $this->asAdminUser()->post('v2/resets', [
+            'id' => $user->id,
+            'type' => PasswordResetType::$forgotPassword,
```
DFurnes: My gut says it's better to use the actual string here, so that we would be notified that we'd broken our "contract" with a failing test if we changed the value for `PasswordResetType::$forgotPassword` (or worse, if an unrelated change caused this property to start returning a different value).

I definitely tend to favor "wetter" (haha, less dry?) tests than most folks, though, so totally up for debate!

DFurnes:
> It also renames the `ResetPassword` mail from #841 as `PasswordReset`, because  `PasswordResetType` felt more readable than `ResetPasswordType` ... and also because the end goal is to post to the `events/user-password-reset` Blink endpoint. I kept going back and forth on this, as Laravel uses both the `PasswordReset` and `ResetPassword` naming within existing classes/functions.

I like this change. 👍 

> I also kept going back and forth on whether to add said `PasswordReset`as a `Notification`, as it was before I refactored it as a `Mailable` in #841 (as Notifications can return a Mailable in the [`toMail`](url)) -- and we could potentially create a [custom notification channel](https://laravel.com/docs/5.7/notifications#custom-channels) to send the notification from Blink... but this seemed potentially like overcomplicating. 

Yeah, hmm! It does seem like a custom notification channel might make more sense since we're effectively just "publishing" an event for the user, and then Customer.io is sending the email. Let's see how we feel when it comes time to make that change?

app/Http/Controllers/ResetController.php
aaronschachter: I've added this -- but a 422 doesn't get thrown as expected, a 200 is returned with the HTML of the Login screen if I try passing an invalid `type` parameter.

app/Http/Controllers/ResetController.php
aaronschachter: Do you mean returning a string `return 'Message sent'` ?  Eventually this endpoint creates a `password_reset` event in Customer.io -- I figured it could make sense to return the payload we post to Blink (but potentially excluding the `body` parameter with rendered HTML)

DFurnes: For example, we return [a message from `DELETE v2/users`](https://github.com/DoSomething/northstar/blob/e1b1e0ceaf8a39dfffb11620f4176572ab41e19b/app/Http/Controllers/Two/UserController.php#L190), that returns this format:

```json
{
   "success": {
        "code": 200,
        "message": "Message sent."
    }
}
```

I'm also good with returning some information for context, but would be worried that returning a link suggests that the application is then responsible for providing it to the user itself.

PR #843, opened 2019-03-05
Forwarding password reset events to Blink

aaronschachter:
#### What's this PR do?

This PR modifies the User `sendPasswordReset` function to no longer send an email via Mandrill, but instead create  a `call_to_action_email` event in Customer.io, via the new `/app/Jobs/SendPasswordResetToCustomerIo`. I've added event triggered campaigns in both our Staging and Production Customer.io instances to send an email with the `call_to_action_email` content.

#### Relevant Tickets
* Fixes https://www.pivotaltracker.com/story/show/163779567
* Refs https://github.com/DoSomething/blink/pull/235

#### Checklist


app/Mail/PasswordReset.php
```diff
@@ -40,4 +42,20 @@ public function build()
...
+
+    /**
+     * Transform the message for Blink.
+     *
+     * @return $array
+     */
+    public function toCustomerIoPayload()
+    {
+        return [
+            'body' => json_encode($this->render()),
```
aaronschachter: If I substitute sending the `body` parameter as `'test'`, the message is sent.

tests/Http/Web/PasswordResetTest.php
```diff
@@ -34,10 +34,11 @@ public function testPasswordResetFlow()
         ]);
 
         // We'll assert that the email was sent & take note of the token for the next step.
-        Mail::assertSent(PasswordReset::class, function ($mail) use ($user, &$token) {
-            $token = $mail->token;
+        Bus::assertNotDispatched(SendPasswordResetToCustomerIo::class, function ($job) use ($user, &$token) {
+            $message = $job->passwordReset;
+            $token = $message->token;
```
aaronschachter: I was assuming that `assertDispatched` would be true, but it's not the case. I also needed to make `passwordReset` public in order to reference these variables.

DFurnes: Huh. Yeah, something definitely seems off here. Does this callback run?

aaronschachter: I'd think the callback must be running, as the rest of the tests for resetting the user password, which would require a valid token? I've been testing this branch by setting the Blink config to true, and verifying messages are sent via Customer.io...

aaronschachter: Confirmed this callback does run by adding an `info` to log that `$token` is updated with a token value.

app/Models/User.php
```diff
@@ -484,19 +484,15 @@ public function sendPasswordResetNotification($token)
...
+     * @return void
      */
     public function sendPasswordReset($token, $type)
     {
-        $email = $this->getEmailForPasswordReset();
-        $message = new PasswordResetMail($email, $token, $type);
-        Mail::to($email)->send($message);
-        // TODO: Send json_encode($message->render()) as body to Blink /events/user-password-reset
-        return $message;
+        SendPasswordResetToCustomerIo::dispatch(new PasswordResetMail($this, $token, $type));
```
aaronschachter: I've been mulling over just sending through a `$user`, `$token`, and `$type` as parameters to the  `SendPasswordResetToCustomerIo` job, skipping creating a `Mailable` altogether. It was used as a way to gradually roll out sending types of Password Reset emails... but mainly handy now for rendering the HTML of the email. What's tricky there is that it's working well in our staging Customer.io instance, but our production layout seems to collide with the styles we send to Blink. [Screenshots](https://gist.github.com/aaronschachter/05b0c13e26cccacfeb90c03d5645091e#gistcomment-2851432)

aaronschachter: Definitely leaning towards this approach - I looked at some of the layouts in Customer.io production and saw how much inline CSS there is . Seems like extra overhead to have to mess with any of this in Northstar code. I'm thinking that we'd be better off saving individual variables like `intro`, `actionUrl`, `actionText` in our Customer.io `password_reset` event instead of `body` (I started this approach in https://github.com/DoSomething/blink/pull/234/commits/ae69a2dc32d274174669d79a4cd37f4714c94336 but then switched over to sending `body` after figuring out how to render the Forgot Password email HTML)

DFurnes: > Definitely leaning towards this approach - I looked at some of the layouts in Customer.io production and saw how much inline CSS there is . Seems like extra overhead to have to mess with any of this in Northstar code. I'm thinking that we'd be better off saving individual variables like `intro`, `actionUrl`, `actionText` in our Customer.io `password_reset` event instead of `body`

I think that's a good call @aaronschachter! 👍 👍

tests/Http/Web/PasswordResetTest.php
aaronschachter: I've tested whether the Queue fake behaves differently, but it's the same result -- this test only passes if I call `Queue::assertNotPushed`, not `assertPushed`.

One shot in the dark as to why the opposite test is passing is because maybe Laravel doesn't queue if its being dispatched from the TestCase [submitForm](https://laravel.com/api/5.2/Illuminate/Foundation/Testing/TestCase.html#method_submitForm) -- but that doesn't help clarify why our ` Mail::assertSent` test was passing as we'd expect.

tests/Http/ResetTest.php
```diff
@@ -41,8 +44,12 @@ public function testCreatePasswordResetLink()
         ]);
         $this->assertResponseStatus(200);
         $this->seeJsonStructure(['success']);
-        $this->blinkMock->shouldHaveReceived('userCallToActionEmail')->once();
-
+        $this->blinkMock->shouldNotHaveReceived('userCallToActionEmail');
```
aaronschachter: Looking into https://github.com/DoSomething/northstar/pull/843#discussion_r263563575, and it makes sense why Blink isn't getting called anymore, it's because I've added a fake queue.

In our test `testPasswordResetFlow` -- we want to mock Queues/Bus so we get access to the Password Reset URL for a created job which we can then use to test resetting the user's password. Since we're already using a Blink mock, it seems we'd be better off not faking the Queue over in `testPasswordResetFlow` - perhaps just by inspecting the real queue instead of using a fake queue? 
 I was searching through the Github repo for  `assertDispatched` and found this thread which seems slightly related: https://github.com/DoSomething/northstar/pull/730#discussion_r182464680

tests/Http/ResetTest.php
```diff
@@ -40,7 +44,12 @@ public function testCreatePasswordResetLink()
         ]);
         $this->assertResponseStatus(200);
         $this->seeJsonStructure(['success']);
-
+        $this->blinkMock->shouldNotHaveReceived('userCallToActionEmail');
+        $resetPasswordUrl = null;
+        Queue::assertNotDispatched(SendCallToActionEmailToCustomerIo::class, function ($job) use (&$resetPasswordUrl) {
```
aaronschachter: This is where it's opposite day - we'd expect our fake queue would have been dispatched here.

aaronschachter:
This is ready for review!

tests/Http/Web/PasswordResetTest.php
aaronschachter: Needed to return a truthy parameter within the callback to get the `assertDispatched` test to pass.

app/Models/User.php
```diff
@@ -484,19 +483,16 @@ public function sendPasswordResetNotification($token)
...
      */
     public function sendPasswordReset($token, $type)
     {
-        $email = $this->getEmailForPasswordReset();
-        $message = new PasswordResetMail($email, $token, $type);
-        Mail::to($email)->send($message);
-        // TODO: Send json_encode($message->render()) as body to Blink /events/user-password-reset
-        return $message;
+        $payload = PasswordResetType::getCallToActionEmailParams($this, $token, $type);
+        SendCallToActionEmailToCustomerIo::dispatch($payload);
```
DFurnes: I would recommend sending at the real `$this, $token, and $type` parameters to the queue job and formatting for Customer.io afterwards – this way we know everything we need to know to retry this job at any point.

For example, we've had scenarios where Customer.io has refused requests due to validation errors. Because we use [`SerializesModels`](https://laravel.com/docs/5.5/queues#class-structure) on this job, if we pass an Eloquent model for the user, it will load the record from the database with any fixes if we retry the same job later. Similarly, we could then change the email copy & any already queued jobs would get the new copy as soon as we deployed.

PR #846, opened 2019-03-11
Fixes email subscription status Customer.io sync

aaronschachter:
#### What's this PR do?

Modifies `Model.setAttribute` to avoid overriding `false` boolean values with `null`, which unblocks fixing updating `email_subscription_status` in Customer.io by checking whether a `email_subscription_status` value is set (vs whether it's truthy)

I've been testing by locally updating my user via API, but also within Aurora to verify all data looks as expected.

#### Relevant Tickets
* Fixes https://www.pivotaltracker.com/story/show/164346693, https://www.pivotaltracker.com/story/show/164346648

#### Checklist

### DIFF
--- a/app/Models/Model.php
+++ b/app/Models/Model.php
@@ -24,7 +24,7 @@ class Model extends BaseModel
     {
         parent::setAttribute($key, $value);
 
-        // Empty strings should be saved as `null` when our attribute is a string.
+        // Empty string values should be saved as `null` if attribute is not boolean.
         $booleanFields = ['email_subscription_status', 'sms_paused'];
         if (empty($this->attributes[$key]) && ! in_array($key, $booleanFields)) {
             $this->attributes[$key] = null;
--- a/app/Models/Model.php
+++ b/app/Models/Model.php
@@ -25,7 +25,7 @@ class Model extends BaseModel
         parent::setAttribute($key, $value);
 
         // Empty string values should be saved as `null` if attribute is not boolean.
-        $booleanFields = ['email_subscription_status', 'sms_paused'];
+        $booleanFields = array_keys($this->casts, 'boolean');
         if (empty($this->attributes[$key]) && ! in_array($key, $booleanFields)) {
             $this->attributes[$key] = null;
         }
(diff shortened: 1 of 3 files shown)


tests/Models/UserTest.php
```diff
@@ -101,4 +101,42 @@ public function it_should_return_password_reset_url_with_token()
 
         $this->assertEquals($result, route('password.reset', [$token, 'email' => $email, 'type' => $type]));
     }
+
+    /** @test */
+    public function it_should_include_unsubscribed_false_in_customerio_payload_if_email_subscription_status_true()
+    {
+        $subscribedStatusUser = factory(User::class)->create([
+            'email' => 'subscribed@example.com',
```
katiecrane: I don't think you should need to include an `email` here, [the user factory should take care of that!](https://github.com/DoSomething/northstar/blob/master/database/factories/ModelFactory.php#L16)

app/Models/Model.php
```diff
@@ -24,8 +24,9 @@ public function setAttribute($key, $value)
     {
         parent::setAttribute($key, $value);
 
-        // Empty strings should be saved as `null`.
-        if (empty($this->attributes[$key])) {
+        // Empty string values should be saved as `null` if attribute is not boolean.
+        $booleanFields = ['email_subscription_status', 'sms_paused'];
```
DFurnes: Rather than hard-coding these field names in the generic Model, let's decide this by [the specified `$casts`](https://github.com/DoSomething/northstar/blob/4f7beffa7a78e31a5915877c2ed0b28b3d774b5c/app/Models/User.php#L164-L173):

```php
$booleanFields = array_keys(self::$casts, 'boolean');
```

aaronschachter: This works if I call `$this->casts`... calling as `self::$casts` throws `Error: Access to undeclared static property: Northstar\Models\Model::$casts`. Refs https://stackoverflow.com/a/27670398/1470725

DFurnes: Oh yup, I totally messed that up!

PR #848, opened 2019-03-15
Update user update command to work with `email_subscription_topics`

katiecrane:
#### What's this PR do?

🎩 Updates `northstar:update` to work with `email_subscription_topics`, which is an array value. We want to make sure that any topics added in through the csv are _added_ to the array and do not overwrite it! We also want to make sure that we don't add duplicate values to the array.

👩‍🎓 Adds a new test for this command for this work. Makes sure we can add `email_subscription_topics` to users who have no topics, have some existing topics, and makes sure we don't add duplicate topics.

Will this allow us to import the [csvs on this card](https://www.pivotaltracker.com/n/projects/2019429/stories/164216160)?

#### Relevant Tickets
[Card](https://www.pivotaltracker.com/story/show/164216160)

#### Checklist

### DIFF
--- a/app/Console/Commands/UpdateUserFieldsCommand.php
+++ b/app/Console/Commands/UpdateUserFieldsCommand.php
@@ -68,6 +68,7 @@ class UpdateUserFieldsCommand extends Command
         $usersToUpdate = $usersCsv->getRecords();
 
         $this->line('northstar:update: Updating '.count($usersCsv).' users...');
+
         $fieldsToUpdate = $this->argument('fields');
 
         $this->totalCount = count($usersCsv);
@@ -95,7 +96,7 @@ class UpdateUserFieldsCommand extends Command
 
                         // Don't add topic if it is already there
                         if (in_array($userToUpdate[$field], $topics)) {
-                            return;
+                            continue;
                         }
 
                         // Add the new topic to our array
--- a/app/Console/Commands/UpdateUserFieldsCommand.php
+++ b/app/Console/Commands/UpdateUserFieldsCommand.php
@@ -88,7 +88,24 @@ class UpdateUserFieldsCommand extends Command
 
             foreach ($fieldsToUpdate as $field) {
                 if (! empty($userToUpdate[$field])) {
-                    $user->{$field} = $userToUpdate[$field];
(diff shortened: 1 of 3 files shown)


app/Console/Commands/UpdateUserFieldsCommand.php
```diff
@@ -88,7 +89,24 @@ public function handle()
 
             foreach ($fieldsToUpdate as $field) {
                 if (! empty($userToUpdate[$field])) {
-                    $user->{$field} = $userToUpdate[$field];
+                    // Special instructions when working with array field
+                    if ($field === 'email_subscription_topics') {
+                        // Get current email topics
+                        $topics = $user->email_subscription_topics ? $user->email_subscription_topics : [];
```
DFurnes: I think we could simplify this a little bit using `array_merge`:

```php
if ($field === 'email_subscription_topics') {
    $existingItems = $user->{$field} ?: [];
    $user->{$field} = array_merge($existingItems, $userToUpdate[$field]);
}
```

tests/Console/UpdateUserFieldsCommandTest.php
```diff
@@ -42,4 +42,41 @@ public function it_should_update_given_fields()
             'created_at' => Carbon::parse('2018-01-01 10:00:00'),
         ]);
     }
+
+    /** @test */
+    public function it_should_update_email_topics()
```
DFurnes: Love, love, love that you added this test! 🙌

app/Console/Commands/UpdateUserFieldsCommand.php
katiecrane: @DFurnes fiddled with this a bit but using `array_merge` is still cause duplicates because [the arrays have numeric keys.](http://php.net/manual/en/function.array-merge.php)
```
If the input arrays have the same string keys, then the later value for that key will overwrite the previous one. If, however, the arrays contain numeric keys, the later value will not overwrite the original value, but will be appended.
```

DFurnes: Ohhh, you're totally right. Gah! 😩

PR #852, opened 2019-03-21
Adds support for Activate Account password resets

aaronschachter:
#### What's this PR do?

Adds [`showResetForm`](https://github.com/laravel/framework/blob/5.5/src/Illuminate/Foundation/Auth/ResetsPasswords.php#L25) to the `ResetPasswordController` to customize the Reset Password form when `type` query parameter includes `activate-account`.

Verify expected content for Forgot Password forms at `/password/reset`, as well as overrides for when the `type` query parameter includes `activate-account`.

#### Relevant Tickets
Fixes https://www.pivotaltracker.com/n/projects/2019429/stories/164780588

#### Checklist

### DIFF
--- a/tests/Http/Web/PasswordResetTest.php
+++ b/tests/Http/Web/PasswordResetTest.php
@@ -19,7 +19,7 @@ class PasswordResetTest extends BrowserKitTestCase
     /**
      * Test that the homepage redirects to login page.
      */
-    public function testPasswordResetFlow()
+    public function testForgotPasswordResetFlow()
     {
         Bus::fake();
 
@@ -29,6 +29,7 @@ class PasswordResetTest extends BrowserKitTestCase
         // The user should be able to request a new password by entering their email.
         $this->visit('/password/reset');
         $this->see('Forgot your password?');
+        $this->see('We\'ve all been there. Reset by entering your email.');
         $this->submitForm('Request New Password', [
             'email' => 'forgetful@example.com',
         ]);
@@ -39,6 +40,7 @@ class PasswordResetTest extends BrowserKitTestCase
 
             return true;
         });
+        info('testForgotPasswordResetFlow '.$resetPasswordUrl);
 
         // The user should visit the link that was sent via email & set a new password.
         $this->visit($resetPasswordUrl);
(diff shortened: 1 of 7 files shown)


DFurnes:
If you have time, it might be nice to add a simple `testPasswordActivationFlow` test to [`PasswordResetTest`](https://github.com/DoSomething/northstar/blob/master/tests/Http/Web/PasswordResetTest.php) that checks that the correct copy is used for that flow.

tests/Http/Web/PasswordResetTest.php
```diff
@@ -79,4 +81,49 @@ public function testPasswordResetRateLimited()
...
+        ]);
+        $this->assertResponseStatus(200);
+        $this->seeJsonStructure(['success']);
+        // We'll assert that the event was created & take note of reset URL for the next step.
+        Bus::assertDispatched(SendPasswordResetToCustomerIo::class, function ($job) use (&$resetPasswordUrl) {
+            $passwordResetUrl = $job->getUrl();
+
+            return true;
+        });
+        */
```
aaronschachter: Having a tough time getting this to work. This commented code runs, and I've debugged to see that `$passwordResetUrl` is populated with a real reset URL (copy/pasted one below). Yet when the `visit` test runs in line 112, it's just finding the regular Northstar homepage for user registration.

DFurnes: Not sure if this was just changed while debugging, but is it [`$passwordResetUrl`](https://github.com/DoSomething/northstar/pull/852/files#diff-27ca59bee4ffdeef3e65ebb7ce276f7eR93) vs. [`$resetPasswordUrl`](https://github.com/DoSomething/northstar/pull/852/files#diff-27ca59bee4ffdeef3e65ebb7ce276f7eR104)?

aaronschachter: Yeah, I had just changed this while debugging... fixing the variable names doesn't do the trick. Line 113 fails once I uncomment out the call to ` $this->asAdminUser()->post` in line 95. I couldn't find anything like a `$this->asAnonUser` to visit the reset URL, wondering if that could have something to do with it... 🤕

tests/Http/Web/PasswordResetTest.php
```diff
@@ -79,4 +81,49 @@ public function testPasswordResetRateLimited()
...
+        // We'll assert that the event was created & take note of reset URL for the next step.
+        Bus::assertDispatched(SendPasswordResetToCustomerIo::class, function ($job) use (&$resetPasswordUrl) {
+            $resetPasswordUrl = $job->getUrl();
+
+            return true;
+        });
+        info('test 2 '.$resetPasswordUrl);
+        */
+
+        $resetPasswordUrl = 'http://localhost/password/reset/15f5b5dbead3e0455e10955febd0458948d088f153016c51f166740bb2d37bfd?email=davon11%40example.org&type=rock-the-vote-activate-account';
```
aaronschachter: This is from the debugging above, definitely setting `$resetPasswordUrl` to a valid password reset URL within our `Bus::assertDispatched` callback

tests/Http/Web/PasswordResetTest.php
```diff
@@ -79,4 +81,49 @@ public function testPasswordResetRateLimited()
...
+
+        $this->visit($resetPasswordUrl);
```
aaronschachter: The thought in https://github.com/DoSomething/northstar/pull/852/files#r268376990 doesn't fix -- tried explicitly calling `$this->actingAs($user)->visit($resetPasswordUrl);` here but the next test fails -- the HTML returned matches the homepage/registration -- not the forgot password HTML.

aaronschachter: Tried adding`$this->get('/logout');` before this line, thinking maybe a permission issue could be causing visiting a valid password reset link to return the registration form... but that's not it either

DFurnes: This was tricky, but you were totally on the right track! It's pretty wild that Laravel doesn't have a simple `$this->logout()` complement to `$this->be(…)`! 😖

Ended up using `$this->refreshApplication()` like [we do in this other test](https://github.com/DoSomething/northstar/blob/ddf98cc07d77bcea0ee572ef6e83c7610e0010be/tests/Http/OAuthTest.php#L500).

aaronschachter: Tricky indeed -- thank you so much for the help! *screams into pillow*

PR #854, opened 2019-03-28
Track password_updated event in Customer.io [pr]

aaronschachter:
#### What's this PR do?

Tracks a `password_updated` event in Customer.io when user updates their password, saving a single `source` attribute -- which there can be 3 values for currently per this branch:

* `forgot-password` - When user clicks link sent from a Forgot Password email
* `rock-the-vote-activate-account` - When an imported Rock The Vote user clicks link sent from an Activate Account email
* `profile` - When user updates their password via profile

This will allow us to configure [Customer.io event triggered campaigns](http://docs.dosomething.org/customer-io#events) to send a confirmation email if an activated user changes their password, or welcome emails to imported users who have activated their account via `/password/reset/rock-the-vote-activate-account/{token}` links (sent from the RTV Activate Account email)

The bad news: This PR breaks existing reset URL's for emails that have been sent, as we're now using `/password/reset/{type}/{token}` instead of `/password/reset/{token}`.  Because the links expire after 24 hours, I don't think this should be a blocker.

Test locally by setting your Customer.io env variables, submit the Forgot Password form, and verify a `password_updated` event is created in Customer.io upon setting a new password.

#### Relevant Tickets
This helps unblock a wishlist item for Rock The Vote Activate Account emails: when a user clicks through an Activate Account email and sets their password, send a follow-up email confirming account creation and welcoming them as a DS member. We'll be able to do this by creating two separate event triggered campaigns:

* 1 - Filter where `source` contains `activate-account`, where we'll welcome users who have set a password (they previously did not have one set, because they were imported)
* 2 - Filter where source does not contain `activate-account`, this will send a standard password change confirmation email - e.g: _You are receiving this because you (or someone else) has changed the password of your DoSomething.org account. If this was you, you can safely ignore this email. If not, please email help@dosomething.org._

#### Checklist

### DIFF
--- a/app/Events/PasswordUpdated.php
+++ b/app/Events/PasswordUpdated.php
@@ -15,22 +15,22 @@ class PasswordUpdated
     public $user;
 
     /**
-     * Source of password update (either profile or type of reset password form).
+     * Where user updated their password from (e.g type of reset password form, profile).
      *
      * @var string
      */
-    public $source;
+    public $updatedVia;
 
     /**
      * Create a new event instance.
      *
      * @param  \Illuminate\Contracts\Auth\Authenticatable  $user
-     * @param  string  $source
+     * @param  string  $updatedVia
      * @return void
      */
-    public function __construct($user, $source)
+    public function __construct($user, $updatedVia)
     {
         $this->user = $user;
-        $this->source = $source;
+        $this->updatedVia = $updatedVia;
     }
 }
--- a/app/Events/PasswordUpdated.php
+++ b/app/Events/PasswordUpdated.php
@@ -25,7 +25,7 @@ class PasswordUpdated
      * Create a new event instance.
      *
      * @param  \Illuminate\Contracts\Auth\Authenticatable  $user
-     * @param  string
+     * @param  string  $source
      * @return void
      */
     public function __construct($user, $source)
--- a/app/Events/PasswordUpdated.php
+++ b/app/Events/PasswordUpdated.php
@@ -1,0 +1,36 @@
+<?php
+
+namespace Northstar\Events;
+
+use Illuminate\Queue\SerializesModels;
+
+class PasswordUpdated
+{
+    use SerializesModels;
+    /**
+     * The user.
+     *
+     * @var \Illuminate\Contracts\Auth\Authenticatable
+     */
+    public $user;
+
+    /**
+     * Source of password update (either profile or type of reset password form).
+     *
+     * @var string
+     */
+    public $source;
+
+    /**
+     * Create a new event instance.
+     *
+     * @param  \Illuminate\Contracts\Auth\Authenticatable  $user
+     * @param  string
+     * @return void
+     */
+    public function __construct($user, $source)
+    {
+        $this->user = $user;
+        $this->source = $source;
+    }
+}
--- a/app/Http/Controllers/Web/ResetPasswordController.php
+++ b/app/Http/Controllers/Web/ResetPasswordController.php
@@ -5,6 +5,7 @@ namespace Northstar\Http\Controllers\Web;
 use Illuminate\Http\Request;
 use Illuminate\Foundation\Auth\ResetsPasswords;
 use Northstar\Events\PasswordUpdated;
+use Northstar\PasswordResetType;
 use Illuminate\Support\Facades\Auth;
 
 class ResetPasswordController extends Controller
@@ -35,7 +36,6 @@ class ResetPasswordController extends Controller
      * Reset the given user's password.
      *
      * @param  \Illuminate\Contracts\Auth\CanResetPassword  $user
-     * @param  string  $password
      * @return void
      */
     protected function resetPassword($user, $password)
@@ -45,23 +45,27 @@ class ResetPasswordController extends Controller
             'remember_token' => str_random(60),
         ])->save();
 
-        // TODO: Pass reset type argument and send through as the PasswordUpdated source.
-        event(new PasswordUpdated($user, 'forgot-password'));
+        event(new PasswordUpdated($user, request()->type));
 
         $this->guard()->login($user);
     }
 
     /**
-     * Display the password reset view for the given token.
+     * Display the password reset view for the given password reset type and token.
      *
      * If no token is present, display the link request form.
      *
      * @param  \Illuminate\Http\Request  $request
+     * @param  string  $type
      * @param  string|null  $token
      * @return \Illuminate\Contracts\View\Factory|\Illuminate\View\View
      */
-    public function showResetForm(Request $request, $token = null)
+    public function showResetForm(Request $request, $type, $token = null)
     {
+        if (! isset($type)) {
+            $type = PasswordResetType::$forgotPassword;
+        }
+
         $data = [
             'title' => 'Forgot Password',
             'header' => trans('auth.forgot_password.header'),
@@ -72,7 +76,7 @@ class ResetPasswordController extends Controller
             'display_footer' => true,
         ];
 
-        if (str_contains(request()->input('type'), 'activate-account')) {
+        if (str_contains($type, 'activate-account')) {
             $data = [
                 'title' => 'Activate Account',
                 'header' => trans('auth.activate_account.header'),
@@ -85,6 +89,7 @@ class ResetPasswordController extends Controller
         }
 
         $data['token'] = $token;
+        $data['type'] = $type;
         $data['email'] = $request->email;
 
         return view('auth.passwords.reset')->with($data);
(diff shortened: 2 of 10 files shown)


aaronschachter:
Added our Customer.io QA environment variables into Wercker as protected variables to get tests to pass.

app/Listeners/ReportPasswordUpdated.php
```diff
@@ -0,0 +1,43 @@
...
+    /**
+     * Handle the event.
+     *
+     * @param PasswordReset $event
+     * @return void
+     */
+    public function handle(PasswordReset $event)
+    {
+        $response = $this->client->post('customers/'.$event->user->id.'/events', [
+            'form_params' => ['name' => 'password_updated'],
```
aaronschachter: I think ideally we need to pass an attribute to specify whether this is a `password_updated` event for an imported user creating a password for the first time (user previously had no password) --where we'd want to send a "You're official" welcome type of email --  or a user who had an existing password, where we'd want to send a confirmation "Your password changed -- if you did this, great, but if not hit us up"

This relates to a feature request @ngjo brought up recently: when an imported user doesn't have a password set and tries to register -- they get a message informing them that their email already has an account. Ideally on the user registration form, if someone tries to create an account for an email that exists -- we'd want to prompt them to create a password if it happens to be an account that was imported, and doesn't have a password set. (This also makes me wonder about potential security concerns though)

app/Providers/EventServiceProvider.php
```diff
@@ -23,6 +25,7 @@ class EventServiceProvider extends ServiceProvider
         Login::class => [ReportSuccessfulAuthentication::class],
         Failed::class => [ReportFailedAuthenticationAttempt::class],
         Throttled::class => [ReportThrottledRequest::class],
+        PasswordReset::class => [ReportPasswordUpdated::class],
```
aaronschachter: This `PasswordReset` Laravel event doesn't get created if a user updates their password from their profile. It seems like it should? I'd imagine we'd want to send the notification to a user when updating via profile (especially since auth sessions don't expire)

aaronschachter: Added in https://github.com/DoSomething/northstar/pull/854/commits/7d4f1b4d732b631062b80b357265f441620a9d02

app/Http/Controllers/Web/ResetPasswordController.php
```diff
@@ -45,7 +45,8 @@ protected function resetPassword($user, $password)
             'remember_token' => str_random(60),
         ])->save();
 
-        event(new PasswordReset($user));
+        // TODO: Pass reset type argument and send through as the PasswordUpdated source.
```
aaronschachter: From what I can tell, to send a password reset's `type` query parameter as a third argument to this `resetPassword` function, we'd need to override the `reset` function (which we're calling in when [posting to `/password/reset`](https://github.com/DoSomething/northstar/blob/84d387a9d83b5ed6e7bcf05899bd05d687b6808a/routes/web.php#L43)).
I think both `GET /password/reset` and `POST /password/reset` routes should require a `type` route parameter that can be used to render the Password Reset form per the Password Reset type... but also send along the type as the `source` of the `password_updated` event in Customer per this TODO

app/Listeners/ReportPasswordUpdated.php
aaronschachter: I've added this in a later commit, passing a `source` attribute to a `password_updated` event. 

Per https://github.com/DoSomething/northstar/pull/482/files#r87455095 - @DFurnes brings up requiring user to confirm current password before changing password, which sounds logical. If we added that, I wonder if we'd need to send the Password Updated email to confirm the password change from within profile... but still seems vaild to keep in, in case a hacker guessed my current PW

app/Listeners/ReportPasswordUpdated.php
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * Handle the event.
+     *
+     * @param PasswordReset $event
+     * @return void
+     */
+    public function handle(PasswordUpdated $event)
+    {
+        $response = $this->client->post('customers/'.$event->user->id.'/events', [
```
aaronschachter: Per [chat with @DFurnes](https://dosomething.slack.com/archives/C02BBRN5A/p1553876905016200) -- I'm going to move this call to the Customer.io API into a new job.

aaronschachter:
Refactored to use a job to send `password_updated` events to Customer.io and verified they're showing in Customer.io. Here's one that was tracked as a result of me clicking through an RTV Activate Account email to create a password.


Screen Shot 2019-03-29 at 10 10 22 AM

app/Http/Controllers/Web/ResetPasswordController.php
```diff
@@ -45,22 +47,32 @@ protected function resetPassword($user, $password)
             'remember_token' => str_random(60),
         ])->save();
 
-        event(new PasswordReset($user));
+        // Pass along the password reset type route parameter as the source of password update.
+        event(new PasswordUpdated($user, request()->type));
```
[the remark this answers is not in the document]
aaronschachter: Yeah, I had a tough time trying to pass through `resetPassword` and couldn't figure out a way.

app/Services/CustomerIo.php
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * Create a new Customer.io API client.
+     */
+    public function __construct()
+    {
+        $config = config('services.customerio');
+
+        $this->client = new \GuzzleHttp\Client([
+            'base_uri' => $config['url'],
+            'auth' => [$config['username'], $config['password']],
```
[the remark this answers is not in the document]
aaronschachter: [Yep!](https://github.com/DoSomething/northstar/blob/84d387a9d83b5ed6e7bcf05899bd05d687b6808a/config/services.php#L65)

aaronschachter:
I've added a [new event triggered campaign to our Customer.io QA instance that sends an email when a `password_updated` event is created (regardless of source)](https://fly.customer.io/env/63547/campaigns/44/overview). Will set this up in production and update our [Events documentation](http://docs.dosomething.org/customer-io#events) once this is merged and ready for deploy.

aaronschachter:
When this is merged and deployed to production, we'll want to set up the relevant Event Triggered Campaigns in our production Customer.io workspace, as well as update our [internal documentation](http://docs.dosomething.org/customer-io#events) to add a new `password_updated` event.

app/Events/PasswordUpdated.php
```diff
@@ -0,0 +1,36 @@
...
+     * @var \Illuminate\Contracts\Auth\Authenticatable
+     */
+    public $user;
+
+    /**
+     * Source of password update (either profile or type of reset password form).
+     *
+     * @var string
+     */
+    public $source;
```
DFurnes: Calling this `$source` makes me think of the `source` field on a user. Would maybe `type` be clearer?

app/Http/Controllers/Web/UserController.php
```diff
@@ -99,6 +100,10 @@ public function update(Request $request, $id)
 
         $user->fill($values)->save();
 
+        if (isset($values['password'])) {
```
DFurnes: Would it make to handle this in a [model event](https://github.com/DoSomething/northstar/blob/84d387a9d83b5ed6e7bcf05899bd05d687b6808a/app/Providers/AppServiceProvider.php#L37-L42), so it'd be fired if the password was changed via other methods? I could see us extracting this form into Phoenix at some point down the road.

app/Jobs/SendPasswordUpdatedToCustomerIo.php
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Execute the job.
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        // Rate limit Customer.io API requests to 10/s.
+        $throttler = Redis::throttle('customerio')->allow(10)->every(1);
```
DFurnes: Now that we've got a couple jobs posting to Customer.io (either through Blink or not), maybe we should extract this `10` limit into `config/services.php` so they all refer to the same value?

app/Events/PasswordUpdated.php
aaronschachter: I had first called it `type` -- but setting password from profile was where source felt a little better.. but for sure was thinking about potential confusion with user.source. It makes me think about something like `updated_from` ?  I could see various profile pages requiring password to edit things like email/SMS, not just a separate change password form (which I started on in #856 but will not likely finish any time this quarter)

app/Http/Controllers/Web/UserController.php
aaronschachter: I do think it makes sense, but not sure how we'd pass along our `updatedVia`, which we need for sending "Account Activated" vs "Password Updated" confirmation emails from Customer.io. Going to hold off on this for now.

app/Jobs/SendPasswordUpdatedToCustomerIo.php
aaronschachter: Could it sweet to add something like a `getThrottler` into the new `Services/CustomerIo` to DRY a lot more of the code besides the just the limit? I'll look into this in a separate branch.

PR #857, opened 2019-03-29
Community backfill command

katiecrane:
#### What's this PR do?

🆘 Adds a helper to easily add a topic to a user's `email_subscription_topics` which takes care of not duplicating a topic.

💬 Adds `northstar:community` command to take all users who have `email_subscription_status` = `true` and who also do not already have `community` in their topics and add `community` to their topics.

🍮 Update the update users command to use the new `addEmailSubscriptionTopic()` helper.

Biggest area of concern: Does the logic in the command update the correct users? Maybe I should add a test for this command!

#### Relevant Tickets
[Card](https://www.pivotaltracker.com/story/show/164216160)

#### Checklist

### DIFF
--- a/app/Console/Commands/SetCommunityTopic.php
+++ b/app/Console/Commands/SetCommunityTopic.php
@@ -1,0 +1,78 @@
+<?php
+
+namespace Northstar\Console\Commands;
+
+use Northstar\Models\User;
+use Illuminate\Console\Command;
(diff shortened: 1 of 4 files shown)


DFurnes:
> Maybe I should add a test for this command!

I do _love_ a good test! If you have some time, definitely can't hurt! 😸

PR #859, opened 2019-04-01
Allow users to update themselves, but no one else

katiecrane:
#### What's this PR do?

🔬 For `v2/users`, removes `update` from the `'role:admin,staff'` middleware because we're gonna use a Gate now.

🏡 Only update a user if the request passes the `'edit-profile'` gate as determined [here](https://github.com/DoSomething/northstar/blob/master/app/Policies/UserPolicy.php#L29-L44).

🔧 Fixes `testV2UpdateProfileAsStaff()` test which was in a "two wrongs make a right" situation. A comment in the test and the code of the test indicated that staff should _not_ be able to update users. The UserPolicy and other documentation (including above this test itself) indicate that staff _should_ be able to update users. The code was a little off, so it wasn't really testing what it said it was because [`$user->fresh()` returns a new model rather than "refreshing" the one you already have](https://laravel.com/docs/5.8/eloquent#retrieving-models). It must be assigned to something in order to do what we think it does! So it this case, `$user` wasn't really refreshed, so no matter what we did it wasn't going to change! I updated this test to search directly in the database because for me that is a lot less confusing and leaves less room for error.

👩‍🎓 Added a new test to make sure users can update their own profile.

⛔️ Added a new test to make sure users can't update someone else's profile (this has a `TODO` in it because I am not getting the expected response code) .

Some very careful 👀 ! Do all these tests really test what I said they do?

#### Relevant Tickets
This is relevant to [this slack thread](https://dosomething.slack.com/archives/C3ASB4204/p1554146339067400).

#### Checklist

### DIFF
--- a/app/Http/Controllers/Two/UserController.php
+++ b/app/Http/Controllers/Two/UserController.php
@@ -40,7 +40,7 @@ class UserController extends Controller
         $this->registrar = $registrar;
         $this->transformer = $transformer;
 
-        $this->middleware('role:admin,staff', ['except' => ['show']]);
+        $this->middleware('role:admin,staff', ['except' => ['show', 'update']]);
         $this->middleware('scope:user');
         $this->middleware('scope:write', ['only' => ['store', 'update', 'destroy']]);
     }
--- a/app/Http/Controllers/Two/UserController.php
+++ b/app/Http/Controllers/Two/UserController.php
@@ -174,7 +174,6 @@ class UserController extends Controller
             $this->registrar->register($request->all(), $user);
         }
 
-
         return $this->item($user);
     }
 
--- a/app/Http/Controllers/Two/UserController.php
+++ b/app/Http/Controllers/Two/UserController.php
@@ -7,6 +7,7 @@ use Northstar\Auth\Role;
 use Northstar\Models\User;
 use Illuminate\Http\Request;
 use Northstar\Auth\Registrar;
+use Illuminate\Support\Facades\Gate;
 use Northstar\Http\Controllers\Controller;
 use Northstar\Http\Transformers\Two\UserTransformer;
 use Northstar\Exceptions\NorthstarValidationException;
@@ -169,7 +170,10 @@ class UserController extends Controller
             Role::gate(['admin']);
         }
 
-        $this->registrar->register($request->all(), $user);
+        if (Gate::allows('edit-profile', $user)) {
+            $this->registrar->register($request->all(), $user);
+        }
+
 
         return $this->item($user);
     }
--- a/app/Http/Controllers/Two/UserController.php
+++ b/app/Http/Controllers/Two/UserController.php
@@ -9,6 +9,7 @@ use Illuminate\Http\Request;
 use Northstar\Auth\Registrar;
 use Illuminate\Support\Facades\Gate;
 use Northstar\Http\Controllers\Controller;
+use Illuminate\Auth\AuthenticationException;
 use Northstar\Http\Transformers\Two\UserTransformer;
 use Northstar\Exceptions\NorthstarValidationException;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
@@ -161,6 +162,10 @@ class UserController extends Controller
     {
         $user = User::findOrFail($id);
 
+        if (! Gate::allows('edit-profile', $user)) {
+            throw new AuthenticationException('This action is unauthorized.');
+        }
+
         // Normalize input and validate the request
         $request = normalize('credentials', $request);
         $this->registrar->validate($request, $user);
@@ -170,9 +175,7 @@ class UserController extends Controller
             Role::gate(['admin']);
         }
 
-        if (Gate::allows('edit-profile', $user)) {
-            $this->registrar->register($request->all(), $user);
-        }
+        $this->registrar->register($request->all(), $user);
 
         return $this->item($user);
     }
--- a/app/Http/Controllers/Two/UserController.php
+++ b/app/Http/Controllers/Two/UserController.php
@@ -4,6 +4,7 @@ namespace Northstar\Http\Controllers\Two;
 
 use Auth;
 use Northstar\Auth\Role;
+use Northstar\Auth\Scope;
 use Northstar\Models\User;
 use Illuminate\Http\Request;
 use Northstar\Auth\Registrar;
@@ -162,7 +163,7 @@ class UserController extends Controller
     {
         $user = User::findOrFail($id);
 
-        if (! Gate::allows('edit-profile', $user)) {
+        if (! (Scope::allows('admin') || Gate::allows('edit-profile', $user))) {
             throw new AuthenticationException('This action is unauthorized.');
         }
 
--- a/tests/Http/UserTest.php
+++ b/tests/Http/UserTest.php
@@ -246,6 +246,31 @@ class UserTest extends BrowserKitTestCase
     }
 
     /**
+     * Test that a machine can update a user's profile.
+     * PUT /v2/users/:id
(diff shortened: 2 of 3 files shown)


app/Http/Controllers/Two/UserController.php
```diff
@@ -169,7 +170,9 @@ public function update($id, Request $request)
             Role::gate(['admin']);
         }
 
-        $this->registrar->register($request->all(), $user);
+        if (Gate::allows('edit-profile', $user)) {
```
DFurnes: I think it might make sense to "flip" this and return an explicit error, so it's clear why fields aren't written:

```php
if (! Gate::allows('edit-profile', $user)) {
    throw new AuthenticationException;
}
```

tests/Http/UserTest.php
```diff
@@ -186,10 +186,64 @@ public function testV2UpdateProfileAsStaff()
...
+     *
+     * @return void
+     */
+    public function testV2UpdateProfileAsSelf()
+    {
+        $user = factory(User::class)->create();
+
+        $this->asUser($user, ['user', 'role:staff', 'write'])->json('PUT', 'v2/users/'.$user->id, [
+            'first_name' => 'Pepper',
+            'last_name' => 'Puppy',
```
chloealee: 🐶

tests/Http/UserTest.php
```diff
@@ -186,10 +186,64 @@ public function testV2UpdateProfileAsStaff()
...
+        $user1 = factory(User::class)->create();
+        $user2 = factory(User::class)->create();
+
+        $this->asUser($user2, ['user', 'role:staff', 'write'])->json('PUT', 'v2/users/'.$user1->id, [
+            'first_name' => 'Burt',
+            'last_name' => 'Macklin',
+        ]);
+
+        // @TO-DO: I would expect at 403 here, but it is getting a 200
+        // $this->assertResponseStatus(403);
```
DFurnes: This should give a 403 with the change suggested above!

chloealee: hm weird! Was there anything the logs for this? Is this needed for merge?

app/Http/Controllers/Two/UserController.php
DFurnes: Another option that might read nicely is to use the [controller helpers](https://laravel.com/docs/5.8/authorization#via-controller-helpers):

```php
public function update($id, Request $request)
{
    $this->authorize('edit-profile', $user);
    
    // ...
```

katiecrane: @DFurnes yes I love this! I was trying to use `$this->authorize` originally but was getting errors that `authorize` didn't exist. I'll fiddle with it again and fall back on the other option if it's still not working, thank you!

tests/Http/UserTest.php
katiecrane: sounds like if I implement the suggestions from @DFurnes in the `UserController` this might start working as expected!

app/Http/Controllers/Two/UserController.php
DFurnes: Ohhh, you might need to include [this trait](https://github.com/laravel/framework/blob/5.6/src/Illuminate/Foundation/Auth/Access/AuthorizesRequests.php) on the controller.

katiecrane: Ahh I'll try that! Thanks!

katiecrane: Okay so with `authorize` when I shouldn't have access I get this error:
```
  "error": {
    "code": 500,
    "message": "This action is unauthorized."
  },
```
which comes from [this line](https://github.com/laravel/framework/blob/5.5/src/Illuminate/Auth/Access/HandlesAuthorization.php#L28).

katiecrane: With the gate method, I get:
```
  "error": {
    "code": 401,
    "message": "Unauthorized."
  },
```
I _think_ we really want a `403` here. 
I tried using `throw new AuthorizationException('This action is unauthorized.');` as well but with that I get:
```
  "error": {
    "code": 500,
    "message": "This action is unauthorized."
  }
```
@DFurnes do you think we're okay to use a `401` here for now?

DFurnes: Wow, funny! Do we need to [add that exception here](https://github.com/DoSomething/northstar/blob/d8b8b37fc09e8fca66189faa63cc86c65f3d3879/app/Exceptions/Handler.php#L74) since our handler is fairly customized? 🤔 If that doesn't help, then I'm also totally good with the Gate & 401!

DFurnes: And a 401 is totally good! I think that's technically actually more correct than a 403.

katiecrane: In that case, I'll leave it at `401` then!

app/Http/Controllers/Two/UserController.php
```diff
@@ -160,6 +162,10 @@ public function update($id, Request $request)
     {
         $user = User::findOrFail($id);
 
+        if (! Gate::allows('edit-profile', $user)) {
```
DFurnes: One more thought before we merge – does this work for "machine" tokens that hit this endpoint too? I remember we had to make a "hack" to the Policy in order to make that work correctly in Rogue…

DFurnes: Another short-term fix (assuming it doesn't work) would be to prepend [`Scope::allows('admin')`](https://github.com/DoSomething/northstar/blob/d8b8b37fc09e8fca66189faa63cc86c65f3d3879/app/Http/Transformers/UserTransformer.php#L32) like we do in the transformers.

DFurnes: And (again assuming that this might not work), we should add a test covering this!

katiecrane: @DFurnes I was able to update a user via my local Aurora, but that was using my admin credentials wasn't it? Would I have to try Chompy in order to test this? Trying to think of how Rogue might update a user and coming up blank.

DFurnes: You can test this with Paw (using the Client Credentials grant and `dev-machine` credentials from Aurora), or by using the `$this->withAccessToken(['admin'])` test helper.

DFurnes: It could make sense to make a `$this->asMachine()` helper that shortcuts to the above `withAccessToken` call, if you're already [in that file](https://github.com/DoSomething/northstar/blob/master/tests/WithAuthentication.php) too!

katiecrane: @DFurnes hmm I wasn't able to get it to work via Paw and I think I'm confused about how `withAccessToken` works. Passing `['admin']` will get a machine token?

DFurnes: Yup! The [two arguments](https://github.com/DoSomething/northstar/blob/4d7c85f756f3c98df3ab97152f6c96f4d1985d2c/tests/WithAuthentication.php#L99-L106) to that helper are a list of scopes, and an (optional) `$user`. Machine tokens are just tokens that don't have a user attached (since they're coming from an application itself, like a Chompy queue or Gambit).

This should end with a token that looks exactly like what the client credentials grant would have given you (either by [hitting the endpoint directly](https://github.com/DoSomething/northstar/blob/master/documentation/endpoints/auth.md#create-token-client-credentials-grant) or via an application).

katiecrane: alright @DFurnes I think this is working now! I added a test and also got it working with Paw (I wasn't requesting a token with the correct scopes before)

PR #862, opened 2019-04-10
Avoid setting duplicates in `email_subscription_topics`

katiecrane:
#### What's this PR do?

Another application might be trying to add a value to a user's `email_subscription_topics` and it would be nice for that application or human to not have to worry about setting duplicates. See an example [here](https://github.com/DoSomething/chompy/pull/83#discussion_r271468629).

👯 Add `setEmailSubscriptionTopicsAttribute()` setter to de-dupe the array being set as `email_subscription_topics` and then use the de-duped array for that field.

✍️ Adds a test to make sure duplicates aren't set!

🦐 Simplifies the `addEmailSubscriptionTopic()` helper to be just one line! We don't need to check for duplicates anymore, so can just merge in the new topic directly with the existing ones.

Is there any way that a duplicate email topic could be added? Can you think of anything else to test? A test specifically for the `addEmailSubscriptionTopic()` helper?

#### Relevant Tickets
[Card](https://www.pivotaltracker.com/story/show/165074676)

#### Checklist

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -552,10 +552,7 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
      */
     public function setEmailSubscriptionTopicsAttribute($value)
     {
-        // De-dupe given array
-        $topics = array_values(array_unique($value));
-
         // Set de-duped array as email_subscription_topics
-        $this->attributes['email_subscription_topics'] = $topics;
+        $this->attributes['email_subscription_topics'] = array_values(array_unique($value));
     }
 }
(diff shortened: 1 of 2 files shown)


app/Models/User.php
```diff
@@ -537,22 +537,25 @@ public function setVoterRegistrationStatusAttribute($status)
...
+     */
+    public function setEmailSubscriptionTopicsAttribute($value)
+    {
+        // De-dupe given array
+        $topics = array_values(array_unique($value));
 
-        // Add the full array of topics to the user
-        $this->email_subscription_topics = $topics;
+        // Set de-duped array as email_subscription_topics
+        $this->attributes['email_subscription_topics'] = $topics;
```
weerd: You could possibly just do this all as a single line w/out the need for the `$topics` variable, but it does read a little clearer this way as a step-wise thing.

katiecrane: ooo I like this suggestion! I think it'll be similarly clear written in one line!

PR #866, opened 2019-04-19
Only log about user updates if not running in console

katiecrane:
#### What's this PR do?

🍃 We don't need these logs if the app is running in console! These are also debug level logs that will not show up on prod or QA anyway! The hope here is that this will help the [Community backfill script](https://github.com/DoSomething/northstar/blob/master/app/Console/Commands/SetCommunityTopic.php) to not run out of memory so quickly.

🚁 Moves the logic that happened in the AppServiceProvider when a user was updating to sanitize the diff into a helper on Model. This way we can still test it even though we're not logging it. Test updated as well!

Will this still log when we want it to?

#### Relevant Tickets
🤠 [Card-ish](https://www.pivotaltracker.com/story/show/165087297)

#### Checklist

### DIFF
--- a/app/Models/Model.php
+++ b/app/Models/Model.php
@@ -113,4 +113,16 @@ class Model extends BaseModel
 
         return true;
     }
+
+    /**
+     * Get the attributes that have been changed.
+     *
+     * @return array
+     */
+    public function getChanged()
+    {
+        $changed = array_replace_keys($this->getDirty(), $this->getHidden(), '*****');
+
+        return $changed;
+    }
 }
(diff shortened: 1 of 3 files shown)


app/Models/Model.php
```diff
@@ -113,4 +113,16 @@ protected function performUpdate(Builder $query, array $options = [])
 
         return true;
     }
+
+    /**
+     * Get the attributes that have been changed.
```
DFurnes: Let's add a little note here about how this redacts "hidden" fields.

PR #867, opened 2019-04-22
Add script to fix Drupal formatting weirdness.

DFurnes:
#### What's this PR do?
This pull request adds a `northstar:dedrupal` script, which can be used to parse fields that have awkward `['value' => 'actual value']` formatted contents into the expected `'actual value'`.

I added a test! Will run on QA before production. 🚥 

#### Relevant Tickets
[#165187491](https://www.pivotaltracker.com/story/show/165187491)

#### Checklist

### DIFF
--- a/app/Console/Commands/FixDrupalFields.php
+++ b/app/Console/Commands/FixDrupalFields.php
@@ -58,8 +58,8 @@ class FixDrupalFields extends Command
                 info('Sanitized object field.', [
                     'id' => $user->id,
                     'field' => $field,
-                    'is_array' => is_array($value),
-                    'parsable' => ! is_null($value),
+                    'before' => $value,
+                    'after' => $user->{$field},
                 ]);
             }
         });
--- a/app/Console/Commands/FixDrupalFields.php
+++ b/app/Console/Commands/FixDrupalFields.php
@@ -1,0 +1,69 @@
+<?php
+
+namespace Northstar\Console\Commands;
+
+use Northstar\Models\User;
+use Illuminate\Console\Command;
+use Illuminate\Support\Collection;
+
+class FixDrupalFields extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'northstar:dedrupal {field}';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Update fields that have nasty Drupal objects if possible.';
+
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        $field = $this->argument('field');
+
+        // Get users where field is not a string:
+        $query = User::whereRaw([
+            $field => [
+                '$exists' => true,
+                '$not' => ['$type' => 'string'],
+            ],
+        ]);
+
+        $progressBar = $this->output->createProgressBar($query->count());
+        $query->chunkById(200, function (Collection $users) use ($field, $progressBar) {
+            foreach ($users as $user) {
+                $value = $user->{$field};
+
+                // If this is an object, it's likely Drupal gook.
+                if (is_array($value)) {
+                    $sanitized = array_get($value, 'value');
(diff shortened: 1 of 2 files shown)


app/Console/Commands/FixDrupalFields.php
```diff
@@ -0,0 +1,69 @@
...
+     */
+    protected $description = 'Update fields that have nasty Drupal objects if possible.';
+
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
```
katiecrane: If this could potentially update a lot of users I think we'll want to make sure we use the low priority queue here:
```
config(['queue.jobs.users' => 'low']);
```

app/Console/Commands/FixDrupalFields.php
```diff
@@ -0,0 +1,69 @@
...
+
+        // Get users where field is not a string:
+        $query = User::whereRaw([
+            $field => [
+                '$exists' => true,
+                '$not' => ['$type' => 'string'],
+            ],
+        ]);
+
+        $progressBar = $this->output->createProgressBar($query->count());
```
katiecrane: I forgot that we figured out that we _can_ use a progress bar with PaperTrail! Excited to see how this looks!

app/Console/Commands/FixDrupalFields.php
```diff
@@ -0,0 +1,69 @@
...
+
+                $progressBar->advance();
+                $user->save();
+
+                info('Sanitized object field.', [
+                    'id' => $user->id,
+                    'field' => $field,
+                    'is_array' => is_array($value),
+                    'parsable' => ! is_null($value),
+                ]);
```
katiecrane: This is BEAUTIFUL! Could it be helpful to show the `new_value` since we won't be getting the usual "user updated" logs because this will be running in the console? Would it be helpful to see the `old_value` too? As a sanity check/peace of mind?

Also can you explain the `parsable` piece? Won't the `$value` always be non-null because we unset `null` values?

tests/Console/FixDrupalFieldsTest.php
```diff
@@ -0,0 +1,39 @@
...
+        $userWithValidField = User::where('email', 'dave@example.com')->first();
+        $userWithParsableField = User::where('email', 'bob@example.com')->first();
+        $userWithNonsenseField = User::where('email', 'hackz@example.com')->first();
+        $newBorkedUsersCount = User::whereRaw(['addr_street1' => $invalidFormatQuery])->count();
+
+        $this->assertEquals(0, $newBorkedUsersCount);
+        $this->assertEquals('19 W 21st St', $userWithValidField->addr_street1);
+        $this->assertEquals('1 Main Street', $userWithParsableField->addr_street1);
+        $this->assertEquals(null, $userWithNonsenseField->addr_street1);
+    }
```
katiecrane: I love this test!!!

app/Console/Commands/FixDrupalFields.php
DFurnes: Good flag! I think we're safe in this case, since each field only has ~200 bad records.

app/Console/Commands/FixDrupalFields.php
DFurnes: Yeah, I wasn't sure if we wanted to drop these into the logs or not. I definitely see how it could make this feel a bit safer to run on production. 👍 The `parsable` line would show whether we were able to get "something" out of the original field (so it'd be `false` in cases where the field wasn't an array or wasn't the right `['value' => x]` format.

DFurnes: Oh but now I'm also seeing that I should've done `! is_null($user->{$field})` oooops. Yeah, you're totally right!

DFurnes: Adjusted these logs based on your comments in 27b12bd.

katiecrane: Love the updated logs, zero cognitive overhead, so clear!

PR #868, opened 2019-04-23
Remove birthdates that are unreasonably far in the past.

DFurnes:
#### What's this PR do?
This pull request adds a script which removes any birthdates that are before the year 1900, because chances are we don't have users over 119 years old. And certainly very few users that are over two thousand years old, as indicated in the ticket.

I added a test! 🚥

#### Relevant Tickets
[#165190368](https://www.pivotaltracker.com/story/show/165190368)

#### Checklist

### DIFF
--- a/app/Console/Commands/RemoveOldBirthdates.php
+++ b/app/Console/Commands/RemoveOldBirthdates.php
@@ -30,7 +30,7 @@ class RemoveOldBirthdates extends Command
      */
     public function handle()
     {
-        // Get users where field is not a string:
+        // Get any users where 'birthdate' is before 1900... a.k.a way too old:
         $query = User::whereRaw([
             'birthdate' => [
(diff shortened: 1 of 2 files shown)


app/Console/Commands/RemoveOldBirthdates.php
```diff
@@ -0,0 +1,58 @@
...
+    protected $description = 'Remove birthdates that are way too old.';
+
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        // Get users where field is not a string:
```
katiecrane: This comment didn't get updated ✍️

DFurnes: Whooopsie! Amended in 76a9000. ✏️

PR #869, opened 2019-04-23
Fix issue where we'd delete all birthdates before 1969!

DFurnes:
#### What's this PR do?
This pull request fixes a funny issue from #868, where `1900-01-01 00:00:00` was getting parsed into `1969-12-06 10:23:31` by [`strtotime`](https://www.php.net/manual/en/function.strtotime.php) (because different timestamp formats), and so we'd clear birthdates for far younger people than we intended to.

I added a failing test in 9ce2c17 and the fix in d947298.

#### Relevant Tickets
[#165190368](https://www.pivotaltracker.com/story/show/165190368)

#### Checklist

### DIFF
--- a/app/Console/Commands/RemoveOldBirthdates.php
+++ b/app/Console/Commands/RemoveOldBirthdates.php
@@ -30,10 +30,12 @@ class RemoveOldBirthdates extends Command
      */
     public function handle()
     {
-        // Get any users where 'birthdate' is before 1900... a.k.a way too old:
+        // Get any users where 'birthdate' is before 1900. Note that strtotime gives
+        // us a traditional UNIX timestamp (in seconds), but UTCDateTime wants a
+        // JavaScript-style timestamp (in milliseconds).
         $query = User::whereRaw([
             'birthdate' => [
-                '$lt' => new UTCDateTime(strtotime('1900-01-01 00:00:00')),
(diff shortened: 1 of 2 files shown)


app/Console/Commands/RemoveOldBirthdates.php
```diff
@@ -30,10 +30,12 @@ class RemoveOldBirthdates extends Command
      */
     public function handle()
     {
-        // Get any users where 'birthdate' is before 1900... a.k.a way too old:
+        // Get any users where 'birthdate' is before 1900. Note that strtotime gives
+        // us a traditional UNIX timestamp (in seconds), but UTCDateTime wants a
+        // JavaScript-style timestamp (in milliseconds).
```
katiecrane: This is so tricky!! And is breaking my brain a bit. If we pass through a larger number (because we're multiplying by 1000 now) we get an _earlier_ time?

DFurnes: Yeah, it's measured in seconds (or in this case milliseconds) before 1970, so we're multiplying a negative number by 1000 to get "earlier". 🙃

DFurnes: This is such a great example of how ridiculous the idea of UNIX time is!

katiecrane: OOOO negatives, love it!

PR #871, opened 2019-05-13
`feature_flags` field on users

katiecrane:
#### What's this PR do?

🎏 New feature flag `DS_BADGES_TEST` which defaults to `false` for whether or not we should be opting folks into the badges test.

🇺🇸 New field on users called `feature_flags`.

📨 For new registrants, give half the feature flag for badges `true` and give half `false`. Only do this if we are currently running the badges test. Example below!

👚 Expose the `feature_flags` attribute in the `v2` user transformer to admins or the user whose account it is.

👩‍🎓 Added tests to make sure `feature_flags['badges']` is assigned appropriately.

🏰 Add `badges` to Customer.io payload if the user is in one of the test groups. ⚠️ ⚠️ ⚠️ _This brings our Customer.io payload up to 29 attributes and we cannot send more than 30_

**Example**
I created a new user who has `feature_flags` as follows:
```
     feature_flags: [
       "badges" => false,
     ],
```

Will this allow us to run the badges test how we want to? Can other services access the `feature_flags` information as they need to?

⚠️ ⚠️ ⚠️ _Please double check my counting and make sure in no cases are we sending more than 30 attributes to Customer.io_

#### Relevant Tickets
[Card](https://www.pivotaltracker.com/story/show/165718950)

#### Checklist

### DIFF
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -198,19 +198,6 @@ class Registrar
             $customizer($user);
         }
 
-        // If the badges test is running, put half of users in badges group and half in control group
-        if (config('features.badges')) {
-            $feature_flags = $user->feature_flags;
-
-            if (rand(0, 1) === 1) {
-                $feature_flags['badges'] = true;
-            } else {
-                $feature_flags['badges'] = false;
-            }
-
-            $user->feature_flags = $feature_flags;
-        }
-
         $user->save();
 
         return $user;
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -202,7 +202,7 @@ class Registrar
         if (config('features.badges')) {
             $feature_flags = $user->feature_flags;
 
-            if (rand(0,1) === 1) {
+            if (rand(0, 1) === 1) {
                 $feature_flags['badges'] = true;
             } else {
                 $feature_flags['badges'] = false;
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -198,6 +198,19 @@ class Registrar
             $customizer($user);
         }
 
+        // If the badges test is running, put half of users in badges group and half in control group
+        if (config('features.badges')) {
+            $feature_flags = $user->feature_flags;
+
+            if (rand(0,1) === 1) {
+                $feature_flags['badges'] = true;
+            } else {
+                $feature_flags['badges'] = false;
+            }
+
+            $user->feature_flags = $feature_flags;
+        }
+
         $user->save();
 
         return $user;
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -436,7 +436,7 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
 
         if (isset($this->feature_flags)) {
             if (array_key_exists('badges', $this->feature_flags)) {
-                $payload['badges'] = $this->feature_flags['badges'];
+                $payload['badges_feature_flag'] = $this->feature_flags['badges'];
             }
         }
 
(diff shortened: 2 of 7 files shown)


app/Auth/Registrar.php
```diff
@@ -198,6 +198,19 @@ public function register($input, $user = null, Closure $customizer = null)
...
+        if (config('features.badges')) {
+            $feature_flags = $user->feature_flags;
+
+            if (rand(0,1) === 1) {
+                $feature_flags['badges'] = true;
+            } else {
+                $feature_flags['badges'] = false;
+            }
+
+            $user->feature_flags = $feature_flags;
```
mendelB: Just confirming that this feature flag logic would only be applied to new users? (of course assuming so based on the method name lol but just noticed the description says (emphasis mine): "Create a new user *OR update an existing user (if given).*")

mendelB:
> ⚠️ ⚠️ ⚠️ Please double check my counting and make sure in no cases are we sending more than 30 attributes to Customer.io

I counted 29 as well!
Assuming we'd be using this in Customer.io to determine to add badge related content in emails?

app/Auth/Registrar.php
katiecrane: I just noticed that as well! I'm gonna check on it, excellent flag!

DFurnes: Yeah, this might make more sense to handle in the [AuthController's customizer callback](https://github.com/DoSomething/northstar/blob/3c64e6898ed8efb1be8ce0b9ce623d059039e09b/app/Http/Controllers/Web/AuthController.php#L202-L221).

app/Models/User.php
```diff
@@ -428,6 +434,12 @@ public function toCustomerIoPayload()
             $payload['unsubscribed'] = (! $this->email_subscription_status);
         }
 
+        if (isset($this->feature_flags)) {
+            if (array_key_exists('badges', $this->feature_flags)) {
+                $payload['badges'] = $this->feature_flags['badges'];
```
DFurnes: Would it make sense to suffix this as `badges_feature_flag` like we do for `email_subscription_status` above?

app/Auth/Registrar.php
```diff
@@ -198,6 +198,19 @@ public function register($input, $user = null, Closure $customizer = null)
             $customizer($user);
         }
 
+        // If the badges test is running, put half of users in badges group and half in control group
+        if (config('features.badges')) {
+            $feature_flags = $user->feature_flags;
+
+            if (rand(0, 1) === 1) {
```
DFurnes: This could be simplified a bit to `$user->feature_flags['badges'] = (bool) rand(0, 1)`.

tests/Http/Web/WebAuthenticationTest.php
```diff
@@ -193,6 +193,49 @@ public function testRegister()
...
+        $this->withHeader('X-Fastly-Country-Code', 'US')
+            ->register();
+
+        $this->seeIsAuthenticated('web');
+
+        /** @var User $user */
+        $user = auth()->user();
+
+        // The user should have a value set for 'badges'
+        $this->assertEquals(true, array_key_exists('badges', $user->feature_flags));
```
DFurnes: This might read more clearly as `$this->assertArrayHasKey('badges', $user->feature_flags)`.

app/Auth/Registrar.php
katiecrane: AMAZING

app/Auth/Registrar.php
katiecrane: That for sure seems like the right place. I feel like we've run into this confusion with this `register` function before too, maybe down the line we can re-name it to something a little clearer.

app/Models/User.php
katiecrane: Love it!

tests/Http/Web/WebAuthenticationTest.php
katiecrane: No "might," it is for sure clearer the way you suggested!

PR #874, opened 2019-05-16
Voter Registration Field Label Height - Bug Fix

mendelB:
#### What's this PR 
This PR fixes a lil bug that popped up with the Voter Registration Status field in the register form which would hide some overflow text on certain breakpoints.

It adds a new `.height-auto` field, which when applied to an element of `.field-label` persuasion, will set `height: auto` to override Forge's `height: 1.5em` applied to `.field-label`s.

Does the field not get hidden anymore as you adjust the viewport size?

I tried to conform to the style style, by adding a new `_field-label.scss` file within which to declare this new class. Was this correct?

#### Relevant Tickets
Subtask in https://www.pivotaltracker.com/story/show/165160297

#### Checklist

📸 

**Before**
image
**After**
image

### DIFF
--- a/resources/assets/scss/_components/_field-label.scss
+++ b/resources/assets/scss/_components/_field-label.scss
@@ -1,0 +1,5 @@
+
+// Overrides the static height definition for .field-label in Forge (https://git.io/fjl59)
+.field-label.height-auto {
+  height: auto;
+}
--- a/resources/assets/scss/app.scss
+++ b/resources/assets/scss/app.scss
@@ -22,6 +22,7 @@
 @import "_components/_divider";
 @import "_components/_password-visibility";
 @import "_components/_text-field";
+@import "_components/_field-label";
 
 // Regions - complete sections of an interface
 @import "_regions/_chrome";
--- a/resources/views/auth/register.blade.php
+++ b/resources/views/auth/register.blade.php
@@ -74,7 +74,7 @@
             </div>
 
             <div class="form-item">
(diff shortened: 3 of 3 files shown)


katiecrane:
@mendelB I haven't really touched styles in Northstar but I see there is a `field-label` something [here in this file](https://github.com/DoSomething/northstar/blob/ddf98cc07d77bcea0ee572ef6e83c7610e0010be/resources/assets/scss/_regions/_container.scss#L59). I'm wondering if it would work or make sense for your change to go in there? The comment above the line does say "Hacky!" so I'm not sure which is a better option but wanted to make sure you saw this spot!

mendelB:
> @mendelB I haven't really touched styles in Northstar but I see there is a `field-label` something [here in this file](https://github.com/DoSomething/northstar/blob/ddf98cc07d77bcea0ee572ef6e83c7610e0010be/resources/assets/scss/_regions/_container.scss#L59). I'm wondering if it would work or make sense for your change to go in there? The comment above the line does say "Hacky!" so I'm not sure which is a better option but wanted to make sure you saw this spot!

Ooh I actually missed that! I was mostly influenced by looking at the pattern of current scss files we have, and particulary the [`_text-field.scss`](https://github.com/DoSomething/northstar/blob/76a6503e7701f072204b77461e0776b58fb2dd91/resources/assets/scss/_components/_text-field.scss) which seems to be doing something structurally similar to this. (adding a modifier class `.is-disabled` to alter the `.text-field`).

What do you think makes the most sense here?

katiecrane:
@mendelB Ah yeah, that example seems to be the exact same idea as what you've done here so let's keep it as is!

PR #877, opened 2019-05-21
Add Google Tag Manager 

mendelB:
#### What's this PR do?
This PR adds the [Google Tag Manager](https://tagmanager.google.com/#/home) script to Northstar and pushes all analytics events into the [Data Layer](https://support.google.com/tagmanager/answer/6164391?hl=en)

I tested this using the [preview mode](https://support.google.com/tagmanager/answer/6107056?hl=en) and it looks like the events are being pushed in as expected!

I mostly copied the setup we used to add this script to Phoenix a while back (https://github.com/DoSomething/phoenix-next/pull/1001) and to track the events to the Data Layer (https://github.com/DoSomething/phoenix-next/pull/1213).

#### Relevant Tickets
[Pivotal ID #166139177](https://www.pivotaltracker.com/story/show/166139177)

#### Checklist


resources/views/layouts/app.blade.php
```diff
@@ -39,6 +39,7 @@
     {{ scriptify($errors->messages(), 'ERRORS') }}
     <script src="{{ elixir('app.js', 'dist') }}"></script>
     @include('layouts.google_analytics')
+    @include('layouts.google_tag_manager')
```
mendelB: Do you think we can bump these to be in the `<head>` of the page, or is there a reason we want these under the rest of our JS on this page? I thought [this article](https://www.bounteous.com/insights/2012/02/09/where-should-google-analytics-tracking-code-be-placed/?ns=l) made a pretty good case for moving this into the `<head>` since this is asynchronously loaded and shouldn't affect site load speed.

katiecrane: Yeah based on that article seems like these should be able to move to the `<head>`...I'm not sure of the history here though, @DFurnes any qualms?

DFurnes: Yeah, I think moving this up to the `head` is totally fine!

mendelB: Moved 'em to the `head` in https://github.com/DoSomething/northstar/pull/877/commits/1c5eb084cd372c03c5c4a0002c51fdf17c68ebd8 (Did I put them in the right place?)

weerd: LOL. I was going to ask y'all the same thing for Phoenix... I'll wait to move it when I get the all clear for removing the GA script and just having the GTM script, but glad to know y'all are on board, haha!

mendelB: Great minds!

PR #882, opened 2019-06-11
Updates the ImportSubStatusFromCio script to accept a CSV

chloealee:
#### What's this PR do?
Updates the `ImportSubStatusFromCio` script to accept a CSV and update users from the CSV instead of all users in the database. 

Does this script work as intended? Does anything break when updating all users? 

It looks like the users in the CSV all have `email_subscription_status` set to `TRUE`. I ran this script on QA and checked random users from the CSV to compare to prod. I couldn't find any users that didn't already have a `email_subscription_status` set to `TRUE` already in the CSV. Have these already been set?  

#### Relevant Tickets
[Pivotal Ticket](https://www.pivotaltracker.com/n/projects/2328687/stories/166361521)

#### Checklist


katiecrane:
> It looks like the users in the CSV all have email_subscription_status set to TRUE. I ran this script on QA and checked random users from the CSV to compare to prod. I couldn't find any users that didn't already have a email_subscription_status set to TRUE already in the CSV. Have these already been set?

@chloealee I'm a little confused about this. Won't this script be setting most of these folks to have `email_subscription_status` as `false`?

chloealee:
@katiecrane ah - you're right! By what the Pivotal Ticket is named, it does seem that it should be setting these values to `false` however from the CSV that I pulled, it looks like all the users in the spreadsheet have `email_subscription_status` set to `TRUE`, which is what they're set already on prod as? 

Also, this script only actually uses the `northstar_id` from the CSV to look up directly what their `email_subscription_status` is set in C.io and then saves that value in northstar. It seems like all of the values are `TRUE` from C.io, but are also already that value in the northstar DB so wondering if this even needs to be run? 

Unless I'm not understanding this card correctly @DFurnes ?

chloealee:
@DFurnes also, this CSV has 16,085 records in it (the Pivotal Card mentions 15,748) - is this ok?

katiecrane:
@chloealee yeah the `email_subscription_status` for these folks on prod are all mistakenly `true`. In c.io, they have `unsubscribed` of `true` and `email_subscription_status` of `true`, which is wrong because those should be opposites. I think the goal is to grab the `unsubscribed` status of these folks from c.io and update `email_subscription_status` to match, so at the end I'd expect all the values for that to be `false`.

It is super confusing to think about because of the double negative with `unsubscribed`.

chloealee:
@katiecrane ahh I see - thanks for explaining that more! The existing code [here](https://github.com/DoSomething/northstar/blob/master/app/Jobs/GetEmailSubStatusFromCustomerIo.php#L54) seems to be grabbing the `unsubscribed` property from the C.io API. This script didn't update and save any of the `email_subscription_status` to `FALSE` so there is something wrong then because we expect all these to be `FALSE` now, right?

katiecrane:
@chloealee yeah! looks like there is an issue in the job it kicks off, `GetEmailSubStatusFromCustomerIo`, seeing this in the logs:
```
Jun 11 13:08:20 dosomething-northstar-qa app/queue.1:  [2019-06-11 20:08:20] qa.INFO: [GetEmailSubStatusFromCustomerIo] For user 5c9597e43e304e00a8414aea got unsubscribed= [] {"user_id":null,"client_id":"northstar","request_id":null} 
Jun 11 13:08:20 dosomething-northstar-qa app/queue.1:  [2019-06-11 20:08:20] qa.INFO: [GetEmailSubStatusFromCustomerIo] For user 5c9597e43e304e00a8414aea set email_subscription_status=1 [] {"user_id":null,"client_id":"northstar","request_id":null} 
```

It got a non-true value for `unsubscribed` but `email_subscription_status` was still set as `true`

chloealee:
a-ha! I'll work on this - thanks @katiecrane !!!

chloealee:
hmm this is strange! When logging out, it looks like the `unsubscribed` values from the C.io calls all return an empty array. However, I tried to hit this same URL for a user (via cURL) with and see that the `unsubscribed` value = `false`. I'm not sure why this would be a different response in the code (with an empty array for `unsubscribed` versus the cURL response?

chloealee:
Also confirmed via the [documentation](https://customer.io/docs/api/) that nothing seemed to change for the API response either?

app/Jobs/GetEmailSubStatusFromCustomerIo.php
```diff
@@ -51,6 +51,7 @@ public function handle()
             // Make request to c.io to get that user's subscription status
             $response = $client->get('/v1/api/customers/'.$this->user->id.'/attributes');
             $body = json_decode($response->getBody());
+            print_r($body);
             $unsubscribed = $body->customer->unsubscribed;
```
chloealee: I tried to grab the `unsubscribed` value from the `attributes` in the response, since it is also returned here in the [api response](https://customer.io/docs/api/) and it seemed to work but not convinced it is the best way because there were some errors in the logs:
```
Undefined property: stdClass::$unsubscribed {"exception":"[object] (ErrorException(code: 0): Undefined property: stdClass::$unsubscribed at /app/app/Jobs/GetEmailSubStatusFromCustomerIo.php:55)"} {"user_id":null,"client_id":"northstar","request_id":null} 
```
but it still updated correctly 🤔 

Confirmed that it was updated in the logs:
`qa.INFO: [GetEmailSubStatusFromCustomerIo] For user *** got unsubscribed=false [] {"user_id":null,"client_id":"northstar","request_id":null} `

but this was a weird error: 
`qa.INFO: [GetEmailSubStatusFromCustomerIo] For user *** set email_subscription_status= [] {"user_id":null,"client_id":"northstar","request_id":null} ` 
(would have thought `email_subscription_status` would be `false`, which when I checked in aurora it was `false`...

app/Jobs/GetEmailSubStatusFromCustomerIo.php
```diff
@@ -51,7 +51,7 @@ public function handle()
             // Make request to c.io to get that user's subscription status
             $response = $client->get('/v1/api/customers/'.$this->user->id.'/attributes');
             $body = json_decode($response->getBody());
-            $unsubscribed = $body->customer->unsubscribed;
+            $unsubscribed = $body->customer->attributes->unsubscribed;
```
chloealee: I tried to grab the `unsubscribed` value from the `attributes` in the response, since it is also returned here in the [api response](https://customer.io/docs/api/) and it seemed to work but not convinced it is the best way because there were some errors in the logs:
```
Undefined property: stdClass::$unsubscribed {"exception":"[object] (ErrorException(code: 0): Undefined property: stdClass::$unsubscribed at /app/app/Jobs/GetEmailSubStatusFromCustomerIo.php:55)"} {"user_id":null,"client_id":"northstar","request_id":null} 
```
but it still updated correctly 🤔 

Confirmed that it was updated in the logs:
`qa.INFO: [GetEmailSubStatusFromCustomerIo] For user *** got unsubscribed=false [] {"user_id":null,"client_id":"northstar","request_id":null} `

but this was a weird error: 
`qa.INFO: [GetEmailSubStatusFromCustomerIo] For user *** set email_subscription_status= [] {"user_id":null,"client_id":"northstar","request_id":null} ` 
(would have thought `email_subscription_status` would be `false`, which when I checked in aurora it was `false`...

DFurnes: It looks like the `unsubscribed` value under the attributes key [in that response](https://customer.io/docs/api/#apibeta-apicustomerscustomers_get_attributes) is the timestamp when a user unsubscribed, and the actual boolean "status" is at the top level (below, in the example response). Is there a reason to use that instead?

DFurnes:
> hmm this is strange! When logging out, it looks like the unsubscribed values from the C.io calls all return an empty array. However, I tried to hit this same URL for a user (via cURL) with and see that the unsubscribed value = false. I'm not sure why this would be a different response in the code (with an empty array for unsubscribed versus the cURL response?

Very strange!

Could we perhaps be sending different headers from the application than you are via cURL? When you see the empty array, is that from something you've parsed out or from the string response body?

app/Jobs/GetEmailSubStatusFromCustomerIo.php
chloealee: @DFurnes so both `$body->customer->attributes->unsubscribed` and `$body->customer->unsubscribed` _should_ return booleans ($body->customer->timestamps->unsubscribed` returns a timestamp) but for some reason, the response for `$body->customer->unsubscribed` is returning an empty array when it is called in the script. 

When accessing the same API endpoint via cURL, it does return a boolean but just not in the script for some reason 🤔 However, accessing this value via `$body->customer->attributes->unsubscribed;` does return a boolean and looks like it updated the users on QA when I deployed the branch there. However, there are a lot of the above errors that doesn't feel right... 

I'm trying to figure out why the API response for `$body->customer->unsubscribed;` doesn't return the expected value from the script. Have you worked with C.io and come across this before?

chloealee: ah sorry just saw your other comment - will comment there!

DFurnes: > @DFurnes so both `$body->customer->attributes->unsubscribed` and `$body->customer->unsubscribed` _should_ return booleans ($body->customer->timestamps->unsubscribed`returns a timestamp) but for some reason, the response for`$body->customer->unsubscribed` is returning an empty array when it is called in the script.

Oh yup, I see I misread that! And the `$body->customer->attributes->unsubscribed` one doesn't have this strange behavior where it sometimes returns an array?

> I'm trying to figure out why the API response for `$body->customer->unsubscribed;` doesn't return the expected value from the script. Have you worked with C.io and come across this before?

I haven't touched this Customer.io API before, unfortunately! I do remember some instances that @ngjo had flagged where customers wouldn't have an `unsubscribed` _attribute_ in Customer.io if they had never been unsubscribed, so we might just need to check for field existence here (and if it's not set, they'd be subscribed still).

chloealee:
> Could we perhaps be sending different headers from the application than you are via cURL? When you see the empty array, is that from something you've parsed out or from the string response body?

@DFurnes confirmed that the headers are the same! The empty array is something from the string body response :-/

app/Jobs/GetEmailSubStatusFromCustomerIo.php
chloealee: > Oh yup, I see I misread that! And the $body->customer->attributes->unsubscribed one doesn't have this strange behavior where it sometimes returns an array?

@DFurnes the `$body->customer->unsubscribed` is the one that doesn't have a value when printing from the script but is an empty array! 

> I haven't touched this Customer.io API before, unfortunately! I do remember some instances that @ngjo had flagged where customers wouldn't have an unsubscribed attribute in Customer.io if they had never been unsubscribed, so we might just need to check for field existence here (and if it's not set, they'd be subscribed still).

I will try this!

chloealee: Actually, thinking about this more, I think this is a list that we want to change their status to `unsubscribed` so all of these should have values? Also, the ` $body->customer->attributes->unsubscribed` all say `false` for these responses so I'm not sure this is the route we want to go? I'll keep digging!

chloealee:
ok weird! This seems to be miraculously working now returning the boolean correctly from the API response! Ready for review!

PR #891, opened 2019-07-03
Add (optional) two-factor authentication for staff users.

DFurnes:
#### What's this PR do?
This pull request adds an optional two-factor authentication step in our login flow (currently only exposed for staff users). Eventually, I think we should require this for all staff accounts.

Staffers (with either `staff` or `admin` role) will see this prompt on identity.dosomething.org:

Screen Shot 2019-07-03 at 4 26 26 PM

Screen Shot 2019-07-03 at 4 20 51 PM

After configuring an authenticator app, future logins will require a code:

<p align=center></p>

Screen Shot 2019-07-08 at 10 41 30 AM

I've added the functionality & tests in 89870fa. This relies on [Spomky-Labs/otphp](https://github.com/Spomky-Labs/otphp) for the actual code provisioning & verification, so the only "custom" code here is to hook things up with our own login flow.

I'll add the ability to "unlink" a two-factor device from your account in a follow-up PR.

#### Relevant Tickets
[#167071844](https://www.pivotaltracker.com/story/show/167071844)

#### Checklist

### DIFF
--- a/app/Http/Controllers/Web/AuthController.php
+++ b/app/Http/Controllers/Web/AuthController.php
@@ -148,6 +148,14 @@ class AuthController extends BaseController
         // If we had stored a destination name, reset it.
         session()->pull('destination');
 
+        // If the user has 2FA enabled, prompt for a code:
+        if ($user->totp) {
+            $this->auth->guard('web')->logout();
+            session(['totp.user' => $user->id]);
+
+            return redirect('/totp');
+        }
+
         return redirect()->intended('/');
     }
 
--- a/app/Http/Controllers/Web/AuthController.php
+++ b/app/Http/Controllers/Web/AuthController.php
@@ -137,7 +137,8 @@ class AuthController extends BaseController
 
         // Attempt to log in the user to Northstar!
         $credentials = $request->only('username', 'password');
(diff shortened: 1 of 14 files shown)


app/Http/Controllers/Web/AuthController.php
```diff
@@ -148,6 +148,15 @@ public function postLogin(Request $request)
         // If we had stored a destination name, reset it.
         session()->pull('destination');
 
+        // If the user has 2FA enabled, redirect them to the code
+        // verification form to complete their authentication:
+        if ($user->totp) {
+            $this->auth->guard('web')->logout();
```
katiecrane: @DFurnes is there any danger in letting the 2FA users exist as "logged in" briefly just before this point?

DFurnes: Hmm, good flag! We definitely would run into trouble if something exploded in between those two steps... so it's probably safer to separate out the actual "login" step from the `attempt()` call above. Lemme see what I can do...

DFurnes: Okay, fixed that up in 128f01b & then added two more quick cleanup commits afterwards!

PR #894, opened 2019-07-08
AuthController cleanup!

DFurnes:
#### What's this PR do?
This pull request includes some more cleanup commits after I peeked into this controller in #891. 

Alongside some less surprising readability tweaks, I decided to use the `Auth` facade rather than the injected factory. 🙀 I think this ultimately makes things a little easier to read & Laravel has added some code hints that make jump-to-definition work better for these (my original concern way back when).

I'd strongly recommend reviewing this one commit-by-commit for full context!

#### Relevant Tickets
N/A

#### Checklist

### DIFF
--- a/app/Http/Controllers/Web/AuthController.php
+++ b/app/Http/Controllers/Web/AuthController.php
@@ -255,6 +255,11 @@ class AuthController extends Controller
      */
     protected function cleanupSession()
     {
-        session()->forget('destination', 'title', 'callToAction', 'coverImage', 'source_detail');
+        $keys = [
+            'authorize_client_id', 'destination', 'title',
+            'callToAction', 'coverImage', 'source_detail'
+        ];
+
+        session()->forget($keys);
     }
 }
--- a/app/Http/Controllers/Web/AuthController.php
+++ b/app/Http/Controllers/Web/AuthController.php
@@ -109,19 +109,18 @@ class AuthController extends Controller
      */
     public function postLogin(Request $request)
     {
-        $this->validate($request, [
+        $credentials = $this->validate($request, [
             'username' => 'required',
             'password' => 'required',
         ]);
 
-        // Check if that user needs to reset their password in order to log in.
+        // Check if the user needs to reset their password in order to log in:
         $user = $this->registrar->resolve(['username' => $request['username']]);
         if ($user && ! $user->hasPassword()) {
             return back()->withInput()->with('request_reset', true);
         }
 
-        // Attempt to log in the user to Northstar!
-        $credentials = $request->only('username', 'password');
+        // Are the given credentials valid?
         if (! Auth::validate($credentials)) {
             return back()->withInput()->withErrors(['username' => 'These credentials do not match our records.']);
         }
@@ -152,9 +151,9 @@ class AuthController extends Controller
      */
     public function getLogout(Request $request)
     {
-        // A custom post-logout redirect can be specified with `/logout?redirect=`
+        // A custom post-logout redirect can be specified with `/logout?redirect=`.
+        // If not provided (or not for a "safe" domain), redirect to the login form.
         $redirect = $request->query('redirect');
-
         if (! $redirect || ! is_dosomething_domain($redirect)) {
             $redirect = 'login';
         }
@@ -242,9 +241,7 @@ class AuthController extends Controller
      */
     public function getCallback()
     {
-        $user = Auth::user();
-
-        return view('auth.callback', compact('user'));
+        return view('auth.callback', ['user' => Auth::user()]);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/Web/AuthController.php
```diff
@@ -118,16 +118,13 @@ public function postLogin(Request $request)
         // Check if that user needs to reset their password in order to log in.
         $user = $this->registrar->resolve(['username' => $request['username']]);
         if ($user && ! $user->hasPassword()) {
-            return back()->withInput($request->only('username'))->with('request_reset', true);
+            return back()->withInput()->with('request_reset', true);
```
katiecrane: Is password excluded here because we already know it doesn't exist?

app/Http/Controllers/Web/AuthController.php
```diff
@@ -118,16 +118,13 @@ public function postLogin(Request $request)
...
 
         // Attempt to log in the user to Northstar!
         $credentials = $request->only('username', 'password');
-        $validCredentials = Auth::validate($credentials);
-        if (! $validCredentials) {
-            return back()
-                ->withInput($request->only('username'))
-                ->withErrors(['username' => 'These credentials do not match our records.']);
+        if (! Auth::validate($credentials)) {
+            return back()->withInput()->withErrors(['username' => 'These credentials do not match our records.']);
```
katiecrane: I'm not sure I see where the default behavior of not including `password` is coming from. In the Laravel docs here under ["Flashing Input Then Redirecting"](https://laravel.com/docs/5.5/requests#old-input) it seems to imply that the password _will_ be passed since in one of the examples it gets manually excluded.

app/Http/Controllers/Web/AuthController.php
```diff
@@ -56,17 +56,20 @@ public function getAuthorize(ServerRequestInterface $request, ResponseInterface
...
-        session(['authorize_client_id' => request()->query('client_id')]);
-
         if (! Auth::check()) {
             $authorizationRoute = request()->query('mode') === 'login' ? 'login' : 'register';
 
+            // Store any context we'll need for login or registration in the session.
+            // NOTE: Be sure to clear these in AuthController@cleanupSession afterwards!
             session([
+                // Store the Client ID so we can set user's source field on registrations:
+                'authorize_client_id' => $client->getIdentifier(),
```
katiecrane: This value is not cleared out in `AuthController@cleanupSession` and I thought it would be based on the note. Is it supposed to be cleared out or should the note be clarified?

app/Http/Controllers/Web/AuthController.php
```diff
@@ -214,17 +214,18 @@ public function postRegister(Request $request)
...
-            $feature_flags['badges'] = (rand(1, 10) < 8);
+                // Give 70% users the badges flag (1-7), 30% in control (8-10)
+                $feature_flags['badges'] = (rand(1, 10) < 8);
+
+                $user->feature_flags = $feature_flags;
+            }
+        });
 
-            $user->feature_flags = $feature_flags;
-        }
```
katiecrane: This seems like a better spot for this but I'm not sure what you mean by `"customizer" callback`?

app/Http/Controllers/Web/AuthController.php
DFurnes: Ahh. I was thinking of [this exception list](https://github.com/laravel/framework/blob/85b2a5218088b0f60bf6e8c3a62891ba275a7269/src/Illuminate/Foundation/Exceptions/Handler.php#L66-L74), but that's only used for validation exceptions! I was tricked when testing this because we don't expose the `old('password')` in the login form like we do for username, but it does end up in the session!

Reverted this change in 98a6550.

app/Http/Controllers/Web/AuthController.php
DFurnes: Reverted this change in 98a6550!

app/Http/Controllers/Web/AuthController.php
DFurnes: Oh, good catch! That 100% should be cleared. 👍

app/Http/Controllers/Web/AuthController.php
DFurnes: Ah sorry! That's just the name I gave [this callback](https://github.com/DoSomething/northstar/blob/99cf91ed406eb53c99fba815531d01749cc3275e/app/Auth/Registrar.php#L197-L199) that lets us modify the user before the Eloquent model is saved to the database.

app/Http/Controllers/Web/AuthController.php
DFurnes: This is cleared up in b88202a.

app/Http/Controllers/Web/AuthController.php
katiecrane: Ah got it! Thanks!

PR #895, opened 2019-07-09
Additional cleanup!

DFurnes:
#### What's this PR do?
This pull request includes more cleanup from a general once-over of this application.

I'd strongly recommend going commit-by-commit, since GitHub's diff otherwise makes this look scarier than it actually is (because a few files have swapped locations with each other).

#### Relevant Tickets
N/A

#### Checklist


app/Http/Controllers/Legacy/MergeController.php
```diff
@@ -1,11 +1,12 @@
 <?php
 
-namespace Northstar\Http\Controllers;
+namespace Northstar\Http\Controllers\Legacy;
```
mendelB: Are these deemed Legacy for any specific reason? (I'm thinking that afaik we're still using the `MergeController` from Aurora, no?)

Gemfile
```diff
@@ -1,5 +0,0 @@
-# Gemfile
-source "https://rubygems.org"
```
mendelB: I was low-key always happy to see some Ruby in our app 😢

app/Http/Controllers/Legacy/MergeController.php
DFurnes: Yeah, MergeController should probably be in the "main" group. I had just moved it in here since it was alongside the other v1 controllers, which are (still in use as well) but solidly deprecated.

katiecrane: Same question from me - this is the only `MergeController` so I think it should maybe be pulled out of `Legacy`?

katiecrane: Whoops more than 2 minutes since my last refresh!

DFurnes: Oh, although it uses the legacy `UserTransformer`, so I take that back! Let's keep this in the "legacy" group so that stays consistent & we can simply copy it over to the `v2/` group (and use the new `UserTransformer`) when we next touch this endpoint.

PR #901, opened 2019-07-22
Add 'northstar:id' command.

DFurnes:
#### What's this PR do?
This pull request adds a `northstar:id` command that can be used to resolve user IDs from a CSV based on an indexed field on the users collection (such as `email` or `mobile`). For example:

```
$ cat example.csv
email
test@dosomething.org
admin@dosomething.org
sporer.winston@example.org
haylee.buckridge@example.org
aniya.carroll@example.org

$ cat example.csv | php artisan northstar:id
email,id
test@dosomething.org,5d36309ffdce2742ff6c64d2
admin@dosomething.org,5d3630a0fdce2742ff6c64d3
sporer.winston@example.org,5d3630a0fdce2742ff6c64d4
haylee.buckridge@example.org,5d3630a0fdce2742ff6c64d5
aniya.carroll@example.org,5d3630a0fdce2742ff6c64d6
```

I'm experimenting with using standard input to provide the CSV to the Artisan command (which should be [supported by `heroku run`](https://devcenter.heroku.com/articles/one-off-dynos#formation-dynos-vs-one-off-dynos)) because that means we won't need to put these CSVs in an S3 bucket in order to process them.

#### How should this be reviewed?
Give it a look! I'll see if I can write up a test before merging this.

#### Relevant Tickets
[#167327823](https://www.pivotaltracker.com/story/show/167327823)

#### Checklist
- [ ] Documentation added for changed endpoints.
- [ ] Tests added for new features/bug fixes.
- [ ] Post a message in #api if this includes something that causes a rebuild!

PR #902, opened 2019-07-29
Update our "delete user" command.

DFurnes:
#### What's this PR do?
This pull request makes some updates to our "delete user" command, now that we'll be running it more often for account deletion requests:
-  I've renamed it to `northstar:delete` since it's not GDPR-specific anymore.
- I added the ability to pipe in a CSV (like in #901), so we don't need temporary storage.
- I've also updated it to remove _all_ fields (unless whitelisted) and set a `deleted_at` flag.
- Finally, it will now delete each user's profile in Customer.io (including all events).

I added a test for this command in b5b15f9.

#### Relevant Tickets
[#167327865](https://www.pivotaltracker.com/story/show/167327865)

#### Checklist

### DIFF
--- a/app/Console/Commands/DeleteUsersCommand.php
+++ b/app/Console/Commands/DeleteUsersCommand.php
@@ -74,5 +74,7 @@ class DeleteUsersCommand extends Command
 
             $this->info('Deleted: '.$user->id);
         }
+
+        $this->info('Done!');
     }
 }
--- a/app/Console/Commands/DeleteUsersCommand.php
+++ b/app/Console/Commands/DeleteUsersCommand.php
@@ -34,12 +34,14 @@ class DeleteUsersCommand extends Command
         $csv = Reader::createFromString($input);
         $csv->setHeaderOffset(0);
 
-        $this->line('Anonymized '.count($csv).' users...');
+        $this->line('Anonymizing '.count($csv).' users...');
 
         foreach ($csv->getRecords() as $record) {
-            $user = User::find($record[$this->option('id_column')]);
+            $id = $record[$this->option('id_column')];
+            $user = User::find($id);
+
             if (! $user) {
-                $this->warn('Skipping: '.$record[$this->option('id_column')]);
+                $this->warn('Skipping: '.$id);
                 continue;
             }
 
--- a/app/Console/Commands/DeleteUsersCommand.php
+++ b/app/Console/Commands/DeleteUsersCommand.php
@@ -59,6 +59,10 @@ class DeleteUsersCommand extends Command
                 $user->drop($fields);
             }
 
+            // Set a flag so we know this user was deleted:
+            $user->deleted_at = now();
+            $user->save();
+
             // Delete refresh tokens to end any active sessions:
             $token = RefreshToken::where('user_id', $user->id)->delete();
 
--- a/app/Console/Commands/DeleteUsersCommand.php
+++ b/app/Console/Commands/DeleteUsersCommand.php
@@ -7,14 +7,14 @@ use Northstar\Models\User;
 use Illuminate\Console\Command;
 use Northstar\Models\RefreshToken;
 
-class AnonymizeUserCommand extends Command
+class DeleteUsersCommand extends Command
 {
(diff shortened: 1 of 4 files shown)


app/Models/Model.php
```diff
@@ -125,4 +125,20 @@ public function getChanged()
...
         return $changed;
     }
+
+    /**
+     * Save the model to the database, without model events.
+     * @see https://stackoverflow.com/a/51301753
+     *
+     * @param array $options
+     */
+    public function saveQuietly(array $options = [])
```
katiecrane: What a cool helper!!

app/Console/Commands/DeleteUsersCommand.php
```diff
@@ -0,0 +1,78 @@
...
+
+            // Delete refresh tokens to end any active sessions:
+            $token = RefreshToken::where('user_id', $user->id)->delete();
+
+            // And finally, delete the user's profile in Customer.io:
+            $customerIo->deleteUser($user);
+
+            $this->info('Deleted: '.$user->id);
+        }
+    }
```
katiecrane: Would it be helpful to print a final "Done anonymizing users" at the end so it's easy to tell if the script is done vs. stuck?

tests/Console/DeleteUsersCommandTest.php
```diff
@@ -0,0 +1,50 @@
...
+        $this->assertAnonymized($user1);
+        $this->assertAnonymized($user2);
+    }
+
+    /**
+     * Assert that the given model has been anonymized.
+     *
+     * @param User $user
+     */
+    protected function assertAnonymized($user)
```
katiecrane: Would it be worth checking the birthday anonymization (I'm being told I've spelled this wrong but this is how I'm seeing it spelled elsewhere!) functionality in here?

app/Console/Commands/DeleteUsersCommand.php
DFurnes: That's a good idea! 👍

tests/Console/DeleteUsersCommandTest.php
DFurnes: Mmm, good idea!

app/Console/Commands/DeleteUsersCommand.php
DFurnes: Added this in 92264633223b9ae0fed66a2302cae48c33b42947.

tests/Console/DeleteUsersCommandTest.php
DFurnes: Added some assertions for this in aa63acd56dfd8e5cadb0050698dcc745913d17d9!

app/Console/Commands/DeleteUsersCommand.php
```diff
@@ -0,0 +1,80 @@
...
+     * @var string
+     */
+    protected $description = 'Delete (anonymize) users, given a CSV of IDs.';
+
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle(CustomerIo $customerIo)
```
katiecrane: Ah sorry just thought of this. This is a pretty destructive command, worth adding a `confirm` or overkill?

DFurnes: Hmmm, yeah that's a good point. Lemme add!

DFurnes: Hmm, it may actually be hard to do this since we're using STDIN to get the command's input. Because of that, we can't prompt the user interactively (because there's no way for the command to differentiate the piped input from a user's action AFAIK)...

katiecrane: Oh hmmm well I guess since all the users to delete have to be in the csv we have pretty minimal chance of something bad happening I'd say

DFurnes: Yeah... it'd still be nice to be able to double-check that we don't accidentally run it on the wrong environment. But alas! Trade-offs!

PR #903, opened 2019-07-31
Update User's Voter Registration Status field validation

mendelB:
#### What's this PR do?
This PR updates some of the validation rules around the `User->voter_registration_status` field. 

- https://github.com/DoSomething/northstar/commit/0c9b9ef93fef4319c9665559538d04a74ff77a47 adds `unregistered` to the `Registerer` validation. 
- https://github.com/DoSomething/northstar/commit/b4f8332a5ca45eca648a2f085793abbb9553132c adds validation for the `voter_registration_status` field for requests from the web registration form to ensure we only accept the expected values.

- we snuck in https://github.com/DoSomething/northstar/pull/903/commits/4236a456ffaf52d1cc6f7e4b8e388ffa72ff91b4 to allow the `sms_status` field to be `nullable` since that's been a todo for a while https://github.com/DoSomething/aurora/pull/219#issuecomment-518291913

🌶And most notably and controversially: https://github.com/DoSomething/northstar/commit/a3df6eff3f0b41885736af4839960921796eb5df is a proposed solution to enabling setting the voter registration status field arbitrarily instead of being tied to the hierarchy. (Specifically towards the effort of enabling admins to edit the `voter_registration_status` field in Aurora [Pivotal ID#165974120](https://www.pivotaltracker.com/story/show/165974120))

This would extract the logic to validate a status against the hierarchy into a separate method on the User model so the automated batch updating scripts like the one we use in [Chompy](https://github.com/DoSomething/chompy/blob/e5a36f7a218a67074ca7209ede1344b3edd6d098/app/Jobs/ImportRockTheVoteRecord.php) can still validate the new status using this method.

[Pivotal ID #160807904](https://www.pivotaltracker.com/n/projects/2019429/stories/160807904) has some context for why we added this validation

❗️One caveat with removing this from the mutator is that previously, a status other than the ones in the hierarchy list would also be invalid, though I don't _think_ that was our intention with this method anyways. But we can, of course, re-add the mutator with such a validation specifically. And either way, I believe API update requests shouldn't make it past [the Registerers validation](https://github.com/DoSomething/northstar/blob/20920964c04a8c67888388dd8b9088a92e808b4e/app/Auth/Registrar.php#L69) to begin with!

❓Another question would be if this method were better off being defined within Chompy itself or some other better place as opposed to on the User model? _Yes (See https://github.com/DoSomething/northstar/pull/903#discussion_r309347379)_

#### Relevant Tickets
[Pivotal ID#165974120](https://www.pivotaltracker.com/story/show/165974120)
A [Slack conversation](https://dosomething.slack.com/archives/C02BBRN5A/p1564162711003300) with some more context on this.

#### Checklist


app/Models/User.php
```diff
@@ -523,27 +523,25 @@ public function sendPasswordReset($token, $type)
...
 
     /**
-     * Set the voter_registration_status based on the hierarchy.
+     * Validates provided status based on current status and hierarchy
+     * returning the new status if valid, or the current status otherwise.
      *
      * @param string $status
      */
-    public function setVoterRegistrationStatusAttribute($status)
+    public function validateHierarchicalVoterRegistrationStatus($status)
```
katiecrane: Doesn't look like this is being used anywhere yet! If this isn't going to be called in Northstar I guess the only option is to extract it into Chompy? This hierarchy has to be enforced on Chompy uploads somehow. I definitely see the problem with this being enforced on ALL `voter_registration_status` updates in Northstar, but I also feel like Chompy shouldn't be the keeper of this "Northstar-related business logic." Or maybe it's "voter registration upload business logic" and it _should_ live in Chompy 🤔 What do you think @mendelB ?

mendelB: Hmm I kinda like your latter point!
> Or maybe it's "voter registration upload business logic" and it should live in Chompy 

If we're indeed cool with Northstar itself allowing arbitrary updates to the voter reg status vis a vis the hierarchy, but we do want _uploads_ to ensure they don't push someone down the hierarchy, then I do kinda feel like it's reasonable to make this a Chompy concern.

mendelB: Ooh and it looks like that's how we deal with hierarchy validation related to updating the post on Chompy!

https://github.com/DoSomething/chompy/blob/e5a36f7a218a67074ca7209ede1344b3edd6d098/app/Jobs/ImportRockTheVoteRecord.php#L304-L326

katiecrane: Oh rad!! I guess we don't need this at all anymore in Northstar then?

mendelB: Ok, I'll 🔥it! I'll make a separate PR in Chompy to add that logic there and will wait to merge this one until that's deployed.

katiecrane: Wait I thought it already is added in Chompy in the link you provided? https://github.com/DoSomething/chompy/blob/e5a36f7a218a67074ca7209ede1344b3edd6d098/app/Jobs/ImportRockTheVoteRecord.php#L304-L326

mendelB: Hmm I was under the impression that that method relates specifically to set the status on posts https://github.com/DoSomething/chompy/blob/e5a36f7a218a67074ca7209ede1344b3edd6d098/app/Jobs/ImportRockTheVoteRecord.php#L231-L234

That being said, I'm realizing that we could update that method to also work for the User status as well! So I think it'll still require an update but a smaller one than I initially figured!

mendelB:
https://github.com/DoSomething/northstar/pull/903/commits/0d0c204dde64730dbd280c3be577c739832ef3cb I snuck in a quick additional update to allow the Registerers `voter_registration_status` field to be `null`. This ensures that on forms like Aurora if this field is unset, to begin with, a staff member can still submit the form without a validation error.

app/Models/User.php
katiecrane: Ah I see okay I guess we can probably re-use that same chunk for setting it on the user?

katiecrane:
@mendelB It shouldn't hurt to allow this field to be null, but I want to note that `null` values in Northstar get unset so it'll be like it never existed. I'm wondering if it could be better to only pass on that part of the request if it was set?

app/Models/User.php
mendelB: Yes, I believe so! I'll test it all out in Chompy

mendelB:
> @mendelB It shouldn't hurt to allow this field to be null, but I want to note that `null` values in Northstar get unset so it'll be like it never existed. I'm wondering if it could be better to only pass on that part of the request if it was set?

Ah ok, and we would _not_ want to allow folks to unset this field? TBF I can't really think of a common scenario where we might practically need to, but maybe just for convenience sake... would it necessarily be a bad thing to allow unsetting this field via the API?

katiecrane:
@mendelB I don't think it's bad! Just a lil food for thought

PR #904, opened 2019-07-31
Restrict Tinker to local development & update dependencies.

DFurnes:
#### What's this PR do?
This pull request makes [`laravel/tinker`](https://github.com/laravel/tinker) a development dependency, so it will no longer be available on production environments. While being able to easily open a REPL can be really helpful, it's also dangerous since it allows developers to execute arbitrary code on production.

You can confirm that Laravel Tinker is not installed or accessible when installing dependencies with `composer install --no-dev` (like our Heroku build process does), or just wait for this to automatically deploy & check there!

#### Relevant Tickets
[#167597101](https://www.pivotaltracker.com/story/show/167597101)

#### Checklist


config/app.php
```diff
@@ -171,7 +171,6 @@
         DoSomething\Gateway\Laravel\GatewayServiceProvider::class,
         DoSomething\StatHat\StatHatServiceProvider::class,
         Laravel\Socialite\SocialiteServiceProvider::class,
-        Laravel\Tinker\TinkerServiceProvider::class,
```
DFurnes: If installed, this provider is [automatically discovered by Laravel](https://github.com/laravel/tinker/blob/eb0075527fdeeb1cc1d68bd4ca7d50256b30a827/composer.json#L38-L40) & so that's why this line can be removed.

katiecrane:
I have questions about this! I use tinker fairly regularly (on QA!) sometimes to:
- set fields for testing (no worries here as I can make sure I can make my updates via Aurora)
- check field values (can also do via Aurora)
- get counts of certain fields (to check on script progress and make sure fields are being set as expected), for example:
`User::whereRaw(['voter_registration_status' => 'unregistered'])->count()`

Is there another way to run queries like that?

DFurnes:
Yeah, it's definitely a handy tool to have! I've done the same things. For the queries you mentioned, you should be able to use one of our `readonly` database accounts (which are automatically provisioned for each of our RDS databases, and manually configured for Atlas).

And I'm definitely down for continuing to think about how we can balance safety & convenience!

katiecrane:
@DFurnes I remember not being able to run queries on Atlas for an unknown reason (I remember we couldn't figure out why it didn't like the syntax or something like that?) so switching to tinker instead. But I just tried a query on Atlas and it worked!

I don't thiiiiink there is a way to get `counts` though?

DFurnes:
Yeah, you'd still have to run counts via the `mongo` CLI, I believe. The `nsreadonly` user credentials for those instances can be found on Lastpass, and you can get the connection string for a given instance from the "Connect" button on Atlas.

katiecrane:
AH got it! I'll have to make sure I'm all set up, thanks Dave!

PR #907, opened 2019-08-08
Add optional fields proof-of-concept.

DFurnes:
#### What's this PR do?
This PR implements "optional fields" in Northstar, from the [Access Control & Auditing](https://docs.google.com/document/d/1OclKWYEtjo0LTI9DzKaVG1dEBY6kfAEEgYZVRzSlD7s/edit) spec. Once this feature is shipped, it will remove fields marked as "sensitive" from the base API response, and allow them to be optionally requested with the `?include=…` query parameter.

For example, if we wanted to display a user's street address in Aurora, we might request:

```
v2/users/5571f4f5a59dbf3c7a8b4569?include=addr_street1,addr_street2
```

When a field is included in this way, we can then create a paper trail (in Papertrail!) for any requests that include more sensitive user information.

This work is behind a `DS_OPTIONAL_FIELDS` feature flag, so can be safely merged.

~I'll add some tests before marking this as "ready to review".~ Done! I also ended up adding two custom assertions, `seeJsonField` and `dontSeeJsonField`, to make it easier to test whether a particular field is included in the JSON response.

#### Relevant Tickets
[#167596587](https://www.pivotaltracker.com/story/show/167596587)

#### Checklist

### DIFF
--- a/app/Http/Transformers/BaseTransformer.php
+++ b/app/Http/Transformers/BaseTransformer.php
@@ -1,0 +1,72 @@
+<?php
+
+namespace Northstar\Http\Transformers;
+
+use League\Fractal\Scope;
+use Northstar\Models\User;
+use Illuminate\Support\Str;
+use Illuminate\Support\Facades\Gate;
+use League\Fractal\TransformerAbstract;
+
+class BaseTransformer extends TransformerAbstract
+{
+    /**
+     * Resources that can be included if requested.
+     *
+     * @return array
+     */
+    public function getAvailableIncludes()
+    {
+        return static::$model::$sensitive;
+    }
+
+    /**
+     * Set the resources that will be included by default.
+     *
+     * @return array
+     */
+    public function getDefaultIncludes()
+    {
+        // If we've enabled the "optional fields" feature flag, then any fields
+        // containing sensitive information must be explicitly requested:
+        if (config('features.optional_fields')) {
+            return [];
+        }
+
+        // Until then, include everything!
+        return $this->getAvailableIncludes();
+    }
+
+    /**
+     * Load any resources specified by the `?include=` query parameter and
+     * automatically register `includeFieldName` handlers for optional
+     * fields on the User model, with authorization & logging.
+     *
+     * @param Scope  $scope
+     * @param string $attribute
+     * @param mixed  $resource
+     *
+     * @throws \Exception
+     *
+     * @return \League\Fractal\Resource\ResourceInterface
+     */
+    protected function callIncludeMethod(Scope $scope, $attribute, $resource)
+    {
+        // Is this a sensitive attribute? If so, check authorization & log access:
+        if (in_array($attribute, static::$model::$sensitive)) {
+            if (! $this->authorize($resource, $attribute)) {
+                return null;
+            }
+
+            // @TODO: Log access to this optional field.
+            // ...
+        }
+
+        // If we don't have a custom "include" method, try default resolver:
+        if (! method_exists($this, 'include'.Str::studly($attribute))) {
+            return $this->primitive($resource->{$attribute});
+        }
+
+        return parent::callIncludeMethod($scope, $attribute, $resource);
+    }
+}
--- a/app/Http/Transformers/BaseTransformer.php
+++ b/app/Http/Transformers/BaseTransformer.php
@@ -56,8 +56,11 @@ class BaseTransformer extends TransformerAbstract
             return null;
         }
 
-        // @TODO: Log access to this optional field.
-        // ...
+        // Log access to this optional field:
+        if (config('features.optional-fields')) {
+            // @TODO: Can we batch up all the included fields in a single log message?
+            info('Sensitive field viewed for '.$resource->id, ['attribute' => $attribute]);
+        }
 
         // If we don't have a custom "include" method, try default resolver:
         if (! method_exists($this, 'include'.Str::studly($attribute))) {
--- a/app/Http/Transformers/BaseTransformer.php
+++ b/app/Http/Transformers/BaseTransformer.php
@@ -28,7 +28,7 @@ class BaseTransformer extends TransformerAbstract
     {
         // If we've enabled the "optional fields" feature flag, then any fields
         // containing sensitive information must be explicitly requested:
-        if (config('features.optional_fields')) {
+        if (config('features.optional-fields')) {
             return [];
         }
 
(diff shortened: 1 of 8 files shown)


tests/BrowserKitTestCase.php
```diff
@@ -86,6 +87,47 @@ public function withHeader($name, $value)
...
     }
 
+    /**
+     * Assert that the JSON response does not have the given field.
+     *
+     * @param  array|null  $structure
+     * @param  array|null  $responseData
+     * @return $this
+     */
+    public function dontSeeJsonField($path)
```
mendelB: I think there's a mismatch with the `@params` declared above? Also, by `$path` we mean the JSON path to the response field we're testing right? That confused me for a second as I thought it might be the $path we're querying the JSON from, but I think that was just me being silly.

tests/Http/UserTest.php
```diff
@@ -168,6 +168,48 @@ public function testV2GetAllDataFromUserAsAdmin()
         ]);
     }
 
+    /**
+     * Test that staffers can request optional fields.
```
mendelB: I think this comment needs to be updated for the Normal user case 👾

app/Http/Transformers/BaseTransformer.php
```diff
@@ -0,0 +1,71 @@
...
+     * @param mixed  $resource
+     *
+     * @throws \Exception
+     *
+     * @return \League\Fractal\Resource\ResourceInterface
+     */
+    protected function callIncludeMethod(Scope $scope, $attribute, $resource)
+    {
+        // Is this a sensitive attribute? If so, check authorization & log access:
+        if (in_array($attribute, static::$model::$sensitive)) {
```
mendelB: Would it make sense to use the `getAvailableIncludes` method instead of referring directly to the `static::$model::$sensitive` or are these technically separate concerns?

app/Http/Transformers/UserTransformer.php
```diff
@@ -17,25 +41,16 @@ public function transform(User $user)
...
         if (Gate::allows('view-full-profile', $user)) {
-            $response['last_name'] = $user->last_name;
-        }
-
-        $response['last_initial'] = $user->last_initial;
-        $response['photo'] = null;
-
-        if (Gate::allows('view-full-profile', $user)) {
-            $response['email'] = $user->email;
-            $response['mobile'] = format_legacy_mobile($user->mobile);
```
mendelB: Do we need an `includeMobile` method to invoke the `format_legacy_mobile` on the field?

tests/BrowserKitTestCase.php
DFurnes: Ahh yes there sure is!

> Also, by `$path` we mean the JSON path to the response field we're testing right?

Yeah! Do you think `$key` might be clearer?

app/Http/Transformers/BaseTransformer.php
```diff
@@ -0,0 +1,71 @@
...
+            if (! $this->authorize($resource, $attribute)) {
+                return null;
+            }
+
+            // @TODO: Log access to this optional field.
```
katiecrane: In your write-up you talked about creating a paper trail, is this `@TODO` for this PR or for later?

app/Http/Transformers/BaseTransformer.php
DFurnes: Oh good catch, definitely!

app/Http/Transformers/UserTransformer.php
DFurnes: Ahh, hmm. We _would_, but I actually think using `format_legacy_mobile` was a mistake here, since we want everything to be consistently formatted as E.164 for the API moving forward.

That said, maybe not the best idea to change this & jet right out on vacation, eh? I'll add an `includeMobile` to keep this behavior, with a `@TODO` to fix it next sprint.

app/Http/Transformers/BaseTransformer.php
DFurnes: For later! Since we'll have this feature toggled off for now, I don't want to flood the logs with all the automatically included fields.

tests/BrowserKitTestCase.php
DFurnes: Updated in 89fab30!

tests/Http/UserTest.php
DFurnes: Gah, classic copy-paste! 🤦‍♂ Fixed in e939585.

app/Http/Transformers/BaseTransformer.php
katiecrane: Ah gotcha! I could see it being helpful when testing this to have the log behind the feature flag so you can see what's going on, but I leave it up to you 😃

DFurnes: Oh hmm, that's a good point! Ok, let me add it in! 💡

app/Http/Transformers/BaseTransformer.php
DFurnes: Was able to simplify this a good deal in b267662! 😌

app/Http/Transformers/UserTransformer.php
DFurnes: Updated in 96fc048!

app/Http/Transformers/BaseTransformer.php
DFurnes: Added some logging in f1e834d. I think I might have to re-think where this happens in the future, since this writes a separate message for each requested field which is a lot!

```
[2019-08-09 15:59:43] local.INFO: Sensitive field viewed for 5d36309ffdce2742ff6c64d2 {"attribute":"email"} {"user_id":"5d3630a0fdce2742ff6c64d3","client_id":"dev-oauth","request_id":null}
[2019-08-09 15:59:43] local.INFO: Sensitive field viewed for 5d36309ffdce2742ff6c64d2 {"attribute":"addr_street1"} {"user_id":"5d3630a0fdce2742ff6c64d3","client_id":"dev-oauth","request_id":null}
```

app/Http/Transformers/BaseTransformer.php
```diff
@@ -0,0 +1,72 @@
...
+    {
+        // If we've enabled the "optional fields" feature flag, then any fields
+        // containing sensitive information must be explicitly requested:
+        if (config('features.optional-fields')) {
+            return [];
+        }
+
+        // Until then, include everything!
+        return $this->getAvailableIncludes();
+    }
```
katiecrane: This function seems to return `[]` no matter what?

DFurnes: It's overridden in `UserTransformer`! The `[]` is just intended as a safe default.

katiecrane: AH the dangers of looking at whatever is closest!

PR #908, opened 2019-08-09
Add some computed "preview" fields for admin interfaces.

DFurnes:
#### What's this PR do?
This pull request adds some "preview" fields, as described in the [Access Control & Auditing spec](https://docs.google.com/document/d/1OclKWYEtjo0LTI9DzKaVG1dEBY6kfAEEgYZVRzSlD7s/edit). These will allow applications (like Aurora or Rogue's admin interface) to provide more helpful "contextual" information without exposing the user's full personal data unnecessarily:

For example, here's how the base `users/:id` response might look for a staff user:

```js
{
  "data": {
    "id": "5d36309ffdce2742ff6c64d2",
    "display_name": "Puppet S.",
    "first_name": "Puppet",
    "last_initial": "S",
    "photo": null,
    "email_preview": "pup...@dosomething.org",
    "mobile_preview": "(950) 968-XXXX",
    "age": 29,
   // and so on...
```

It might be easiest to review this PR commit-by-commit. I also added some tests in a109013 demonstrating how these fields work for anonymous & staff viewers.

#### Relevant Tickets
[#167596587](https://www.pivotaltracker.com/story/show/167596587)

#### Checklist

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -4,6 +4,7 @@ use Carbon\Carbon;
 use Illuminate\Support\Str;
 use Northstar\Models\Client;
 use Northstar\Auth\Normalizer;
+use libphonenumber\PhoneNumber;
 use Illuminate\Support\HtmlString;
 use libphonenumber\PhoneNumberUtil;
 use libphonenumber\PhoneNumberFormat;
@@ -303,4 +304,32 @@ function stringify_object($object)
 {
     // e.g. utm_source:test,utm_medium:internet,utm_campaign:uconn_lady_huskies
     return str_replace('=', ':', http_build_query($object, '', ','));
+}
+
+/**
+ * Parse a phone number into a PhoneNumber object.
+ *
+ * @param string $string
+ * @return PhoneNumber
+ */
+function parse_mobile($string): ?PhoneNumber
+{
+    try {
+        $parser = PhoneNumberUtil::getInstance();
+        $number = $parser->parse($string, 'US');
+
+        return $number;
+    } catch (\libphonenumber\NumberParseException $e) {
+        return null;
+    }
+}
+
+/**
+ * Format the given PhoneNumber as a string.
+ */
+function format_mobile(PhoneNumber $number, $format): string
+{
+    $parser = PhoneNumberUtil::getInstance();
+
+    return $parser->format($number, $format);
 }
--- a/tests/Http/UserTest.php
+++ b/tests/Http/UserTest.php
@@ -108,20 +108,24 @@ class UserTest extends BrowserKitTestCase
      */
     public function testV2GetPublicDataAsAuthenticatedUser()
     {
-        $user = factory(User::class)->create();
+        $user = factory(User::class)->create([
+            'first_name' => 'Puppet',
+            'last_name' => 'Sloth',
+        ]);
         $viewer = factory(User::class)->create();
 
         // Test that we can view public profile as another user.
         $this->asUser($viewer, ['user', 'user:admin'])->get('v2/users/'.$user->id);
         $this->assertResponseStatus(200);
+        $this->seeJsonField('data.first_name', 'Puppet');
+        $this->seeJsonField('data.display_name', 'Puppet S.');
 
         // And test that private profile fields are hidden for the other user.
-        $data = $this->decodeResponseJson()['data'];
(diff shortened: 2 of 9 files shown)


DFurnes:
Hmm, all the tests are passing on my local! (Famous last words.)

I'll have to look into what could be going on here post-vacay.

DFurnes:
Ah, it turns out [imap_rfc822_parse_adrlist](https://www.php.net/manual/en/function.imap-rfc822-parse-adrlist.php) requires the [IMAP](https://www.php.net/manual/en/book.imap.php) PHP extension. To avoid requiring a heavy & otherwise unused extension, I'm going to try this [`email-parse`](https://github.com/mmucklo/email-parse) package instead.

app/helpers.php
```diff
@@ -304,3 +305,31 @@ function stringify_object($object)
...
     return str_replace('=', ':', http_build_query($object, '', ','));
 }
+
+/**
+ * Parse a phone number into a PhoneNumber object.
+ *
+ * @param string $string
+ * @return PhoneNumber
+ */
+function parse_mobile($string): ?PhoneNumber
```
DFurnes: This includes two "new" PHP features that I'm really excited about, [return types](https://www.php.net/manual/en/migration70.new-features.php#migration70.new-features.return-type-declarations) and [nullable types](https://www.php.net/manual/en/migration71.new-features.php#migration71.new-features.nullable-types). This allows us to define and enforce the type of object this function returns (rather than just a comment that might get out of sync).

katiecrane: Whoa I was just reading about this concept but was like "I don't think PHP has this" but lo and behold!!

tests/Http/UserTest.php
```diff
@@ -132,18 +136,27 @@ public function testV2GetPublicDataAsAuthenticatedUser()
...
         // Check that public & private profile fields are visible
-        $this->seeJsonStructure([
-            'data' => [
-                'id', 'email', 'first_name', 'last_name', 'facebook_id',
-            ],
-        ]);
+        $this->seeJsonField('data.first_name', 'Puppet');
+        $this->seeJsonField('data.display_name', 'Puppet S.');
+        $this->seeJsonField('data.last_name', 'Sloth');
+        $this->seeJsonField('data.email', 'puppet.sloth@dosomething.org');
```
katiecrane: @DFurnes I've looked back through https://github.com/DoSomething/northstar/pull/907 and also this PR and I'm confused about how "email" and the other sensitive fields are included in any responses. It's not explicitly listed in the transformer anymore, but is in `$sensitive` on the user, but I think I'm confused about how that is actually being used? I'm missing a link somewhere

DFurnes: Yeah, it's kinda tucked away now which is definitely a downside.

The optional fields are listed by [`getAvailableIncludes`](https://github.com/DoSomething/northstar/blob/de922204853beea37e58e77f1563d3901eef90be/app/Http/Transformers/UserTransformer.php#L10-L18) (which is how [Fractal](https://fractal.thephpleague.com/) decides what resources can be "included" with a response). This is where we hook the transformer up to that "sensitive" field on the user.

Usually, it would then look for a `includeFieldName` method for any available & requested includes (like we do for [includes in Rogue](https://github.com/DoSomething/rogue/blob/7862a4a6ef5f787678de07f1e2c00e1f3852996b/app/Http/Transformers/PostTransformer.php#L77-L114)) but that would get super tedious with how many optional fields we have here!

So instead, for fields like `email`, we include [a "default resolver"](https://github.com/DoSomething/northstar/blob/de922204853beea37e58e77f1563d3901eef90be/app/Http/Transformers/BaseTransformer.php#L65-L68) that just reads that field directly from the model if a custom `includeFieldName` method isn't set for that field. This means we only need to define [custom include methods](https://github.com/DoSomething/northstar/blob/de922204853beea37e58e77f1563d3901eef90be/app/Http/Transformers/UserTransformer.php#L28-L47) when there's something special happening (like ISO date formatting).

DFurnes: I added a comment summarizing this in 124b15628a7631f5e85d98f741780b77da57854e so hopefully it's more clear in the future!

katiecrane: Ahhh I think the `callIncludeMethod()` is the piece I was missing, thanks Dave!

PR #912, opened 2019-09-13
Update user model

lindsayrmaher:
#### What's this PR do?
This PR adds the causes array to the user, adds it to the transformer to define what acceptable causes can be passed and adds the information to send to customer.io! Also, minimal testing added :)

Test to make sure customer.io information is set correctly per cause and that the updates follow a similar pattern to newsletter subscription topics!!

#### Relevant Tickets
[Pivotal Ticket #168008868](https://www.pivotaltracker.com/story/show/168008868)

#### Checklist

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -59,7 +59,7 @@ use Northstar\Jobs\SendPasswordResetToCustomerIo;
  *
  * 
  * Causes and Interests
- * @property array $user_causes_and_interests
+ * @property array $causes
  *
  * Fields for Make a Voting Plan
  * @property string $voting_plan_status
@@ -98,7 +98,7 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
         'email', 'mobile', 'password', 'role',
 
         // Profile:
-        'first_name', 'last_name', 'birthdate', 'voter_registration_status',
+        'first_name', 'last_name', 'birthdate', 'voter_registration_status', 'causes',
 
         // Address:
         'addr_street1', 'addr_street2', 'addr_city', 'addr_state', 'addr_zip',
@@ -196,6 +196,7 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
         'birthdate' => 'date',
         'sms_paused' => 'boolean',
         'email_subscription_status' => 'boolean',
+        'causes' => 'array',
     ];
 
     /**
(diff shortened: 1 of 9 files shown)


tests/Models/UserTest.php
```diff
@@ -42,6 +42,18 @@ public function it_should_send_new_users_to_blink()
             'lifestyle_email_subscription_status' => isset($user->email_subscription_topics) ? in_array('lifestyle', $user->email_subscription_topics) : false,
             'community_email_subscription_status' => isset($user->email_subscription_topics) ? in_array('community', $user->email_subscription_topics) : false,
             'scholarship_email_subscription_status' => isset($user->email_subscription_topics) ? in_array('scholarships', $user->email_subscription_topics) : false,
+            'animal_welfare' => in_array('animal_welfare', $this->causes) ? true : false,
```
DFurnes: This is erroring out in your test since `$this->causes` is reading from `UserModelTest` (this class) rather than the user. You can swap the test assertion to check against `$user->causes` instead to clear this up.

It might actually make more sense to hard-code some causes (and email subscription topics, for that matter) in the test, so we don't end up duplicating the implementation (`in_array` yadda yadda) in the test. Something like this:

```php
// line 13, above:
    'birthdate' => '1/2/1990',
    'causes' => ['bullying', 'environment'],
]);

// ...and then down here in our assertion:
'bullying' => true,
'education' => false,
'environment' => true,
```

DFurnes: This is a trade-off, of course, since we lose some "coverage" of different possible combinations of subscriptions or causes. However, if there ended up being a bug in our `in_array` or `isset` logic, this would be more likely to help us catch it in the test!

lindsayrmaher: 🤦‍♀ thanks for the help with this error! i was wondering what the better approach would be re: harding coding the values, i think updating to your suggestion makes sense

app/Models/User.php
```diff
@@ -641,4 +655,21 @@ public function setEmailSubscriptionTopicsAttribute($value)
         // Set de-duped array as email_subscription_topics
         $this->attributes['email_subscription_topics'] = array_values(array_unique($value));
     }
+
+    public function getCausesAttribute($value)
```
DFurnes: Can you add a docblock explaining this function? e.g.

```php
/**
 * Accessor for the `causes` attribute.
 *
 * @param  mixed value
 * @return string[]
 */
```

resources/views/profiles/about/edit.blade.php
```diff
@@ -0,0 +1,3 @@
+@extends('layouts.app')
```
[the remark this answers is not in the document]
lindsayrmaher: how does one remove a file from a PR/commit? or should i just remove it on this branch and then push it up from the branch i'm working on this file with locally?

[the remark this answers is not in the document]
DFurnes: Yep! Deleting the file locally, committing the change & pushing up to this branch will work.

PR #914, opened 2019-09-17
Routes profile

lindsayrmaher:
#### What's this PR do?

This PR is doing a lot...

- adds routes for `profile/about` and `profile/subscriptions`
- adds controllers for `about` and `subscription`
- adds a `profile` blade template to make the form experience the more uniform
- starts building out a very generic form (this is still very much a WIP) for the `about page`
- adds some base styling to the `profile` template

It also adjusts the functionality around setting the `-extended` class for the view content in `app.blade.php` which handled adding a cover image and half-width wrapper styling - to now account for a new `$coverImage` variable, and not blanket apply to any `-extended` view. (Differentiating the new full-width forms and the current register form,)

These files should be the base of all the work on the registration pages, and once merged will give us a good jumping off point for the rest of the workflow.

#### Relevant Tickets
[Pivotal Ticket 168009111](https://www.pivotaltracker.com/story/show/168009111)
[Pivotal Ticket 168353804](https://www.pivotaltracker.com/story/show/168353804)

#### Checklist


app/Http/Controllers/Web/ProfileSubscriptionsController.php
```diff
@@ -0,0 +1,40 @@
...
+        return view('profiles.subscriptions.edit', ['user' => auth()->guard('web')->user()]);
+    }
+
+    /**
+     * Handle submissions of the User subscriptions form.
+     *
+     * @param Request $request
+     * @return \Illuminate\Http\RedirectResponse
+     */
+    public function store()
```
mendelB: Should these be `update` instead of `store`?

lindsayrmaher: I suppose if we're using a `patch` instead of `post` that would make more sense?

lindsayrmaher: I had named them `store` after an initial walk through with @weerd, but i can see the case for keeping things consistent

mendelB: My thinking re `patch` and re `update` vs `post` and `store` was that we're only updating the existing user at this point in the onboarding flow since these are essentially just glorified profile editing forms - so I think it feels more appropriate to be updating via `patch` as opposed to storing via `post`.

lindsayrmaher: I think that makes sense! This also makes sense to me considering they could potentially hit this route if/when they want to `update` their profile preferences.

resources/views/profiles/profile.blade.php
```diff
@@ -0,0 +1,12 @@
+@extends('layouts.app', ['extended' => true])
+
+@section('content')
+    {{-- @TODO: add percentage based width to the tailwind config & adjust lg:pl-24 to 1/3 --}}
+    <div class="pt-12 md:pt-24 lg:pt-0 lg:pl-24 " style="background-image: url(@yield('form-image-url'))">
```
DFurnes: Oh man, it's happening!! 🌬

lindsayrmaher: https://gph.is/29PBTDn

PR #915, opened 2019-09-18
Feature flag for Refer A Friend

aaronschachter:
#### What's this PR do?

This PR extends the `feature_flags` User field introduced in #871 to store a new `refer-friends` boolean property, setting it to `true` when a new `DS_REFER_FRIENDS_TEST` config var has been set to `'true'`.

This PR also tweaks some logic introduced in #871: Now that there are two `feature_flags` values to support, they should operate independently from each other (e.g. if `DS_BADGES_TEST` is set to `false`, it shouldn't affect the setting of the `refer-friends` feature flag value based on `DS_REFER_FRIENDS_TEST`).

Upon deploying this to production, we'll make corresponding changes in Phoenix to display the Referral Page Banner in an Affirmation if the user's `feature_flags['refer-friend']` value is set to `true` (replacing our hacky feature-flag check via `refer-friends` query string - refs https://github.com/DoSomething/phoenix-next/pull/1551/files#r316379229)

Test new user registrations locally on this branch:

- Without setting the `DS_REFER_FRIENDS_TEST` config variable, users will only receive a `feature_flags` value of `badges` based on the `DS_BADGES_TEST` config var value, as per current production behavior

- When setting the `DS_REFER_FRIENDS_TEST` config variable to `'true'`, users will receive a `feature_flags` value of `true`. This will happen independently of whatever the `DS_BADGES_TEST` config variable is set to.

#### Relevant Tickets

These requirements were defined in https://www.pivotaltracker.com/story/show/168489148/comments/206716996.

#### Checklist
- [x] Documentation added for changed endpoints. -  See `.env.example`
- [ ] ~~Post a message in #api if this includes something that causes a rebuild!~~ - Not applicable


app/Http/Controllers/Web/AuthController.php
```diff
@@ -221,16 +221,20 @@ public function postRegister(Request $request)
                 $user->source_detail = stringify_object($sourceDetail);
             }
 
+            $feature_flags = $user->feature_flags;
+
             // If the badges test is running, sort users into badges group control group
             // (while ensuring that we completely exclude any 'club' referrals).
```
aaronschachter: Per https://www.pivotaltracker.com/story/show/168489148/comments/206811849 -- this needs to be modified so that any club referrals do not receive the `refer-friends` feature flag  (mirroring what we're doing for the `badges` feature flag). I'll update this PR and comment when ready for final review.

aaronschachter:
I've updated the clubs logic per https://github.com/DoSomething/northstar/pull/915#discussion_r325770996 -- this PR is ready for review.

app/Http/Controllers/Web/AuthController.php
```diff
@@ -221,16 +221,22 @@ public function postRegister(Request $request)
...
+                    // Give 70% users the badges flag (1-7), 30% in control (8-10)
+                    $feature_flags['badges'] = (rand(1, 10) < 8);
+                }
+                // If the refer-friends test is running, give all users the refer-friends flag.
+                if (config('features.refer-friends')) {
+                    $feature_flags['refer-friends'] = true;
+                }
             }
+
+            $user->feature_flags = $feature_flags;
```
mendelB: Can we set this field within the above conditional, since otherwise there'd be no new value?

aaronschachter: Yes, good call!

PR #917, opened 2019-09-24
About controller validation

lindsayrmaher:
#### What's this PR do?
This PR adds backend validation and updates the user with the correct information in their profile. It also updates the form in order to accommodate for the data formatting we expect in the northstar db.

after logging into `northstar.test` or creating a new account, you should be able to navigate to `northstar.test/profile/about`.

You should be able to update 1-3 fields and hit next, which will bring you to the next step of registration. If you choose skip instead, you will also be able to move to the subscriptions step of the reg process. 

You should then be able to check the user in tinker and find the appropriate fields updated. 

#### Relevant Tickets
[pivotal ticket 168009111](https://www.pivotaltracker.com/story/show/168009111)

#### Checklist


app/Http/Controllers/Web/ProfileAboutController.php
```diff
@@ -35,22 +34,33 @@ public function edit()
...
         $user = auth()->guard('web')->user();
 
-        // dd([
-        //     $request,
-        //     $request->all(),
-        //     auth()->guard('web')->user()
-        // ]);
+        $request['birthdate'] = implode('/', $request['birthdate']);
+
+        if (strlen($request['birthdate'] < 3)) {
```
DFurnes: Is this to handle cases where a user doesn't fill in all three fields, but it'd still parse as a date (incorrectly)? If so, let's add a little comment so it's clear for someone returning to this file in the future.

app/Http/Controllers/Web/ProfileAboutController.php
```diff
@@ -35,22 +34,33 @@ public function edit()
...
+            $request['birthdate'] = null;
+        }
+
+        $this->registrar->validate($request, null, [
+            'birthdate' => 'nullable|date|date:now',
```
DFurnes: Should this be `before:now` instead of `date:now`?

app/Http/Controllers/Web/ProfileAboutController.php
```diff
@@ -35,22 +34,33 @@ public function edit()
...
+            'email' => 'email|nullable|unique:users',
+            'mobile' => 'mobile|nullable|unique:users',
+        ]);
+
+        if ($request['birthdate']) {
```
DFurnes: We could use `$user->fill` (and the fields we'd previously set as mass-assignable) to simplify this a bit:

```php
$completedFields = array_filter($request->all());
$user->fill($completedFields);
```

app/Http/Controllers/Web/ProfileAboutController.php
```diff
@@ -35,22 +34,33 @@ public function edit()
...
+            $user->birthdate = $request['birthdate'];
+        }
+
+        if ($request['voter_registration_status']) {
+            $user->voter_registration_status = $request['voter_registration_status'];
+        }
+
+        if ($request['causes']) {
+            $user->causes = array_merge($user->causes, $request['causes']);
```
DFurnes: Do we actually want to merge these with the existing value? I feel like, as a user, I'd expect to see my existing causes pre-checked on the form & then removed from my profile if I chose to uncheck them here.

resources/views/profiles/about/edit.blade.php
```diff
@@ -62,55 +62,56 @@
         {{-- @TODO: Refactor to loop through list of causes to create the different options --}}
         <div class="form-item">
             <label for="cause_areas" class="field-label height-auto">{{"What cause areas do you care about most?"}}</label>
```
DFurnes: Oh, missed this before but the `{{` and `}}` around this string can be removed!

resources/views/profiles/about/edit.blade.php
```diff
@@ -62,55 +62,56 @@
...
-                <span>Racial Justice & Equity</span>
-            </div>
-             <div>
-                <input type="checkbox" name="sexual_harassment_assault" value="sexual_harassment_assault">
-                <span>Sexual Harassment & Assault</span>
-            </div>
+            <fieldset>
+                <div>
+                    <input type="checkbox" name="causes[]" value="animal_welfare">
+                    <span>Animal Welfare</span>
```
DFurnes: Since there's a few of these, it might be nice to make a reusable partial!

Here's an example of where we do that in Rogue ([partial](https://github.com/DoSomething/rogue/blob/master/resources/views/forms/option.blade.php) and [form](https://github.com/DoSomething/rogue/blob/82c40227f127788a7e448ac4d1dd19052127cb7c/resources/views/campaign-ids/edit.blade.php#L30-L36)).

lindsayrmaher:
@DFurnes I pushed up changes based on your feedback!! Lemme know when you have a chance to look through everything.

app/Http/Controllers/Web/ProfileAboutController.php
```diff
@@ -2,29 +2,29 @@
...
     /**
      * Add Authentication Middleware.
      */
-    public function __construct()
+    public function __construct(Registrar $registrar)
     {
+        $this->registrar = $registrar;
         $this->middleware('auth:web');
         $this->middleware('role:admin,staff');
+        $this->causes = ['animal_welfare'=> 'Animal Welfare', 'bullying'=>'Bullying', 'education'=>'Education', 'environment' => 'Environment', 'gender_rights_equality' => 'Gender Rights & Equality', 'homelessness_poverty'=> 'Homelessness & Poverty', 'immigration_refugees'=> 'Immigration & Refugees', 'lgbtq_rights_equality' => 'LGBTQ+ Rights & Equality', 'mental_health' => 'Mental Health', 'physical_health' => 'Physical Health', 'racial_justice_equity' => 'Racial Justice & Equity', 'sexual_harassment_assault' => 'Sexual Harassment & Assault'];
```
DFurnes: It might be nice to put these on separate lines for readability! 📄

```php
$this->causes = [
    'animal_welfare'=> 'Animal Welfare',
    'bullying'=>'Bullying',
    // ...
];
```

app/Http/Controllers/Web/ProfileAboutController.php
```diff
@@ -2,29 +2,29 @@
...
 
     /**
      * Display the User Details Form
-     *
-     * @return \Illuminate\Http\Response
      */
     public function edit()
     {
-        return view('profiles.about.edit', ['user' => auth()->guard('web')->user()]);
+        return view('profiles.about.edit', ['user' => auth()->guard('web')->user(), 'causes' => $this->causes]);
```
DFurnes: Same deal! This line is looking a little cramped now. ↔️

app/Http/Controllers/Web/ProfileAboutController.php
DFurnes: @lindsayrmaher Just bumping this one in case you missed it! 🏐

lindsayrmaher: yes! if a user leaves the fields blank, line 39 ends up making a string with just a `/`, so the conditional is making sure it's set to null rather than that string. i added some comments with this last commit

PR #918, opened 2019-09-24
Profile Subscriptions Form

mendelB:
#### What's this PR do?
This PR fleshes out most of the Profile Subscriptions form, with some minimal initial styling. It also updates the ProfileSubscriptionsController to be more or less fully set up to handle form submission, validate & update the user, and redirect to the intended destination.

👁 

#### Relevant Tickets
https://www.pivotaltracker.com/story/show/168353804

#### Checklist

### DIFF
--- a/app/Http/Controllers/Web/ProfileSubscriptionsController.php
+++ b/app/Http/Controllers/Web/ProfileSubscriptionsController.php
@@ -46,7 +46,7 @@ class ProfileSubscriptionsController extends Controller
     {
         $user = auth()->guard('web')->user();
 
-        $this->registrar->validate($request, $user, ['mobile' => 'mobile|nullable|unique:users']);
+        $this->registrar->validate($request, $user);
         $this->registrar->register($request->all(), $user);
 
         return redirect()->intended('/');
--- a/app/Http/Controllers/Web/ProfileSubscriptionsController.php
+++ b/app/Http/Controllers/Web/ProfileSubscriptionsController.php
@@ -2,7 +2,6 @@
 
 namespace Northstar\Http\Controllers\Web;
 
-// use Northstar\Models\User;
 use Illuminate\Http\Request;
 use Northstar\Auth\Registrar;
 use Northstar\Http\Controllers\Controller;
@@ -41,6 +40,11 @@ class ProfileSubscriptionsController extends Controller
      */
     public function update(Request $request)
     {
-        //update stuff
+        $user = auth()->guard('web')->user();
+
+        $this->registrar->validate($request, $user, ['mobile' => 'mobile|nullable|unique:users']);
+        $this->registrar->register($request->all(), $user);
+
+        return redirect()->intended('/');
     }
 }
--- a/app/Http/Controllers/Web/ProfileSubscriptionsController.php
+++ b/app/Http/Controllers/Web/ProfileSubscriptionsController.php
@@ -4,15 +4,21 @@ namespace Northstar\Http\Controllers\Web;
 
 // use Northstar\Models\User;
 // use Illuminate\Http\Request;
+use Northstar\Auth\Registrar;
 use Northstar\Http\Controllers\Controller;
 
 class ProfileSubscriptionsController extends Controller
 {
     /**
-     * Set auth middleware.
+     * Make a new ProfileSubscriptionsController,
+     * inject dependencies, and set auth middleware.
+     *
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/Web/ProfileSubscriptionsController.php
```diff
@@ -33,8 +42,13 @@ public function edit()
...
      * @return \Illuminate\Http\RedirectResponse
      */
-    public function update()
+    public function update(Request $request)
     {
-        //update stuff
+        $user = auth()->guard('web')->user();
+
+        $this->registrar->validate($request, $user);
+        $this->registrar->register($request->all(), $user);
```
mendelB: I'm realizing now that this diverges from the pattern in the [`ProfileAboutController`](https://github.com/DoSomething/northstar/blob/caade42a8a3f44072814e1f7007f0f14010324b1/app/Http/Controllers/Web/ProfileAboutController.php#L70-#L74) and [`Web/UserController`](https://github.com/DoSomething/northstar/blob/caade42a8a3f44072814e1f7007f0f14010324b1/app/Http/Controllers/Web/UserController.php#L100-L103) where we don't use the `registrar` to save the user. 

Any reason _not_ to use the registrar? It _seems_ like a nice convenience.

katiecrane: I dug around and I see no reason not to use the registrar, especially because it centralizes all the business logic!

resources/views/profiles/subscriptions/edit.blade.php
```diff
@@ -3,21 +3,71 @@
...
+    '/images/registration-v2-03.svg'
 @endsection
 
 @section('profile-title')
-    <h2 class="text-black">A Title!</h2>
+    <h2 class="text-black">Choose your contact method</h2>
 @endsection
 @section('profile-subtitle')
-    <p>A subtitle!<p>
+    <p>What’s the best way to reach you -- email? Phone? Carrier pigeon? (Jkjk.) Let us know and we’ll send you all the best stuff, when and where you want it.<p>
```
katiecrane: 😂

mendelB: lol directly from the creative mind of @lkpttn

app/Http/Controllers/Web/ProfileSubscriptionsController.php
```diff
@@ -24,7 +30,10 @@ public function __construct()
      */
     public function edit()
     {
-        return view('profiles.subscriptions.edit', ['user' => auth()->guard('web')->user()]);
+        return view('profiles.subscriptions.edit', [
+            'user' => auth()->guard('web')->user(),
+            'intended' => session()->pull('url.intended') ?: '/',
```
lindsayrmaher: what does the 'intended variable' here add? I can make an assumption but I haven't seen it used thus far!

app/Http/Controllers/Web/ProfileSubscriptionsController.php
lindsayrmaher: I'm not against it but I'm not sure how this works! I'd be happy to go either way if you think it would make more sense. is it basically just doing what the `save` method does but with some additional validations?

app/Http/Controllers/Web/ProfileSubscriptionsController.php
mendelB: Yeah sorry, I should have clarified! This is for the "Skip" button. Since this is the last onboarding form page, that button would need to lead to the user's intended destination before being directed to authenticate (or defaults to the home page).
https://github.com/DoSomething/northstar/blob/d2358570fa4281443501cba81dc3daba63e4ac0c/resources/views/profiles/subscriptions/edit.blade.php#L63

We pull it from the session and pass it along to the view. (I'm also realizing now that this could be slightly simplified!)

```suggestion
            'intended' => session()->pull('url.intended', '/'),
```

app/Http/Controllers/Web/ProfileSubscriptionsController.php
mendelB: Exactly! It's really not much different than manually updating & saving as you were -quite reasonablely- doing in the subscription controller but it does save us a line or two of code and as @katiecrane mentioned - has the logic centralized (so is perhaps a bit DRYer?)

PR #922, opened 2019-10-07
Complete profile styling

lindsayrmaher:
#### What's this PR do?

This PR adds responsive styling to match the mocks for the registration pages!

Test in browser responsively and see screenshots added to the ticket!

#### Relevant Tickets
[Ticket #168009111](https://www.pivotaltracker.com/story/show/168009111)

#### Checklist


lindsayrmaher:
Full Size
Mobile 1
Mobile 2
Tablet

resources/views/profiles/about/edit.blade.php
```diff
@@ -3,7 +3,7 @@
 @section('title', 'Edit Profile | DoSomething.org')
 
 @section('form-image-url')
-    '/images/registration-v2-02.svg'
+    '/images/pattern-02.png'
```
mendelB: I think it'd be helpful for these pattern image files to have descriptive names (maybe similar to the ones before? (`registration-form-cover`, `about-form-cover`, `subscriptions-form-cover`? ugh I'm bad at names, something like that...)

I _think_ the `Icons-[1-3]` images could also use more descriptive naming. (but again I hit my creative wall -- maybe `progress-icon-[1-3]`?)

Also, was there a specific reason for the switch to `png`?

resources/views/profiles/about/edit.blade.php
```diff
@@ -14,43 +14,57 @@
...
+    @endif
+
     <form method="POST" action="{{ url('profile/about') }}">
         <input type="hidden" name="_method" value="PATCH">
         {{ csrf_field() }}
 
-        <div class="form-item">
-            <label for="birthdate" class="field-label">{{ trans('auth.fields.birthday') }}</label>
-            <div class="form-item -reduced">
+        <div class="form-item flex flex-wrap justify-start justify-between md:justify-start">
```
mendelB: Is the first `justify-start` here extra since it'd overwritten by the `justify-between`, and then set specifically for `md` afterward?

resources/views/profiles/about/edit.blade.php
```diff
@@ -61,18 +75,33 @@
...
+                <img src="/images/icons-2.svg" />
+            </div>
+            <div class="w-2/3 flex justify-around md:justify-end p-2">
+                <div class="w-1/4 sm:w1/2">
+                    <a href="{{ url('profile/subscriptions') }}" class="button">Skip</a>
+                </div>
+                <div class="w-1/4 sm:w1/2">
+                    <input type="submit" id="register-submit" class="button" value="Next">
+                </div>
+            </div>
```
mendelB: Nice! When I add this to the subscription form, I can see if it indeed makes sense to move this to the partial!

resources/views/profiles/about/edit.blade.php
```diff
@@ -14,43 +14,57 @@
...
+
+    @if (count($errors) > 0)
+        <div class="validation-error fade-in-up">
+            <h4>{{ trans('auth.validation.issues') }}</h4>
+            <ul class="list -compacted">
+                @foreach ($errors->all() as $error)
+                    <li>{{ $error }}</li>
+                @endforeach
+            </ul>
+        </div>
```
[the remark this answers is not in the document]
lindsayrmaher: aw yuss, i shall update!

resources/views/profiles/about/edit.blade.php
lindsayrmaher: So I struggled with this a bit. I want there to be `justify-start` and `justify-between` for small sizes, but only `justify-start` on md & lg screen sizes. I ended up needing to define it twice. I tried just defining `justify-between` on sm but that didn't work the way I wanted it to. I think is a result of how TW is handling the media queries.

resources/views/profiles/about/edit.blade.php
lindsayrmaher: I looked at making this a partial once you pointed this out, I was just struggling with adding it as a partial inside the form. If you want to take a look that works for me, otherwise I could try refactoring for this PR!

resources/views/profiles/about/edit.blade.php
lindsayrmaher: I will def rename!

app/Http/Controllers/Web/ProfileAboutController.php
```diff
@@ -39,7 +41,8 @@ public function edit()
     {
         return view('profiles.about.edit', [
             'user' => auth()->guard('web')->user(),
-            'causes' => $this->causes,
+            'causes1' => $this->causes1,
+            'causes2' => $this->causes2,
```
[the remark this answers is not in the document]
lindsayrmaher: I like the former!! Great idea to keep them all together as one variable.

resources/views/profiles/about/edit.blade.php
mendelB: Oh hmm I would've thought that the `justify-between` would overwrite the `justify-start` since they're both affecting the same `justify-content` rule 🤔

> I tried just defining justify-between on sm but that didn't work the way I wanted it to.

So something like `sm:justify-between md:justify-start` would _not_ work?

resources/views/profiles/about/edit.blade.php
mendelB: I think it's fine as-is for now! I even wonder if it's just overkill to have this as a partial for now, but I think I can just see how it is once I implement this in the subscriptions form

resources/views/profiles/about/edit.blade.php
lindsayrmaher: AH, you right. I updated to just `justify-between md:justify-start`

lindsayrmaher:
@mendelB Lemme know what you think of the updates!

resources/views/profiles/about/edit.blade.php
```diff
@@ -61,18 +68,33 @@
...
-            <a href="{{ url('profile/subscriptions') }}" class="button">Skip</a>
-        </div>
-        <div class="form-actions -padded -right">
-            <input type="submit" id="register-submit" class="button" value="Next">
+        <div class="flex pt-4">
+            <div class="w-1/3 flex justify-between md:justify-start">
+                <img src="/images/about-form-icon.svg" />
+            </div>
+            <div class="w-2/3 flex justify-around md:justify-end p-2">
+                <div class="w-1/4 sm:w1/2">
```
mendelB: Do these individual buttons need their own width set, or can they just rely on the spacing assigned by the parent div, and we can just give them some padding here?

resources/views/profiles/about/edit.blade.php
```diff
@@ -61,18 +68,33 @@
...
-        <div class="form-actions -padded -left">
-            <a href="{{ url('profile/subscriptions') }}" class="button">Skip</a>
-        </div>
-        <div class="form-actions -padded -right">
-            <input type="submit" id="register-submit" class="button" value="Next">
+        <div class="flex pt-4">
+            <div class="w-1/3 flex justify-between md:justify-start">
+                <img src="/images/about-form-icon.svg" />
+            </div>
+            <div class="w-2/3 flex justify-around md:justify-end p-2">
```
mendelB: Is there a reason we need `justify-around` on small screens as opposed to `justify-end` all around?

resources/views/profiles/about/edit.blade.php
```diff
@@ -61,18 +68,33 @@
...
+            </div>
         </div>
 
-        <div class="form-actions -padded -left">
-            <a href="{{ url('profile/subscriptions') }}" class="button">Skip</a>
-        </div>
-        <div class="form-actions -padded -right">
-            <input type="submit" id="register-submit" class="button" value="Next">
+        <div class="flex pt-4">
+            <div class="w-1/3 flex justify-between md:justify-start">
```
mendelB: Similarly here - I'm wondering about the `justify-between` for small screens?

resources/views/profiles/about/edit.blade.php
```diff
@@ -14,43 +14,50 @@
...
                 <label for="day">Day</label>
-                <input class="text-field" name="birthdate[]" type="text" placeholder="DD" maxlength="2" />
+                <input class="text-field " name="birthdate[]" type="text" placeholder="DD" maxlength="2" />
             </div>
-            <div class="form-item -reduced">
+            <div class="form-item -reduced w-1/5">
                 <label for="year">Year</label>
                 <input class="text-field" name="birthdate[]" type="text" placeholder="YYYY" maxlength="4" />
             </div>
         </div>
```
mendelB: Just confirming - this whole field is set for some construction to make it one single input, right?

resources/views/profiles/about/edit.blade.php
```diff
@@ -14,43 +14,50 @@
...
 
-        <div class="form-item">
-            <label for="voter_registration_status" class="field-label height-auto">Are you registered to vote at your current address?</label>
-            <div class="form-item -reduced">
+        <div class="form-item flex flex-wrap justify-between md:justify-start">
```
mendelB: What's the reason for the `justify-between` on small screens here?

resources/views/profiles/about/edit.blade.php
lindsayrmaher: The buttons end up overlapping one another on small screens if I leave it just at `justify-end` for all sizes

resources/views/profiles/about/edit.blade.php
lindsayrmaher: I think we could probably go either way, this just felt like a simple solution for now to use the built-in TW classes!

PR #924, opened 2019-10-08
New Register Page

mendelB:
#### What's this PR do?
Adds a beta register form for the new onboarding flow. It updates the AuthController's register methods to handle the new flow and adds some routes to access the new page.

Confirm that we're comfortable adding publicly accessible routes for this new form? Currently, this will try redirecting the user to the next part of the onboarding flow but will redirect to the homepage since they're not staff/admin. We could also remove that role restriction so that - in case this is accessed - they'll still be able to go through the whole onboarding flow, which should technically work now?

#### Relevant Tickets
https://www.pivotaltracker.com/story/show/168354105

#### Checklist

📸 
- [Large](https://user-images.githubusercontent.com/12417657/66593340-cec84880-eb64-11e9-9093-cb1ba7c0e36b.png)
- [Medium](https://user-images.githubusercontent.com/12417657/66593341-cec84880-eb64-11e9-8463-7a914ccd3998.png)
- [Small](https://user-images.githubusercontent.com/12417657/66593342-cec84880-eb64-11e9-846c-43e12e37aeb4.png)

### DIFF
--- a/resources/views/auth/register-beta.blade.php
+++ b/resources/views/auth/register-beta.blade.php
@@ -60,5 +60,5 @@
 
     @include('auth.facebook')
 
-    <p class="text-gray-500">Already have an account? <a href="{{ url('login') }}">Log In</a></p>
+    <p class="text-gray-500 mt-5">Already have an account? <a href="{{ url('login') }}">Log In</a></p>
 @endsection
--- a/resources/views/auth/register-beta.blade.php
+++ b/resources/views/auth/register-beta.blade.php
@@ -22,7 +22,7 @@
         {{ csrf_field() }}
 
         <div class="md:flex md:flex-wrap md:justify-between">
-            <div class="form-item md:w-1/2 pr-4">
+            <div class="form-item md:w-1/2 md:pr-4">
                 <label for="first_name" class="field-label">First Name</label>
                 <input name="first_name" type="text" id="first_name" class="text-field required js-validate" placeholder="First Name" value="{{ old('first_name') }}" autofocus data-validate="first_name" data-validate-required />
             </div>
--- a/resources/views/auth/register-beta.blade.php
+++ b/resources/views/auth/register-beta.blade.php
@@ -47,7 +47,7 @@
         </div>
 
         <div class="form-item">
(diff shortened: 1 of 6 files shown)


resources/views/auth/register-beta.blade.php
```diff
@@ -0,0 +1,62 @@
...
+@section('profile-form')
+    @if (count($errors) > 0)
+        @include('forms.errors', ['errors' => $errors])
+    @endif
+
+    <form method="POST" action="{{ url('register-beta')}}">
+        {{ csrf_field() }}
+
+        <div class="form-item">
+            <label for="first_name" class="field-label">{{ trans('auth.fields.first_name') }}</label>
```
mendelB: What's our policy on using translated labels for these forms? We seem to have them for a bunch of our labels, but not all of them. Is this something that just got carried over from hashtag global? Should we keep using these / and updating these for the new forms we're making? (I'd assume that we should be consistent either way.)

lindsayrmaher: I was wondering this same thing!

lindsayrmaher:
> I figured it was technically ready to go sans finalized styling / front end validations!

That's fair, I would say that is about where it's at. And just finalizing the data persistence setup which I should also be able to wrap up today.

resources/views/auth/register-beta.blade.php
DFurnes: Yeah, I think it's worth removing this. As you mentioned, we don't translate everything on the form, we don't translate _anything_ elsewhere on Phoenix, and I don't believe we're investing any resources into more translations moving forward (at least in the short-term).

DFurnes:
> Confirm that we're comfortable adding publicly accessible routes for this new form? Currently, this will try redirecting the user to the next part of the onboarding flow but will redirect to the homepage since they're not staff/admin. We could also remove that role restriction so that - in case this is accessed - they'll still be able to go through the whole onboarding flow, which should technically work now?

Yeah, I'm good with that! 👍 And no red flags for removing the role restriction from me.

mendelB:
UPDATE:
I've updated the register page to be fully compliant with the mocks. 

https://github.com/DoSomething/northstar/pull/924/commits/8a4014c19c5e5e8db82f942222e6afb9635fb4a3 https://github.com/DoSomething/northstar/pull/924/commits/805dfbaebce34007eef4eb891e0b8bd00263ac5e Removed the translatable fields (also from the about form while I was at it cc @lindsayrmaher) as per https://github.com/DoSomething/northstar/pull/924#discussion_r333091522
And removed the auth scoping from the profile controllers in https://github.com/DoSomething/northstar/pull/924/commits/0740ccbc630d5dc4c0b45a730db4c5386b9bc97d per https://github.com/DoSomething/northstar/pull/924#issuecomment-540096221

PR #927, opened 2019-10-10
Starts on Google authentication (sans birthdate)

aaronschachter:
#### What's this PR do?

This PR starts on OAuth Google authentication using Laravel Socialite. Following our Facebook integration by example, this PR adds:

* New user field:`google_id` to save the user's Google ID upon authentication

* New `app/Http/Controllers/Web/GoogleController`, to be used by new routes:

    * `/google/continue` - redirects to the Google OAuth consent screen
    * `/google/verify` - receives callback from Google OAuth after authorization

Upon authorization, a new Northstar user is created and logged in, populating first name, last name, google ID, and email from the Google user.  No birthdate yet.

From what I can tell, I don't think there's a way to read a Google user's birthday via Socialite. We're asking for it in scopes, but don't have access to it in the [API response we get back from `getUserByToken`](https://github.com/laravel/socialite/blob/65f474ac5ed5b24d1dfd5daca0fae7b383530f78/src/Two/GoogleProvider.php#L61). [This Laravel News post](https://laravel-news.com/google-api-socialite) walks through using the Socialite in combination with the Google API client to read a list of a user's contacts -- I think we'll need to do something similar by making an API request to`/people/:id` with our authenticated user's ID, [specifying `birthdays` as a `personFields`](https://developers.google.com/people/api/rest/v1/people/get?apix_params=%7B%22resourceName%22%3A%22people%2Fme%22%2C%22personFields%22%3A%22birthdays%22%7D) in order to populate our Northstar user's `birthdate` field. That, my friends, is for a future PR.

I've been testing by using ngrok to make my local Northstar instance public, and then creating a new project in Google Cloud that uses OAuth, configured with my ngrok URL for allowed domains and the redirect URL. 

I will setup our Northstar Dev project to use dev.dosomething.org. Will also need to do this for QA and prod eventually. I think the best bet for testing this PR before merging would be to deploy this branch to our dev instance, if we wanted to sanity check that it works before merging into master.

#### Relevant Tickets

https://www.pivotaltracker.com/story/show/167675336

#### Checklist


aaronschachter:
Re: the Northstar dev instance, it may not be feasible to have 3 different cloud projects for our 3 different Northstar environments (see [PT comment](https://www.pivotaltracker.com/story/show/169024616/comments/207561666)).

This is an example of my local setup / personal Google Cloud project asking for birthday access after I authorized:

Screen Shot 2019-10-09 at 2 45 59 PM

which I got working by manually adding `https://www.googleapis.com/auth/user.birthday.read` to the scopes in the OAuth Consent screen configuration in the Google Cloud project:

Screen Shot 2019-10-09 at 9 17 04 PM

app/Http/Controllers/Web/GoogleController.php
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        // Aggregate public profile fields
+        $fields = [
+            'google_id' => $googleUser->id,
+            'first_name' => $googleUser->user['given_name'],
+            'last_name' => $googleUser->user['family_name'],
+        ];
+
+        $northstarUser = User::where('email', '=', $googleUser->email)->first();
```
DFurnes: Let's update this to use the Registrar's [`resolve`](https://github.com/DoSomething/northstar/blob/cd6fa1ce383c5f56f793dd00d76f2e40fa3ab249/app/Auth/Registrar.php#L86-L122) method instead, since that normalizes the email address before querying (so we wouldn't have a "false negative" if we had different casing than Google).

```php
$this->registrar->resolve(['email' => $googleUser->email]);
```

DFurnes: And maybe make the same change [in FacebookController](https://github.com/DoSomething/northstar/blob/cd6fa1ce383c5f56f793dd00d76f2e40fa3ab249/app/Http/Controllers/Web/FacebookController.php#L84) while it's top of mind!

lindsayrmaher:
@aaronschachter @DFurnes I have a question about getting the birthdate field populated. In the new registration flow that @mendelB and I have been working on, birthdate is no longer required on initial sign up. So I would assume then that you don't have to worry about getting that functionality set up with the google auth process? And possibly with our current Facebook implementation? Just wanted to make sure we're not adding additional work for Aaron that would be necessary in the near future!

app/Http/Controllers/Web/GoogleController.php
```diff
@@ -0,0 +1,102 @@
...
+
+        if ($northstarUser) {
+            $northstarUser->fillUnlessNull($fields);
```
katiecrane: This would overwrite a user's `first_name` and `last_name` right? Do we do that with Facebook users?

katiecrane:
@lindsayrmaher I think we still consider it valuable to have a user's birthday (esp. for voter reg stuff), so if we can get it without them having to fill out a form I think we want to do that 🎂

app/Http/Controllers/Web/GoogleController.php
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * Redirect the user to the Google authentication page.
+     *
+     * @return Response
+     */
+    public function redirectToProvider()
+    {
+        return Socialite::driver('google')
+            ->scopes(['profile', 'https://www.googleapis.com/auth/user.birthday.read'])
```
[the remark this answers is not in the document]
aaronschachter: This lets Google know that our application wants access to read the user's birthday, but we still need to make an additional API request to Google to fetch the user's birthday (details and examples linked in the PR description)

app/Http/Controllers/Web/GoogleController.php
aaronschachter: Looks like it would - and yes we do same for Facebook. Most of this controller (and its tests) is copied verbatim from `FacebookController`:
https://github.com/DoSomething/northstar/blob/e34f43904406e2351270899239428223d6cb73a9/app/Http/Controllers/Web/FacebookController.php#L87

aaronschachter:
As far as importing user's birthday, it definitely would simplify the Northstar work -- we'd pretty be much done with Google OAuth per this PR. Importing birthday is the only missing piece left to build, and requires an extra Google API call, so it's not exactly just like adding an element to array in `fields` like we get to do in the Facebook Socialite:
https://github.com/DoSomething/northstar/blob/e34f43904406e2351270899239428223d6cb73a9/app/Http/Controllers/Web/FacebookController.php#L57


Diego D suggests we wouldn't need to import birthday per these registration over in https://www.pivotaltracker.com/story/show/167675336/comments/207584176. I'm personally fine with not importing the birthday if we're going to ask for it later, although it sounds like we shouldn't go live with the Google sign-in until the new registration forms are live. 

Sure, there's a UX win in having it pre-populated for Google users, but since Google needs to verify our OAuth consent form before it goes live -- it'd be nice to skip querying for the birthday. We can potentially still keep it within the scopes we ask for (so Google wouldn't need to re-verify if we decide to import later)... but there's a UX loss there (hey, I granted you access to my birthday and now I need to manually enter it)

katiecrane:
oh cool! I didn't realize that there had already been a call made to _not_ get the birthdate. that works for me!

app/Http/Controllers/Web/GoogleController.php
katiecrane: Perfect! Just wanted to make sure that was consistent 👍

app/Http/Controllers/Web/GoogleController.php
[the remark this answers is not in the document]
katiecrane: Gotcha - I guess this url will change if we don't want the birthday?

aaronschachter:
We're still figuring out the call re: birthday import over in [#team-applications](https://dosomething.slack.com/archives/CP2D7UGAU/p1570726508006700)

PR #929, opened 2019-10-10
Temporary feature flag redirecting to new register flow

mendelB:
#### What's this PR do?
 🚩Adds a temporary feature flag - via query parameter - to redirect users to use the new registration flow. We'll use this from the Phoenix end, to allow testing the new flow end-to-end without any disruption on production.

EDIT: Updated re https://github.com/DoSomething/northstar/pull/929#discussion_r334029291 in https://github.com/DoSomething/northstar/pull/929/commits/85c62100c6369295da49d85ee8bab8ca582e0829 to simply utilize the `mode` to affect the authorization route. Way cleaner ☺️ 

This won't work just yet! 
- I'll add a corresponding PR on Phoenix to send over this param
- This relies on the temporary `register-beta` route defined in #924 which hasn't been merged yet. 

👁 

#### Relevant Tickets
We discussed this idea [here on Slack](https://dosomething.slack.com/archives/C02BBRN5A/p1570732864009800)

### DIFF
--- a/app/Http/Controllers/Web/AuthController.php
+++ b/app/Http/Controllers/Web/AuthController.php
@@ -67,6 +67,8 @@ class AuthController extends Controller
                 'title' => request()->query('title', trans('auth.get_started.create_account')),
                 'callToAction' => request()->query('callToAction', trans('auth.get_started.call_to_action')),
                 'coverImage' => request()->query('coverImage', asset('members.jpg')),
+                // Temporary 'feature flag' to redirect to the new registration flow (for testing).
+                'registerBeta' => request()->query('register_beta'),
                 // Store any provided UTMs, Referrer User ID, or Contentful ID for user's source_detail:
                 'source_detail' => array_filter([
                     'contentful_id' => request()->query('contentful_id'),
--- a/app/Http/Controllers/Web/AuthController.php
+++ b/app/Http/Controllers/Web/AuthController.php
@@ -67,8 +67,6 @@ class AuthController extends Controller
                 'title' => request()->query('title', trans('auth.get_started.create_account')),
                 'callToAction' => request()->query('callToAction', trans('auth.get_started.call_to_action')),
                 'coverImage' => request()->query('coverImage', asset('members.jpg')),
-                // Temporary 'feature flag' to redirect to the new registration flow (for testing).
-                'registerBeta' => request()->query('register_beta'),
                 // Store any provided UTMs, Referrer User ID, or Contentful ID for user's source_detail:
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/Web/AuthController.php
```diff
@@ -67,6 +67,8 @@ public function getAuthorize(ServerRequestInterface $request, ResponseInterface
                 'title' => request()->query('title', trans('auth.get_started.create_account')),
                 'callToAction' => request()->query('callToAction', trans('auth.get_started.call_to_action')),
                 'coverImage' => request()->query('coverImage', asset('members.jpg')),
+                // Temporary 'feature flag' to redirect to the new registration flow (for testing).
+                'registerBeta' => request()->query('register_beta'),
```
DFurnes: NBD since this is temporary, but would it make more sense to add another option to our [existing `mode` parameter](https://github.com/DoSomething/northstar/blob/8b06f5287642bf43495bc61e2118b3ed0a2658b6/app/Http/Controllers/Web/AuthController.php#L82-L83) instead? (So you could select `login`, `register`, or `register-beta`?)

mendelB: Oh wait sorry, I may have misunderstood your comment in that thread that we're cool with just this feature flag since we're not A/B testing? I can totally make the switch here and on Phoenix if you think that's a better approach! Would we similarly just be setting the `mode` to `register-beta` over on Phoenix based on the query parameter?

DFurnes: I'm good either way!

If we want to just use a feature flag, it might be simpler to have a single environment variable that we swap on and off in Northstar (rather than sending it along from Phoenix). If we want to be able to swap it on and off per-request, I think using `mode` makes sense.

DFurnes: But that said, I also don't know if it's worth changing at this point!

mendelB: Hmm ok gotcha! Since we do want to swap it on and off per-request, I'll go ahead and update the PR's to use mode. (If I see it start taking too long though I'll just merge as is!)

mendelB: @DFurnes Updated in https://github.com/DoSomething/northstar/pull/929/commits/85c62100c6369295da49d85ee8bab8ca582e0829 and updated over on Phoenix as well https://github.com/DoSomething/phoenix-next/pull/1639. It definitely feels way better this way!

PR #930, opened 2019-10-12
Fetch Google user birthday

aaronschachter:
#### What's this PR do?

This PR fetches an authenticated Google user's birthday from the People API, and saves it to the corresponding Northstar user `birthdate` field.

* 👀 

* Upon deploy, visiting `/google/continue` and allowing access should save your Google user's birthday to your Northstar user's `birthdate` field

* In a future PR, I'd like to document the steps I've taken for setting up the Google Cloud project, as well as using ngrok for local development.

#### Relevant Tickets

https://www.pivotaltracker.com/n/projects/2401401/stories/167675336

#### Checklist
- [ ] Documentation added for changed endpoints. - for future PR

### DIFF
--- a/app/Http/Controllers/Web/GoogleController.php
+++ b/app/Http/Controllers/Web/GoogleController.php
@@ -55,14 +55,12 @@ class GoogleController extends Controller
         try {
             $googleUser = Socialite::driver('google')->user();
 
-            // @see https://developers.google.com/people/api/rest/v1/people/get
             $client = new Google($googleUser->token);
 
             $data = $client->getProfile();
 
             // TODO: Loop through $data->birthdays array, checking if each object entry has a data property that contains year. Use that as birthday.
-
-            info(print_r($data->birthdays, true));
+            // info(print_r($data->birthdays, true));
         } catch (RequestException | ClientException | InvalidStateException $e) {
             logger()->warning('google_token_mismatch');
 
--- a/app/Http/Controllers/Web/GoogleController.php
+++ b/app/Http/Controllers/Web/GoogleController.php
@@ -50,11 +50,24 @@ class GoogleController extends Controller
      */
     public function handleProviderCallback()
     {
-        // Grab the user's profile using their Google OAuth token.
+        // Fetch user's birthday using their Google OAuth token.
         try {
             $googleUser = Socialite::driver('google')->user();
-            // TODO: Make API request with $googleUser->token to get user birthday.
-            // @see https://developers.google.com/people/api/rest/v1/people/get?apix_params=%7B%22resourceName%22%3A%22people%2Fme%22%2C%22personFields%22%3A%22birthdays%22%7D
+
+            // @see https://developers.google.com/people/api/rest/v1/people/get
+            $client = new \GuzzleHttp\Client([
+                'base_uri' => 'https://people.googleapis.com/v1/',
+                'headers' =>  [
+                    'Authorization' => 'Bearer '.$googleUser->token,
+                ],
+            ]);
+
+            $response = $client->get('people/me?personFields=birthdays');
+            $data = json_decode($response->getBody()->getContents());
+
+            // TODO: Loop through $data->birthdays array, checking if each object entry has a data property that contains year. Use that as birthday.
+
+            info(print_r($data->birthdays, true));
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/Web/GoogleController.php
```diff
@@ -50,11 +51,16 @@ public function redirectToProvider()
...
-        // Grab the user's profile using their Google OAuth token.
+        // Fetch user's birthday using their Google OAuth token.
         try {
             $googleUser = Socialite::driver('google')->user();
-            // TODO: Make API request with $googleUser->token to get user birthday.
-            // @see https://developers.google.com/people/api/rest/v1/people/get?apix_params=%7B%22resourceName%22%3A%22people%2Fme%22%2C%22personFields%22%3A%22birthdays%22%7D
+
+            $client = new Google($googleUser->token);
+
+            $data = $client->getProfile();
```
aaronschachter: If I comment this line out, the tests in `GoogleTest` work as expected (introduced and passing in #927)

app/Http/Controllers/Web/GoogleController.php
```diff
@@ -50,11 +51,16 @@ public function redirectToProvider()
...
     public function handleProviderCallback()
     {
-        // Grab the user's profile using their Google OAuth token.
+        // Fetch user's birthday using their Google OAuth token.
         try {
             $googleUser = Socialite::driver('google')->user();
-            // TODO: Make API request with $googleUser->token to get user birthday.
-            // @see https://developers.google.com/people/api/rest/v1/people/get?apix_params=%7B%22resourceName%22%3A%22people%2Fme%22%2C%22personFields%22%3A%22birthdays%22%7D
+
+            $client = new Google($googleUser->token);
```
DFurnes: In order to mock things, they need to come from the [service container](https://laravel.com/docs/5.5/container) (since otherwise, there's no way for Laravel to swap this out with your mock). To do that, you can either inject it via the constructor (like we do for `Registrar`), or use the [`app`](https://laravel.com/docs/5.5/helpers#method-app) helper.

Since that means we'll be letting Laravel instantiate this class, it probably makes sense to instead accept a `$token` in the `getProfile()` method and override the headers there.

aaronschachter:
Thank you for the 👀 @katiecrane and @DFurnes - this is now ready for review when you have a chance.

app/Http/Controllers/Web/GoogleController.php
```diff
@@ -68,11 +74,18 @@ public function handleProviderCallback()
             return redirect('/register')->with('status', 'We need your email to contact you if you win a scholarship.');
         }
 
-        // Aggregate public profile fields
+        // Some date properties in this array may not contain a year property.
```
katiecrane: So there can be multiple `birthdays` on the user, but only one of them will contain a year?

app/Services/Google.php
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    /**
+     * Returns authenticated user profile.
+     * @param string $token
+     * @return object
+     */
+    public function getProfile($token)
+    {
+        $response = $this->client->get('people/me?personFields=birthdays', [
```
katiecrane: Does this request return more than just the birthdays?

app/Http/Controllers/Web/GoogleController.php
aaronschachter: I'm not sure exactly. The test payload provided in `GoogleTest` was from the results of the API explorer when using my own work gmail account, where I saw one item without the year. When I used my personal gmail, only one item was returned, that included year.

The docs don't give much information about how multiple birthdays end up here: https://developers.google.com/people/api/rest/v1/people#Person.Birthday

katiecrane: Oooo interesting. Well this approach seems good since we know it's possible to have one without a year!

app/Services/Google.php
aaronschachter: It returns a `resourceName` and `etag`, per https://developers.google.com/people/api/rest/v1/people and whatever `personFields` you pass through. I considered calling this `getBirthday`, but there was a question via Slack about potentially capturing other user info like mobile number or address, so I figured I'd go general with `getProfile`.

katiecrane: That sounds good to me!

PR #933, opened 2019-10-16
Remove 'last_name' field with 'display_name' in C.io.

DFurnes:
#### What's this PR do?
This pull request replaces the `last_name` field with `display_name` for future Customer.io updates. We can reach out to the team at Customer.io to clear out old values for this field (since it'll no longer be updated when it changes).

I [confirmed with Anthony](https://dosomething.slack.com/archives/C03T8SDJW/p1571257426052700) that we don't use this, so just a quick 👁 is good!

#### Relevant Tickets
[#167596958](https://www.pivotaltracker.com/story/show/167596958)

#### Checklist

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -510,7 +510,7 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
             'facebook_id' => $this->facebook_id,
             'google_id' => $this->google_id,
             'first_name' => $this->first_name,
-            'last_name' => $this->last_name,
+            'display_name' => $this->display_name,
             'birthdate' => optional($this->birthdate)->timestamp,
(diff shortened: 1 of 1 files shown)


aaronschachter:
@DFurnes Looks like we need to change up our User Model test:
https://github.com/DoSomething/northstar/blob/329ba38d004216d66d5c816e6ab84c815ef18e26/tests/Models/UserTest.php#L22

mendelB:
> @DFurnes Looks like we need to change up our User Model test:
> https://github.com/DoSomething/northstar/blob/329ba38d004216d66d5c816e6ab84c815ef18e26/tests/Models/UserTest.php#L22

Good catch! I wonder why the build didn't fail on that test? 🤔

PR #934, opened 2019-10-16
New Onboarding Flow - Analytics Events

mendelB:
#### What's this PR do?
🔬 This PR updates our general analytics logic to account (_account - get it?_) for the new registration and onboarding flow.

We're adding some new events, as listed on [this doc](https://docs.google.com/spreadsheets/d/1lm-fGrIm85nUTxSojqyCt_Ehmm1zEbViFhKpxcJiz1A/edit#gid=406441516) in the _green_ cells.

We're also updating the general logic for many of our events to be more general, accounting for both register and onboarding flows, notably:
- 🎯  17e605df8c15c26be3a28028a98424e702659a30, 17e605df8c15c26be3a28028a98424e702659a30 allow setting the `target` for `.login-links` click events, to account for legacy buttons tracked as links (see clarifying comment in https://github.com/DoSomething/northstar/commit/f50b286f64a4fb2136beecaa96719a5d1a14d815)
- ®️ 6e5cf6962c780a75057e73a306ca44797dd5b252, 67a8b43a4807e4fbe5ec54cba7a5e489a6641ff4 help us grab the `category` for events based on the URL, so that we can differentiate between `onboarding` and `authentication` events (`/register` is clearly `authentication`, `/profile/[about|subscriptions]` would be `onboarding`)
- 👁 🇩🇪  d1a5983d977d3664ce19fe9894e50a6987ccd959 some of our submission events were hardcoded to `register` (as an adjective describing the form, or just the `noun` to begin with), this adds a helper to dynamically grab the form type using the `id` attribute, which should follow what I think is a reasonable convention (which is based on an existing id) of `profile-[form type]-form`.
-👂 4fb0ec5935ebcd818a76705b5047c28536345a14 will add a temporary event listener for the onboarding forms, to allow us to track the form submission in the absence of a `dosomething-validation` published event, which only is triggered when the form contains _required_ or _populated_ inputs. See [this sumptuous Slack thread](sumptuous) for more context. (@weerd's apt alternative recommendation to update the `dosomething-validation` package can be tackled in a cleanup sprint (@ngjo took an oath that we'd be able to) since my attempt to do it that way was blocked by what look like some tricky breaking dependency updates we need to make for security which should probably be tackled before we update the package (or so I assumed.))

👟 
I snuck in [one](https://github.com/DoSomething/northstar/commit/bf66cfc8534db10818d2b5a64889d1098769148d) or [two](https://github.com/DoSomething/northstar/commit/2fb35528a207e51dfa8de603a4b7a6bbbae55395) easy wins, while I was poking around.

I've manually tested all expected analytics events, along with the new ones - and things seem spry. That being said, some of this logic was tricky to finesse so call out anything suspicious! There's certainly some controversial logic implemented here, so I'd venture to suggest that some larger suggested updates based on feedback could wait for a post launch cleanup, and I'd advocate for a larger cleanup of this analytics flow in general since we've certainly outgrown the boutique artisanal analytics setup we started with, and have demonstrably reached a point of complexity deserving of some 👀 and rethinking. That being said, anything unconscionable quite obviously needs to be updated before this merges.

NAMING! I didn't have much time for good creative naming ideation -- so any better naming ideas for these events/matadata/helpers are most welcome!

#### Relevant Tickets
https://www.pivotaltracker.com/story/show/169013232

See [#topic-web-analytics](https://dosomething.slack.com/archives/CJU7WCQG4/p1570816458009600) in Slack for more context on the new categories etc. 

#### Checklist

### DIFF
--- a/resources/assets/js/utilities/Analytics.js
+++ b/resources/assets/js/utilities/Analytics.js
@@ -490,8 +490,8 @@ function init() {
       });
     });
 
-    $('.login-link').on('click', (e) => {
-      const target = e.target.dataset.target || 'button';
+    $('.login-link').on('click', (element) => {
+      const target = element.target.dataset.target || 'button';
 
       // Tracks clicking on any of the 'Log in' buttons and links.
       trackAnalyticsEvent({
--- a/resources/assets/js/utilities/Analytics.js
+++ b/resources/assets/js/utilities/Analytics.js
@@ -389,6 +389,41 @@ function init() {
       trackInputFocus(inputName);
     }
 
+    // Specifically tracks the #profile subscriptions & about forms to ensure submission is
+    // tracked irrespective of the validation package registering the submission sans required, populated input fields.
+    // @TODO replace this @HACK with the updated doSomething/validation package to register
+    // these submissions to begin with.
+    $('#profile-subscriptions-form, #profile-about-form').on('submit', function() {
+      // The following copies logic used by the dosomething-validation package,
+      // which determines weather there are fields to be validated.
+      var $form = $(this);
+
+      var $validationFields = $form.find("[data-validate]");
+
+      $validationFields = $validationFields.map(function() {
+        var $this = $(this);
+        if(typeof $this.attr("data-validate-required") !== "undefined" || $this.val() !== "") {
+          return $this;
+        }
+      });
+
+      // If there are fields to be validated, we defer to the dosomething-validation package
+      // which will publish a submission event which we've subscribed to, to track analytics.
+      if ($validationFields.length) {
+        return;
+      }
+
+      // Otherwise, track when an inline validation error free submission is made.
+      trackAnalyticsEvent({
+        metadata: {
+          category: getCategoryFromPath(),
+          noun: getFormTypeFromId($form.attr('id')),
+          target: 'form',
+          verb: 'submitted',
+        },
+      });
+    });
+
     // Tracks when user focuses on form field.
     $('input').on('focus', (element) => {
       const inputName = element.target.name;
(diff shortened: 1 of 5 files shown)


resources/assets/js/utilities/Analytics.js
```diff
@@ -81,6 +81,31 @@ export function getAdditionalContext() {
   };
 }
 
+/**
+ * Get the category for an event based on current pathname.
+ *
+ * @return {String}
+ */
+const getCategoryFromPath = () => {
+  const pathToCategoryMap = {
```
DFurnes: Neat! I like this idea. 👍

resources/assets/js/utilities/Analytics.js
```diff
@@ -81,6 +81,31 @@ export function getAdditionalContext() {
...
+  return pathToCategoryMap[window.location.pathname] || 'authentication';
+}
+
+/**
+ * Parse an id value to discern the form type (with a sensible default).
+ *
+ * @param {String} [id='']
+ * @return {String}
+ */
+const getFormTypeFromId = (id = '') => id.split('-')[1] || 'form';
```
DFurnes: It might help to include an example in the docblock... I'm assuming this is intended to parse `register` from `#profile-register-form`? If so, do we have any concerns about the "implicit" format we're imposing on IDs?

mendelB: Ah wise, I'll add an example! Ya know, I had the same reservation, but I wasn't quite sure of a great way to dynamically grab the 'type' of form we're dealing with in e.g. [submission events](https://github.com/DoSomething/northstar/pull/934/files#diff-63c2a55da3832f72e090747b5fc49408R360) where all we have is the ID of the form 🤔

mendelB: My thinking was that this is a specific helper for this specific scenario, so while we _are_ imposing this "implicit" format on IDs it's just kind of part of the contract of using this helper. (Maybe adding the example will help clarify that!)

mendelB: I'm going to merge this PR (to unblock @sheyd on QA-ing these events) pending feedback here since the event names themselves should remain irrespective of potential follow up PR's here if we update this logic!

PR #939, opened 2019-10-17
Sms Status in Subscriptions Onboarding

mendelB:
#### What's this PR do?
This PR ensures a user's `sms_status` is set to `'active'` if they add or update their `mobile` number in the onboarding flow.

(The logic is there to ensure we don't unintentionally resubscribe a user if they merely submitted the form without updating their existing unsubscribed number.)

👁 

#### Relevant Tickets
https://www.pivotaltracker.com/story/show/169193734 (See comments for discussion confirming this logic.)

#### Checklist

### DIFF
--- a/app/Http/Controllers/Web/ProfileSubscriptionsController.php
+++ b/app/Http/Controllers/Web/ProfileSubscriptionsController.php
@@ -45,7 +45,17 @@ class ProfileSubscriptionsController extends Controller
         $user = auth()->guard('web')->user();
 
         $this->registrar->validate($request, $user);
-        $this->registrar->register($request->all(), $user);
+
+        $currentMobile = $user->mobile;
+
+        $this->registrar->register($request->all(), $user, function ($user) use ($currentMobile) {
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/Web/ProfileSubscriptionsController.php
```diff
@@ -45,7 +45,17 @@ public function update(Request $request)
...
+        $currentMobile = $user->mobile;
+
+        $this->registrar->register($request->all(), $user, function ($user) use ($currentMobile) {
+            // Set the sms_status if we're adding or updating the user's mobile.
+            if ($user->mobile && $user->mobile !== $currentMobile) {
+                $user->sms_status = 'active';
+            }
+        });
+
+        dd($user->sms_status);
```
lindsayrmaher: we should probs remove the debug before merging 🐛

mendelB: whooooops good catch!

PR #943, opened 2019-10-17
DRY values for web registrations 

aaronschachter:
#### What's this PR do?

Branching from #942, this PR adds a `get_default_web_registration_fields` helper to DRY setting default values for web registrations (account created via the register page, Facebook login, or Google login)

👀 

#### Checklist

### DIFF
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -203,4 +203,29 @@ class Registrar
 
         return $user;
     }
+
+    /**
+     * Create a new user OR update an existing user (if given).
+     *
+     * @param array $input - Profile fields
+     * @param User $user - Optionally, user to update
+     * @param Closure $customizer - Customize the user instance before saving.
+     * @return User|null
+     */
+    public function registerViaWeb($input, $user = null, Closure $customizer = null)
+    {
+        return $this->register($input, $user, function() use ($user) {
+            if (! is_null($customizer)) {
+                $customizer($user);
+            }
+            // Set the user's country code by Fastly geo-location header.
+            $user->country = country_code();
+            // Set language based on locale (either 'en', 'es-mx').
+            $user->language = app()->getLocale();
+            // Sign the user up for email messaging & give them the "community" topic.
+            $user->email_subscription_status = true;
+            $user->email_subscription_topics = ['community'];
+            // TODO: Check user source detail and assign feature flags if not clubs member.
+        });
+    }
 }
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -205,27 +205,34 @@ class Registrar
     }
 
     /**
-     * Create a new user OR update an existing user (if given).
+     * Create a new user via web.
      *
      * @param array $input - Profile fields
-     * @param User $user - Optionally, user to update
      * @param Closure $customizer - Customize the user instance before saving.
      * @return User|null
      */
-    public function registerViaWeb($input, $user = null, Closure $customizer = null)
+    public function registerViaWeb($input, Closure $customizer = null)
     {
-        return $this->register($input, $user, function () use ($user) {
-            if (! is_null($customizer)) {
-                $customizer($user);
-            }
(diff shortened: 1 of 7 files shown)


app/helpers.php
```diff
@@ -333,3 +333,21 @@ function format_mobile(PhoneNumber $number, $format): string
 
     return $parser->format($number, $format);
 }
+
+/**
+ * Returns array of fields and values to use for a new user registering via web.
+ *
+ * @return array
+ */
+function get_default_web_registration_fields()
```
DFurnes: This feels like a "kitchen sink" helper, in that it's only in `helpers.php` because it couldn't find a better home. Would this make more sense as a method in the Registrar itself? That'd let us keep all the logic around registration in one place.

Thinking something like this…

```php
public function registerViaWeb($input, $user = null, Closure $customizer = null)
{
    return $this->register($input, $user, function() use ($user) {
        $user->country = country_code();
        $user->language = app()->getLocale();

        $user->email_subscription_status = true;
        $user->email_subscription_topics = ['community'];

        $customizer();
    });
}
```

aaronschachter: I think that makes sense. So our AuthController, FacebookController, and GoogleController would all call `registerViaWeb` instead of `register`. I'll make this change.

This also feels quite applicable to the work @Vmack is looking into for https://www.pivotaltracker.com/n/projects/2401401/stories/169046001 -- ideally the feature flags experiments assignments should go here as well to DRY.

aaronschachter: I ran into an undefined `$customizer` trying to get the sample code to run but I figured that it's simple enough to just assume we're always using `registerViaWeb` to create a new user, and removed the call to `$this->register`.

DFurnes: Ah, oops! That should've been `use ($user, $customizer)` to make that variable available in the closure, but I think splitting them out also makes sense!

app/Http/Controllers/Web/AuthController.php
```diff
@@ -197,23 +197,14 @@ public function postRegister(Request $request)
...
-            $user->email_subscription_topics = ['community'];
-
+        $user = $this->registrar->registerViaWeb($editableFields, function ($user) {
             // Set source_detail, if applicable.
             $sourceDetail = session('source_detail');
             if ($sourceDetail) {
                 $user->source_detail = stringify_object($sourceDetail);
             }
 
+            // TODO: Move this into registerViaWeb.
```
aaronschachter: Refs #946

app/Auth/Registrar.php
```diff
@@ -203,4 +203,37 @@ public function register($input, $user = null, Closure $customizer = null)
...
+
+        // Set the user's country code by Fastly geo-location header.
+        $user->country = country_code();
+        // Set language based on locale (either 'en', 'es-mx').
+        $user->language = app()->getLocale();
+        // Sign the user up for email messaging & give them the "community" topic.
+        $user->email_subscription_status = true;
+        $user->email_subscription_topics = ['community'];
+
+        // TODO: Check user->source_detail and assign feature flags if not clubs member.
```
aaronschachter: We can move the feature flags assignment from the AuthController into this function, so registrations via the register form, Google login, and Facebook login can share the same code.

tests/Http/Web/FacebookTest.php
```diff
@@ -104,6 +104,8 @@ public function testFacebookVerify()
         $this->assertEquals($user->email, 'test@dosomething.org');
         $this->assertEquals($user->source, 'northstar');
         $this->assertEquals($user->source_detail, 'facebook');
+        $this->assertEquals($user->country_code, country_code());
+        $this->assertEquals($user->language, app()->getLocale());
```
aaronschachter: Once we've moved the feature flags assignment into `registerViaWeb` in #946 - we can check to make sure a user's feature flag is set at a minimum by testing:
```
$this->assertObjectHasAttribute('feature_flags', $user);
```

PR #951, opened 2019-10-28
Google Auth birthdate bug fix

mendelB:
#### What's this PR do?
This PR adds a safety check to ensure we received a `birthdays` attribute from the Google /peoples API before running the logic to parse and set the `birthdate` field on the user.

This is a bit of an edge case, but it seems that when a Google profile does _not_ have their birthdate set (though currently, this seems to be a mandatory field on new google accounts, this could be the case for legacy accounts or org domain accounts), we wouldn't receive the `birthdays` field from Google, leading to errors when trying to parse and set the field on the user profile.

^ Disclaimer that this is all something of a hypothesis, validated by trying to login with accounts _without_ a birthday set, it breaking, and then setting the birthday on the account and seeing it work! This should also fix some other edge case related to this field not being returned, but wouldn't inform us if there's something else suspicious going on with the API or whatnot. But figured this is a good easy fix for now!

👁 This should be straightforward logic!

#### Relevant Tickets
https://www.pivotaltracker.com/story/show/169398251

#### Checklist

### DIFF
--- a/app/Http/Controllers/Web/GoogleController.php
+++ b/app/Http/Controllers/Web/GoogleController.php
@@ -68,19 +68,31 @@ class GoogleController extends Controller
         }
 
         $email = $googleUser->email;
-        // Some date properties in this array may not contain a year property.
-        $birthdaysWithYear = array_filter($googleProfile->birthdays, function ($item) {
-            return isset($item->date->year);
-        });
-        $birthday = Arr::first($birthdaysWithYear)->date;
+
+        $birthday = null;
+        // If birthdate is not set on the google profile, we won't receive a 'birthdays' attribute.
+        if (array_key_exists('birthdays', $googleProfile)) {
+            // Some date properties in this array may not contain a year property.
+            $birthdaysWithYear = array_filter($googleProfile->birthdays, function ($item) {
+                return isset($item->date->year);
+            });
+            $birthday = Arr::first($birthdaysWithYear)->date;
+        }
 
         // Aggregate Google profile fields.
         $fields = [
             'google_id' => $googleUser->id,
             'first_name' => $googleUser->user['given_name'],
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/Web/GoogleController.php
```diff
@@ -68,20 +68,32 @@ public function handleProviderCallback()
...
         $email = $googleUser->email;
-        // Some date properties in this array may not contain a year property.
-        $birthdaysWithYear = array_filter($googleProfile->birthdays, function ($item) {
-            return isset($item->date->year);
-        });
-        $birthday = Arr::first($birthdaysWithYear)->date;
+
+        $birthday = null;
+        // If birthdate is not set on the google profile, we won't receive a 'birthdays' attribute.
+        if (array_key_exists('birthdays', $googleProfile)) {
```
weerd: Just curious but why does a user have multiple birthdays from Google?

katiecrane: This was something that Aaron was seeing at least with his account (multiple birthdays, at least one of which did not have a year) so we decided to account for it

weerd: Yeah, just curious whyyyy? Seems weird it's returning multiple birthdays and wondering what the explanation is from Google... 🤔

katiecrane: An excellent question! Here's [the original thread](https://github.com/DoSomething/northstar/pull/930#discussion_r335201032) about it which includes a link to Google's docs (different from Google docs 😂 ) but they don't really explain it

weerd: Interesting... 🤔 Ok, thanks for the info @katiecrane 👍

PR #952, opened 2019-10-28
Google Authentication - Confirm Required User Attributes

mendelB:
#### What's this PR do?
Aborts the Google authentication flow if the Google API returns a user missing any key attributes.

This should help prevent any breakage resulting from a user missing any of these attributes we grab when creating new accounts. (E.g. you can unset your surname on your Google account and the `family_name` would be missing from the User payload).

It also moves the redirect logic into a `protected` function since we're using it twice now.

Does this make sense? Did I put the protected function in the right place?

#### Relevant Tickets
https://www.pivotaltracker.com/story/show/169398251

#### Checklist

### DIFF
--- a/app/Http/Controllers/Web/GoogleController.php
+++ b/app/Http/Controllers/Web/GoogleController.php
@@ -28,9 +28,9 @@ class GoogleController extends Controller
      *
      * @return \Illuminate\Http\RedirectResponse
      */
-    protected function redirectUnsuccessfulRequest()
+    protected function redirectUnsuccessfulRequest($message = null)
     {
-        return redirect('/register')->with('status', 'Unable to verify Google account.');
+        return redirect('/register')->with('status', $message ?: 'Unable to verify Google account.');
     }
 
     /**
@@ -86,7 +86,7 @@ class GoogleController extends Controller
 
         // If this is a new registration, ensure we've received the required profile fields.
         if (! $northstarUser && (! $firstName || ! $lastName)) {
-            return $this->redirectUnsuccessfulRequest();
+            return $this->redirectUnsuccessfulRequest('We need your first and last name to create your account! Please confirm that these are set on your Google profile and try again.');
         }
 
         $birthday = null;
--- a/app/Http/Controllers/Web/GoogleController.php
+++ b/app/Http/Controllers/Web/GoogleController.php
@@ -26,7 +26,7 @@ class GoogleController extends Controller
     /**
      * Redirect unsuccessful authentication requests.
      *
-     * @return Redirect
+     * @return \Illuminate\Http\RedirectResponse
      */
     protected function redirectUnsuccessfulRequest()
     {
--- a/app/Http/Controllers/Web/GoogleController.php
+++ b/app/Http/Controllers/Web/GoogleController.php
@@ -24,6 +24,16 @@ class GoogleController extends Controller
     protected $registrar;
 
     /**
+     * Redirect unsuccessful authentication requests.
+     *
+     * @return Redirect
+     */
+    protected function redirectUnsuccessfulRequest()
+    {
+        return redirect('/register')->with('status', 'Unable to verify Google account.');
+    }
+
+    /**
      * Make a new GoogleController, inject dependencies,
      * and set middleware for this controller's methods.
      *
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/Web/GoogleController.php
```diff
@@ -64,7 +74,14 @@ public function handleProviderCallback()
...
             logger()->warning('google_token_mismatch');
 
-            return redirect('/register')->with('status', 'Unable to verify Google account.');
+            return $this->redirectUnsuccessfulRequest();
+        }
+
+        // Ensure all required user attributes are available on the given profile.
+        foreach (['given_name', 'family_name'] as $requiredAttribute) {
+            if (! array_key_exists($requiredAttribute, $googleUser->user)) {
+                return $this->redirectUnsuccessfulRequest();
```
aaronschachter: It might potentially be more helpful to indicate "We need your first name" vs Unable to verify here.

I'm also thinking that if a Google user is able to unset their last name, we shouldn't require it - as we don't require on our register form. cc @ngjo for confirmation

I was following by example from our FacebookController for the Google login, but we began changing up some logic (e.g. #942). I'd bet Facebook requires last name so we don't need to make the change there.

mendelB: > It might potentially be more helpful to indicate "We need your first name" vs Unable to verify here.

That's a good idea! 

> I'm also thinking that if a Google user is able to unset their last name, we shouldn't require it - as we don't require on our register form. cc @ngjo for confirmation

We do indeed require it on the register form https://github.com/DoSomething/northstar/blob/a49c138dbc757b534cb37022f61187cf758deaf2/app/Http/Controllers/Web/AuthController.php#L193

Though that does raise the issue of us redirecting an _existing_ user who's missing these properties on Google, and / or is merely logging in!

mendelB: > It might potentially be more helpful to indicate "We need your first name" vs Unable to verify here.

That's a good idea! 

> I'm also thinking that if a Google user is able to unset their last name, we shouldn't require it - as we don't require on our register form. cc @ngjo for confirmation

We do indeed require it on the register form https://github.com/DoSomething/northstar/blob/a49c138dbc757b534cb37022f61187cf758deaf2/app/Http/Controllers/Web/AuthController.php#L193

Though that does raise the issue of us redirecting an _existing_ user who's missing these properties on Google, and / or is merely logging in!

ngjo: > I'm also thinking that if a Google user is able to unset their last name, we shouldn't require it - as we don't require on our register form. cc @ngjo for confirmation

We actually do require last name now per a request from management. Here's where we made that change: https://dosomething.slack.com/archives/C0KDV4YES/p1555606467015900

aaronschachter: Oops -- the times they are a-changin. Thanks for taking on this work when I was out @mendelB !

mendelB:
**Updates:**

- Ensuring we only validate the required first and last name fields for _new_ users registering in https://github.com/DoSomething/northstar/pull/952/commits/8d6d5e80c07de0bc31ff003a73e32ad06ae1f74b
- Update the validation flash message for these missing profile fields to be more helpful https://github.com/DoSomething/northstar/pull/952/commits/1f7bd7c37487af191269d821fa44ae7c77064d84
- I noticed [an error in Papertrail](https://my.papertrailapp.com/groups/10447062/events?focus=1126007075316183057&q=%28sender%3Adosomething-northstar%20%28production.ERROR%20OR%20status%3D5%29%29%20OR%20%28sender%3Afastly%20status%3D5%20app%3Ddosomething-northstar%29&selected=1126007075316183057) related to us [retrieving the `date` field](https://github.com/DoSomething/northstar/blob/d9fbd47b39d3ad3926e55c14005259e685fdbfcb/app/Http/Controllers/Web/GoogleController.php#L79) from the filtered `birthdays`: `Trying to get property 'date' of non-object` (presumabely, this'd happen when we get _no_ year at all within the birthdays array), so safely retrieving the field in https://github.com/DoSomething/northstar/pull/952/commits/8c18b59d1e9ec298408fb97a3c077b10cb8e920d
- And finally, added test cases for the missing profile and birthdays fields in https://github.com/DoSomething/northstar/pull/952/commits/d4dd7cddd918b64e29ee226e87fd6db283c2082c (Since these required specialized mocking, I couldn't really make use of the [`defaultMocks`](https://github.com/DoSomething/northstar/blob/d4dd7cddd918b64e29ee226e87fd6db283c2082c/tests/Http/Web/GoogleTest.php#L92-L104) helper, so there's a bit of reppetition happening for the mocking in these tests. I spent some time trying to finesse the helpers to be more flexible but didn't feel great about my results so leaving this static for now.)

app/Http/Controllers/Web/GoogleController.php
```diff
@@ -64,26 +74,36 @@ public function handleProviderCallback()
...
         $email = $googleUser->email;
 
+        $northstarUser = $this->registrar->resolve(['email' => $email]);
+
+        $firstName = data_get($googleUser->user, 'given_name');
+        $lastName = data_get($googleUser->user, 'family_name');
+
+        // If this is a new registration, ensure we've received the required profile fields.
+        if (! $northstarUser && (! $firstName || ! $lastName)) {
+            return $this->redirectUnsuccessfulRequest('We need your first and last name to create your account! Please confirm that these are set on your Google profile and try again.');
```
aaronschachter: Beautiful 🖼

PR #954, opened 2019-10-31
Adds school_id field

aaronschachter:
#### What's this PR do?

Adds a `school_id` text field to be used by the School Finder component in Phoenix (https://github.com/DoSomething/phoenix-next/pull/1668)

👀 

#### Relevant Tickets

https://www.pivotaltracker.com/n/projects/2401401/stories/169294661/

#### Checklist

### DIFF
--- a/app/Http/Transformers/UserTransformer.php
+++ b/app/Http/Transformers/UserTransformer.php
@@ -73,6 +73,7 @@ class UserTransformer extends BaseTransformer
 
             $response['interests'] = [];
             $response['age'] = $user->age;
+            $response['school_id'] = $user->school_id;
 
             $response['addr_city'] = $user->addr_city;
             $response['addr_state'] = $user->addr_state;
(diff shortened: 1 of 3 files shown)


aaronschachter:
That's correct! We'll save the Great Schools `gsid` as the user `school_id` if they select a school, and save a string like `not-available` if they are unable to find their school in the dropdown.

mshmsh5000:
Here's a database view:

Image 2019-10-31 at 11 59 59 AM

aaronschachter:
I've been assuming we should save `gs-id`, as we've historically done. Universal ID is a new player in the mix, do you think we're ok with sticking with `gs-id` @mshmsh5000 ?

mshmsh5000:
In the Great Schools data dictionary, `gs-id` is

>Unique within each state; also included in GreatSchools' school/district overview page URL string

whereas `universal-id` seems to be unique within the GS data set:

>ID for data matching between feed products

The full data dictionary is on the Shared Drive: Data -> Vendors -> Great Schools -> 2019-10

PR #957, opened 2019-11-07
Update Oauth redirects

lindsayrmaher:
#### What's this PR do?

We are adding an additional check to redirect users to complete their profile after initially signing up for the site via Oauth. 

Will an existing user be redirected as expected to the homepage, while a new user gets directed to the second step of the new registration flow to complete their profile?

#### Relevant Tickets
Ticket Number [169460212](https://www.pivotaltracker.com/story/show/169460212)

#### Checklist


tests/Http/Web/FacebookTest.php
```diff
@@ -125,7 +125,7 @@ public function testFacebookPublicProfileFieldsAreSet()
     {
         $this->defaultMock();
 
-        $this->visit('/facebook/verify')->seePageIs('/');
+        $this->visit('/facebook/verify')->seePageIs('/profile/about');
```
mendelB: I wonder if it's worth throwing a `->seePageIs('/')` in the `testFacebookAccountMerge` test [below](https://github.com/DoSomething/northstar/pull/957/files#diff-98d29e4c6b77c7c945f5c07d7cf853fbR173) and same in the Google test, to make sure our redirects are working properly?

lindsayrmaher: i don't think it would hurt. probably smart to double check!

aaronschachter:
Somewhat related to this PR -- one tweak to make on the new profile/registration page per Google auth is pre-populating the user's birthday field if it's set (see [Slack thread](https://dosomething.slack.com/archives/CP2D7UGAU/p1572462928039900?thread_ts=1572459806.031100&cid=CP2D7UGAU)).

lindsayrmaher:
@aaronschachter 

if we're setting the value of this field to their old birthdate, it should pre-populate with their b-day after the oauth is successful and redirects right? Or is there more work we would need to do for updating this?

https://github.com/DoSomething/northstar/blob/c3f632d81c4d80465790dbb4a88f15376ad8ed96/resources/views/profiles/about/edit.blade.php#L29

aaronschachter:
Yup, I don't think there's any additional work besides setting default value if user already has a birthdate.

PR #959, opened 2019-11-13
Subscriptions endpoint

katiecrane:
#### What's this PR do?
Creates the `/subscriptions` endpoint in Northstar!

- This endpoint is public, no auth
- Rate-limited to 10 requests per hour per IP
- Accepts: `email`, `email_subscription_topics`, `source`, `source_detail`
    - Tries to find a user with the `email` and if it does adds the given `email_subscription_topics`
    - If no user is found, creates a user with the given `email`, `email_subscription_topics`, `source`, `source_detail`
 
New users also receive an activate account email of type `pays-to-do-good-activate-account`. In order to send the account activation email, I had to steal the `createTokenRepository()` function from the `ResetController`. I considered extracting it out to a trait, but I find traits kind of hard to use because it can be tricky to find where the functions are defined. I also considered pull it out as a helper, but not sure it makes sense as a helper either? I'd love thoughts on this!

There are tests and docs added for this endpoint as well!

For use with Newsletter CTAs in Phoenix!

There is a question in the area above, but I think the tests should have us covered so make sure those get a good 👀 

#### Relevant Tickets
[Card](https://www.pivotaltracker.com/story/show/169330972)

#### Checklist
- [X] Documentation added for changed endpoints.

### DIFF
--- a/app/Http/Controllers/SubscriptionController.php
+++ b/app/Http/Controllers/SubscriptionController.php
@@ -42,7 +42,7 @@ class SubscriptionController extends Controller
      * @param Request $request
      * @return array
      */
-    public function create(Request $request) 
+    public function create(Request $request)
     {
         $this->validate($request, [
             'email' => 'required|email',
@@ -57,7 +57,7 @@ class SubscriptionController extends Controller
         // If the user already exists, only update the email topics
         if ($existingUser) {
             foreach ($request->get('email_subscription_topics') as $topic) {
-                $existingUser->addEmailSubscriptionTopic($topic);   
+                $existingUser->addEmailSubscriptionTopic($topic);
             }
 
             $existingUser->save();
--- a/app/Http/Controllers/SubscriptionController.php
+++ b/app/Http/Controllers/SubscriptionController.php
@@ -46,8 +46,7 @@ class SubscriptionController extends Controller
     {
         $this->validate($request, [
             'email' => 'required|email',
-            'email_subscription_topics' => 'required',
-            'email_subscription_topics.*' => 'in:news,scholarships,lifestyle,community',
+            'email_subscription_topic' => 'required|in:news,scholarships,lifestyle,community',
             'source' => 'required',
             'source_detail' => 'required',
         ]);
@@ -56,9 +55,7 @@ class SubscriptionController extends Controller
 
         // If the user already exists, only update the email topics
         if ($existingUser) {
-            foreach ($request->get('email_subscription_topics') as $topic) {
-                $existingUser->addEmailSubscriptionTopic($topic);
-            }
+            $existingUser->addEmailSubscriptionTopic($request->get('email_subscription_topic'));
 
             $existingUser->save();
 
@@ -67,6 +64,7 @@ class SubscriptionController extends Controller
 
         $newUser = $this->registrar->register($request->all());
 
+        $newUser->email_subscription_topics = [$request->get('email_subscription_topic')];
         $newUser->email_subscription_status = true;
         $newUser->source = $request->get('source');
         $newUser->source_detail = $request->get('source_detail');
--- a/app/Http/Controllers/SubscriptionController.php
+++ b/app/Http/Controllers/SubscriptionController.php
@@ -6,7 +6,6 @@ use Northstar\Models\User;
 use Illuminate\Http\Request;
 use Northstar\Auth\Registrar;
 use Northstar\Http\Transformers\UserTransformer;
-use Jenssegers\Mongodb\Auth\DatabaseTokenRepository;
 
 class SubscriptionController extends Controller
 {
@@ -90,21 +89,5 @@ class SubscriptionController extends Controller
         }
 
         return $this->item($newUser, 201);
-    }
-
-    /**
-     * Create a token repository instance based on the given configuration.
-     *
-     * @return DatabaseTokenRepository
-     */
-    protected function createTokenRepository()
-    {
-        return new DatabaseTokenRepository(
-            app('db')->connection(),
-            app('hash'),
-            config('auth.passwords.users.table'),
-            config('app.key'),
-            config('auth.passwords.users.expire')
-        );
     }
 }
--- a/app/Http/Controllers/SubscriptionController.php
+++ b/app/Http/Controllers/SubscriptionController.php
@@ -23,8 +23,8 @@ class SubscriptionController extends Controller
      */
     protected $transformer;
 
-     /**
-     * Make a new ResetController, inject dependencies,
+    /**
+     * Make a new SubscriptionController, inject dependencies,
      * and set middleware for this controller's methods.
      */
     public function __construct(Registrar $registrar, UserTransformer $transformer)
@@ -42,7 +42,8 @@ class SubscriptionController extends Controller
      * @param Request $request
      * @return array
      */
-    public function create(Request $request) {
+    public function create(Request $request) 
+    {
         $this->validate($request, [
             'email' => 'required|email',
             'email_subscription_topics' => 'required',
@@ -50,12 +51,12 @@ class SubscriptionController extends Controller
             'source' => 'required',
             'source_detail' => 'required',
         ]);
-        
+
         $existingUser = $this->registrar->resolve($request->only('email'));
 
         // If the user already exists, only update the email topics
         if ($existingUser) {
-            foreach($request->get('email_subscription_topics') as $topic) {
+            foreach ($request->get('email_subscription_topics') as $topic) {
                 $existingUser->addEmailSubscriptionTopic($topic);   
             }
 
@@ -68,7 +69,7 @@ class SubscriptionController extends Controller
 
         $newUser->email_subscription_status = true;
         $newUser->source = $request->get('source');
-        $newUser->source_detail =  $request->get('source_detail');
+        $newUser->source_detail = $request->get('source_detail');
 
         $newUser->save();
 
--- a/app/Http/Controllers/SubscriptionController.php
+++ b/app/Http/Controllers/SubscriptionController.php
@@ -1,0 +1,75 @@
+<?php
+
+namespace Northstar\Http\Controllers;
+
+use Northstar\Models\User;
+use Illuminate\Http\Request;
+use Northstar\Auth\Registrar;
+use Northstar\Http\Transformers\UserTransformer;
+
+class SubscriptionController extends Controller
+{
+    /**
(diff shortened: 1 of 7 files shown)


app/Http/Controllers/SubscriptionController.php
```diff
@@ -0,0 +1,99 @@
...
+        $newUser->email_subscription_status = true;
+        $newUser->source = $request->get('source');
+        $newUser->source_detail = $request->get('source_detail');
+
+        $newUser->save();
+
+        // Send activate account email to new user
+        $tokenRepository = $this->createTokenRepository();
+        $token = $tokenRepository->create($newUser);
+        $message = $newUser->sendPasswordReset($token, 'pays-to-do-good-activate-account');
```
aaronschachter: This should reference `PasswordResetType::$paysToDoGoodActivateAccount` to DRY:
https://github.com/DoSomething/northstar/blob/c3f632d81c4d80465790dbb4a88f15376ad8ed96/app/PasswordResetType.php#L33

As author of that helper, and the email subscription import in Chompy, I'm confused why we're sending this template for any subscription. Shouldn't the email change based on the topic a user is subscribing to?

Per the example docs, if subscribing to `lifestyle`, per the Chompy import I'd expect user to receive the `boost-activate-account` reset type:
https://github.com/DoSomething/chompy/blob/b15cbe0829eb8b6ab42ae9693e012c0012e2bcbd/config/import.php#L15

aaronschachter: Per your question about repeating code from the `ResetController`, I wonder about adding a function named something like `send` into the PasswordResetType class to DRY, e.g: 
```
function send($user, $passwordResetType) {
    $tokenRepository = $this->createTokenRepository();
    $token = $tokenRepository->create($user);
    return $user->sendPasswordReset($token, $passwordResetType);
}

File: app/Http/Controllers/SubscriptionController.php [hunk 2]
```diff
@@ -0,0 +1,99 @@
...
+     * @return DatabaseTokenRepository
+     */
+    protected function createTokenRepository()
+    {
+        return new DatabaseTokenRepository(
+            app('db')->connection(),
+            app('hash'),
+            config('auth.passwords.users.table'),
+            config('app.key'),
+            config('auth.passwords.users.expire')
```
aaronschachter: This could be moved into the helper that'd be used across both files as well to DRY?

documentation/endpoints/subscriptions.md
```diff
@@ -0,0 +1,85 @@
+# Subscriptions Endpoint
+
+This endpoint has no authentication, but is rate limited to 10 requests per hour per IP.
+
+## Add a subscription topic to a user by email
+
+Tries to find a user by email, and creates a new user if one is not found. If the user already exists, the given `email_subscription_topics` are added to the user. If a new user was created, the given `email_subscription_topics`, `source`, and `source_detail` are set on the user.
+
```
aaronschachter: We should also spell out that this will send an activate account email to new users.

app/Http/Controllers/SubscriptionController.php
aaronschachter: Thining about this more and what if we moved the creating the DatabaseTokenRepository and generating token into `sendPasswordReset` to DRY (instead of the `PasswordResetType` -- which is currently more of just a lookup table)

katiecrane: Oh shoot @aaronschachter you're right about sending a different `PasswordResetType` based on the topic added - I have it with just `paysToDoGoodActivateAccount` right now because that's all we'll be receiving for now. BUT since we should only be sending one activate account email, I'm now thinking this endpoint should only be able to receive ONE topic, rather than multiple topics. What are your thoughts on this?

Do you mean you're leaning towards moving the DatabaseTokenRepository and generating token into the `SendPasswordResetToCustomerIo` job?

documentation/endpoints/subscriptions.md
katiecrane: GREAT CATCH!!

app/Http/Controllers/SubscriptionController.php
aaronschachter: >  I'm now thinking this endpoint should only be able to receive ONE topic, rather than multiple topics

Agreed!  Now that all email subscription topics have Activate Account emails configured, we changed the Chompy import from in https://github.com/DoSomething/chompy/pull/111 to only send a single topic to avoid any multiple email mistakes. It'd be to make Northstar the source of truth for the subscription-topic / activate-account-email template configuration, and refactor Chompy to post to this endpoint instead (although ideally the widgets should deprecate the CSV import anyway?)

For where to create a `$token` and DRY, sorry - I should have linked to the User  `sendPasswordReset` method:
https://github.com/DoSomething/northstar/blob/0a8fa7d90fd64b57930015ec20a71f22e62f83a9/app/Models/User.php#L648

We could alter this method to make `$token` an optional argument, and generate the token here if we don't receive one.  `sendPasswordReset` gets called from the Forgot Password page on the web (via overriding the User  [`sendPasswordResetNotification`](https://laravel.com/api/5.8/Illuminate/Auth/Passwords/CanResetPassword.html#method_sendPasswordResetNotification)) and the  `POST /resets` and `POST /subscriptions` API requests, where we need to generate a token in order to call `sendPasswordReset` . What do you think about that?

It got me wondering whether a `POST /subscriptions` essentially deprecates our `POST /resets` endpoint.. but resets was around to be able to trigger a password reset email via the mobile app. So even if we no longer need the Chompy email subscription email import,I could see life for `POST /resets` should we ever want to support non web-apps triggering a Forgot Password email (Gambit?)

katiecrane: Ah thank you! Okay so this endpoint will accept only ONE email topic per request.

I LOVE the idea about using `sendPasswordReset` that way. That way the token repository stuff can all happen right there within that method if it needs a token yeah?

Yeah I think it'd be cool to keep `/resets` around - I could for sure see that being used down the line in Gambit to help get people over to the web.

documentation/endpoints/subscriptions.md
katiecrane: updated!

app/Http/Controllers/SubscriptionController.php
katiecrane: updated - everything is now in `sendPasswordReset` on the user model!

katiecrane:
@aaronschachter I think I got all the updates we discussed in now - tests are all passing so I don't think any of the refactoring mucked anything up!

tests/Http/SubscriptionsTest.php
```diff
@@ -0,0 +1,141 @@
...
+    {
+        $this->json('POST', 'v2/subscriptions', [
+            'email' => 'topics@dosomething.org',
+            'email_subscription_topic' => 'scholarships',
+            'source' => 'phoenix-next',
+            'source_detail' => 'test_source_detail',
+        ]);
+
+        $this->assertResponseStatus(201);
+        $this->blinkMock->shouldHaveReceived('userCallToActionEmail')->once();
```
aaronschachter: Nice to have for a future PR could be to test that the relevant template is sent for each topic.

tests/Http/SubscriptionsTest.php
```diff
@@ -0,0 +1,141 @@
...
+                'email' => 'topics'.$i.'@dosomething.org',
+                'email_subscription_topic' => 'news',
+                'source' => 'phoenix-next',
+                'source_detail' => 'test_source_detail',
+            ]);
+
+            $this->assertResponseStatus(201);
+        }
+
+        // Get a "Too Many Attempts" response when trying for an 11th post within an hour
```
aaronschachter: Very nice 💯

app/Http/Controllers/SubscriptionController.php
katiecrane: AH yes - I got the one in `ResetController` but missed this one 🙈

tests/Http/SubscriptionsTest.php
katiecrane: Agreed, I fiddle a bit trying to get that to work but didn't figure it out!

app/Http/Controllers/SubscriptionController.php
```diff
@@ -0,0 +1,110 @@
...
+        $newUser->email_subscription_status = true;
+        $newUser->source = $request->get('source');
+        $newUser->source_detail = $request->get('source_detail');
+
+        $newUser->save();
+
+        // Send activate account email to new user
+        switch ($topic) {
+            case 'scholarships':
+                $newUser->sendPasswordReset('pays-to-do-good-activate-account');
```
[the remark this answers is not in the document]
katiecrane: Ah yeah, that's the whole point! Updated this!

documentation/endpoints/subscriptions.md
```diff
@@ -0,0 +1,89 @@
...
+  /* The email subscription topics to add to the user.
+   *
+   * Valid topics:
+   * - 'lifestyle'
+   * - `news`
+   * - 'scholarships'
+   * - `community`
+   */
+  email_subscription_topics: Array;
```
aaronschachter: 1️⃣ - `email_subscription_topic`

documentation/endpoints/subscriptions.md
```diff
@@ -0,0 +1,89 @@
...
+<summary><strong>Example Request</strong></summary>
+
+```sh
+  curl -X "POST" "http://northstar.test/v2/subscriptions" \
+     -H 'Accept: application/json' \
+     -H 'Content-Type: application/json; charset=utf-8' \
+     -d $'{
+  "email": "funner@dosomething.org",
+  "source_details": "subscription-page",
+  "email_subscription_topics": [
```
aaronschachter: 1️⃣

katiecrane:
> For vLater, it'd be sweet to add support for optional first_name parameter, as well as a Chompy override for throttling. We could 🔥all the code over in the ImportEmailSubscription job by posting to this new endpoint instead.

@aaronschachter yeah! it'd be cool to add `first_name` or whatever else we want to collect. 

OOO That Chompy cleanup/throttle override sounds like a real good tech debt card - I added making that card to my to-do list.

app/Http/Controllers/SubscriptionController.php
katiecrane: I made a card for using this endpoint from Chompy: https://www.pivotaltracker.com/n/projects/2401401/stories/169786620#

PR #963, opened 2019-11-22
Refer Friends Feature Flag - Scholarship Edition!

mendelB:
#### What's this PR do?
Updates all references to the `refer-friends` feature flag to the `refer-friends-scholarship` feature flag!

I figured it'd be neater to just replace for now, and if we ever wanted to do a side by side test or wtvr down the pike, we can just add it back in instead of both cluttering our codebase.

1. Toggle the `DS_REFER_FRIENDS_SCHOLARSHIP_TEST` to true in `.env` 
2. Create a new user via web
3. Do they have the `refer-friends-scholarship` feature flag?
4. Did Customer.io get this field in the Payload?

#### Relevant Tickets

[Pivotal card](https://www.pivotaltracker.com/story/show/169376190)

#### Checklist

- [ ] Documentation added for new or changed endpoints.
- [x] Tests added for new features/bug fixes/API updates.
- [ ] Post a message in #dev-northstar if this includes something that causes a rebuild!


ngjo:
I have a silly question -- would this mean that people who had `refer-friends` be opted into this scholarship experience too and how would we distinguish between the two groups (`refer-friends` = gift card and these people are scholarship entries)?

mendelB:
> I have a silly question -- would this mean that people who had `refer-friends` be opted into this scholarship experience too and how would we distinguish between the two groups (`refer-friends` = gift card and these people are scholarship entries)?

Fair question! Thankfully, this will only apply to _new_ web registrants, so no existing users should be affected.

mendelB:
Though, this made me

> > I have a silly question -- would this mean that people who had `refer-friends` be opted into this scholarship experience too and how would we distinguish between the two groups (`refer-friends` = gift card and these people are scholarship entries)?
> 
> Fair question! Thankfully, this will only apply to _new_ web registrants, so no existing users should be affected.

Though, this made me realize that if we remove the `refer-friends` from the C.io blink payload, that flag won't get send to c.io if we ever updated a user manually with that flag, or backfilled or something. I think I need to add that field back in @DFurnes ?

DFurnes:
Ah, yeah, good point. I didn't think about opting people into that flag after the fact!

mendelB:
> Ah, yeah, good point. I didn't think about opting people into that flag after the fact!

Like I soooort of doubt we would, but I guess it's a fair possibility so doesn't hurt? (Out of curiosity @DFurnes , if we ever update a user and their attributes are sent to c.io but missing a field that's already been set on c.io, would that field be deleted or is it just a merge they're doing, and the field would remain un-updated?) 

Anyways, I'll add it back in, and merge!

DFurnes:
> Out of curiosity @DFurnes , if we ever update a user and their attributes are sent to c.io but missing a field that's already been set on c.io, would that field be deleted or is it just a merge they're doing, and the field would remain un-updated?

Unless we explicitly send `null` for that field, it just merges with the existing profile. 😌

mendelB:
> > Out of curiosity @DFurnes , if we ever update a user and their attributes are sent to c.io but missing a field that's already been set on c.io, would that field be deleted or is it just a merge they're doing, and the field would remain un-updated?
> 
> Unless we explicitly send `null` for that field, it just merges with the existing profile. 😌

Ah golden! Ok, so I guess worth it just for the manual opting-in... It can't hurt

PR #969, opened 2019-12-13
Expand badges configuration - allow campaigns to be opted out

katiecrane:
### What's this PR do?

This pull request creates a new environment variable `DS_CONTENTFUL_IDS_FOR_CAMPAIGNS_WITH_NO_BADGES`. The variable takes a comma separated list of Contentful IDs.

Currently when a user registers from a campaign page, we save the Contentful ID of the campaign in their `source_detail`. Now, if their `source_detail` contains one of these excluded Contentful IDs, we will NOT opt them in to the badges experiment. They will have NO `feature_flags` value for `badges`.

Is this the best way to accomplish this goal? Does this provide a way to test on QA and also configure on prod?

We don't want to give teachers badges.

### Relevant tickets

References [Pivotal #170284773](https://www.pivotaltracker.com/story/show/170284773).

### Checklist


katiecrane:
@DFurnes I'd like to get your thoughts on this as well given our feature flag discussion in retro yesterday!

config/features.php
```diff
@@ -23,4 +23,6 @@
     'optional-fields' => env('DS_OPTIONAL_FIELDS', false),
 
     'refer-friends-scholarship' => env('DS_REFER_FRIENDS_SCHOLARSHIP_TEST', false),
+
+    'no-badge-campaigns' => explode(',', env('DS_CONTENTFUL_IDS_FOR_CAMPAIGNS_WITH_NO_BADGES', null)),
```
DFurnes: This feels a little awkward, since it's sorta an "anti" feature flag. I wonder if it'd make sense to group with the `badges` one above, as like `badges-excludes` (so it's clearer that they're related to each other)?

It could also make sense to have each feature flag be an object that allowed multiple customizable parameters, but that's maybe more refactoring than we want to do right before a holiday break:

```php
'refer-friends-scholarship' => [
    'enabled' => env('DS_REFER_FRIENDS_SCHOLARSHIP_TEST', false),
],

'badges' =>  [
    'enabled' => env('DS_BADGES_TEST', false),
    'excluded-campaigns' => explode(',', env('DS_CONTENTFUL_IDS_FOR_CAMPAIGNS_WITH_NO_BADGES', null)),
]
```

app/Auth/Registrar.php
```diff
@@ -233,11 +233,21 @@ public function registerViaWeb($input, Closure $customizer = null)
         // Exclude any 'clubs' referrals from our feature flag tests.
         if (! str_contains($user->source_detail, 'utm_source:clubs')) {
             $feature_flags = [];
-            // If the badges test is running, sort users into badges group control group
-            if (config('features.badges')) {
+
+            // Did the user come from a campaign that should not have badges?
+            $badgesEligible = true;
+            foreach (config('features.no-badge-campaigns') as $noBadgeCampaigns) {
```
DFurnes: I think we can remove this `foreach`, since you can pass the array directly to [`str_contains`](https://laravel.com/docs/5.5/helpers#method-str-contains):

```php
$badgesExcluded = str_contains($user->source_detail, config('features.no-badge-campaigns'));
$badgesEligible = ! $badgesExcluded;
```

.env.example
```diff
@@ -16,6 +16,7 @@ DS_ENABLE_RATE_LIMITING=false
 DS_OPTIONAL_FIELDS=false
 DS_BADGES_TEST=false
 DS_REFER_FRIENDS_SCHOLARSHIP_TEST=false
+DS_CONTENTFUL_IDS_FOR_CAMPAIGNS_WITH_NO_BADGES=
```
DFurnes: I love how verbose & clear this environment variable is. 😌

katiecrane: lol I was like well this is long but we'll definitely know what it is!!

config/features.php
katiecrane: I really like this grouping idea, I'm adding this to our list of cleanup work for the start of the year (since I agree it's more than we want to do today).

katiecrane: and added!

app/Auth/Registrar.php
katiecrane: ah I thought that function only took 2 strings, but after testing it does work with arrays! rad, thank you!

config/features.php
ngjo: documenting that this work is ticketed [here](https://www.pivotaltracker.com/n/projects/2401401/stories/170558705)

PR #970, opened 2020-01-07
Remove Puck 😭

mendelB:
### What's this PR do?

This pull request removes the Puck library and analytics tracking from Northstar.

Should be straightforward! Just make sure I didn't muck anything else up through my tears.

🥃 
We've moved on to tracking with Snowplow and GTM!

### Relevant tickets

References [Pivotal #167633809](https://www.pivotaltracker.com/story/show/167633809).
See https://github.com/DoSomething/phoenix-next/pull/1837

### Checklist

### DIFF
--- a/resources/assets/js/utilities/Analytics.js
+++ b/resources/assets/js/utilities/Analytics.js
@@ -1,7 +1,6 @@
 const $ = require('jquery');
 const queryString = require('query-string');
 const Validation = require('dosomething-validation');
-const { Engine } = require('@dosomething/puck-client');
 const {
   flattenDeep,
   isNil,
@@ -14,9 +13,6 @@ const {
 
 // App name prefix used for analytics event naming.
 const APP_PREFIX = 'northstar';
-
-// Variable that stores the instance of PuckClient.
-let puckClient = null;
 
 /**
  * Get the query-string value at the given key.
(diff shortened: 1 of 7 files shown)


resources/assets/js/utilities/Analytics.js
```diff
@@ -222,17 +190,16 @@ export function analyzeWithSnowplow(name, category, action, label, data) {
  */
 const sendToServices = (name, category, action, label, data, service) => {
   switch (service) {
-    case 'ga':
+    case 'google':
```
katiecrane: I didn't see any other updates in the code that changed this - are we currently never sending `ga` as the service anywhere?

mendelB: Ah sorry, I didn't provide context for this change -- this is kind of unrelated but brings this method up to par with [Phoenix analytics](https://github.com/DoSomething/phoenix-next/blob/master/resources/assets/helpers/analytics.js#L128)
(And may we merit to see the day when this is consolidated into its own package so we don't need to manually keep these consistent🤞 🙏  )

katiecrane: Perfect, thanks for keeping things consistent 👍 !!

mendelB:
> I feel like I need to apologize for approving this 😭

🙌Class act. 😢

PR #978, opened 2020-02-04
Include source_detail session params for FB/Google registrations

aaronschachter:
### What's this PR do?

This pull request moves the check for a session `source_detail` into the  `registerViaWeb` method , which has been refactored to accept an optional `auth_source` property via the Facebook and Google controllers.  This keeps the `source_detail` of users who authenticate via FB or Google to be consistent with users who register via the Northstar registration form (a stringified object with optional UTM, Contentful, Referrer params) .

👀 

ℹ️ 

### Relevant tickets

References [Pivotal #169213761](https://www.pivotaltracker.com/n/projects/2401401/stories/169213761/).

### Checklist

### DIFF
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -217,9 +217,15 @@ class Registrar
 
         $user->fill($input);
 
-        // The user's source_detail is often set via this customizer parameter.
         if (! is_null($customizer)) {
             $customizer($user);
+        }
+
+        // Set source_detail to the session source_detail if we haven't set one yet.
+        $sourceDetail = session('source_detail');
+
+        if ($sourceDetail && ! isset($user->source_detail)) {
+            $user->source_detail = stringify_object($sourceDetail);
         }
 
         // Set the user's country code by Fastly geo-location header.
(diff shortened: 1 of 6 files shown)


aaronschachter:
@katiecrane @ngjo -- After working in #979 and getting to know the `SubscriptionController` a bit,  it seems we'll be missing these UTM parameters in the `source_detail` of any new registrations via the Newsletter Subscription widget. We wouldn't be able to call this`registerViaWeb` there, since we're not redirecting the user to the registration page.. so maybe it'd be best resolved by the [Phoenix component passing over the stringified UTM params, as well as its current value, in the `source_detail`when it makes a `POST /subscriptions` request?](https://github.com/DoSomething/phoenix-next/blob/8066e69184b3be00e0b9489d2c73ca6907dfdc35/resources/assets/components/utilities/CtaPopover/CtaPopoverEmailForm.js#L47)

ngjo:
Is the Newsletter Subscription widget the overlay on like Easy Scholarships? 

Either way, yeah, good 👀 ! We definitely need to continue tracking the UTM parameters in the `source_detail`s

PR #979, opened 2020-02-05
Set email subscription status to true if adding topics

aaronschachter:
### What's this PR do?

This pull request ensures that if a user is created or updated via API request and has any `email_subscription_topics` array items, their `email_subscription_status` is set as `true`.

- 👀 
- Could also pull this branch down and test expected updates via Aurora or API requests

`app/Auth/Registar` seemed like an appropriate place to add the check, especially since we're explicitly setting these fields below in the `registerViaWeb` method. The tests felt a little more home in the User HTTP test vs the Auth Registar test though... but the logic didn't feel like it'd make sense to add only into an API controller (wasn't sure about a post-save type of check in the Model, like [Mongoose middleware](https://mongoosejs.com/docs/middleware.html#pre)).

### Relevant tickets

References [Pivotal #170599403](https://www.pivotaltracker.com/n/projects/2401401/stories/170599403).

### Checklist

### DIFF
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -199,6 +199,16 @@ class Registrar
             $customizer($user);
         }
 
+        /**
+         * If any email subscription topics exist, ensure that email subscription status is true.
+         *
+         * Note: We're explicitly not yet checking for inverse of unsubscribing when topics is empty.
+         * @see https://www.pivotaltracker.com/n/projects/2401401/stories/170599403/comments/211127349.
+         */
+        if (isset($user->email_subscription_topics) && count($user->email_subscription_topics) && ! $user->email_subscription_status) {
+            $user->email_subscription_status = true;
+        }
+
         $user->save();
 
         return $user;
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -202,7 +202,7 @@ class Registrar
         /**
          * If any email subscription topics exist, ensure that email subscription status is true.
          *
-         * Note: We're explicitly not yet checking for inverse of unsubscribing when topics is empty.
+         * Note: We're intentionally not checking for inverse of unsubscribing if topics is empty,
          * @see https://www.pivotaltracker.com/n/projects/2401401/stories/170599403/comments/211127349.
          */
         if (isset($user->email_subscription_topics) && count($user->email_subscription_topics) && ! $user->email_subscription_status) {
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -199,16 +199,6 @@ class Registrar
             $customizer($user);
         }
 
-        /**
-         * If any email subscription topics exist, ensure that email subscription status is true.
-         *
-         * Note: We're intentionally not checking for inverse of unsubscribing if topics is empty,
-         * @see https://www.pivotaltracker.com/n/projects/2401401/stories/170599403/comments/211127349.
-         */
-        if (isset($user->email_subscription_topics) && count($user->email_subscription_topics) && ! $user->email_subscription_status) {
-            $user->email_subscription_status = true;
-        }
-
         $user->save();
 
         return $user;
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -226,9 +226,8 @@ class Registrar
         $user->country = country_code();
         // Set language based on locale (either 'en', 'es-mx').
         $user->language = app()->getLocale();
-        // Sign the user up for email messaging & give them the "community" topic.
-        $user->email_subscription_status = true;
-        $user->email_subscription_topics = ['community'];
+        // Subscribe user to the community email topic.
+        $user->addEmailSubscriptionTopic('community');
 
         // Exclude any 'clubs' referrals from our feature flag tests.
         if (! str_contains($user->source_detail, 'utm_source:clubs')) {
--- a/tests/Http/UserTest.php
+++ b/tests/Http/UserTest.php
@@ -863,4 +863,24 @@ class UserTest extends BrowserKitTestCase
             'email_subscription_topics' => ['news'],
         ]);
     }
+
+    /**
+     * Test that user email_subcription_status is set to true if adding email_subscription_topics.
+     * PUT /v2/users/:id
+     *
+     * @return void
+     */
+    public function testV2UpdateEmailSubscriptionStatusWhenAddingTopics()
+    {
+        $nullStatusUser = factory(User::class)->create();
+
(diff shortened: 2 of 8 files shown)


tests/Http/UserTest.php
```diff
@@ -863,4 +863,47 @@ public function testV2UpdateEmailSubscriptionTopicsWithNoDupesAsAdmin()
...
+
+    /**
+     * Test that user email_subcription_status is set to true if adding email_subscription_topics.
+     * PUT /v2/users/:id
+     *
+     * @return void
+     */
+    public function testEmailSubscriptionStatusWhenAddingTopics()
+    {
+        $nullStatusUser = factory(User::class)->create();
```
katiecrane: I know we don't set `email_subscription_status` in the User factory currently, but I think it could be worth explicitly setting that field to `null` here? In case we ever change that in the factory, this test could become unreliable.

app/Auth/Registrar.php
```diff
@@ -199,6 +199,16 @@ public function register($input, $user = null, Closure $customizer = null)
...
 
+        /**
+         * If any email subscription topics exist, ensure that email subscription status is true.
+         *
+         * Note: We're intentionally not checking for inverse of unsubscribing if topics is empty,
+         * @see https://www.pivotaltracker.com/n/projects/2401401/stories/170599403/comments/211127349.
+         */
+        if (isset($user->email_subscription_topics) && count($user->email_subscription_topics) && ! $user->email_subscription_status) {
+            $user->email_subscription_status = true;
+        }
```
katiecrane: I think this _is_ the right place to do this, but [the `/subscriptions` endpoint is not currently calling this `register` function](https://github.com/DoSomething/northstar/blob/master/app/Http/Controllers/SubscriptionController.php#L56-L74) and I believe that it should be (which is my bad!)

tests/Http/UserTest.php
aaronschachter: I purposely did not set the `email_subscription_status` to mimic the default value (`null`) passed when a user is created and the field is not specified. Your question makes me wonder whether we'd want to set it by default in the factory though, vs always specifying when we're creating a fake user...

katiecrane: Always explicitly set it to `null` or do a situation where we randomly choose between `null` and `false` and `true`?

aaronschachter: I think it'd make sense for the factory to set the field to `null` by default -- as we're not always guaranteed that we'll be setting the field at all when creating a user. An example of this is when [Gambit creates a new user for a mobile number that's never texted in before](https://github.com/DoSomething/gambit/blob/8688e1eb78e1359cfa10b054b9cd7ae8a0947f1e/lib/helpers/user.js#L267). So this line of code would stay as is (not setting anything), and we'd make the change to the User factory.

katiecrane: I like that idea!

aaronschachter: This was a little tricker to do because our factory could potentially be setting `email_subscription_topics`. I've updated this branch with new User factory states to clean this up.

katiecrane: So we're now guaranteed that the base user will not have `email_subscription_status` set?

app/Auth/Registrar.php
aaronschachter: I'm not sure what best practices are for making updates to users via the registrar or a simple `save`. That SubscriptionsController is currently causing the same bug if an unsubscribed user happens to fill out the Newsletter form. We could fix this bug by manually setting `email_subscription_status` to `true` from that controller when we check for an existing user... but it's where I was hoping Laravel had something like the Mongo middleware. I've seen some posts/comments suggest using a SQL trigger, but wonder if Laravel model events could be a way to go too? cc @DFurnes

katiecrane: Oops what I linked to was the subscriptions controller - did you mean that that's the same bug that this work fixes as well? But in a different place? 

Seems to me like we should be using the registrar everywhere, but I'm into the idea of a model event! Curious to see what Dave thinks too!

aaronschachter: I don't think this work does fix that bug, I can update the controller to use `registrar` or manually set the `email_subscription_status` -- but ideally model event or SQL trigger (which i've never heard us discuss) sounds like the better fix..

katiecrane: Right okay I think we've been on the same page all along!

tests/Http/UserTest.php
aaronschachter: Just updated the base factory to explicitly set it to `null`.

app/Auth/Registrar.php
DFurnes: Model events are great! We use them in Northstar [here](https://github.com/DoSomething/northstar/blob/master/app/Observers/UserObserver.php) – to automatically set `source` and other things that should always happen when a user is created/updated (and I'm actually using them in my upcoming PR to make sure we always perform a set of tasks when deleting users, too). The best thing about them is it's nearly impossible to "forget" to use them! (But somewhat easier to forget they're triggering.)

I agree that it seems sensible to use the `Registrar` here (since it's our too cleverly named `UserRepository`). And for anything that anything happens upon creation for _any_ user (not just web users), I think it could definitely be smart to pull those bits out into the [`creating`](https://github.com/DoSomething/northstar/blob/d5492492b830538a83fd97a54188f0b76df4d7f6/app/Observers/UserObserver.php#L10-L20) event.

aaronschachter: It sounds like then maybe we should move the logic out of the Registrar and into both `creating` and `updating` events of the UserObserver? That way code like the SubscriptionsController doesn't really need to worry about whether its calling the model `save` or Registrar `register` to add a user's subscription topic (although I'm still fuzzy which is best to use)

DFurnes: Yeah, I think that sounds good!

Re: model `save` vs. Registrar's `register` – It feels nice to have a single place where we handle all the ways we might register a user, but it doesn't look like we do much of anything there in actuality (it's all in `registerViaWeb`)...

I think I'd be happy enough using `save` and model events here.

DFurnes: I think the `Registrar` could honestly use some rethinking (it's largely been untouched for four years now, and feels like it's become a bit of a "grab bag" of functionality). I'm gonna make a note to revisit during the next cleanup sprint!

aaronschachter:
Updated this branch to DRY logic into the Model event per https://github.com/DoSomething/northstar/pull/979#pullrequestreview-355385742, and also added logic/test to clear subscription topics when unsubscribing per our [discussion in Slack today](https://dosomething.slack.com/archives/C02BBRN5A/p1581632559005700).

Issue #840, opened 2019-02-27
Experiment: Simplified authentication for client-side applications.


DFurnes:
### The Problem
We currently use the [Authorization Code grant](https://github.com/DoSomething/northstar/blob/master/documentation/endpoints/auth.md#create-token-authorization-code-grant) to log users in to our web platform (on Phoenix, Ashes, Rogue admin, Aurora, etc.). This has been super handy as we've deconstructed our monolith, but has some downsides:

1. Once you're logged in, each front-end application manages its own session. This means that **logging out one application will not log you out of others** (so, say, logging out of Phoenix would not invalidate your Rogue admin session).
2. Our current approach **requires a server-side application** that can store a "secret". This was a reasonable assumption in 2016. But today, Phoenix, Rogue admin, and our [GraphQL Playground](https://graphql.dosomething.org) are  almost entirely client-side experiences, and so "trapping" sessions on the server-side is awkward.
3. These client-side applications often directly hit OAuth-protected APIs, but have no way to refresh tokens without [refreshing the entire page](https://github.com/DoSomething/phoenix-next/pull/1276). This leads to an awkward user-experience & the potential for [things to get out of "sync"](https://github.com/DoSomething/phoenix-next/issues/1268).

### The Solution
We really should be thinking about these client-side applications as applications in their own right, and in that case they should be able to handle authentication themselves!

I've put together a proof-of-concept in DoSomething/graphql#66 and DoSomething/northstar#839 demonstrating a client-side application that handles login 100% in the user's browser. This uses an [OAuth2/OpenID Connect package](https://github.com/IdentityModel/oidc-client-js) to then configure & handle [the authorization flow](https://oauth2.thephpleague.com/authorization-server/implicit-grant/) – relying almost entirely on the standards to handle wiring everything up!

This allows users to log in to this application _without_ the need to attach a local users table or session store to this Lambda function. It also keeps track of how long until the user's token expires & raises an event to silently refresh it via an iframe (or refresh token if PKCE flow is supported).

And since this makes token refresh more transparent to the user, we can then safely reduce the token expiration to 5-10 minutes, which reduces the window a compromised token can be misused & would allow us to effectively log a user out of multiple apps by just ending their Northstar session.

### Next Steps
[...]

PR #981, opened 2020-02-19
Update dependencies, including 'league/oauth2-server' to 8.x.

DFurnes:
### What's this PR do?

This pull request updates this application's dependencies, including a bump from 'league/oauth2-server' from 7.x to 8.x, following notes in the [changelog](https://github.com/thephpleague/oauth2-server/blob/master/CHANGELOG.md#800---released-2019-07-13) and [upgrade guide](https://oauth2.thephpleague.com/upgrade-guide/).

I've broken this into bite-sized commits, referencing the changelog for why each change is made. We'll also probably want to do a little extra testing on QA before pushing this to production.

This unblocks some authentication experiments within our upcoming server-side rendering work (see #840) and makes it a little simpler for Northstar to issue it's own access tokens to itself in my upcoming self-serve account deletion PR.

### Relevant tickets

References [Pivotal #171358766](https://www.pivotaltracker.com/story/show/171358766).

### Checklist

### DIFF
--- a/app/Auth/Entities/AccessTokenEntity.php
+++ b/app/Auth/Entities/AccessTokenEntity.php
@@ -46,8 +46,8 @@ class AccessTokenEntity implements AccessTokenEntityInterface
      * Generate a JWT from the access token. We override this method from
      * AccessTokenTrait so that we can add the `iss` and `role` claims.
      *
-     * @param \League\OAuth2\Server\CryptKey $privateKey
-     * @return string
+     * @param CryptKey $privateKey
+     * @return Token
      */
     public function convertToJWT(CryptKey $privateKey)
     {
--- a/app/Auth/Entities/ClientEntity.php
+++ b/app/Auth/Entities/ClientEntity.php
@@ -34,6 +34,18 @@ class ClientEntity implements ClientEntityInterface
     }
 
     /**
+     * Is this client able to keep a secret?
+     *
+     * @return bool
+     */
+    public function isConfidential()
+    {
+        // TODO: When we add support for PKCE flow, we'll want to
+        // swap this to 'true' for clients that support it.
+        return true;
+    }
+
+    /**
      * Make a new ClientEntity from an Eloquent model.
      *
      * @param Client $client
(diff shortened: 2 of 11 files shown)


app/Auth/Entities/AccessTokenEntity.php
```diff
@@ -46,8 +46,8 @@ public function setRole($role)
      * Generate a JWT from the access token. We override this method from
      * AccessTokenTrait so that we can add the `iss` and `role` claims.
      *
-     * @param \League\OAuth2\Server\CryptKey $privateKey
-     * @return string
+     * @param CryptKey $privateKey
+     * @return Token
```
katiecrane: Was this already outdated? I'm confused how this is now returning a `Token` instead of a `string` when the function itself didn't change?

app/Auth/Entities/ClientEntity.php
```diff
@@ -33,6 +33,18 @@ public function __construct($client_id, $client_name, $scopes, $redirect_uri = '
         $this->redirectUri = $redirect_uri;
     }
 
+    /**
+     * Is this client able to keep a secret?
```
katiecrane: giphy

app/Auth/Entities/AccessTokenEntity.php
DFurnes: Ah yeah, this one's sorta magical! They added a [`__toString`](https://github.com/thephpleague/oauth2-server/blob/e6af93c58a625dc5e478e8e5d3d7b4ca97129a44/src/Entities/Traits/AccessTokenTrait.php#L56-L62) method to this entity's trait that then calls this, so we can return a `Token` here but automatically cast it to a string when we want to use it that way.

DFurnes: I think you're right that this docblock was wrong before!

katiecrane: oooooooo that is magical for sure! thanks for explaining!

PR #982, opened 2020-02-20
When deleting users, remove them from other services as well.

DFurnes:
### What's this PR do?

This pull request updates our user deletion model event to also delete a user's data from Rogue, Gambit, and CustomerIo. This super-charges our existing Artisan commands to handle all of our services (sans Quasar) in one fell swoop!

### How should this be reviewed?

I split this work into individual commits for each component, which may be easier to review.

### Any background context you want to provide?

I'll set the new `ROGUE_URL`, `GAMBIT_URL`, `GAMBIT_USERNAME` and `GAMBIT_PASSWORD` environment variables on each environment before merging this, and then enable the `DS_ENABLE_DELETE_APIS` feature flag on dev & QA to test.

### Relevant tickets

References [Pivotal #170953571](https://www.pivotaltracker.com/story/show/170953571).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [ ] Documentation added for new features/changed endpoints.
- [ ] Added appropriate feature/unit tests.
- [ ] If new attributes were added to users, there is a corresponding PR to surface these in Aurora.
- [ ] If new attributes were added to users, then the data team already knows about these changes.

PR #983, opened 2020-02-21
Fix up some goofs from #982.

DFurnes:
### What's this PR do?

This pull request fixes up some little goofs I made in #982:

⚙️ Adds `gambit` and `rogue` sections to `services.php` so that we can actually read these URLs and credentials in the API clients. 8d7d501

🔍 Updates `Gambit` client to handle 404s as a non-fatal error (since this is an undocumented but expected error code for this [Gambit endpoint](https://github.com/DoSomething/gambit/blob/master/documentation/endpoints/users.md) if a user doesn't have messages). 15fceac

🚥 Asserts that we make these API calls in our test suite. 4ccb8d2

✏️ Adds this flag to `.env.example` and updates an incorrect docblock. a2efdcc &  a2efdcc

👀

Should've tested this more carefully the first time around, oops!

### Relevant tickets

References [Pivotal #170953571](https://www.pivotaltracker.com/story/show/170953571).

### Checklist


weerd:
> ✏️ Adds this flag to .env.example and updates an incorrect docblock. a2efdcc & a2efdcc

Just fyi, you linked to the same commit twice :)

DFurnes:
> Just fyi, you linked to the same commit twice :)

Oops, pretend that second one was 8bc4e5d!

PR #984, opened 2020-02-21
Add new "deletion request" APIs.

DFurnes:
### What's this PR do?

This pull request adds new endpoints to allow users to mark their account for deletion (or if they change their mind, unmark it) via new `/v2/users/{id}/deletion` endpoints. These can also be used by staffers to mark any arbitrary account for deletion. 🔥 

It may be easier to review commit-by-commit – up to you! I added tests in ae7f8e1.

This simply sets a "flag" on the user, and the cron job introduced in #980 actually deletes them.

### Relevant tickets

References [Pivotal #170953317](https://www.pivotaltracker.com/story/show/170953317).

### Checklist

### DIFF
--- a/app/Policies/UserPolicy.php
+++ b/app/Policies/UserPolicy.php
@@ -75,4 +75,16 @@ class UserPolicy
     {
         return $this->isSuperuser($viewer) || $this->isOwner($viewer, $target);
     }
+
+    /**
+     * Determine if the authorized user can request deletion for the target user.
+     *
+     * @param User $viewer
+     * @param User $target
+     * @return bool
+     */
+    public function delete(?User $viewer, User $target)
+    {
+        return $this->isSuperuser($viewer);
+    }
 }
--- a/app/Policies/UserPolicy.php
+++ b/app/Policies/UserPolicy.php
@@ -77,7 +77,7 @@ class UserPolicy
     }
 
     /**
-     * Determine if the authorized user can immediately delete the target user.
+     * Determine if the authorized user can request deletion for the target user.
      *
      * @param User $viewer
      * @param User $target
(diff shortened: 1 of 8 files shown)


app/Policies/UserPolicy.php
```diff
@@ -11,52 +11,92 @@ class UserPolicy
...
+    }
+
+    /**
+     * Is the current user the same person as this profile?
+     *
+     * @param User $viewer
+     * @param User $target
+     * @return bool
+     */
+    protected function isOwner(?User $viewer, User $target)
```
katiecrane: I think I'm not familiar with the `?` syntax here - does that mean this parameter is optional?

tests/Http/DeletionRequestTest.php
```diff
@@ -0,0 +1,78 @@
...
+     * @return void
+     */
+    public function testNormalUsersCantMarkOthersForDeletion()
+    {
+        $villain = factory(User::class)->create();
+        $user = factory(User::class)->create();
+
+        $this->asUser($villain, ['user', 'write'])->delete('v2/users/'.$user->id.'/deletion');
+
+        $this->assertResponseStatus(403);
```
katiecrane: I think this is my favorite test ever

app/Policies/UserPolicy.php
DFurnes: Yep! [Nullable types](https://www.php.net/manual/en/migration71.new-features.php#migration71.new-features.nullable-types) were a new feature in PHP 7.1. 🐘

tests/Http/DeletionRequestTest.php
DFurnes: 🦹🏻‍♂️

PR #989, opened 2020-02-28
Log deletion requests & send deletion timestamp to Customer.io.

DFurnes:
### What's this PR do?

This pull request adds some logging when users request to delete their accounts (or change their mind) so that we can alert ourselves if there are any unexpected spikes!

I've also added `deletion_requested_at` to our Customer.io payload so we can use it to send "your account will be deleted" reminder messaging.

👀 

🔥

### Relevant tickets

References Pivotal [#170957143](https://www.pivotaltracker.com/story/show/170957143) and [#170956399](https://www.pivotaltracker.com/story/show/170956399)

### Checklist

### DIFF
--- a/app/Http/Controllers/DeletionRequestController.php
+++ b/app/Http/Controllers/DeletionRequestController.php
@@ -40,14 +40,14 @@ class DeletionRequestController extends Controller
 
         $user->deletion_requested_at = now();
 
-        info('created_deletion_request', ['id' => $user->id]);
-
         // We'll also automatically unsubscribe users from marketing:
         $user->email_subscription_status = false;
         $user->email_subscription_topics = [];
         $user->sms_status = 'stop';
 
         $user->save();
+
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/DeletionRequestController.php
```diff
@@ -60,6 +62,8 @@ public function destroy(User $user)
     {
         $this->authorize('request-deletion', $user);
 
+        info('undo_deletion_request', ['id' => $user->id]);
+
```
katiecrane: This is very picky, but in `store` the logging happens after we successfully set a `deletion_requested_at` and here we're logging before we successfully null out that field, so we may end up seeing this logged for destroy requests that were not successful.

DFurnes: Oh, that's a good point. Let me move it down below.

PR #990, opened 2020-03-04
Email subscription resource

lkpttn:
### What's this PR do?

This PR adds a new Controller and routes that allows us to update and remove email topic subscriptions individually by sending requests to `/v2/users/:id/subscriptions/:topic`.

POST will add a topic to the array
DELETE will remove a topic to the array

Normal 👀 and run the tests!

This sets us up to toggle subscriptions individually in the user profile, coming sooooon!

### Relevant tickets

References [Pivotal #171542935](https://www.pivotaltracker.com/story/show/171542935).

### Checklist

### DIFF
--- a/app/Http/Controllers/SubscriptionUpdateController.php
+++ b/app/Http/Controllers/SubscriptionUpdateController.php
@@ -27,7 +27,7 @@ class SubscriptionUpdateController extends Controller
         $this->middleware('scope:write');
     }
 
-    public function update(User $user, string $topic)
+    public function store(User $user, string $topic)
     {
       $this->authorize('edit-profile', $user);
 
--- a/app/Http/Controllers/SubscriptionUpdateController.php
+++ b/app/Http/Controllers/SubscriptionUpdateController.php
@@ -8,16 +8,16 @@ use Northstar\Http\Transformers\UserTransformer;
 class SubscriptionUpdateController extends Controller
 {
     /**
-      * @var UserTransformer
-      */
+     * @var UserTransformer
+     */
     protected $transformer;
 
     /**
-      * Make a new SubpscriptionUpdateController, inject dependencies,
-      * and set middleware for this controller's methods.
-      *
-      * @param UserTransformer $transformer
-      */
+     * Make a new SubpscriptionUpdateController, inject dependencies,
+     * and set middleware for this controller's methods.
+     *
+     * @param UserTransformer $transformer
+     */
     public function __construct(UserTransformer $transformer)
     {
         $this->transformer = $transformer;
@@ -51,5 +51,4 @@ class SubscriptionUpdateController extends Controller
 
         return $this->item($user);
     }
-
 }
--- a/app/Http/Controllers/SubscriptionUpdateController.php
+++ b/app/Http/Controllers/SubscriptionUpdateController.php
@@ -8,16 +8,16 @@ use Northstar\Http\Transformers\UserTransformer;
 class SubscriptionUpdateController extends Controller
 {
     /**
-    * @var UserTransformer
-    */
+      * @var UserTransformer
+      */
     protected $transformer;
 
     /**
-     * Make a new SubpscriptionUpdateController, inject dependencies,
-     * and set middleware for this controller's methods.
-     *
-     * @param UserTransformer $transformer
-     */
+      * Make a new SubpscriptionUpdateController, inject dependencies,
+      * and set middleware for this controller's methods.
+      *
+      * @param UserTransformer $transformer
+      */
     public function __construct(UserTransformer $transformer)
     {
         $this->transformer = $transformer;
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/SubscriptionUpdateController.php
```diff
@@ -0,0 +1,56 @@
...
+        $this->middleware('scope:write');
+    }
+
+    public function store(User $user, string $topic)
+    {
+      $this->authorize('edit-profile', $user);
+
+      if (!in_array($topic, ['news', 'scholarships', 'community', 'lifestyle'])) {
+        abort(404, 'That subscription does not exist.');
+      }
```
mendelB: I wonder if it makes more sense to have this logic in the router kinda how we do this [in Phoenix](https://github.com/DoSomething/phoenix-next/blob/27bb66ee4d14a6bbdd3134c03a755b94f93b273c/routes/web.php#L57-L60) with page cateogires, where we _only_ allow this route to work for a predetermined list of `topics` e.g.:
`$this->post('users/{user}/subscriptions/{topic}', 'SubscriptionUpdateController@store')->where('topic', 'news|scholarships|community|lifestyle');`

(Same for the logic in `destroy` below).

app/Http/Controllers/SubscriptionUpdateController.php
```diff
@@ -0,0 +1,56 @@
...
+
+class SubscriptionUpdateController extends Controller 
+{
+    /**
+    * @var UserTransformer
+    */
+    protected $transformer;
+
+    /**
+     * Make a new UserController, inject dependencies,
```
mendelB: *`SubscriptionUpdateController `

app/Http/Controllers/SubscriptionUpdateController.php
```diff
@@ -0,0 +1,56 @@
...
+
+    public function store(User $user, string $topic)
+    {
+      $this->authorize('edit-profile', $user);
+
+      if (!in_array($topic, ['news', 'scholarships', 'community', 'lifestyle'])) {
+        abort(404, 'That subscription does not exist.');
+      }
+
+      $user->push('email_subscription_topics', $topic, true);
```
mendelB: Do we need to still save the user (`$user->save()`) for this to take effect in the database? (I _believe_ the tests would still be passing because we're returning this local instance of the user which we've updated.)

Same below in `destroy`.

tests/Http/SubscriptionUpdateTest.php
```diff
@@ -0,0 +1,69 @@
...
+
+        $this->assertResponseStatus(200);
+        $this->seeJsonField('data.email_subscription_topics', ["news"]);
+    }
+
+    public function testAddExistingSubscription()
+    {
+        $user = factory(User::class)->create();
+
+        $this->asUser($user, ['user', 'write'])->post('v2/users/'.$user->id.'/subscriptions/news');
```
mendelB: I _think_, instead of mocking the extra API request, we can just initialize the user above with the existing topic:
```php
  'email_subscription_topics' => ['news'],
]);
```

tests/Http/SubscriptionUpdateTest.php
```diff
@@ -0,0 +1,69 @@
...
+     * Test that a user can mark remove email subscriptions.
+     * POST /v2/users/:id/subscriptions/:topic
+     *
+     * @return void
+     */
+    public function testRemoveSubscription()
+    {
+        $user = factory(User::class)->create();
+
+        $this->asUser($user, ['user', 'write'])->post('v2/users/'.$user->id.'/subscriptions/news');
```
mendelB: I think we can do the same thing as above where we initialize the user with the existing topic instead of using the API?

tests/Http/SubscriptionUpdateTest.php
```diff
@@ -0,0 +1,69 @@
...
+        $this->asUser($user, ['user', 'write'])->post('v2/users/'.$user->id.'/subscriptions/news');
+
+        $this->assertResponseStatus(200);
+        $this->seeJsonField('data.email_subscription_topics', ["news"]);
+    }
+
+    /**
+     * Test that a user can mark remove email subscriptions.
+     * POST /v2/users/:id/subscriptions/:topic
```
mendelB: *`DELETE`

tests/Http/SubscriptionUpdateTest.php
```diff
@@ -0,0 +1,69 @@
...
+        $this->asUser($user, ['user', 'write'])->post('v2/users/'.$user->id.'/subscriptions/news');
+        $this->asUser($user, ['user', 'write'])->delete('v2/users/'.$user->id.'/subscriptions/news');
+
+        $this->assertResponseStatus(200);
+        $this->seeJsonField('data.email_subscription_topics', []);
+    }
+
+
+    /**
+     * Test that a user can mark remove email subscriptions.
```
mendelB: I think this description should be updated? (e.g. `Test that a user cannot update other user's subscriptions`)

app/Http/Controllers/SubscriptionUpdateController.php
DFurnes: Oh nice! I like that approach. 👍

lkpttn: I just tried this out and don't really understand/like the new error we're getting. If you try to subscribe to `dogs` it just returns a 405 error with no message attached, less useful than the 404 we can set from within the method.

```
{
  "error": {
    "code": 405,
    "message": ""
  },
```

mendelB: Hmm that's confusing! This should be returning a `404` as far as I'm aware 🤔

app/Http/Controllers/SubscriptionUpdateController.php
mendelB: Ohhh neato!!

PR #991, opened 2020-03-09
Add sms frequency to reg

lindsayrmaher:
### What's this PR do?

This pull request adds the sms preferences to our subscriptions page of registration in northstar, updates the controller to properly check what preference a user has selected and updates our tests to check that this has been updated correctly.

👀 

Quick note about our subscription tests. While I was adding `sms_status` to the list of form fields that get updated, I noticed that the test wasn't correctly updating the user. And it turns out the old phone number we were passing in the test was actually not passing our validation tests, and consequently not updating our sms_status!! This PR should cover the updates needed to resolve that.

### Relevant tickets

References [Pivotal # 168908600](https://www.pivotaltracker.com/story/show/168908600).

### Checklist

### DIFF
--- a/app/Http/Controllers/Web/ProfileSubscriptionsController.php
+++ b/app/Http/Controllers/Web/ProfileSubscriptionsController.php
@@ -55,7 +55,7 @@ class ProfileSubscriptionsController extends Controller
             if ($user->mobile && $user->mobile !== $currentMobile) {
                 if ($user->sms_status !== 'less') {
                     $user->sms_status = 'active';
-                };
+                }
             }
         });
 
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/Web/ProfileSubscriptionsController.php
```diff
@@ -48,10 +48,14 @@ public function update(Request $request)
 
         $currentMobile = $user->mobile;
 
+        // dd($currentMobile);
```
weerd: oop! Debug comment 👾

lindsayrmaher: Good catch! Just pushed up a fix 😎

PR #994, opened 2020-03-12
User Causes Endpoint

lindsayrmaher:
### What's this PR do?

This pull request adds two endpoints so we can directly update our users cause preferences and sets the scope of those endpoints so that only the user that those causes below to (or an admin) can update them!

This PR also adds tests and updates the Users API docs to show a sample request and response for the two endpoints.

Let me know if the documentation could be clearer, or anything could be cleaned up in the controller.

This is pre-work for creating a Mutation in Graphql that we'll be using in a users Phoenix profile. 

### Relevant tickets

References [Pivotal # 171749230](https://www.pivotaltracker.com/story/show/171749230).

### Checklist


app/Http/Controllers/CauseUpdateController.php
```diff
@@ -0,0 +1,54 @@
...
+
+    public function destroy(User $user, string $cause)
+    {
+        $this->authorize('edit-profile', $user);
+
+        if (! in_array($cause, ['animal_welfare', 'bullying', 'education', 'environment', 'gender_rights_equality', 'homelessness_poverty', 'immigration_refugees', 'lgbtq_rights_equality', ' mental_health', 'physical_health', 'racial_justice_equity', 'sexual_harassment_assault'])) {
+            abort(404, 'That cause does not exist.');
+        }
+
+        $user->pull('causes', $cause);
```
katiecrane: I feel like I did _not_ know about this `pull()` helper, so cool!

documentation/endpoints/v2/users.md
```diff
@@ -443,6 +443,74 @@ curl -X PUT \
 
 </details>
 
+## Update a Users Cause Preferences
```
katiecrane: nitpick: User's 😬

documentation/endpoints/v2/users.md
```diff
@@ -443,6 +443,74 @@ curl -X PUT \
...
+
+Update a user resources cause preferences, retrieved with the user's Northstar ID. This requires the `user` scope and the `write` scope.
```
katiecrane: resource's?

documentation/endpoints/v2/users.md
```diff
@@ -443,6 +443,74 @@ curl -X PUT \
...
+
+```
+DELETE /v2/users/:user_id/causes/:cause
+```
+
+<details>
+<summary><strong>Example Request</strong></summary>
+
+```sh
+curl -X POST \
```
katiecrane: I think this should be DELETE

tests/Http/CauseUpdateTest.php
```diff
@@ -0,0 +1,91 @@
...
+    public function testAddExistingCause()
+    {
+        $user = factory(User::class)->create([
+            'causes' => ['bullying'],
+        ]);
+
+        $this->asUser($user, ['user', 'write'])->post('v2/users/'.$user->id.'/causes/bullying');
+
+        $this->assertResponseStatus(200);
+        $this->seeJsonField('data.causes', ['bullying']);
```
katiecrane: Does this explicitly check that the value of `data.causes` is `['bullying']` or is it just seeing if the array contains `'bullying'`?

tests/Http/CauseUpdateTest.php
```diff
@@ -0,0 +1,91 @@
...
+        $this->assertResponseStatus(404);
+    }
+
+    /**
+     * Test that a user can remove a cause.
+     * DELETE /v2/users/:id/causes/:cause
+     *
+     * @return void
+     */
+    public function testRemoveCause()
```
katiecrane: I would love a test similar to this where we start a user with multiple causes, and when they remove one, we make sure the others are still there. What do you think?

tests/Http/CauseUpdateTest.php
lindsayrmaher: I think it's checking that it exactly matches `["bullying"]`

tests/Http/CauseUpdateTest.php
lindsayrmaher: Yah! I'll update it so we start with more than one cause in the array

tests/Http/CauseUpdateTest.php
katiecrane: rad! perfect!

tests/Http/CauseUpdateTest.php
katiecrane: I think having both could still be helpful - this one is good in case something strange happens when we try to remove the last cause or something

lindsayrmaher: good point good point! i added a secondary one that checks that an empty array gets returned in addition to just removing a single cause.

PR #998, opened 2020-03-19
Adds User sms_subscription_topics field

aaronschachter:
### What's this PR do?

This pull request adds a new `sms_subscription_topics` array field to our User model, similar to the `email_subscription_topics` field. We currently only have one topic defined, `voting`, which is what we'll set in the Chompy Rock The Vote import if a user opts-in to SMS messaging from DoSomething. We'll use this value for segmenting over in Customer.io -- to only send voting related broadcasts to these members (excluding them from general campaign, scholarship broadcasts we send).

Create or update a user with valid `sms_subscription_topics`:
```
curl --location --request POST 'http://northstar.test/v2/users' \
--header 'Content-Type: application/json' \
--header 'Authorization: Bearer [token]
--data-raw '{
"addr_state": "CA",
"email": "your-email@dosomething.org",
"sms_subscription_topics": ["voting"]
}'
```
and verify expected response:
```
{
    "data": {
        "id": "5e73eec9fdce2708bd2d4696",
        "email_preview": "asc...@dosomething.org",
        ...
        "sms_subscription_topics": [
            "voting"
        ],
        ...
        "updated_at": "2020-03-19T22:14:33+00:00",
        "created_at": "2020-03-19T22:14:33+00:00",
    }
}
```

This shouldn't block merge but I'm very confused about how these `POST /users` requests get validated. If I send an array with value other than `voting`, or a string value -- I don't get a 422 back but instead a 200 response with the HTML of the "Create Account" page. I'm also finding that if I create a user with an invalid mobile number, I can successfully create the user but their mobile is null. 

### Relevant tickets

References [Pivotal #171847123](https://www.pivotaltracker.com/story/show/171847123).

### Checklist

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -110,7 +110,12 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
 
         // External profiles:
         'mobilecommons_id', 'mobilecommons_status', 'facebook_id', 'google_id',
-        'sms_status', 'sms_paused', 'email_subscription_status', 'email_subscription_topics', 'last_messaged_at',
+
+        // SMS Subscription:
+        'sms_status', 'sms_paused', 'sms_subscription_topics', 'last_messaged_at',
+
+        // Email Subscription:
+        'email_subscription_status', 'email_subscription_topics',
 
         // Voting Plan:
         'voting_plan_status', 'voting_plan_method_of_transport', 'voting_plan_time_of_day', 'voting_plan_attending_with',
@@ -537,6 +542,7 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
             'lifestyle_email_subscription_status' => isset($this->email_subscription_topics) ? in_array('lifestyle', $this->email_subscription_topics) : false,
             'community_email_subscription_status' => isset($this->email_subscription_topics) ? in_array('community', $this->email_subscription_topics) : false,
             'scholarship_email_subscription_status' => isset($this->email_subscription_topics) ? in_array('scholarships', $this->email_subscription_topics) : false,
+            'voting_sms_subscription_status' => isset($this->sms_subscription_topics) ? in_array('voting', $this->sms_subscription_topics) : false,
             'animal_welfare' => in_array('animal_welfare', $this->causes) ? true : false,
             'bullying' => in_array('bullying', $this->causes) ? true : false,
             'education' => in_array('education', $this->causes) ? true : false,
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -709,6 +709,17 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
     }
 
     /**
+     * Mutator to ensure no duplicates in the SMS topics array.
+     *
+     * @param array $value
+     */
+    public function setSmsSubscriptionTopicsAttribute($value)
+    {
+        // Set de-duped array as sms_subscription_topics
+        $this->attributes['sms_subscription_topics'] = array_values(array_unique($value));
+    }
+
+    /**
      * Accessor for the `causes` attribute.
      *
      * @param  mixed value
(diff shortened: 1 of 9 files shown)


documentation/endpoints/users.md
```diff
@@ -316,7 +316,10 @@ PUT /v1/users/drupal_id/<drupal_id>
   role: String; // Can only be modified by admins. Either 'user' (default), 'staff', or 'admin'.
   sms_status: String; // Either 'active', 'stop', less', 'undeliverable', 'pending', or 'unknown'
   sms_paused: Boolean; // Whether a user is in a support conversation.
-  email_subscription_status: Boolean; // Whether a user can recieve emails or not.
+  sms_subscription_topics: Array; // Valid values: 'voting'
+  email_subscription_status: Boolean; // Whether a user is subscribed to receive emails.
+  email_subscription_topics: Array; // Valid values: 'news', 'scholarships', 'lifestyle', 'community'
```
aaronschachter: Do we use this `v1/users` endpoint for anything anymore? As I wrote the question, I thought of the answer -- Gambit still [posts to v1 via northstar-js](https://github.com/DoSomething/northstar-js/blob/903d494b3add91b70f7a53231abafb5cc80d4623/.env.example#L1) -- would be so sweet to deprecate the use of that library for GraphQL

app/Models/User.php
```diff
@@ -110,7 +110,12 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
 
         // External profiles:
         'mobilecommons_id', 'mobilecommons_status', 'facebook_id', 'google_id',
-        'sms_status', 'sms_paused', 'email_subscription_status', 'email_subscription_topics', 'last_messaged_at',
+
+        // SMS Subscription:
+        'sms_status', 'sms_paused', 'sms_subscription_topics', 'last_messaged_at',
+
+        // Email Subscription:
+        'email_subscription_status', 'email_subscription_topics',
```
katiecrane: YES this is so much easier to look at!

katiecrane:
>This shouldn't block merge but I'm very confused about how these POST /users requests get validated. If I send an array with value other than voting, or a string value -- I don't get a 422 back but instead a 200 response with the HTML of the "Create Account" page. I'm also finding that if I create a user with an invalid mobile number, I can successfully create the user but their mobile is null.

I combed through the `email_subscription_topics` work and as far as I can tell, your validation is happening in the exact same way and the `email_subscription_topics` validation is working for me so it might be something deeper. All I can think of for the 200 and getting the HTML back is maybe not having the `Content-Type: application/json` header but you are using that so I'm not sure about that. 

For the mobile validation I wonder if something is messed up with our validation string of `'mobile|nullable|unique:users,mobile,'.$existingId.',_id|required_without:email'` (not the simplest validation!) and we don't seem to have a test specific for that, but I'd be curious what happens if you write a test for creating a user with an invalid mobile number. Do you only see the strange behavior when you also supply an email or is it all the time?

katiecrane:
Everything looks fine to the naked eye, but I'd be hesitant to merge until the validation on the new `sms_subscription_topics` field is working!

aaronschachter:
I must be doing something silly when testing via Postman that I'm missing... because per your great suggestion, I've added tests for creating a user with both for `sms_subscription_topics` and `mobile` fields and working as expected (which is good!)

tests/Http/UserTest.php
```diff
@@ -619,6 +619,57 @@ public function testV2SetCreatedAtField()
...
+        ]);
+
+        $this->assertResponseStatus(422);
+
+        $this->asAdminUser()->json('POST', 'v2/users', [
+            'email' => 'test@example.com',
+            'sms_subscription_topics' => 'bugs',
+        ]);
+
+        $this->assertResponseStatus(500);
```
aaronschachter: This is a 500 because of the method I've added post-review, `setSmsSubscriptionTopicsAttribute`, throws an ErrorExcepetion:`array_unique() expects parameter 1 to be array, string given {"exception":"[object] (ErrorException(code: 0): array_unique() expects parameter 1 to be array, string given at /home/vagrant/Code/northstar/app/Models/User.php:719)`

katiecrane: I think that's cool, same thing happens with `email_subscription_topics`

PR #999, opened 2020-03-20
fixes a space in the mental_health name that was breaking the mutation

lindsayrmaher:
### What's this PR do?

This pull request fixes a space I didn't notice before one of the cause names that is causing a break in profile page!

👀 

naming needs to be an exact match in order for our mutations to work as expected.

### Relevant tickets

References [Pivotal # 171439769](https://www.pivotaltracker.com/story/show/171439769).

### Checklist

### DIFF
--- a/app/Http/Controllers/CauseUpdateController.php
+++ b/app/Http/Controllers/CauseUpdateController.php
@@ -30,7 +30,7 @@ class CauseUpdateController extends Controller
     {
         $this->authorize('edit-profile', $user);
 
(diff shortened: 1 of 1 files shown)


mendelB:
This does highlight that it might be prudent for us to do something similar in Northstar as [in Rogue](https://github.com/DoSomething/rogue/blob/b32e35f8a7ac909de15c52e0ba81033861f3dc0e/app/Types/Cause.php) where we formalize the list of Causes somewhere. We seem to be hardcoding this in quite a few places now!

PR #1005, opened 2020-04-06
Adds new values for voter_registration_status

aaronschachter:
### What's this PR do?

This pull request updates the validation rule for `voter_registration_status` with the new values we'll be passing from Chompy, once https://github.com/DoSomething/chompy/pull/153 is deployed to production.

👀 

@DFurnes @katiecrane -- Do you think we should we be sending over these values with underscores (`step_1`) instead of hyphens (`step-1`) to keep Northstar values looking consistent?

 Rogue already saves hyphenated status values like `complete-OVR`, but I can see we're using underscores for cause values too -- would have guessed we'd be using hyphens though per what seems be our normal standard in other apps..

### Relevant tickets

References [Pivotal #171737617](https://www.pivotaltracker.com/story/show/171737617).

### Checklist

### DIFF
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -67,7 +67,11 @@ class Registrar
             'last_messaged_at' => 'date',
             'email_subscription_status' => 'boolean',
             'email_subscription_topics.*' => 'in:news,scholarships,lifestyle,community',
-            'voter_registration_status' => 'nullable|in:uncertain,ineligible,unregistered,confirmed,registration_complete',
+            /**
+             * Includes current values sent from Rock The Vote import, as well as older values for
+             * backwards compatability.
+             */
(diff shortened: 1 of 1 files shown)


DFurnes:
> @DFurnes @katiecrane -- Do you think we should we be sending over these values with underscores (`step_1`) instead of hyphens (`step-1`) to keep Northstar values looking consistent?
> 
> Rogue already saves hyphenated status values like `complete-OVR`, but I can see we're using underscores for cause values too -- would have guessed we'd be using hyphens though per what seems be our normal standard in other apps..

Yeah, we've gotten ourselves in a little formatting bind. I think we're generally using `kebab-case` for things like this & so that'd be my vote (and we can maybe clean up causes to follow suit).

Another idea might be to use GraphQL-styled `ENUM_CASE` for all of these since it'd remove some context switching from that REST-to-GraphQL transformation layer... but that's something we'd probably want to consider across the board later on.

PR #1006, opened 2020-04-09
Upgrade to Laravel 6.0

DFurnes:
### What's this PR do?

This pull request updates Northstar to use Laravel 6.0, following the steps in the upgrade guides (linked in each commit). While somewhat tedious, most of these changes are simple (e.g. "rename this method" or "add this new parameter").

Since you can't link directly to specific sub-headings in the docs, I made one commit per "guide" and then split other fixes I had to make on my own into their own commits. This is going to be a bit of a monster, so apologies in advance!

We're jumping from one "long-term service" release to the next, since Laravel 5.5 is nearing [the end of its support period](https://laravel.com/docs/5.7/releases#support-policy). Laravel 6.0 is supported until September 2021.

We're also hoping that being able to upgrade to a newer [jenssegers/laravel-mongodb](https://github.com/jenssegers/laravel-mongodb) release [will fix up some of the query performance issues](https://dosomething.slack.com/archives/CUQMU4Q6B/p1586281304002600?thread_ts=1585863975.003800&cid=CUQMU4Q6B) we've been seeing on this application.

### Relevant tickets

References [Pivotal #171217731](https://www.pivotaltracker.com/story/show/171217731).

### Checklist


config/app.php
```diff
@@ -106,23 +106,6 @@
...
-    | you a variety of powerful log handlers / formatters to utilize.
-    |
-    | Available Settings: "single", "daily", "syslog", "errorlog"
-    |
-    */
-
-    'log' => env('APP_LOG', 'single'),
-
-    'log_level' => env('APP_LOG_LEVEL', 'debug'),
-
```
DFurnes, line 125: These are replaced by `config/logging.php`, below.

config/logging.php
```diff
@@ -0,0 +1,61 @@
...
+    | Default Log Channel
+    |--------------------------------------------------------------------------
+    |
+    | This option defines the default log channel that gets used when writing
+    | messages to the logs. The name specified in this option should match
+    | one of the channels defined in the "channels" configuration array.
+    |
+    */
+
+    'default' => env('LOG_CHANNEL', 'stack'),
```
DFurnes, line 16: Since we're defaulting to the `stack` log driver on Heroku instances, we won't need to set this before deploying.

resources/views/auth/callback.blade.php
```diff
@@ -22,13 +22,13 @@
         <div class="container__block -centered">
             <figure class="figure -medium">
                 <div class="figure__media">
-                    <img class="avatar" alt="avatar" src="{{ $user->avatar or asset('avatar-placeholder.png') }}" />
+                    <img class="avatar" alt="avatar" src="{{ $user->avatar ?? asset('avatar-placeholder.png') }}" />
```
DFurnes, line 25: This custom `or` operator in Blade is just being replaced with the more "standard" PHP `??`.

phpunit.xml
```diff
@@ -7,8 +7,7 @@
          convertNoticesToExceptions="true"
          convertWarningsToExceptions="true"
          processIsolation="false"
-         stopOnFailure="false"
-         syntaxCheck="false">
```
DFurnes, line 11: This `syntaxCheck` option was removed in PHPUnit 6.x.

app/Providers/AppServiceProvider.php
```diff
@@ -29,13 +29,6 @@ public function boot()
      */
     public function register()
     {
-        // Register custom Gate with anonymous authorization support:
-        $this->app->singleton(GateContract::class, function ($app) {
-            return new CustomGate($app, function () use ($app) {
-                return call_user_func($app['auth']->userResolver());
-            });
-        });
```
DFurnes, line 47: Support for authorizing anonymous requests is now included in the framework, so we don't need this override.

tests/HelpersTest.php
```diff
@@ -42,19 +42,24 @@ public function testRouteHasAttachedMiddleware()
...
+        $this->withLegacyApiKey($client)->getJson('/status');
         $this->assertEquals(client_id(), 'legacy_client');
+    }
 
-        // And make a request using OAuth.
-        $this->asNormalUser()->json('GET', '/status');
+    /** @test */
+    public function testGetClientIdFromOAuth()
+    {
+        $this->asNormalUser()->getJson('/status');
```
DFurnes, line 62: These didn't strictly _need_ to be split out. I did this while debugging an issue with our OAuth parsing middleware during this upgrade step to make it easier to sort out which part of the test was failing, and kept since it's a wee bit cleaner.

DFurnes:
Okay, this is finally ready for review!! 😌 For the most part, the diff actually isn't too bad... most changes are just changed interfaces (e.g. swapping a `protected` method to `public`, or adding a new argument that wasn't there before, or renaming a method call).

One of the trickiest parts of this was actually dependency resolution. Since PHP only allows _one_ instance of a given class, I ended up in a few situations where I got in a stuck in a "something changed" dependency loop (because there were dependencies that Composer _couldn't_ make compatible). I solved these by removing all of our dependencies & re-adding them one-by-one.

DFurnes:
One thing I may do in a follow-up PR is compare this application to the [`laravel/laravel`](https://en.wikipedia.org/wiki/Boolean_satisfiability_problem) starter repository to see if there are any other interesting changes we want to pull in (for example, moving `resources/assets` up into `resources/`).

app/Providers/OAuthServiceProvider.php
```diff
@@ -39,20 +40,14 @@ class OAuthServiceProvider extends ServiceProvider
...
-        $auth = $this->app['auth'];
-
         // Register our custom user provider
-        $auth->provider('northstar', function ($app, array $config) {
+        Auth::provider('northstar', function ($app, array $config) {
             return new NorthstarUserProvider($app[Registrar::class], $app['hash'], $config['model']);
         });
 
         // Register our custom token Guard implementation
-        $auth->extend('northstar-token', function ($app, $name, array $config) use ($auth) {
```
weerd, line 54: Nice the `use` isn't needed now.

database/migrations/2015_10_26_193708_AddDrupalIDIndex.php
```diff
@@ -24,7 +24,7 @@ public function up()
     public function down()
     {
         Schema::table('users', function ($collection) {
-            $collection->dropIndex('drupal_id');
+            $collection->dropIndex('drupal_id_1');
```
weerd, line 27: I'm moving through this a file-at-a-time, so maybe you mention something further down, but why do these fields have `_1` suffix appended?

DFurnes: Yeah, this was an undocumented change I stumbled upon. Before, we provided the indexed column name. After this upgrade, we have to provide the index name (gotten via `db.users.getIndexes()` on the MongoDB CLI).

weerd: Ah! ok, thanks for clarifying!

resources/views/auth/callback.blade.php
weerd: ah nice! I'm a fan of the null coalescing operator!

weerd: that and the elvis operator `?:` 🕺

routes/api.php
```diff
@@ -10,68 +10,68 @@
...
 // Discovery
-$router->group(['prefix' => '.well-known'], function () {
-    $this->get('openid-configuration', 'DiscoveryController@index');
+Route::group(['prefix' => '.well-known'], function () {
+    Route::get('openid-configuration', 'DiscoveryController@index');
 });
 
 // Simple health check endpoint
-$router->get('/status', function () {
+Route::get('/status', function () {
```
weerd, line 75: 😱 have you changed your stance?!

tests/CreatesApplication.php
```diff
@@ -73,7 +73,7 @@ public function createApplication()
      *
      * @return void
      */
-    public function setUp()
+    public function setUp(): void
```
weerd, line 76: 👀

routes/api.php
DFurnes: Yeah, I think I prefer following the docs more closely to make it easier to onboard (even if I still think facades can be a little confusing if you don't know what's going on with them).

PR #1008, opened 2020-04-15
Set default SMS subscription topics

aaronschachter:
### What's this PR do?

This pull request adds the following changes: 

* Sets `general` and `topic` as the default values for `sms_subscription_topics` field if not provided when creating a user with a  subscribed `sms_status` value of `active` or `less`. 

* Sets `sms_subscription_topics` to empty if updating a user with an unsubscribed `sms_status` value of `stop` or `undeliverable`. 

* Sets default `sms_subscription_topics` as `general` and `topic` if updating a user's `sms_status` field to an subscribed value of `active` or `less` and user currently does not have any `sms_subscription_topics`

👀 

I could not figure out how to get Aurora to stop from clearing the updated `sms_subscription_topics` when changing from unsubscribed to subscribed. To maintain sanity, I've removed it from the edit form and will open a PR accordingly.

### Relevant tickets

References [Pivotal #172209330](https://www.pivotaltracker.com/story/show/172209330).

### Checklist

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -725,7 +725,11 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
      */
     public function setSmsSubscriptionTopicsAttribute($value)
     {
-        // Set de-duped array as sms_subscription_topics
+        if (! is_array($value)) {
+            $value = [$value];
+        }
+
+        // Set de-duped array as sms_subscription_topics.
         $this->attributes['sms_subscription_topics'] = array_values(array_unique($value));
     }
 
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -246,7 +246,7 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
      *
      * @return bool
      */
-    public function getIsSmsSubscribedAttribute()
+    public function isSmsSubscribed()
     {
         return isset($this->sms_status) && self::isSubscribedSmsStatus($this->sms_status);
     }
@@ -256,7 +256,7 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
      *
      * @return bool
      */
-    public function getHasSmsSubscriptionTopicsAttribute()
+    public function hasSmsSubscriptionTopics()
     {
         return isset($this->sms_subscription_topics) && count($this->sms_subscription_topics);
     }
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -248,7 +248,7 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
      */
     public function getIsSmsSubscribedAttribute()
     {
-        return isset($user->sms_status) && self::isSmsSubscribedStatus($user->sms_status);
+        return isset($this->sms_status) && self::isSubscribedSmsStatus($this->sms_status);
     }
 
     /**
@@ -258,7 +258,7 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
      */
     public function getHasSmsSubscriptionTopicsAttribute()
     {
-        return isset($user->sms_subscription_topics) && count($user->sms_subscription_topics);
+        return isset($this->sms_subscription_topics) && count($this->sms_subscription_topics);
     }
 
     /**
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -214,6 +214,33 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
     ];
 
     /**
+     * Constants for SMS subscriptions.
+     */
+    const DEFAULT_SMS_SUBSCRIPTION_TOPICS = ['general', 'voting'];
+    const SUBSCRIBED_SMS_STATUSES = ['active', 'less'];
+    const UNSUBSCRIBED_SMS_STATUSES = ['stop', 'undeliverable'];
+
+    /**
+     * Whether user has a subscribed SMS status.
+     *
+     * @return boolean
+     */
+    public function getIsSmsSubscribedAttribute()
+    {
+        return isset($user->sms_status) && in_array($user->sms_status, SUBSCRIBED_SMS_STATUSES);
+    }
+
+    /**
+     * Whether user has any SMS subscription topics.
+     *
+     * @return boolean
+     */
+    public function getHasSmsSubscriptionTopics()
+    {
+        return isset($user->sms_subscription_topics) && count($user->sms_subscription_topics);
+    }
+
+    /**
      * Computed last initial field, for public profiles.
      *
      * @return string
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -214,54 +214,6 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
     ];
 
     /**
-     * Default SMS subscription topics to assign when user is created or updated with a subscribed
-     * SMS status and topics have not been provided.
-     */
-    const DEFAULT_SMS_SUBSCRIPTION_TOPICS = ['general', 'voting'];
-
-    /**
-     * Whether a SMS status value is a subscribed SMS status.
-     *
-     * @param string $smsStatusValue
-     * @return bool
-     */
-    public static function isSubscribedSmsStatus($smsStatusValue)
-    {
-        return in_array($smsStatusValue, ['active', 'less']);
-    }
-
-    /**
-     * Whether a SMS status value is a unsubscribed SMS status.
-     *
-     * @param string $smsStatusValue
-     * @return bool
-     */
-    public static function isUnsubscribedSmsStatus($smsStatusValue)
-    {
-        return in_array($smsStatusValue, ['stop', 'undeliverable']);
-    }
-
-    /**
-     * Whether user has a subscribed SMS status.
-     *
-     * @return bool
-     */
-    public function isSmsSubscribed()
-    {
-        return isset($this->sms_status) && self::isSubscribedSmsStatus($this->sms_status);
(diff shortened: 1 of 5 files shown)


app/Models/User.php
```diff
@@ -725,8 +725,8 @@ public function setEmailSubscriptionTopicsAttribute($value)
      */
     public function setSmsSubscriptionTopicsAttribute($value)
     {
-        // Set de-duped array as sms_subscription_topics
-        $this->attributes['sms_subscription_topics'] = array_values(array_unique($value));
+        // Set de-duped array as sms_subscription_topics.
+        $this->attributes['sms_subscription_topics'] = array_values(array_unique($value ? $value : []));
```
aaronschachter, line 729: This was needed to avoid an error that occurred in `Tests/Http/MergeTest.php`, where a `null` value was being passed to this function.

weerd: not necessary for merge but if you do so desire you could use the `?:` elvis operator here to slightly simplify 🙂:

```php
$this->attributes['sms_subscription_topics'] = array_values(array_unique($value ?: []));
```

app/Observers/UserObserver.php
```diff
@@ -9,6 +9,9 @@
 
 class UserObserver
 {
+    public static $subscribedSmsStatuses = ['active', 'less'];
+    public static $defaultSmsSubscriptionTopics = ['general', 'voting'];
```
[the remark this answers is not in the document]
aaronschachter: It's mainly to avoid hardcoding string values all over the codebase, especially for these arrays that appear in multiple functions. There's a related thread over in https://github.com/DoSomething/chompy/pull/156#discussion_r408901036, where @DFurnes mentions using class constants instead of the static variables initially used to DRY constants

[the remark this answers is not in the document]
weerd: Ah, interesting. Is that something we could try here then?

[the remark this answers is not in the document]
DFurnes: Yeah, we haven't done this much in the past but I can see the value of naming and extracting constants for these. I think I miiiight put them on the `User` instead, but that's somewhat subjective.

app/Observers/UserObserver.php
```diff
@@ -22,6 +25,11 @@ public function creating(User $user)
             $user->email_subscription_status = true;
         }
 
+        // Populate default topics if subscribing to SMS without any topics provided.
+        if (isset($user->sms_status) && in_array($user->sms_status, static::$subscribedSmsStatuses) && ! isset($user->sms_subscription_topics)) {
```
weerd, line 29: This kinda feels like a mouthful 😖 It took me a few times to read through it to grok it...

Maybe breaking this out with a method on the class that clearly states what it's looking for could help:

```php
// returns a boolean
public function hasSmsSubscription(User $user) {
    return isset($user->sms_status) && in_array($user->sms_status, static::$subscribedSmsStatuses);
}

public function creating(User $user)
{
    //...

    if (hasSmsSubscription($user) && ! isset($user->sms_subscription_topics)) {
        $user->sms_subscription_topics = static::$defaultSmsSubscriptionTopics;
    }

    //...
}
```

Could be a better name for the method, but intention is something where reading the method clearly states what it does and helps with understanding the further logic.

aaronschachter: Maybe a method on the User model, `$user->isSmsSubcriber`. It could check the constants for the subscribed `sms_status` values (which I agree with @DFurnes would make sense to add in that class - https://github.com/DoSomething/northstar/pull/1008#discussion_r409769189)

weerd: Ya! I really dig that 👍 Makes sense to me and a good combo w/ the constants 👏 Also reads much nicer chained off the `user` object!

aaronschachter: Although `$user->isSmsSubcriber` (or `isSmsSubscribed`, i like better -- for a sibling `isSmsUnsubscribed`) wouldn't make sense to check for topics. so it could look like `$user->isSmsSubscribed && isset($user->sms_subscription_topics)`

weerd: ya, agree with not checking the topics, which is what I have in example code :)

oh ya. Maybe `$user->isSubscribedToSms()` and `$user->isUnsubscribedToSms()`? That feels like it reads pretty clearly. Thoughts?

aaronschachter: Would it be better as a "magic" attribute vs a function? I liked it more as an attribute, but wonder if related it's why I'm having a tough time getting the truthy unit test to pass...

tests/Models/UserTest.php
```diff
@@ -140,4 +140,73 @@ public function it_should_exclude_unsubscribed_in_customerio_payload_if_email_su
...
         $this->assertFalse(isset($result['unsubscribed']));
     }
+
+    public function testIsSmsSubscribedisTrueIfSmsStatusIsActive()
+    {
+        $user = factory(User::class)->create([
+            'sms_status' => 'active',
+        ]);
+        // @TODO: Why have you forsaken me
+        //$this->assertTrue($user->is_sms_subscribed);
```
aaronschachter, line 150: Per https://github.com/DoSomething/northstar/pull/1008#discussion_r409864718 - I'm having trouble figuring out why this fails with
```
Failed asserting that false is true.
```

aaronschachter:
This should be ready for another a look, thanks for reviewing!

app/Observers/UserObserver.php
```diff
@@ -22,6 +22,11 @@ public function creating(User $user)
             $user->email_subscription_status = true;
         }
 
+        // Populate default topics if subscribing to SMS without providing any topics.
+        if ($user->is_sms_subscribed && ! $user->has_sms_subscription_topics) {
+            $user->sms_subscription_topics = User::DEFAULT_SMS_SUBSCRIPTION_TOPICS;
+        }
```
weerd, line 28: Kind of confused why your're using an [accessor](https://laravel.com/docs/7.x/eloquent-mutators#defining-an-accessor) here and set it up as such. When we discussed it before it was just a simple method that checked the same logic you had before, `$user->isSubscribedToSms()` is just a simple method that returns true/false.

app/Observers/UserObserver.php
```diff
@@ -67,14 +72,19 @@ public function updating(User $user)
...
-        if (isset($changed['sms_status']) && $changed['sms_status'] == 'stop') {
-            $user->sms_subscription_topics = [];
+        if (isset($changed['sms_status'])) {
+            /**
+             * If we're unsubscribing from SMS, clear all topics.
+             *
+             * Note: We don't allow users to set their own SMS subscription topics yet, so there
+             * isn't a need to change sms_status if an unsubscribed user adds a SMS topic.
+             */
+            if (User::isUnsubscribedSmsStatus($changed['sms_status'])) {
```
DFurnes, line 82: Could this use the accessor on the user?

```php
if (! $user->is_sms_subscribed) {
    $user->sms_subscription_topics = [];
} 
```

app/Observers/UserObserver.php
weerd: I think getting and setting the values with attributes _does_ make sense but I think I'm confused a bit with the approach around deciding whether they need to be set or not 🤔

DFurnes: I think I agree that a method would be a little clearer here (since a reader might otherwise assume these are fields that can be written to, like `$user->is_sms_subscribed = false`. Not a big deal either way, though!

weerd: Mulling it over a bit more, and I think to me it makes sense to use an attribute accessor if you're planning on using the custom attribute in output in html, for example if you have `firstName` and `lastName` attributes and you want a more convenient `fullName` attribute to easily output the full name in a form or some other html (and yes, that is a super contrived example, lol, but gets the point across)

But since this is more of a check in a condition to decide later logic it seems more readable/less confusing to use a method for checking truthiness, which also seems more useful in general for many other logic checks.

app/Models/User.php
```diff
@@ -213,6 +213,54 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
...
+
+    /**
+     * Whether user has any SMS subscription topics.
+     *
+     * @return bool
+     */
+    public function getHasSmsSubscriptionTopicsAttribute()
+    {
+        return isset($this->sms_subscription_topics) && count($this->sms_subscription_topics);
+    }
```
weerd, line 262: Another reason why it might make more sense for these to be simple methods on the `User` class is so that they'll read a lot easier.... `isSubscribedToSms() {...}` and `hasSmsSubscription() {...}` is pretty straightforward compared to `getIs...` and `getHas...` which sounds peculiar.

app/Observers/UserObserver.php
aaronschachter: We need to check the changed value of the `sms_status` -- isn't `$changed['sms_status']` not equal to the user's `sms_status` property at this point (in the `updating` event)?

aaronschachter:
Alrighty! third time's a charm? ☘️

app/Observers/UserObserver.php
```diff
@@ -22,6 +22,11 @@ public function creating(User $user)
             $user->email_subscription_status = true;
         }
 
+        // Populate default topics if subscribing to SMS without any topics provided.
+        if ($user->isSmsSubscribed() && ! $user->hasSmsSubscriptionTopics()) {
+            $user->addDefaultSmsSubscriptionTopics();
+        }
```
weerd, line 28: Noice! ✨ This reads very clearly! 👌

PR #1009, opened 2020-04-16
Adds console command to set default SMS subscription topics

aaronschachter:
### What's this PR do?

This pull request follows #857 and #861 by example to add a `northstar:default-sms-topics` console command, used to populate the `sms_subscription_topics` with `general` and `voting` values for all users with `sms_status` equal to a subscribed value (`active`, `less`, and `pending`).

👀 

We want to run this backfill before we start updating user SMS subscriptions via the Chompy Rock the Vote import (see https://github.com/DoSomething/chompy/pull/156) -- which may remove the `voting` topic from existing subscribers if they choose to opt-out of voting messaging from DS on the voter registration form (the checkbox is hidden on the RTV site until this backfill is done and the Chompy feature flag is removed to begin updating existing users' SMS preferences).

We'll also likely need to wait until we upgrade Laravel in #1006 to avoid database timeouts.

### Relevant tickets

References [Pivotal #171848124](https://www.pivotaltracker.com/story/show/171848124).

### Checklist

### DIFF
--- a/app/Console/Commands/SetDefaultSmsSubscriptionTopics.php
+++ b/app/Console/Commands/SetDefaultSmsSubscriptionTopics.php
@@ -1,0 +1,81 @@
+<?php
+
+namespace Northstar\Console\Commands;
+
+use Northstar\Models\User;
+use Illuminate\Console\Command;
+use Illuminate\Support\Collection;
+
+class SetDefaultSmsSubscriptionTopics extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'northstar:default-sms-topics';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Add "general" and "voting" SMS topics for each user who is subscribed to SMS.';
+
+    /**
+     * The number of users updated so far.
+     *
+     * @var string
+     */
+    protected $currentCount = 0;
+
+    /**
+     * Create a new command instance.
+     *
+     * @return void
+     */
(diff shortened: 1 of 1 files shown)


app/Console/Commands/SetDefaultSmsSubscriptionTopics.php
```diff
@@ -0,0 +1,81 @@
...
+     */
+    public function handle()
+    {
+        // Use low priority queue for these updates
+        config(['queue.jobs.users' => 'low']);
+
+        // Grab users who are subscribers and do not have topics set.
+        $query = (new User)->newQuery();
+        $query = $query->whereIn('sms_status', ['active', 'less', 'pending']);
+        $query = $query->where('sms_subscription_topics', null);
```
DFurnes, line 55: We may have to add an index on these fields... let's wait and see when we run it on QA!

app/Console/Commands/SetDefaultSmsSubscriptionTopics.php
```diff
@@ -0,0 +1,81 @@
...
+
+            return;
+        }
+
+        $query->chunkById(200, function (Collection $users) use ($totalCount) {
+            $users->each(function (User $user) use ($totalCount) {
+                $user->sms_subscription_topics = ['general', 'voting'];
+                $user->save();
+
+                $this->line('northstar:default-sms-topics - User '.$user->id.' given "general" and "voting" topics.');
```
DFurnes, line 70: Since this is going to run on a couple million people, let's remove this log line to save some log usage.

PR #1012, opened 2020-04-20
VR prompt in reg flow

lindsayrmaher:
### What's this PR do?

This pull request adds a prompt to the registration flow to encourage members who said "No" when signing up to register to vote via our partnership with RTV!

👀  - could the JS for the toggle be any cleaner? I'm not a jquery aficionado 

This is a V1 for prompting new users to register to vote as a part of their DS registration process.

### Relevant tickets

References [Pivotal # 172094902](https://www.pivotaltracker.com/story/show/172094902).

### Checklist

### DIFF
--- a/resources/assets/js/app.js
+++ b/resources/assets/js/app.js
@@ -18,9 +18,9 @@ window.Drupal = {
 
 // Utilities
 import Analytics from './utilities/Analytics';
-import ContentVisibility from './utilities/ContentVisibility';
 import DeLorean from './utilities/DeLorean';
 import PasswordVisibility from './utilities/PasswordVisibility';
+import VoterRegistrationContent from './utilities/VoterRegistrationContent';
 import './validators/auth';
 
 // Display environment badge on local, dev, or QA:
@@ -35,8 +35,8 @@ if (document.documentElement.lang !== 'en') {
 // Initialize analytics.
 Analytics.init();
 
-//Initialize content visibility toggle.
-ContentVisibility.init();
+//Initialize Voter Registration Content Toggle.
+VoterRegistrationContent.init();
 
 // Initialize routing back to last page.
 DeLorean.init();
--- a/resources/assets/js/app.js
+++ b/resources/assets/js/app.js
@@ -20,7 +20,7 @@ window.Drupal = {
 import Analytics from './utilities/Analytics';
 import DeLorean from './utilities/DeLorean';
 import PasswordVisibility from './utilities/PasswordVisibility';
-import VoterRegistrationContent from './utilities/VoterRegistrationContent';
+import VoterRegistrationCtaToggle from './utilities/VoterRegistrationCtaToggle';
 import './validators/auth';
 
 // Display environment badge on local, dev, or QA:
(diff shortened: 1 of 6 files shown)


resources/assets/js/app.js
```diff
@@ -34,6 +35,9 @@ if (document.documentElement.lang !== 'en') {
 // Initialize analytics.
 Analytics.init();
 
+//Initialize content visibility toggle.
+ContentVisibility.init();
```
mendelB, line 39: LMK watcha think -- I had a thought about making this a bit more explicitly specific to the voter registration toggle it applies to by either:
- putting this JS into a `<script>` and importing it in the form where it's applicable (this way we're not globally invoking this listener on all forms). But I'm admittedly not too proficient with our build/bundling process, so not positive a `<script>` would be a win here... or
- renaming this and it's internal functions to be a bit more explicit to the voter registration toggle it's running? If we want it open-ended bc we anticipate other toggles, then maybe just the internal functions. (But mayyyybe renaming this to be like `ContentVisibilityToggle`?) 

I'm just thinking since we're using pretty static logic applied to the ID's and values of the voter-reg toggle it might be good to label 'em more specifically

resources/views/profiles/about/edit.blade.php
```diff
@@ -55,6 +55,11 @@
             </div>
         </div>
 
+        <div id="vr-details" class="pb-4 hidden">
```
mendelB, line 58: what do you think of `vr-cta` for this ID?

aaronschachter: I think that suggestion makes sense @mendelB -- I'd also prefer to see the prefix spelled out as `voter-reg`  instead of `vr` for readability.

resources/assets/js/app.js
lindsayrmaher: I was at first thinking maybe we could make this more re-usable, but for right now i agree we could make the naming more explicit

lindsayrmaher:
@mendelB (&& @aaronschachter) made some updates on 210d913 & 3ac006c based on your feedback!!

resources/assets/js/app.js
```diff
@@ -34,6 +35,9 @@ if (document.documentElement.lang !== 'en') {
 // Initialize analytics.
 Analytics.init();
 
+//Initialize Voter Registration Content Toggle.
+VoterRegistrationContent.init();
```
mendelB, line 39: Sorry, another naming gripe 🙃-- what do you think of calling this `VoterRegistrationCtaToggle`?

resources/assets/js/utilities/VoterRegistrationContent.js
```diff
@@ -0,0 +1,28 @@
+const $ = require('jquery');
+
+/**
+ * Utility script to enable toggling visibility of voter registration content
```
mendelB, line 4: Similar to above re the naming, can we amend this to: 
```suggestion
 * Utility script to enable toggling visibility of voter registration call to action on '/profile/about' form
```

resources/assets/js/utilities/VoterRegistrationContent.js
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+function clickHandlerToggleContent(event) {
+    const target = event.target;
+    const content = document.getElementById('voter-reg-cta');
+
+    if(target.value === 'unregistered') {
```
mendelB, line 11: Could it be more readable to just invoke `event.target.value` here instead of defining the variable up top?

resources/assets/js/app.js
lindsayrmaher: let's do it!

PR #1015, opened 2020-04-22
Fixes an issue deserializing date-times before 1970.

DFurnes:
### What's this PR do?

This pull request fixes an issue where any users whose birthdates were before 1970 would not be able to load their profile – requests to `v2/users/{id}` would explode with a 500 error! 💣

When representing a date before [January 1st 1970](https://en.wikipedia.org/wiki/Unix_time), the UNIX timestamp is a negative number (e.g. August 8th 1965 turns into `-138844800`). So far, so good. The bug we're running into here is introduced when _milliseconds_ come into the picture...

For some reason, milliseconds (before 1970) are _also_ a negative number in PHP, so the package would [write the millisecond timestamp](https://github.com/jenssegers/laravel-mongodb/blob/master/src/Jenssegers/Mongodb/Eloquent/Model.php#L102) as `-138844800-324` (rather than `-138844800324`).

I added a failing test test case for this bug in jenssegers/laravel-mongodb#2028. If that gets reviewed and merged, we can remove this overridden method on our own `Model` class.

I don't really understand why these dates have milliseconds on them at all, to be honest. My best guess is that was due to some sort of quirk either in this library or our application code once upon a time. We should probably write a quick script to clean these up sometime.

### Relevant tickets

References [Pivotal #172439494](https://www.pivotaltracker.com/story/show/172439494).

### Checklist

### DIFF
--- a/app/Models/Model.php
+++ b/app/Models/Model.php
@@ -3,6 +3,8 @@
 namespace Northstar\Models;
 
 use Carbon\Carbon;
+use MongoDB\BSON\UTCDateTime;
+use Illuminate\Support\Facades\Date;
 use Illuminate\Database\Eloquent\Builder;
 use Jenssegers\Mongodb\Eloquent\Model as BaseModel;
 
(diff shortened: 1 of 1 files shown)


DFurnes:
Yeah, it was tricky! Newly-saved dates seem to be totally fine, so it's only an issue with data that somehow got milliseconds appended on it in the past (weird import? old app bug? not sure!).

I was able to track down the pattern by looking at affected records in the database & then reproduce locally by directly setting one of those "bad" date on a user record in `mongo`:

```
db.users.insert({"email": "broken@example.com", "birthdate": ISODate("1968-01-17T14:19:49.324Z")});
```

PR #1018, opened 2020-04-24
Cause preference errors

lindsayrmaher:
### What's this PR do?

This pull request changes the approach for how we're removing the causes from a users interests.

👀 

This was an issue for users who had chosen causes when they initially signed up, but then attempted to remove them from their phoenix profile because the method we were using was expecting an array rather than an obj. This is part of some ongoing fixes to work around the underlying problem being created by the storage of causes as objects at sign up rather than the expected array. 

### Relevant tickets

References [Pivotal # 172363507](https://www.pivotaltracker.com/story/show/172363507).

### Checklist

### DIFF
--- a/app/Http/Controllers/CauseUpdateController.php
+++ b/app/Http/Controllers/CauseUpdateController.php
@@ -49,7 +49,13 @@ class CauseUpdateController extends Controller
             abort(404, 'That cause does not exist.');
         }
 
-        $user->pull('causes', $cause);
+        $causesArray = $user->causes;
+
+        $causesArray->pull($cause);
+
+        $user->causes = $causesArray;
+
+        $user->save();
 
         return $this->item($user);
     }
--- a/app/Http/Controllers/CauseUpdateController.php
+++ b/app/Http/Controllers/CauseUpdateController.php
@@ -50,8 +50,7 @@ class CauseUpdateController extends Controller
         }
 
         $causesArray = $user->causes;
-
-        $causesArray->pull($cause);
+        collect($causesArray)->pull($cause);
 
         $user->causes = $causesArray;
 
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/CauseUpdateController.php
```diff
@@ -49,7 +49,13 @@ public function destroy(User $user, string $cause)
             abort(404, 'That cause does not exist.');
         }
 
-        $user->pull('causes', $cause);
+        //using the getter method to pull the array of causes rather than stored obj
```
katiecrane, line 52: Nitpick - Can you updated to `// Using ...`?

app/Http/Controllers/CauseUpdateController.php
```diff
@@ -49,7 +49,13 @@ public function destroy(User $user, string $cause)
...
+        $causesArray = $user->causes;
+
+        // reassigning it to be an array with the item removed
```
katiecrane, line 55: another nit - and uppercase start to this one too please 😬

katiecrane:
I remember at some point talking about a `removeCause()` helper, curious why you decided to not go that route?

app/Http/Controllers/CauseUpdateController.php
lindsayrmaher: updated on b89cde2!

app/Http/Controllers/CauseUpdateController.php
lindsayrmaher: updated on b89cde2!

lindsayrmaher:
> I remember at some point talking about a `removeCause()` helper, curious why you decided to not go that route?

I just felt like this was already super readable and adding a helper wasn't super necessary! I could've gone either way though since we added the `addCause` helper

katiecrane:
@lindsayrmaher works for me especially since we only use it here so far - just something to keep in mind if we need to do any cause removin' elsewhere!

PR #1022, opened 2020-05-13
Formalize Sixpack Experiment

Vmack:
### What's this PR do?

This pull request moves social auth buttons to the top of the registration form permanently 

👀

Users convert ~0.7% more when social signup buttons appear above registration form

### Relevant tickets

References [Pivotal #171413064](https://www.pivotaltracker.com/story/show/171413064).

### Checklist

### DIFF
--- a/app/Http/Controllers/Web/AuthController.php
+++ b/app/Http/Controllers/Web/AuthController.php
@@ -199,8 +199,6 @@ class AuthController extends Controller
         $editableFields = $request->except(User::$internal);
         $user = $this->registrar->registerViaWeb($editableFields);
 
-        convert('social-auth-position');
-
         $this->cleanupSession();
 
         Auth::login($user, true);
--- a/app/Http/Controllers/Web/AuthController.php
+++ b/app/Http/Controllers/Web/AuthController.php
@@ -178,6 +178,7 @@ class AuthController extends Controller
     {
         return view('auth.register-beta');
     }
+    
     /**
      * Handle submissions of the registration form.
      *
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/Web/AuthController.php
```diff
@@ -176,11 +176,10 @@ public function getLogout(Request $request)
      */
     public function getRegister()
     {
-        $positionSocialAuth = participate('social-auth-position', ['position_bottom', 'position_top']);
+        $positionSocialAuth = participate('social-auth-position', ['position_top']);
```
katiecrane, line 179: We can remove this whole line! `participate()` is what actually puts someone in an a/b test so we don't need it anymore 🔥

Vmack: ahah! ok gotcha!

resources/views/auth/register-beta.blade.php
```diff
@@ -69,22 +69,6 @@
         </div>
     </form>
 
-    @if ($social_auth_position === 'position_bottom')
```
katiecrane, line 72: There's an `if` statement wrapped around the social share buttons up at the top - we can remove the conditional and just always display the buttons now!

katiecrane:
This looks great so far! There are a few more remnants left in the codebase. When someone converts on an a/b test we need to track it so we have a few:
```js
convert('social-auth-position');
```
one in AuthController and then one in GoogleController and FacebookController too! So those can all be removed as well

PR #1023, opened 2020-05-15
Adds Voter id back to the link for analytics

lindsayrmaher:
### What's this PR do?

Adds an id to the link in the onboarding flow for voter registration. We need this for our analytics to recognize that the tracking event needs to get triggered.

👀 

This was added initially but it got removed in a styling/refactoring PR and we stopped tracking!

### Relevant tickets

References [Pivotal # 172863740](https://www.pivotaltracker.com/story/show/172863740).

### Checklist

### DIFF
--- a/resources/views/profiles/about/edit.blade.php
+++ b/resources/views/profiles/about/edit.blade.php
@@ -55,7 +55,7 @@
             </div>
             <div id="voter-reg-cta" class="w-full hidden">
                 <p>Make your voice heard on the issues that matter to you. Take 2 minutes and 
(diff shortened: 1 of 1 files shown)


lindsayrmaher:
> I wonder if we shouldn't start implementing the pattern of `data-test/id` as we do in Phoenix 🤔 (not to hold you back here, just thinking aloud!)

Yeah i thought about that as well, i'm kind of torn whether to start changing that pattern in NS when we follow the id/class pattern for all of our other analytics tracking right now.

PR #1024, opened 2020-05-22
Adds an additional prompt to the reg flow for uncertain voters

lindsayrmaher:
### What's this PR do?

This pull request adds a different prompt for uncertain voters to the question during registration regarding their voter reg status. It also starts adding some test coverage for these prompts and adds analytics tracking to the new link.

👀 

We want to reach this portion of our new members!

I initially added test coverage for all 3 scenarios of voter registration options being selected, but ran into an issue with not being able to account for run time functionality with Laravel. ([see slack thread](https://dosomething.slack.com/archives/CUQMU4Q6B/p1590508302008300) and commit removing the tests a7dd996)

We decided to reduce scope creep for this work, but added a [collective icebox ticket](https://www.pivotaltracker.com/story/show/173020906) for adding support in Northstar for this kind of testing in the future. 

uncertainvoterregprompt

### Relevant tickets

References [Pivotal # 172848257](https://www.pivotaltracker.com/story/show/172848257).

### Checklist

### DIFF
--- a/resources/assets/js/utilities/VoterRegistrationCtaToggle.js
+++ b/resources/assets/js/utilities/VoterRegistrationCtaToggle.js
@@ -12,14 +12,11 @@ function clickHandlerToggleContent(event) {
 
     if(value === 'unregistered') {
         unregisteredContent.classList.remove('hidden')
-        if(!uncertainContent.classList.contains('hidden')) {
-            uncertainContent.classList.add('hidden')
-        }
+        uncertainContent.classList.add('hidden')
+    
     } else if(value === 'uncertain') {
         uncertainContent.classList.remove('hidden')
-        if(!unregisteredContent.classList.contains('hidden')) {
-            unregisteredContent.classList.add('hidden')
-        }
+        unregisteredContent.classList.add('hidden')
     } else {
         unregisteredContent.classList.add('hidden');
         uncertainContent.classList.add('hidden');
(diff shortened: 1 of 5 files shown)


resources/assets/js/utilities/VoterRegistrationCtaToggle.js
```diff
@@ -6,12 +6,23 @@ const $ = require('jquery');
...
+    const unCertainContent = document.getElementById('voter-reg-cta-uncertain');
+    const value = event.target.value
 
-    if(event.target.value === 'unregistered') {
-        content.classList.remove('hidden')
+    if(value === 'unregistered') {
+        unRegisteredContent.classList.remove('hidden')
+        if(!unCertainContent.classList.contains('hidden')) {
+            unCertainContent.classList.add('hidden')
+        }
```
mendelB, line 17: My fault for not realizing this - but I think the `add` will only add the class _if_ it's not already in the list which means we don't need the conditional 🎊

resources/assets/js/utilities/VoterRegistrationCtaToggle.js
```diff
@@ -6,12 +6,23 @@ const $ = require('jquery');
  */
 
 function clickHandlerToggleContent(event) {
-    const content = document.getElementById('voter-reg-cta');
+    const unRegisteredContent = document.getElementById('voter-reg-cta-unregistered');
+    const unCertainContent = document.getElementById('voter-reg-cta-uncertain');
```
aaronschachter, line 10: Nit, can we rename to `unregisteredContent` and `uncertainContent` to stay consistent with the `unregistered` and `uncertain` values we save for this field?

aaronschachter:
Looking good so far!

PR #1027, opened 2020-06-24
Store referrer_user_id on the User instead of source_details

mendelB:
### What's this PR do?

This pull request stores incoming auth requests with `?referrer_user_id` query params in the session directly (not nested in the `source_detail`), then sets the registring user's `referrer_user_id` field to this value.

👀 

- If we add more fields to the top level, it could be nice to nest them under one session key and assign them to the user more elegantly instead of individually, but figured for this one field this works fine. (`source_detail` contains it's own mechanics so they definitely can't be grouped together).
- I asked a question [in Slack](https://dosomething.slack.com/archives/CUQMU4Q6B/p1593023731003100) about testing the session storage part of this flow since it doesn't look like we have anything set up yet. For now, I've at least added the applicable test to the `WebAuthenticationTest` for the Registrar's hand in this.

### Relevant tickets

References [Pivotal #172981900](https://www.pivotaltracker.com/story/show/172981900).

### Checklist

Screen Shot 2020-06-24 at 2 49 08 PM

### DIFF
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -233,6 +233,11 @@ class Registrar
         }
 
         $user->setSource(null, $sourceDetail ? stringify_object($sourceDetail) : null);
+
+        if (session('referrer_user_id')) {
+            $user->referrer_user_id = session('referrer_user_id');
+        }
+
         // Set the user's country code by Fastly geo-location header.
         $user->country = country_code();
         // Set the user's zip code by Fastly geo-location header.
--- a/app/Http/Controllers/Web/AuthController.php
+++ b/app/Http/Controllers/Web/AuthController.php
@@ -67,14 +67,15 @@ class AuthController extends Controller
                 'title' => request()->query('title', trans('auth.get_started.create_account')),
                 'callToAction' => request()->query('callToAction', trans('auth.get_started.call_to_action')),
                 'coverImage' => request()->query('coverImage', asset('members.jpg')),
-                // Store any provided UTMs, Referrer User ID, or Contentful ID for user's source_detail:
+                // Store any provided UTMs or Contentful ID for user's source_detail:
                 'source_detail' => array_filter([
                     'contentful_id' => request()->query('contentful_id'),
-                    'referrer_user_id' => request()->query('referrer_user_id'),
                     'utm_source' => request()->query('utm_source'),
                     'utm_medium' => request()->query('utm_medium'),
                     'utm_campaign' => request()->query('utm_campaign'),
                 ]),
+                // Store the provided Referrer User ID for user's referrer_user_id field:
+                'referrer_user_id' => request()->query('referrer_user_id'),
             ]);
 
             // Optionally, we can override the default authorization page using `?mode=login`.
--- a/tests/Http/Web/WebAuthenticationTest.php
+++ b/tests/Http/Web/WebAuthenticationTest.php
@@ -198,6 +198,21 @@ class WebAuthenticationTest extends BrowserKitTestCase
     }
 
     /**
+     * Test that a referrar_user_id in session is attached to the registering user.
+     */
+    public function testRegisterBetaWithReferrerUserId()
+    {
+        // Mock a session for the user with ?referrer_user_id=x param, indicating a referred user.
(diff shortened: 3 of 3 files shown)


app/Auth/Registrar.php
```diff
@@ -233,6 +233,11 @@ public function registerViaWeb($input, $authSource = null)
         }
 
         $user->setSource(null, $sourceDetail ? stringify_object($sourceDetail) : null);
+
+        if (session('referrer_user_id')) {
+            $user->referrer_user_id = session('referrer_user_id');
```
mendelB, line 238: Do we need to validate this field at all? Is there a vulnerability in this value being from a query param that can be manipulated on the Web side?

aaronschachter: We don't seem to be sanitizing any of the other query parameters...but I could see it not exactly hurting us to save only alphanumeric values. I wonder if the data team has an opinion on whether to discard referrer user id's that aren't valid...  which reminds me that I need to figure out how to validate whether I fixed https://www.pivotaltracker.com/story/show/172465010 in https://github.com/DoSomething/phoenix-next/pull/2114 which is a bit off topic.

app/Http/Controllers/Web/AuthController.php
```diff
@@ -67,14 +67,15 @@ public function getAuthorize(ServerRequestInterface $request, ResponseInterface
...
+                // Store any provided UTMs or Contentful ID for user's source_detail:
                 'source_detail' => array_filter([
                     'contentful_id' => request()->query('contentful_id'),
-                    'referrer_user_id' => request()->query('referrer_user_id'),
                     'utm_source' => request()->query('utm_source'),
                     'utm_medium' => request()->query('utm_medium'),
                     'utm_campaign' => request()->query('utm_campaign'),
                 ]),
+                // Store the provided Referrer User ID for user's referrer_user_id field:
+                'referrer_user_id' => request()->query('referrer_user_id'),
```
aaronschachter, line 78: Nit but it could be nice to define these array keys alphabetically (`title` throws it off, but maybe could get moved to this line instead)

tests/Http/Web/WebAuthenticationTest.php
```diff
@@ -197,6 +197,21 @@ public function testRegisterBeta()
         $this->assertEquals(['community'], $user->email_subscription_topics);
     }
 
+    /**
+     * Test that a referrar_user_id in session is attached to the registering user.
```
aaronschachter, line 201: rar 🦁

app/Auth/Registrar.php
aaronschachter: I think this is fine to merge as-is, and revisit with better (or any) guidance for whether to validate.

mendelB: That makes sense! I'll flag this Q to the data team first while I inform them of this change.

app/Http/Controllers/Web/AuthController.php
mendelB: They seem to be grouped logically per that comment on top of the registration customization fields
https://github.com/DoSomething/northstar/blob/80e122237bd296921eb58f1bcaf86815069c728d/app/Http/Controllers/Web/AuthController.php#L65

Do you think it's worth leaving as-is to maintain this logical grouping vs alphabetical sorting?

tests/Http/Web/WebAuthenticationTest.php
mendelB: Oh spelling, what art thou. (fixed in https://github.com/DoSomething/northstar/pull/1027/commits/11c0b4be5617101aacebac045a084c3dd1d2adfa)

app/Http/Controllers/Web/AuthController.php
aaronschachter: Ohh, I see -- keeping as-is sounds fine.

app/Auth/Registrar.php
mendelB: And just following up here, we created [Pivotal #173514593](https://www.pivotaltracker.com/story/show/173514593) to address the validation.

PR #1029, opened 2020-07-01
Adds Mutator to Ensure that Null is not Returned When User Re-Subscribes

Vmack:
### What's this PR do?

This pull request is a bug fix to a problem we are having when global unsubscribe link is clicked. This commit fixes the "null" error returned when a user attempts to re-subscribe.

Is this function formatted correctly? Are the comments fitting?

See [thread](https://dosomething.slack.com/archives/CUQMU4Q6B/p1593187616006000) 

### Relevant tickets

References [Pivotal #172444740](https://www.pivotaltracker.com/story/show/172444740).

### Checklist

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -725,7 +725,7 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
      */
     public function getEmailSubscriptionTopicsAttribute($value)
     {
-        //Set empty value equal to array of email_subscription_topics
+        //Ensure we always return an array value for the email_subscription_topics attribute.
         return empty($value) ? [] : $value;
     }
 
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -719,6 +719,17 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
     }
 
     /**
+     * Mutator to ensure null is not returned when user re-subscribes.
+     * 
+     * @param array $value
+     */
+    public function getEmailSubscriptionTopicsAttribute($value)
+    {   
+        //Set empty value equal to array of email_subscription_topics
+        return empty($value) ? [] : $value;
+    }
+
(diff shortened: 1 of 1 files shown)


app/Models/User.php
```diff
@@ -718,6 +718,17 @@ public function setEmailSubscriptionTopicsAttribute($value)
         $this->attributes['email_subscription_topics'] = array_values(array_unique($value));
     }
 
+    /**
+     * Mutator to ensure null is not returned when user re-subscribes.
```
mendelB, line 722: I think it would be helpful to throw in a link to our discussion so we have that context, and hopefully get around to fixing this `null` bug!

app/Models/User.php
```diff
@@ -718,6 +718,17 @@ public function setEmailSubscriptionTopicsAttribute($value)
...
+     *
+     * @param array $value
+     */
+    public function getEmailSubscriptionTopicsAttribute($value)
+    {
+        //Set empty value equal to array of email_subscription_topics
```
mendelB, line 728: Total nit, but maybe something like?
```suggestion
        // Ensure we always return an array value for the email_subscription_topics attribute.
```

(I also wonder if this comment is kinda covered by our method docblock description, but I see this pattern in other method too).

app/Models/User.php
Vmack: Hey @mendelB it's there actually, in the "more background context" section.

app/Models/User.php
Vmack: sounds good!

app/Models/User.php
mendelB: Ah neat! I think in this case it's also useful to have it in the codebase itself in addition to the PR, so that we kinda have the built-in TODO to reevaluate with the context for this method.

PR #1030, opened 2020-07-06
Adding a voter reg prompt for members who say "yes"

lindsayrmaher:
# What's this PR do?

This pull request adds a prompt for users who choose the "confirmed" option in response to whether they are registered to vote while they fill out member profile information.

### How should this be reviewed? 

Is there a way to simplify the JS? I went to for the easiest way forward, which feels a little repetitive but simple.

### Any background context you want to provide?

We've seen positive numbers from the other two prompts that have been added to the membership flow!

Large Screen
Screen Shot 2020-07-06 at 1 31 00 PM

Medium Screen
Screen Shot 2020-07-06 at 1 31 19 PM

Kapture 2020-07-06 at 13 33 12

### Relevant tickets

References [Pivotal # 173282289](https://www.pivotaltracker.com/story/show/173282289).

### Checklist

- [ ] This PR has been added to the relevant Pivotal card.
- [ ] Documentation added for new features/changed endpoints.
- [ ] Added appropriate feature/unit tests.
- [ ] If new attributes were added to users, there is a corresponding PR to surface these in Aurora.
- [ ] If new attributes were added to users, then the data team already knows about these changes.

PR #1031, opened 2020-07-06
Adds analytics event for confirmed users

lindsayrmaher:
### What's this PR do?

This pull request is a follow up to #1030 and adds analytics tracking for a user who chooses "yes" as their voter registration status.

👀 

This prompt is serving as a reminder for users to double check their status, hence the naming choice.

### Relevant tickets

References [Pivotal # 173282289](https://www.pivotaltracker.com/story/show/173282289).

### Checklist

### DIFF
--- a/resources/assets/js/utilities/Analytics.js
+++ b/resources/assets/js/utilities/Analytics.js
@@ -547,6 +547,20 @@ function init() {
       });
     });
 
+    $('#voter-reg-status-link-confirmed').on('click', () => {
+      // Tracks clicking on the Check Registration Status Link in the Onboarding flow for confirmed users.
+      trackAnalyticsEvent({
+        metadata: {
+          adjective: 'double_check_voter_registration',
+          category: 'onboarding',
+          noun: 'link_action',
+          target: 'link',
+          verb: 'clicked',
+          label: 'voter_registration',
+        },
+      });
+    });
+
     // Check for and track validation errors returned from the backend after form submission.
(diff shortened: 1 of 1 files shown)


resources/assets/js/utilities/Analytics.js
```diff
@@ -547,6 +547,20 @@ function init() {
...
+    $('#voter-reg-status-link-confirmed').on('click', () => {
+      // Tracks clicking on the Check Registration Status Link in the Onboarding flow for confirmed users.
+      trackAnalyticsEvent({
+        metadata: {
+          adjective: 'double_check_voter_registration',
+          category: 'onboarding',
+          noun: 'link_action',
+          target: 'link',
+          verb: 'clicked',
+          label: 'voter_registration',
```
katiecrane, line 559: Is "label" needed here? Asking because the above "uncertain" event doesn't have this and they seem similar!

katiecrane: Oh right this is a follow-up to that work so makes sense that they're similar

lindsayrmaher: The label was included on the pivotal ticket, but they should be very similar so I can double check on that.

lindsayrmaher: @katiecrane checked in with Tej and we decided to keep it consistent with the other events. Dropping the label! Thanks for flagging

PR #1033, opened 2020-07-07
Use `addTopic` Helper in Northstar When Adding One Email Subscription Topic

Vmack:
### What's this PR do?

This pull request adds the `addEmailSubscriptionTopic` helper so that email_subscription_topic will properly update after a user unsubscribes instead of returning null when a user re-subscribes.

Here's the [thread](https://dosomething.slack.com/archives/CUQMU4Q6B/p1594069313011300) where it was addressed for more context.

👀

This is a bug fix to [#2201](https://github.com/DoSomething/phoenix-next/pull/2201)

### Relevant tickets

References [Pivotal #173691415](https://www.pivotaltracker.com/story/show/173691415).

### Checklist

### DIFF
--- a/app/Http/Controllers/SubscriptionUpdateController.php
+++ b/app/Http/Controllers/SubscriptionUpdateController.php
@@ -36,6 +36,8 @@ class SubscriptionUpdateController extends Controller
 
         $user->addEmailSubscriptionTopic($topic);
 
+        $user->save();
+
         return $this->item($user);
     }
 
--- a/app/Http/Controllers/SubscriptionUpdateController.php
+++ b/app/Http/Controllers/SubscriptionUpdateController.php
@@ -34,7 +34,7 @@ class SubscriptionUpdateController extends Controller
             abort(404, 'That subscription does not exist.');
         }
 
-        $user->push('email_subscription_topics', $topic, true);
+        $user->addEmailSubscriptionTopic($topic);
 
         return $this->item($user);
     }
--- a/tests/Http/SubscriptionUpdateTest.php
+++ b/tests/Http/SubscriptionUpdateTest.php
@@ -54,6 +54,7 @@ class SubscriptionUpdateTest extends BrowserKitTestCase
 
         $this->assertResponseStatus(200);
         $this->seeJsonField('data.email_subscription_topics', []);
+        $this->assertEquals(['news'], $user->fresh()->email_subscription_topics);
     }
 
     /**
--- a/tests/Http/SubscriptionUpdateTest.php
+++ b/tests/Http/SubscriptionUpdateTest.php
@@ -18,6 +18,7 @@ class SubscriptionUpdateTest extends BrowserKitTestCase
 
         $this->assertResponseStatus(200);
         $this->seeJsonField('data.email_subscription_topics', ['news']);
+        $this->assertEquals(['news'], $user->fresh()->email_subscription_topics);
     }
 
     /**
@@ -36,6 +37,7 @@ class SubscriptionUpdateTest extends BrowserKitTestCase
 
         $this->assertResponseStatus(200);
         $this->seeJsonField('data.email_subscription_topics', ['news']);
+        $this->assertEquals(['news'], $user->fresh()->email_subscription_topics);
     }
 
     /**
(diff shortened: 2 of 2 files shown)


app/Http/Controllers/SubscriptionUpdateController.php
```diff
@@ -34,7 +34,7 @@ public function store(User $user, string $topic)
             abort(404, 'That subscription does not exist.');
         }
 
-        $user->push('email_subscription_topics', $topic, true);
+        $user->addEmailSubscriptionTopic($topic);
```
mendelB, line 37: I think, similarly to the `CauseUpdateController#store` method, we'd need to also add `$user->save()` after this to actually persist our change in the database since we don't get the automatic update that comes with using `push` anymore.

DFurnes: Good eyes Mendel! I was a bit surprised our test suite didn't catch this, since we do [have test coverage](https://github.com/DoSomething/northstar/blob/master/tests/Http/SubscriptionUpdateTest.php) for this controller. The problem is that we only test the "in-memory" `$user` returned from the response, though, and not what's persisted to the database! It'd be nice to add a check for that to each test method, like this:

```diff
  $this->assertResponseStatus(200);
  $this->seeJsonField('data.email_subscription_topics', ['news']);
+ $this->assertEquals(['news'], $user->fresh()->email_subscription_topics);
```

mendelB: Oooh interesting! Funny, I wouldn't intuitively think that checking the db itself would be necessary helpful... The More You Know™️

Vmack: oh that's very important! Thanks guys!

Vmack: hey guys let me know if this looks good @mendelB @DFurnes

tests/Http/SubscriptionUpdateTest.php
```diff
@@ -54,6 +54,7 @@ public function testRemoveSubscription()
 
         $this->assertResponseStatus(200);
         $this->seeJsonField('data.email_subscription_topics', []);
+        $this->assertEquals(['news'], $user->fresh()->email_subscription_topics);
```
mendelB, line 57: Ooh tricky, looks like this test is failing. I think we'd actually want this new assertion in the 
`testAddSubscription` test above:
https://github.com/DoSomething/northstar/blob/250e7f9fd503cdb48410aa76eb32131d666d4a32/tests/Http/SubscriptionUpdateTest.php#L7-L21

Here, we're testing the _removal_ of the subscription so we'd actually expect this to _not_ be in the database. 

Though, it could be nice to even leave this in while we have it per Dave's point, and instead assert that this does _not_ show up, and that the `email_subscription_topics` list is now empty 🤔 
```suggestion
        $this->assertEquals([], $user->fresh()->email_subscription_topics);
```

DFurnes: Yep – this one should assert that the array is empty `[]`. The `testAddSubscription` and `testAddExistingSubscription ` tests above should both assert that the array is `['news']`.

Vmack: aha! thanks for the clarification guys, makes a lot of sense 🙌🏽

Vmack: So if I'm clear I'm adding the same test to `TestAddSubscription` and `testAddExistingSubscription` but the `testRemoveSubscription` test should be the only one with an empty `[]` because we're expecting it to not be tracked in the db?

DFurnes: Right!

- In `testAddSubscription` we're adding a `'news'` subscription to a user who has no current subscriptions, so we'd expect their user profile to have `['news']` after the request is completed.
- In `testAddExistingSubscription`, we're adding a `'news'` subscription but the user already has it (and no other subscriptions), so we'd expect their user profile to have `['news']` after the request is completed.
- Here, in `testRemoveSubscription`, we're _removing_ a `'news'` subscription from a user that already has it (and no other subscriptions), so we'd expect their user profile to have `[]` no subscriptions after the request is completed.

Vmack: Excellent! This is so clear!

PR #1036, opened 2020-07-17
Adding a pre registration prompt for users who may be eligible

lindsayrmaher:
### What's this PR do? 

This pull request adds a second JS function to toggle the pre registration prompt for voters when they enter the correct pre reg age. It also suppresses the unregistered prompt since they are in a specific group. 

Let me know if this seems to make sense and covers the age check accurately. It seems to work as expected, but there should perhaps be a test that enters in the expected dates. I'm also seeing this JS error when I'm on the initial log in or register step. I don't see it for the other event listener so i feel like i might have defined it slightly off. 

Screen Shot 2020-07-17 at 1 16 14 PM

Behavior in App:
Kapture 2020-07-17 at 13 19 11

Last level of VR prompt in the registration flow to see if there are any more users we can catch at this stage of sign up.

### Relevant tickets

References [Pivotal # 172094857](https://www.pivotaltracker.com/story/show/172094857).

### Checklist

### DIFF
--- a/resources/assets/js/utilities/VoterRegistrationCtaToggle.js
+++ b/resources/assets/js/utilities/VoterRegistrationCtaToggle.js
@@ -17,7 +17,6 @@ function togglePreregistrationPrompt() {
     if(birthdate.length === 10 && new Date(birthdate)) {
         const newBirthdate = new Date(birthdate);
         const age = calculateAge(newBirthdate);
-        console.log('this is the age', age)
         if(16 <= age && age < 18) {
             preRegisteredContent.classList.remove('hidden');
         }
--- a/resources/assets/js/utilities/VoterRegistrationCtaToggle.js
+++ b/resources/assets/js/utilities/VoterRegistrationCtaToggle.js
@@ -5,8 +5,8 @@ const $ = require('jquery');
  * based on birthdate on '/profile/about' form
  */
 
-function togglePreregistrationPrompt() {
-    const birthdate = document.getElementById('birthdate').value;
+function togglePreregistrationPrompt(event) {
+    const birthdate = event.target.value;
     const preRegisteredContent = document.getElementById('voter-reg-cta-pre-registration');
 
     const calculateAge = (date) => {
@@ -17,6 +17,7 @@ function togglePreregistrationPrompt() {
     if(birthdate.length === 10 && new Date(birthdate)) {
         const newBirthdate = new Date(birthdate);
         const age = calculateAge(newBirthdate);
+
         if(16 <= age && age < 18) {
             preRegisteredContent.classList.remove('hidden');
         }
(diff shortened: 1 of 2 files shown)


weerd:
> Let me know if this seems to make sense and covers the age check accurately. It seems to work as expected, but there should perhaps be a test that enters in the expected dates. I'm also seeing this JS error when I'm on the initial log in or register step. I don't see it for the other event listener so i feel like i might have defined it slightly off.

@lindsayrmaher do you mean it specifically started showing after the addition you added or error had always been there?

lindsayrmaher:
> @lindsayrmaher do you mean it specifically started showing after the addition you added or error had always been there?

I can't say for sure, but it's throwing an error specifically around the addition I made for this PR.

resources/assets/js/utilities/VoterRegistrationCtaToggle.js
```diff
@@ -1,17 +1,44 @@
...
+
+    const calculateAge = (date) => {
+        const now = new Date();
+        return now.getFullYear() - date.getFullYear();
+    }
+
+    if(birthdate.length === 10 && new Date(birthdate)) {
+        const newBirthdate = new Date(birthdate);
+        const age = calculateAge(newBirthdate);
+        if(16 <= age && age < 18) {
```
weerd, line 20: 👾 add some breathing room between the variable declarations and the conditional 😌

resources/assets/js/utilities/VoterRegistrationCtaToggle.js
```diff
@@ -1,17 +1,44 @@
 const $ = require('jquery');
 
+/**
+ * Utility script to enable toggling visibility of pre registration for voting 
+ * based on birthdate on '/profile/about' form
+ */
+
+function togglePreregistrationPrompt() {
```
weerd, line 8: I think maybe the error is related to not passing in the target event into the function. You don't really need to grab it from the `document` again since it will be passed as part of the argument to the function:

```js
function togglePreregistrationPrompt(e) {
  const birthdate = e.target.value;

  // etc...
}

Comment (weerd) [reply resources/assets/js/utilities/VoterRegistrationCtaToggle.js]: @lindsayrmaher did this change fix the console error or help at all?

Comment (lindsayrmaher) [reply resources/assets/js/utilities/VoterRegistrationCtaToggle.js]: this didn't change the error, but i added a check in to the `init` function which stopped it from trying to add the event listener before it was on the page where that field actually exists.

File: resources/assets/js/utilities/VoterRegistrationCtaToggle.js [hunk 3]
```diff
@@ -1,17 +1,45 @@
...
 
 function clickHandlerToggleContent(event) {
     const confirmedContent = document.getElementById('voter-reg-cta-confirmed');
     const uncertainContent = document.getElementById('voter-reg-cta-uncertain');
     const unregisteredContent = document.getElementById('voter-reg-cta-unregistered');
+    const preRegisteredContent = document.getElementById('voter-reg-cta-pre-registration');
     const value = event.target.value
 
-    if(value === 'unregistered') {
+    if(value === 'unregistered' && window.getComputedStyle(preRegisteredContent).display === 'none') {
```
weerd, line 42: Would this work?

```js
if(value === 'unregistered' && preRegisteredContent.classList.contains("hidden")) { ... }
```

if it does it would simplify the statement and use the same "hidden" class being toggled in `togglePreregistrationPrompt()` :)

PR #1040, opened 2020-07-22
Fixes bug in the age checker

lindsayrmaher:
### What's this PR do?

This pull request adds some additional checks on the calculateAge function for the pre-registration prompt to account for edge cases.

👀 

We noticed this bug in QA, and this mostly just uses more of the logic we use for the input validator.

### Relevant tickets

References [Pivotal # 172094857](https://www.pivotaltracker.com/story/show/172094857).

### Checklist

### DIFF
--- a/resources/assets/js/utilities/VoterRegistrationCtaToggle.js
+++ b/resources/assets/js/utilities/VoterRegistrationCtaToggle.js
@@ -11,7 +11,12 @@ function togglePreregistrationPrompt(event) {
 
     const calculateAge = (date) => {
         const now = new Date();
-        return now.getFullYear() - date.getFullYear();
+        let age = now.getFullYear() - date.getFullYear();
+        const m = now.getMonth() - date.getMonth();
+        if (m < 0 || (m === 0 && now.getDate() < date.getDate())) {
+            age--;
+        }
+        return age;
     }
 
     if(birthdate.length === 10 && new Date(birthdate)) {


resources/assets/js/utilities/VoterRegistrationCtaToggle.js
```diff
@@ -11,7 +11,12 @@ function togglePreregistrationPrompt(event) {
 
     const calculateAge = (date) => {
         const now = new Date();
-        return now.getFullYear() - date.getFullYear();
+        let age = now.getFullYear() - date.getFullYear();
+        const m = now.getMonth() - date.getMonth();
+        if (m < 0 || (m === 0 && now.getDate() < date.getDate())) {
+            age--;
+        }
+        return age;
```
weerd, line 19: Looks good. But also curious, don't we have a birthday validation function? Could that be used or is it too specific and doesn't apply to this scenario?

lindsayrmaher: We do have a birthday validation function, but it specifically returns true or false whether the age that was entered is valid vs just returning the age itself. this logic is pulled from one of the checks in the validation.

weerd: Ok, right. I suspected it might not entirely apply, but was curious 😄

PR #1045, opened 2020-08-10
Update Northstar to PHP 7.4.

DFurnes:
### What's this PR do?

This pull request updates Northstar to use PHP 7.4. This allows us to use new language features, like arrow functions, and prepares us for the end of PHP 7.3's [active support window](https://www.php.net/supported-versions.php).

I've split this work into separate commits for each change & bug fix:

🆙 Updating to PHP 7.4 in 8580e9c, 602f098, and 508fa99.

🚁 Removing the `laravel/helpers` package in d501a6d, for consistency with other apps & because the `str_contains` helper we were using there conflicts with a new PHP 8.0 polyfill function.

👻 Updates tests for a change in introduced in a newer `league/oauth2-server` release where the `redirect_uri` parameter is now required when using the authorization code grant in 67b15c9. (This should already be something we're sending along, but I'll test carefully in QA before this hits prod!)

🗝️ The `array_key_exists` method can no longer be used on objects/classes, and so I had to swap the way we check for a property on Google's API responses in 5379957.

📅 There was a change to how MongoDB dates are serialized (inside JSON objects, where they don't seem to be cast-able), so I removes the no longer used `['date']` key in fc14391.

⛽ I've updated [Trusted Proxy](https://github.com/fideloper/TrustedProxy) to 4.x and updated it's configuration (in the middleware) to be consistent with our other applications in 6d4b501 and a09677f.

Here's the high-level [PHP 7.4 release announcement](https://www.php.net/releases/7_4_0.php) and [migration guide](https://www.php.net/manual/en/migration74.php).

### Relevant tickets

References [Pivotal #173629223](https://www.pivotaltracker.com/story/show/173629223).

### Checklist


DFurnes:
This was a tricky one – because Symfony's [PHP 8.0 polyfill](https://github.com/symfony/polyfill-php80) is now being included, it's overriding Laravel's own `str_contains` helper from [laravel/helpers](https://github.com/laravel/helpers) with the upcoming PHP 8.0 function that has the same name (and different behavior). I've opted to remove the helpers package in 8277e32.

DFurnes:
This last one 67b15c9 is interesting – it seems like the `redirect_uri` parameter is required after [this change](https://git.io/JJ98Q) (despite the PR implying otherwise). I believe we send this, regardless, so _shouldn't_ be a breaking change outside of the test suite (but will want to test to be sure)!

PR #1046, opened 2020-08-19
Add Club ID to User Profile

mendelB:
### What's this PR do?

This pull request adds a new club_id attribute to the User model, along with an index migration since we'll be querying & filtering on this field, as well as the accompanying payload attributes & test updates.

I'd recommend a commit-by-commit for a nice piecemeal review!

(50f03b1f0730efa9dc95db5a3297e8c9d3672491 contains some information on the `sparse` index.)

This is per the new Clubs table in Rogue! https://github.com/DoSomething/rogue/pull/1102

Adding a new attribute was deceptively tricky -- I wasn't quite sure of _every_ place to update, and kinda poked around some old PR's and searched on attribute names to try and comprehensively add this attribute in. (I'm not quite sure how to make this clearer but just noting!)

One question that occurred to me was whether this attribute needs to be marked as 'sensitive'. The `school_id` is, and a Rogue club will optionally contain a `school_id`, so perhaps this is a sensitive trail of Personally Identifiable Information? (Though this field _is_ only returned for authorized users, so I felt like that should be enough?)

Our documentation for the returned payload for /users in the API seems to not include the sensitive field inclusion rules. I can open a freezer cube for us to update!

### Relevant tickets

References [Pivotal #174339687](https://www.pivotaltracker.com/story/show/174339687).

### Checklist

- [x] If new attributes were added to users, there is a corresponding PR to surface these in Aurora. _([#174106253](https://www.pivotaltracker.com/story/show/174106253))_

### DIFF
--- a/app/Http/Transformers/UserTransformer.php
+++ b/app/Http/Transformers/UserTransformer.php
@@ -56,7 +56,6 @@ class UserTransformer extends BaseTransformer
             'id' => $user->_id,
             'display_name' => $user->display_name,
             'first_name' => $user->first_name,
-            'display_name' => $user->display_name,
             'last_initial' => $user->last_initial,
             'photo' => null,
         ];
--- a/app/Http/Transformers/UserTransformer.php
+++ b/app/Http/Transformers/UserTransformer.php
@@ -90,7 +90,7 @@ class UserTransformer extends BaseTransformer
             // Email subscription status
             $response['email_subscription_status'] = (bool) $user->email_subscription_status;
 
-            //Cause Areas
+            // Cause Areas
             $response['causes'] = $user->causes;
 
             // Voter registration status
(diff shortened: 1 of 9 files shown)


app/Http/Transformers/UserTransformer.php
```diff
@@ -56,7 +56,6 @@ public function transform(User $user)
             'id' => $user->_id,
             'display_name' => $user->display_name,
             'first_name' => $user->first_name,
-            'display_name' => $user->display_name,
```
DFurnes, line 59: I think this might have been removed by mistake!

DFurnes:
> One question that occurred to me was whether this attribute needs to be marked as 'sensitive'. The `school_id` is, and a Rogue club will optionally contain a `school_id`, so perhaps this is a sensitive trail of Personally Identifiable Information? (Though this field _is_ only returned for authorized users, so I felt like that should be enough?)

Hmm, yeah. I think this is fine to keep on the normal authorized response for now.

> Our documentation for the returned payload for /users in the API seems to not include the sensitive field inclusion rules. I can open a freezer cube for us to update!

Gah! Yeah, we should absolutely document that.

app/Http/Transformers/UserTransformer.php
mendelB: I thought this was a duplicate of line 57?

PR #1048, opened 2020-08-20
Add Clubs Email Subscription Topic

mendelB:
### What's this PR do?

This pull request adds a new `clubs` email subscription topic. (by adding `'clubs'` to the list of valid topics).

Am I missing anything?

I was able to successfully sign a user up for the clubs topic via the API:
image

We'll be signing up any member who uses the [Club Finder](https://www.pivotaltracker.com/story/show/174340202) to join a club, for this new `'clubs'` newsletter.

We'll be using the [Subscriptions Update endpoint](https://github.com/DoSomething/northstar/blob/dbcf7e98ba0d6476a2a3259909936ced3794734f/routes/api.php#L33) for this via GraphQL. But I also added this topic to the general list in the `Registrar#validate` method.

### Relevant tickets

References [Pivotal #174339806](https://www.pivotaltracker.com/story/show/174339806).

### Checklist

- [x] If new attributes were added to users, there is a corresponding PR to surface these in Aurora. ([#174106253](https://www.pivotaltracker.com/story/show/174106253))

### DIFF
--- a/app/Http/Controllers/SubscriptionUpdateController.php
+++ b/app/Http/Controllers/SubscriptionUpdateController.php
@@ -15,7 +15,7 @@ class SubscriptionUpdateController extends Controller
     /**
      * @var array
      */
-    protected $emailSubscriptionTopics = ['news', 'scholarships', 'community', 'lifestyle'];
+    protected $emailSubscriptionTopics = ['news', 'scholarships', 'community', 'lifestyle', 'clubs'];
 
     /**
      * Make a new SubpscriptionUpdateController, inject dependencies,
--- a/app/Http/Controllers/SubscriptionUpdateController.php
+++ b/app/Http/Controllers/SubscriptionUpdateController.php
@@ -13,6 +13,11 @@ class SubscriptionUpdateController extends Controller
     protected $transformer;
 
     /**
+     * @var array
+     */
+    protected $emailSubscriptionTopics = ['news', 'scholarships', 'community', 'lifestyle'];
+
+    /**
      * Make a new SubpscriptionUpdateController, inject dependencies,
      * and set middleware for this controller's methods.
      *
@@ -30,7 +35,7 @@ class SubscriptionUpdateController extends Controller
     {
         $this->authorize('edit-profile', $user);
 
-        if (! in_array($topic, ['news', 'scholarships', 'community', 'lifestyle'])) {
+        if (! in_array($topic, $this->emailSubscriptionTopics)) {
             abort(404, 'That subscription does not exist.');
         }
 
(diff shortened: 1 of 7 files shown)


app/Models/User.php
```diff
@@ -551,6 +551,7 @@ public function toCustomerIoPayload()
             'lifestyle_email_subscription_status' => isset($this->email_subscription_topics) ? in_array('lifestyle', $this->email_subscription_topics) : false,
             'community_email_subscription_status' => isset($this->email_subscription_topics) ? in_array('community', $this->email_subscription_topics) : false,
             'scholarship_email_subscription_status' => isset($this->email_subscription_topics) ? in_array('scholarships', $this->email_subscription_topics) : false,
+            'clubs_email_subscription_status' => isset($this->email_subscription_topics) ? in_array('clubs', $this->email_subscription_topics) : false,
```
mendelB, line 554: Although the `scholarships` topic is plural, we use the singular value in this c.io attribute. I wasn't sure if this was done intentionally or a typo. I thought keeping it consistent with the internal name might be a good bet. But the flip side is an inconsistency in the attributes. 🤔

app/Http/Controllers/SubscriptionUpdateController.php
```diff
@@ -12,6 +12,11 @@ class SubscriptionUpdateController extends Controller
      */
     protected $transformer;
 
+    /**
+     * @var array
+     */
+    protected $emailSubscriptionTopics = ['news', 'scholarships', 'community', 'lifestyle', 'clubs'];
```
aaronschachter, line 18: This definitely beats hardcoding the list of topics repeatedly. Nothing needed for merge but I wonder about formalizing this into a class to DRY validation with the `Auth/Registrar`. I added something like this for the different [types of password reset we support](https://github.com/DoSomething/northstar/blob/master/app/PasswordResetType.php), and at the time wasn't aware of a lot of the [`Types` classes in Rogue](https://github.com/DoSomething/rogue/tree/master/app/Types) that I'd likely model it after now, knowing they exist.

mendelB: Heck yes -- agreed! I was tempted to try something like this tbh, but I was just nervous about the impending Clubs launch, so opted to keep it simple. I'll add a ticket to the icebox -- I think this would be neat for the next cleanup sprint!

app/Models/User.php
weerd: Maybe a future cleanup task? Feels like something the `Types` classes could help with 🤔

mendelB: Totes! It could be neat to be able to just map over the Types and DRY up explicitly defining each attribute. I could add this note to [the ticket](https://www.pivotaltracker.com/story/show/174421934)!

PR #1050, opened 2020-09-02
Club ID Update Customer.io Event

mendelB:
### What's this PR do?

This pull request adds a Customer.io event dispatched when a user's `club_id` attribute is updated.

Does it make sense not to dispatch the event in case of an invalid club? Is this executed properly?
How are the tests for this? 
Is the event name sound? 

We're interested in two emails being triggered through this event (two ['Campaigns'](https://customer.io/docs/campaigns-in-customerio) in customer.io). One, using the provided ID will be to the user or `customer` welcoming them to the club. (No specific attributes needed for that beyond the club name). Two, to the club's leader informing them that the user joined their club (hence those leader attributes).

I first took a page out of @aaronschachter's work in Rogue (https://github.com/DoSomething/rogue/pull/1018) to add a general `CreateCustomerIoEvent` job so we can easily track individual events. I also imitated the flow from those referral events in that work.

A pain point here was some of the test mocking. For one, I had trouble balancing the general mock in `CreatesApplication` for all the other tests involving `club_id` updates to work, while specifically testing against the arguments/return values in our individual tests. We seem to need to re-mock the GraphQL class / Customer.io class in order for this to work.

### Relevant tickets

References [Pivotal #174106264](https://www.pivotaltracker.com/story/show/174106264).

### Checklist

- [ ] Documentation added for new features/changed endpoints. - _do we document these events anywhere_?

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -615,7 +615,7 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
     public function getClubIdUpdatedEventPayload($clubId)
     {
         $club = app(GraphQL::class)->getClubById($clubId);
-        $clubLeader = app(User::class)->find(array_get($club, 'leaderId'));
+        $clubLeader = app(User::class)->find(Arr::get($club, 'leaderId'));
 
         if (! $club || ! $clubLeader) {
             return;
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -607,6 +607,30 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
     }
 
     /**
+     * Get payload for a club_id_updated event.
+     *
+     * @param  number $clubId
+     * @return array
+     */
+    public function getClubIdUpdatedEventPayload($clubId)
+    {
+        $club = app(GraphQL::class)->getClubById($clubId);
+        $clubLeader = app(User::class)->find(array_get($club, 'leaderId'));
+
+        if (!$club || !$clubLeader) {
+            return;
+        }
+
+        return [
+            'club_name' => $club['name'],
+            'club_leader_id' => $club['leaderId'],
+            'club_leader_first_name' => $clubLeader->first_name,
+            'club_leader_display_name' => $clubLeader->display_name,
+            'club_leader_email' => $clubLeader->email,
+        ];
+    }
+
+    /**
      * Scope a query to get all of the users in a group.
      *
      * @return \Illuminate\Database\Eloquent\Builder
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -617,7 +617,7 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
         $club = app(GraphQL::class)->getClubById($clubId);
         $clubLeader = app(User::class)->find(array_get($club, 'leaderId'));
 
-        if (!$club || !$clubLeader) {
+        if (! $club || ! $clubLeader) {
             return;
         }
 
--- a/app/Services/GraphQL.php
+++ b/app/Services/GraphQL.php
@@ -19,7 +19,7 @@ class GraphQL
     public function __construct()
     {
         $this->client = ClientBuilder::build(config('services.graphql.url'), [
-            'headers' => ['apollographql-client-name' => 'northstar']
+            'headers' => ['apollographql-client-name' => 'northstar'],
         ]);
     }
 
--- a/app/Services/GraphQL.php
+++ b/app/Services/GraphQL.php
@@ -44,7 +44,7 @@ class GraphQL
     public function getClubById($clubId)
     {
         $query = '
-        query GetClubById($clubId: Int!) {
+        query GetClubQuery($clubId: Int!) {
           club(id: $clubId) {
             name
             leaderId
--- a/app/Services/GraphQL.php
+++ b/app/Services/GraphQL.php
@@ -18,7 +18,9 @@ class GraphQL
      */
     public function __construct()
     {
-        $this->client = ClientBuilder::build(config('services.graphql.url'));
+        $this->client = ClientBuilder::build(config('services.graphql.url'), [
+            'headers' => [ 'apollographql-client-name' => 'northstar' ]
+        ]);
     }
 
     /**
--- a/app/Services/GraphQL.php
+++ b/app/Services/GraphQL.php
@@ -19,7 +19,7 @@ class GraphQL
     public function __construct()
     {
         $this->client = ClientBuilder::build(config('services.graphql.url'), [
-            'headers' => [ 'apollographql-client-name' => 'northstar' ]
+            'headers' => ['apollographql-client-name' => 'northstar']
         ]);
     }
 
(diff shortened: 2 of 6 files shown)


tests/Models/UserTest.php
```diff
@@ -234,4 +236,59 @@ public function addsDefaultSmsSubscriptionTopicsIfChangingToSubscribed()
...
+
+        $eventPayload = $user->getClubIdUpdatedEventPayload($newClubId);
+
+        // Our event payload attributes should contain the club and club leader values.
+        $this->assertEquals($eventPayload['club_name'], $newClubName);
+        $this->assertEquals($eventPayload['club_leader_first_name'], $clubLeader->first_name);
+        $this->assertEquals($eventPayload['club_leader_display_name'], $clubLeader->display_name);
+        $this->assertEquals($eventPayload['club_leader_email'], $clubLeader->email);
+
+        $this->mock(CustomerIo::class)->shouldReceive('trackEvent')->once()->withSomeOfArgs('club_id_updated', $eventPayload);
```
mendelB, line 272: I couldn't for the life of me figure out how to successfully get the `$user` into the expected args, so I used `withSomeOfArgs` to at least check the main ones for this event specifically.

app/Models/User.php
```diff
@@ -606,6 +606,30 @@ public function toCustomerIoPayload()
...
+    /**
+     * Get payload for a club_id_updated event.
+     *
+     * @param  number $clubId
+     * @return array
+     */
+    public function getClubIdUpdatedEventPayload($clubId)
+    {
+        $club = app(GraphQL::class)->getClubById($clubId);
+        $clubLeader = app(User::class)->find(Arr::get($club, 'leaderId'));
```
mendelB, line 618: We're using the Facade here so we can mock this response in the tests to make sure we're returning the expected values.

mendelB:
I'm a bit confused about these [test failures](https://app.circleci.com/pipelines/github/DoSomething/northstar/387/workflows/b5166c4e-045f-4bf3-9941-3ba7e9c57c5e/jobs/395). I don't see how this new work would cause our Blink service's `userCreate` to be invoked an extra time 🤔

app/Models/User.php
```diff
@@ -606,6 +606,30 @@ public function toCustomerIoPayload()
...
+        if (! $club || ! $clubLeader) {
+            return;
+        }
+
+        return [
+            'club_name' => $club['name'],
+            'club_leader_id' => $club['leaderId'],
+            'club_leader_first_name' => $clubLeader->first_name,
+            'club_leader_display_name' => $clubLeader->display_name,
+            'club_leader_email' => $clubLeader->email,
```
aaronschachter, line 629: What are we sending this along for -- would this appear in the body of the email somewhere?

app/Services/GraphQL.php
```diff
@@ -35,6 +35,29 @@ public function query($query, $variables)
         return $response ? $response->getData() : [];
     }
 
+    /**
+     * Query for a Club by ID.
+     *
+     * @param  $clubId Int
+     * @return array
+     */
+    public function getClubById($clubId)
```
aaronschachter, line 44: I know this follows `getSchoolById` by example... but I wonder if it'd be worth using the same format we use in Phoenix -- `GetClubQuery`.

Because we're not using Apollo client and can't send along an app name, using `Northstar` somewhere in the query name could potentially be a good or terrible idea... e.g. `NorthstarGetClubQuery`

app/Models/User.php
mendelB: We'll be sending an email directly to the club leader, and instead of two separate events for both the user and the leader, the thought is that we can just build a c.io campaign to send the email to `{{event->club_leader_email}}`.

app/Services/GraphQL.php
mendelB: Totally, how about keeping the method name but updating the actual `query` name below?

mendelB: In terms of client naming, I wonder if we can somehow add the [client name header](https://www.apollographql.com/docs/apollo-server/monitoring/metrics/#identifying-distinct-clients) on our own using the [`guzzleOptions` parameter](https://github.com/softonic/graphql-client/blob/69ae17b8222c1de277d8a7660ca016e4259447fa/src/ClientBuilder.php#L10) for our GraphQL client 🤔

mendelB: Oooh looks like this could work according to the [Guzzle docs](http://docs.guzzlephp.org/en/stable/request-options.html#headers), we can pass through default headers when instantiating the client, so something like 
```php
$this->client = ClientBuilder::build(config('services.graphql.url', [
  'headers' => [ 'apollographql-client-name' => 'northstar' ]
]));
```

would do the trick in [our GraphQL service](https://github.com/DoSomething/northstar/blob/a149d56743a6f49650294f53987c8ae3728ff2c2/app/Services/GraphQL.php#L21)!

aaronschachter: Ah, didn't realize this is a header we can send. It'll be great to add this into Rogue and Gambit too, nice find! Also - I meant to comment on the query name, not the method name.

app/Models/User.php
aaronschachter: I thought that might be the case, and initially was wondering about including email in the payload... but realized we send it anyway for user updates. I can understand using one event if it makes things simpler on the Customer.io side.. there's just something that feels a cleaner / DRY to me about sending two different events. Not a biggie though.

I could see potentially a scenario where we'd want to notify club leaders that a user has left a club by changing clubs.. but would also guess letting users change clubs isn't in the cards anytime soon.

aaronschachter (approved):

app/Models/User.php
mendelB: That's fair. I was thinking about it from the event perspective, and it felt like one declarative event for what changed per the user, and then the deduction on the customer.io side to enact two campaign per this event. But like I totally also do see how two separate events is cleaner or at least more explicit!

I suppose for now we can keep this, and nothing stopping us from adding a separate event in the future!

app/Services/GraphQL.php
mendelB: Ah perfect! Updated in https://github.com/DoSomething/northstar/pull/1050/commits/6983104b497bab8f15c4944537ab7369cb7514b6

mendelB: 😱 [It worked!](https://studio.apollographql.com/graph/dosomething-graphql-dev-lambda/clients). Updated in https://github.com/DoSomething/northstar/pull/1050/commits/1add971dd2238fc13cfc4afbbe8278d09d14cdd3. I'll add a Pivotal ticket for us to address in the other apps.

app/Services/GraphQL.php
```diff
@@ -35,6 +37,29 @@ public function query($query, $variables)
...
+
+        $variables = [
+            'clubId' => $clubId,
+        ];
+
+        return $this->query($query, $variables)['club'];
+    }
+
     /**
      * Query for a School by ID.
```
aaronschachter, line 64: I know it's out of scope for this PR, but should we change name of the school GQL query too to match before this goes live?

mendelB: Sure, I think that's a good idea! I'll address in a quick follow-up before we deploy

PR #1051, opened 2020-09-02
Use NIST password rules.

DFurnes:
### What's this PR do?

This pull request updates our password validation rules to match the [current best-practices from NIST](https://pages.nist.gov/800-63-3/sp800-63b.html#sec5), using the [laravel-nist-password-rules](https://github.com/langleyfoxall/laravel-nist-password-rules) package. This includes the following checks:

- **Changed:** The password must be at least 8 characters long, rather than our previous 6 character minimum.
- **New:** The password cannot be a dictionary word (e.g. `volunteer`).
- **New:** The password can't include (or be too close to) the user's email or "DoSomething.org".
- **New:** The password can't consist of only repeated or sequential characters (e.g. `aaaaaaaa` or `abcdefgh`).
- **New:** The password can't have been found in a third-party data breach (via the ["Have I Been Pwned" API](https://haveibeenpwned.com/Passwords)).

We customize the rules a little bit in our own `PasswordRules` class, and then apply them where needed.

#### Open Question: Preventing Passwords from Data Breaches

> **New:** The password can't have been found in a third-party data breach.

I'm curious how people feel about this rule. The idea is that if a password has been included in a data breach, we'd let someone know it was not safe to use. (This prevents someone from accidentally using a password that's floating around on the "dark web" and leaving their account open to hackers).

From a security standpoint, I think this is a smart check & safely implemented – the ["Have I Been Pwned" API](https://haveibeenpwned.com/Passwords) is maintained by a reputable security professional, is actively used by software like 1Password, and only matches based on _hashes_ of passwords (so it's not like we're sending user's passwords "in the clear" to a third party).

_That said_, I'm not sure from a UX perspective if this might cause confusion. Is "The password was found in a third party data breach, and can not be used" clear enough to help users understand why we're saying they need to use a different password? Is there a better way of communicating that? cc: @ngjo @ddonizeth @hbghidey 

The intent here is to strike a balance between rules that encourage strong passwords, without adding draconian policies (like adding a certain number of digits or symbols) that prevent users from using [passphrases](https://xkcd.com/936/). Hopefully these rules should not ever affect most users since they're really specific "edge cases" for especially poor passwords.

These rules will _only_ take effect for _new_ passwords (either new registrations or the next time a user changes their password).

I'll be following up in another PR to [update the "Change Password" form to require a user's current password](https://www.pivotaltracker.com/story/show/173717920)

### Relevant tickets

References [Pivotal #173718892](https://www.pivotaltracker.com/story/show/173718892).

### Checklist

### DIFF
--- a/app/Auth/PasswordRules.php
+++ b/app/Auth/PasswordRules.php
@@ -1,0 +1,43 @@
+<?php
+
+namespace Northstar\Auth;
+
+use LangleyFoxall\LaravelNISTPasswordRules\PasswordRules as NISTPasswordRules;
+use LangleyFoxall\LaravelNISTPasswordRules\Rules\BreachedPasswords;
+use LangleyFoxall\LaravelNISTPasswordRules\Rules\ContextSpecificWords;
+use LangleyFoxall\LaravelNISTPasswordRules\Rules\DerivativesOfContextSpecificWords;
+use LangleyFoxall\LaravelNISTPasswordRules\Rules\DictionaryWords;
+use LangleyFoxall\LaravelNISTPasswordRules\Rules\RepetitiveCharacters;
+use LangleyFoxall\LaravelNISTPasswordRules\Rules\SequentialCharacters;
+
+abstract class PasswordRules extends NISTPasswordRules
+{
+    /**
+     * The base rules for setting a new password. These are used to construct
+     * the "change password" rules. (We omit the "confirm" rule here, since
+     * we don't ask users to confirm their password when registering.)
+     */
+    public static function register($username)
+    {
+        return [
+            'required',
+            'string',
+            'min:8',
+            'max:512', // We enforce this maximum to prevent long passwords from bogging down hashing.
+            new SequentialCharacters(),
+            new RepetitiveCharacters(),
+            new DictionaryWords(),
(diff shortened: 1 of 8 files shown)


mendelB:
> That said, I'm not sure from a UX perspective if this might cause confusion. Is "The password was found in a third party data breach, and can not be used" clear enough to help users understand why we're saying they need to use a different password? Is there a better way of communicating that? cc: @ngjo @ddonizeth @hbghidey

I agree that this rule makes sense to include! I also agree that the messaging _might_ be somewhat intimidating or confusing. I'll humbly defer to the UX folks on this one! (Looks like the Slack thread is generating some good content).

DFurnes:
The feedback from the [Slack thread](https://dosomething.slack.com/archives/C09ANFQLA/p1599065392020800) was that telling users their password was found in a data breach would be too scary and confusing, and so the UX recommendation was to instead just say it's "not secure". I've made that change in ac6e1c6.

I'm hoping that whenever we next do some user testing on the registration flow we can see if this message is clear enough for users to take action on, since it may be easy to interpret as a "format" thing, e.g. needs to be longer, have special characters, etc. The good news - changes like that would also probably mean it wouldn't be a leaked password anymore!

PR #1054, opened 2020-09-03
Cleanup for new password reset rules.

DFurnes:
### What's this PR do?

This pull request includes follow-up from #1051:

🍃 Removes an old validation class from our `$dontReport` array that no longer exists in this Laravel release. 47b2d7b

🎋 Write validation errors to the application log. This will help us keep an eye on how often these occur. d8be3c0

*️⃣ Updates the placeholder for the registration form's password field to specify new length requirement. 08a7bad

⛑️ Applies new password validation rules to reset form as well. aa6a19b

I'd recommend reviewing commit-by-commit!

I've got the changes to separate out the "change password" form staged, but holding for another PR to keep this small.

### Relevant tickets

References [Pivotal #173718892](https://www.pivotaltracker.com/story/show/173718892).

### Checklist

### DIFF
--- a/app/Http/Controllers/Web/ResetPasswordController.php
+++ b/app/Http/Controllers/Web/ResetPasswordController.php
@@ -5,6 +5,7 @@ namespace Northstar\Http\Controllers\Web;
 use Illuminate\Support\Str;
 use Illuminate\Http\Request;
 use Illuminate\Foundation\Auth\ResetsPasswords;
+use Northstar\Auth\PasswordRules;
 use Northstar\Events\PasswordUpdated;
 use Northstar\PasswordResetType;
 use Illuminate\Support\Facades\Auth;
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/Web/ResetPasswordController.php
```diff
@@ -35,6 +36,20 @@ public function __construct()
...
+     */
+    protected function rules()
+    {
+        return [
+            'token' => 'required',
+            'email' => 'required|email',
+            'password' => PasswordRules::changePassword(request()->email),
+        ];
+    }
+
```
mendelB, line 52: Where do these rules actually get applied -- is there internal functionality that pulls in the `rules` which is now just customized for this controller?

mendelB (approved):
🎱

app/Http/Controllers/Web/ResetPasswordController.php
DFurnes: Yep! Most of the logic for this controller comes from the `ResetsPasswords` trait, which uses these rules [here](https://github.com/laravel/framework/blob/d2e513a608c34fcea668bdabe9f38c8ed0b335d4/src/Illuminate/Foundation/Auth/ResetsPasswords.php#L40).

DFurnes: It does make me wonder if this is a convention we'd want to adopt for our own shared validation rules in controllers. We tend to store these on a `$this->rules` property, if anything...

mendelB: ohhh cool -- I dig it!

PR #1056, opened 2020-09-16
Fix issues with 'is_dosomething_domain' helper.

DFurnes:
### What's this PR do?

This pull request fixes some issues with the `is_dosomething_domain` helper, and adds tests to verify the correct behavior.

I've added passing tests for what _was_ working in 6b610ca ✅ and some failing tests in a8fef74 ❌ for the reported issue and another bug that I noticed while doing my own testing. I then fixed both of those issues in d01588f. ✅ 

These were interesting bugs! I'll add some more context to the problems & solution in the diff.

### Relevant tickets

References [Pivotal #173719172](https://www.pivotaltracker.com/story/show/173719172).

### Checklist

### DIFF
--- a/tests/HelpersTest.php
+++ b/tests/HelpersTest.php
@@ -62,4 +62,15 @@ class HelpersTest extends BrowserKitTestCase
         $this->asNormalUser()->getJson('/status');
         $this->assertEquals(client_id(), 'phpunit');
     }
+
+    /** @test */
+    public function testIsDoSomethingDomain()
+    {
+        $this->assertTrue(is_dosomething_domain('https://dosomething.org'), 'It should recognize our base domain.');
+        $this->assertTrue(is_dosomething_domain('https://identity.dosomething.org'), 'It should recognize one of our subdomains.');
+        $this->assertTrue(is_dosomething_domain('https://www.dosomething.org/campaigns/teens-for-jeans'), 'It should recognize a nested path.');
+
+        $this->assertFalse(is_dosomething_domain('https://www.google.com'), 'It should reject a non-DoSomething hostname.');
+        $this->assertFalse(is_dosomething_domain('https://dosomething.org.evil.com'), 'It should reject a non-DoSomething hostname with valid "prefix".');
+    }
 }
--- a/tests/HelpersTest.php
+++ b/tests/HelpersTest.php
@@ -72,5 +72,7 @@ class HelpersTest extends BrowserKitTestCase
 
         $this->assertFalse(is_dosomething_domain('https://www.google.com'), 'It should reject a non-DoSomething hostname.');
         $this->assertFalse(is_dosomething_domain('https://dosomething.org.evil.com'), 'It should reject a non-DoSomething hostname with valid "prefix".');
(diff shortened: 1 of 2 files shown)


tests/HelpersTest.php
```diff
@@ -62,4 +62,17 @@ public function testGetClientIdFromOAuth()
...
+    /** @test */
+    public function testIsDoSomethingDomain()
+    {
+        $this->assertTrue(is_dosomething_domain('https://dosomething.org'), 'It should recognize our base domain.');
+        $this->assertTrue(is_dosomething_domain('https://identity.dosomething.org'), 'It should recognize one of our subdomains.');
+        $this->assertTrue(is_dosomething_domain('https://www.dosomething.org/campaigns/teens-for-jeans'), 'It should recognize a nested path.');
+
+        $this->assertFalse(is_dosomething_domain('https://www.google.com'), 'It should reject a non-DoSomething hostname.');
+        $this->assertFalse(is_dosomething_domain('https://dosomething.org.evil.com'), 'It should reject a non-DoSomething hostname with valid "prefix".');
+        $this->assertFalse(is_dosomething_domain('https://www.dontdosomething.org'), 'It should reject a non-DoSomething hostname with valid "suffix".');
```
DFurnes, line 75: This was just a silly oversight – since we only checked if the hostname ended in `dosomething.org`, it used to be possible to sneak an unapproved domain, like `dontdosomething.org`, through as long as it _ended_ in that suffix.

tests/HelpersTest.php
```diff
@@ -62,4 +62,17 @@ public function testGetClientIdFromOAuth()
...
+        $this->assertFalse(is_dosomething_domain('https://cobalt.io\@admin.dosomething.org'), 'It should reject a URL with username hack.');
```
DFurnes, line 76: This one was trickier – `parse_url` correctly(?) parses out `cobalt.io\` as the username in this URL and `admin.dosomething.org` as the hostname (which, then, passes our check).

However, Chrome and Firefox see the `\` escape character on the `@` and instead parse this as a `cobalt.io` hostname with a `@admin.dosomething.org` path. (Which is maybe more correct? I'm not really sure, and truthfully I'm surprised that in this case it doesn't parse `cobalt.io\@admin.dosomething.org` as one long (invalid) domain.

Anyways, this would have let someone sneakily hijack this redirect. Since there's no reason to support redirects that include HTTP basic authentication, I've just blanket denied anything that includes a username in the URL.

tests/HelpersTest.php
```diff
@@ -62,4 +62,17 @@ public function testGetClientIdFromOAuth()
         $this->asNormalUser()->getJson('/status');
         $this->assertEquals(client_id(), 'phpunit');
     }
+
+    /** @test */
+    public function testIsDoSomethingDomain()
+    {
+        $this->assertTrue(is_dosomething_domain('https://dosomething.org'), 'It should recognize our base domain.');
```
mendelB, line 69: Love these test descriptions!

tests/HelpersTest.php
mendelB: 🤯

mendelB (approved):
Fascinating stuff! Thanks for the helpful context.

PR #1058, opened 2020-09-24
increase token life span

Dinnall:
### What's this PR do?

This pull request switched the user's password token life span from 24 hours to 72 hours 

...

...

### Relevant tickets

References [Pivotal #174368567](https://www.pivotaltracker.com/n/projects/2401401/stories/174368567).

### Checklist

### DIFF
--- a/config/auth.php
+++ b/config/auth.php
@@ -106,7 +106,7 @@ return [
             'provider' => 'users',
             'email' => 'auth.emails.password',
             'table' => 'password_resets',
-            'expire' => 1440,
+            'expire' => 4320,
         ],
     ],
 


katiecrane:
@Dinnall this looks good! you can go ahead and switch this from draft to regular PR

config/auth.php
```diff
@@ -106,7 +106,7 @@
             'provider' => 'users',
             'email' => 'auth.emails.password',
             'table' => 'password_resets',
-            'expire' => 1440,
+            'expire' => 4320,
```
mendelB, line 109: Maybe worth throwing in a comment? (I'm just really bad at math 🙃 )
```suggestion
            // 72 hours.
            'expire' => 4320,
```

mendelB (approved):
⌛

config/auth.php
katiecrane: oh good call @mendelB ! then we can see at a glance what this seemingly random number means!

PR #1059, opened 2020-09-25
Update auth.php

Dinnall:
### What's this PR do?

This pull request adds a comment to give more explanation regarding a user's requested token lifespan

...

...

### Relevant tickets

Related to user token [PR](https://github.com/DoSomething/northstar/pull/1058).
PR 
### Checklist

- [  x ] This PR has been added to the relevant Pivotal card.

### DIFF
--- a/config/auth.php
+++ b/config/auth.php
@@ -101,6 +101,15 @@ return [
     |
     */
 
+    /*
+    |--------------------------------------------------------------------------
+    | Users requested token lifespan
+    |--------------------------------------------------------------------------
+    | 'expire' => 1440 // requested token will expire in 1440 minutes/ 24 hours
+    | 'expire' => 4320 // requested token will expire in 4320 minutes/ 72 hours
+    |
+    */
+
     'passwords' => [
         'users' => [
             'provider' => 'users',
--- a/config/auth.php
+++ b/config/auth.php
@@ -101,21 +101,12 @@ return [
     |
     */
 
-    /*
-    |--------------------------------------------------------------------------
-    | Users requested token lifespan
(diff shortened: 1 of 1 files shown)


config/auth.php
```diff
@@ -101,6 +101,15 @@
...
     */
 
+    /*
+    |--------------------------------------------------------------------------
+    | Users requested token lifespan
+    |--------------------------------------------------------------------------
+    | 'expire' => 1440 // requested token will expire in 1440 minutes/ 24 hours
+    | 'expire' => 4320 // requested token will expire in 4320 minutes/ 72 hours
+    |
+    */
```
DFurnes, line 111: Rather than placing this as a block comment, what if we did this as an inline comment next to the `expire` key down below? These (especially with the big ASCII-line headers) are usually used to denote "sections" in the file.

```diff
    'passwords' => [
         'users' => [
             'provider' => 'users',
            'email' => 'auth.emails.password',
            'table' => 'password_resets',
+           'expire' => 4320, // requested token will expire in 4320 minutes / 72 hours
        ],
    ],
```

DFurnes: Another option to make this a bit more readable (and easily changeable in the future!) would be to do some inline math. That way if we, say, wanted to extend this to more days we could just change `3` to `7`!

```php
'expire' => 60 * 24 * 3; // reset link will expire in 3 days (4320 minutes).
```

Dinnall: Updated 👍.  This makes sense, especially since it's specific to `expire` and not the user as a whole.

DFurnes (approved):

PR #1060, opened 2020-09-25
Directly send profile updates to Customer.io.

DFurnes:
### What's this PR do?

This pull request updates Northstar to send profile updates directly to [Customer.io](https://customer.io), rather than routing them through Blink for an extra round of validation, queueing, and rate-limiting. (We're already doing that here!) 

I've split this up into individual commits to make it easier to review:

💬 I added a method to talk directly to Customer.io's ["update customer" endpoint](https://customer.io/docs/api/#apitrackcustomerscustomers_update) in  564826c.

🏭 I've updated our transformer to format fields for Customer.io (e.g. turning all dates into timestamps and renaming `mobile` to `phone`) in b75a292. This handles transformations previously applied in Blink's [`CustomerIoUpdateCustomerMessage.js`](https://github.com/DoSomething/blink/blob/8e7fa6592976354305fe1d2eb7da83863b97979a/src/messages/CustomerIoUpdateCustomerMessage.js).

🏒 Makes Customer.io API calls in our queued event, rather than Blink's event API. 841af7e

📥 _Cleanup:_ I inlined the `CUSTOMER_IO_URL` environment variable in c7a2228f1333f7781d9b6cb4c1d94dbc53441e77. This isn't something we ever need to customize (like we do with our own services, which have multiple environments), so it makes more sense to hard-code to reduce the chance of configuration error.

🚥 _Tests:_ Finally, I've updated test assertions to expect the `CustomerIo` class to be called instead of `Blink` in 7d478be.

For more context, see the [Northstar/Rogue Queue Simplification](https://docs.google.com/document/d/1Mu48WFJqBosA4nzPKCd21ZejV_tQA0GrxWav4cZaIOA/edit#heading=h.kz5h7hwnv2z2) spec.

### Relevant tickets

References Pivotal [#174827016](https://www.pivotaltracker.com/story/show/174827016).

### Checklist

### DIFF
--- a/app/Services/CustomerIo.php
+++ b/app/Services/CustomerIo.php
@@ -2,6 +2,8 @@
 
 namespace Northstar\Services;
 
+use Northstar\Models\User;
+
 class CustomerIo
 {
     /**
@@ -25,6 +27,16 @@ class CustomerIo
     }
 
     /**
+     * Is Customer.io enabled for this app?
+     *
+     * @return bool
+     */
+    protected function enabled(): bool
+    {
+        return config('features.blink');
+    }
+
+    /**
      * Track Customer.io event for given user with given name and data.
      * @see https://customer.io/docs/api/#apitrackeventsevent_add
      *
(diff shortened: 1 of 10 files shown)


DFurnes:
Tests are still failing on the mock assertion for some reason, even though they _should_ be updated in 7d478be... 🤔 I'm going to keep digging into that on Monday, but let me know if anyone sees what I could've missed!

DFurnes:
Okay, this should be good to go now! The issue was that we were calling `$this->mock(CustomerIo::class)` in a bunch of places, which would reset the "global" mock that handles the `updateCustomer` calls that happen on the model observer. I've fixed the test failures by updating these to reference `$this->customerIoMock` property like our other pre-configured mocks.

app/Services/CustomerIo.php
```diff
@@ -24,6 +26,16 @@ public function __construct()
...
     }
 
+    /**
+     * Is Customer.io enabled for this app?
+     *
+     * @return bool
+     */
+    protected function enabled(): bool
+    {
+        return config('features.blink');
```
DFurnes, line 36: Note: I'll rename this & the underlying feature flag in a follow-up PR!

aaronschachter (approved):
✂️

PR #1063, opened 2020-10-01
Send "call to action" password reset events directly to Customer.io.

DFurnes:
### What's this PR do?

This pull request updates Northstar to send the `track_call_to_action_email` password reset/account activation emails directly to Customer.io, rather than proxying them through Blink for a second round of rate-limiting and validation.

I've split this work up into atomic commits:

🍃 I cleaned up the `trackEvent` method to: first, check if Customer.io is enabled for the environment we're in (so we don't need that logic in every job that sends events); use their `json` payload format to remove need for a for-loop; and check response status to throw an exception if something is wrong. ba01f78

🎯 Sends the `track_call_to_action_email` event directly to Customer.io. This was previously sent to Blink's `callToAction` queue which routed it to the correct Customer.io event name [here](https://github.com/DoSomething/blink/blob/dfe6d5ef138bd2fb3d66a105d626f4ab6f62d6c5/src/workers/CustomerIoCallToActionEmailWorker.js#L9). 6882e6b

🗑️ Removes unused configuration and mocks for Blink, now that we no longer use this API here. cd68b94

📄 Updates documentation on the `toCustomerIoPayload` method to explain some API limitations. b3ae448

This fully removes this application's reliance on Blink! 🥍 

### Relevant tickets

References [Pivotal #174827019](https://www.pivotaltracker.com/story/show/174827019).

### Checklist

### DIFF
--- a/app/Jobs/SendPasswordResetToCustomerIo.php
+++ b/app/Jobs/SendPasswordResetToCustomerIo.php
@@ -4,6 +4,7 @@ namespace Northstar\Jobs;
 
 use Northstar\Models\User;
 use Illuminate\Bus\Queueable;
+use Northstar\Services\CustomerIo;
 use Illuminate\Support\Facades\Redis;
 use Illuminate\Queue\SerializesModels;
 use Illuminate\Queue\InteractsWithQueue;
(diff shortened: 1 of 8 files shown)


app/Jobs/SendPasswordResetToCustomerIo.php
```diff
@@ -65,24 +66,16 @@ public function getUrl()
...
-    public function handle()
+    public function handle(CustomerIo $customerIo)
     {
-        // Rate limit Blink/Customer.io API requests to 10/s.
+        // Rate limit Customer.io API requests to 10/s.
         $throttler = Redis::throttle('customerio')->allow(10)->every(1);
-        $throttler->then(function () {
-            $payload = [
+        $throttler->then(function () use ($customerIo) {
+            $customerIo->trackEvent($this->user, 'track_call_to_action_email', [
```
aaronschachter, line 74: Shouldn't this be `call_to_action_email`?
```suggestion
            $customerIo->trackEvent($this->user, 'call_to_action_email', [
```

app/Models/User.php
```diff
@@ -514,8 +514,11 @@ public function indexes()
     }
 
     /**
-     * Transform the user model for Blink.
-     * WARNING: THIS PAYLOAD CAN ONLY INCLUDE 300 ATTRIBUTES!!
+     * Transform the user model for Customer.io's profile schema.
+     *
+     * These values must be simple strings or integers (no objects or arrays).
+     * Dates should be represented as UNIX timestamps (other) in order to use
+     * them in triggers. (Note: this payload is limited to 300 attributes).
```
aaronschachter, line 521: 💯 (x 3)

app/Jobs/SendPasswordResetToCustomerIo.php
DFurnes: Oh, thanks for catching that. I got confused by [this line](https://github.com/DoSomething/blink/blob/dfe6d5ef138bd2fb3d66a105d626f4ab6f62d6c5/src/workers/CustomerIoCallToActionEmailWorker.js#L9) in Blink. Is [this](https://github.com/DoSomething/blink/blob/dfe6d5ef138bd2fb3d66a105d626f4ab6f62d6c5/src/messages/CallToActionEmailMessage.js#L12) where the actual event name is set?

(I want to make sure I'm looking in the right place in the future!)

aaronschachter: Yup, that's where the event name is set. Later on in that file, it's used to send the Customer.io request:
https://github.com/DoSomething/blink/blob/dfe6d5ef138bd2fb3d66a105d626f4ab6f62d6c5/src/messages/CallToActionEmailMessage.js#L37-L41

aaronschachter (approved):

PR #1064, opened 2020-10-02
if statement to redirect users activating accounts

Dinnall:
### What's this PR do?

This pull request send users to Northstar reg `page /about`, after they activate their account

👁️ 
...

We want to redirect a user to `profile/about` IF they've just activated their account.
...

### Relevant tickets

References [Pivotal #174675565](https://www.pivotaltracker.com/n/projects/2401401/stories/174675565).

### Checklist

### DIFF
--- a/app/Http/Controllers/Web/ResetPasswordController.php
+++ b/app/Http/Controllers/Web/ResetPasswordController.php
@@ -137,9 +137,8 @@ class ResetPasswordController extends Controller
     public function redirectPath()
     {
         if (Str::contains(request()->type, 'activate-account')) {
-            return 'profile/about';
+        return 'profile/about';
         }
-        
         return config('services.phoenix.url').'/next/login';
     }
 }
--- a/app/Http/Controllers/Web/ResetPasswordController.php
+++ b/app/Http/Controllers/Web/ResetPasswordController.php
@@ -136,7 +136,8 @@ class ResetPasswordController extends Controller
      */
     public function redirectPath()
     {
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/Web/ResetPasswordController.php
```diff
@@ -136,6 +136,10 @@ protected function guard()
      */
     public function redirectPath()
     {
+        if (Str::contains(request()->type, 'activate-account')) {
+            return 'profile/about';
+        }
+
```
katiecrane, line 142: Looks like StyleCI wants this to be a plain newline with no spaces, so if you remove the spaces from this line (but keep the blank line) it should be good!

Dinnall: I removed the space and moved the return to a new line. Currently still fail withing StyleCI

app/Http/Controllers/Web/ResetPasswordController.php
```diff
@@ -136,6 +136,10 @@ protected function guard()
      */
     public function redirectPath()
     {
+        if (Str::contains(request()->type, 'activate-account')) 
+        {
+        return 'profile/about';
+        }
```
katiecrane, line 142: I think this is what it wants:
```suggestion
        if (Str::contains(request()->type, 'activate-account')) {
            return 'profile/about';
        }

```

katiecrane (approved):
looks good! I think to test on QA we'll have to trigger the activation via the API!

PR #1066, opened 2020-10-23
Escape user-provided input in Customer.io payload.

DFurnes:
### What's this PR do?

This pull request fixes an issue where an evildoer could put HTML in a user-editable field, like `first_name`, and it would be rendered into our Customer.io email templates. This could be used in a phishing attempt to trick someone into clicking a link that we didn't provide them. 🎣 

We want to escape anything that is user-provided (e.g. by a free-form text field). Anything I missed?

I wish that Liquid escaped content by default, like Laravel's default-escaped `{{ $user->first_name }}` vs. unsafe `{!! $user->first_name !!}`! We could use Liquid's [`strip_html` filter](https://customer.io/docs/liquid-tag-master-list), but that'd involve remembering to use `{{ customer.first_name | strip_html }}` everywhere, which is tedious & easy to forget.

### Relevant tickets

References [Pivotal #]().

### Checklist

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -516,30 +516,33 @@ class User extends Model implements AuthenticatableContract, AuthorizableContrac
     /**
      * Transform the user model for Customer.io's profile schema.
      *
-     * These values must be simple strings or integers (no objects or arrays).
-     * Dates should be represented as UNIX timestamps (other) in order to use
-     * them in triggers. (Note: this payload is limited to 300 attributes).
-     *
      * @return array
      */
     public function toCustomerIoPayload()
     {
+        // Please keep in mind the following restrictions when adding fields:
+        //
+        //  - These values may only be strings or integers (no objects or arrays).
+        //  - Any user-provided strings must be sanitized by 'strip_tags'.
+        //  - All dates should be formatted as UNIX timestamps.
+        // 
+        // (Note: this payload is limited to 300 attributes).
         $payload = [
             'id' => $this->id,
             'email' => $this->email,
             'phone' => $this->mobile,
             'sms_status' => $this->sms_status,
-            'sms_status_source' => (isset($this->audit['sms_status']['source'])) ? $this->audit['sms_status']['source'] : null,
+            'sms_status_source' => data_get($this->audit, 'sms_status.source'),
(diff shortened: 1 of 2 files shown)


app/Models/User.php
```diff
@@ -575,12 +581,13 @@ public function toCustomerIoPayload()
...
-            'voting_method' => $this->voting_method,
-            'voting_plan_status' => $this->voting_plan_status,
-            'voting_plan_method_of_transport' => $this->voting_plan_method_of_transport,
-            'voting_plan_time_of_day' => $this->voting_plan_time_of_day,
-            'voting_plan_attending_with' => $this->voting_plan_attending_with,
+            'voting_method' => strip_tags($this->voting_method),
+            'voting_plan_status' => strip_tags($this->voting_plan_status),
+            'voting_plan_method_of_transport' => strip_tags($this->voting_plan_method_of_transport),
+            'voting_plan_time_of_day' => strip_tags($this->voting_plan_time_of_day),
+            'voting_plan_attending_with' => strip_tags($this->voting_plan_attending_with),
```
katiecrane, line 590: I thought some of these were multiple choice? But `voting_plan_attending_with` sounds like it wouldn't be and I guess it can't hurt to run `strip_tags` on things even if it's not strictly necessary?

tests/Models/UserTest.php
```diff
@@ -294,4 +294,16 @@ public function testUpdatingClubIdWithInvalidClub()
 
         $user->update(['club_id' => 123]);
     }
+
+    /** @test */
+    public function it_should_sanitize_user_input()
+    {
+        $evildoer = factory(User::class)->create([
+            'first_name' => '<a href="evil.com">click here</a>',
```
katiecrane, line 302: ah, my favorite website!

katiecrane (approved):
left one non-blocking Q, looks good!

app/Models/User.php
DFurnes: Yeah I think you're right! I wasn't 100% sure so thought it couldn't hurt to be safe! 🙈

tests/Models/UserTest.php
DFurnes: 👹

PR #1069, opened 2020-12-01
Configure Prettier & PHP CS Fixer.

DFurnes:
### What's this PR do?

This pull request configures [Prettier](https://prettier.io) and [PHP CS Fixer](https://cs.symfony.com) for Northstar, following Diego's lead configuring this for Rogue and Aurora! You can read more of the "why" behind why we're doing this in the [PHP Code Formatting RFC](https://github.com/DoSomething/rfcs/blob/master/011-php-code-formatting.md).

I configured the tooling for this in 0adb2e0, and then ran each "formatter" in a follow-up commit.

I _strongly_ recommend reviewing this commit-by-commit because otherwise it's _very_ overwhelming, even though I didn't actually write much code here! I also don't think we need to review the "format" commits all too carefully because they're just automatic formatting & shouldn't (unless there's a Prettier bug!) affect any logic.

Finally, I removed "route comments" (like `POST /users`) from PHPUnit docblocks because PHP CS Fixer started including periods at the end of these "sentences" (awkward) and I don't believe they provide all the much value & weren't consistently included in docblocks.

I'm picking this work up now because we're planning to use this application as our "base" for our new unified backend (just because that makes order-of-operations for this "merger" as simple as possible)! By setting this up ahead of time, we'll keep our nice formatting on anything we pull in here. 💅

### Relevant tickets

References [Pivotal #174652836](https://www.pivotaltracker.com/story/show/174652836).

### Checklist


DFurnes:
Hmm, ignore the failing StyleCI check. Somehow it re-enabled itself. 🤷🏻‍♂️

aaronschachter (approved):
💅 💇 🎨

PR #1070, opened 2020-12-02
Rename 'Northstar\' namespace to 'App\'.

DFurnes:
### What's this PR do?

This pull request renames the `Northstar\` custom namespace to `App\`. 

This is really just a simple find-replace! The PSR-4 namespace was changed in `composer.json`'s `"psr-4"` key, and then I just updated the `namespace` for each file & any references that used the old `Northstar\` prefix to reference those classes.

You can see the original rationale behind this change in the [Remove Custom App Namespace RFC](https://github.com/DoSomething/rfcs/blob/master/012-remove-custom-app-namespace.md).

Another nice benefit of this change (alongside formatting changes from #1069) – you can now easily find any of the classes we've written ourselves grouped at the top of the imports list, since `App` is (almost) always going to be alphabetically first!

### Relevant tickets

References [Pivotal #175983239](https://www.pivotaltracker.com/story/show/175983239).

### Checklist

### DIFF
--- a/app/Auth/DrupalPasswordHash.php
+++ b/app/Auth/DrupalPasswordHash.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace Northstar\Auth;
+namespace App\Auth;
 
 define('DRUPAL_MIN_HASH_COUNT', 7);
 define('DRUPAL_MAX_HASH_COUNT', 30);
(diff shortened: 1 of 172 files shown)


Dinnall (approved):
Looks good to me 🦎

PR #1077, opened 2020-12-10
Add ability to manage redirects to Northstar's admin panel.

DFurnes:
### What's this PR do?

This pull request adds Aurora's Fastly redirect management interface to Northstar's admin panel. This allows staffers to configure redirects for paths on `www.` without a developer's assistance. I've placed these routes under a feature flag so we can wait to "flip the switch" on them until after we come back from break.

These files are more-or-less copied over from their equivalents on Aurora: [`RedirectsController.php`](https://github.com/DoSomething/aurora/blob/2bf89f3a9360d358dd4494ba150323c527869487/app/Http/Controllers/RedirectsController.php), [`Fastly.php`](https://github.com/DoSomething/aurora/tree/2bf89f3a9360d358dd4494ba150323c527869487/app/Services/Fastly.php), [`Resources/*.php`](https://github.com/DoSomething/aurora/tree/2bf89f3a9360d358dd4494ba150323c527869487/app/Resources), and [`views/redirects/*.php`](https://github.com/DoSomething/aurora/tree/2bf89f3a9360d358dd4494ba150323c527869487/resources/views/redirects).

This was the easiest piece to extract since it doesn't have any local data dependencies!

### Relevant tickets

References [Pivotal #175943296](https://www.pivotaltracker.com/story/show/175943296).

### Checklist

### DIFF
--- a/app/Http/Controllers/Web/Admin/RedirectsController.php
+++ b/app/Http/Controllers/Web/Admin/RedirectsController.php
@@ -1,0 +1,141 @@
+<?php
+
+namespace App\Http\Controllers\Web\Admin;
+
+use App\Http\Controllers\Controller;
+use App\Services\Fastly;
+use App\Services\Resources\Redirect;
+use Illuminate\Http\Request;
+
+class RedirectsController extends Controller
+{
+    /**
+     * The Fastly API client.
+     * @var Fastly
+     */
+    protected $fastly;
+
+    /**
+     * Create a RedirectsController.
+     *
+     * @param Fastly $fastly
+     */
+    public function __construct(Fastly $fastly)
+    {
+        $this->fastly = $fastly;
+
+        $this->middleware('auth:web');
+        $this->middleware('role:admin');
+    }
+
+    /**
+     * Display a listing of the resource.
+     *
+     * @param Request $request
+     * @return \Illuminate\Http\Response
+     */
+    public function index(Request $request)
+    {
+        $redirects = $this->fastly->getAllRedirects();
+
+        return view('admin.redirects.index', ['redirects' => $redirects]);
+    }
+
+    /**
+     * Show redirect details.
+     *
+     * @param Redirect $redirect
+     * @return \Illuminate\Http\Response
+     */
+    public function show(Redirect $redirect)
+    {
+        return view('admin.redirects.show', ['redirect' => $redirect]);
+    }
+
+    /**
+     * Create a new redirect.
+     *
+     * @return \Illuminate\Http\Response
+     */
+    public function create()
+    {
+        return view('admin.redirects.create');
+    }
+
+    /**
+     * Store a new redirect.
+     *
+     * @param Request $request
+     * @return \Illuminate\Http\Response
+     */
+    public function store(Request $request)
+    {
+        $this->validate(
+            $request,
+            [
+                'path' => 'required|string|regex:/^[^?]+$/',
+                'target' => 'required|url',
+            ],
+            [
+                'path.regex' => 'Paths cannot contain query strings.',
+            ],
+        );
+
+        $redirect = $this->fastly->createRedirect(
+            $request->path,
+            $request->target,
+        );
+
+        return redirect()->route('admin.redirects.show', $redirect->id);
+    }
+
+    /**
+     * Edit redirect details.
+     *
+     * @param Redirect $redirect
+     * @return \Illuminate\Http\Response
+     */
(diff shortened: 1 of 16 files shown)


app/Services/Fastly.php
```diff
@@ -0,0 +1,168 @@
...
+     * @param array $overrides
+     */
+    public function __construct($overrides = [])
+    {
+        $this->apiKeyHeader = 'Fastly-Key';
+        $this->apiKey = config('services.fastly.api_key');
+
+        // Resources that we'll be working with in Fastly:
+        $this->frontendServiceId = config('services.fastly.services.frontend');
+        $this->backendServiceId = config('services.fastly.services.backend');
```
DFurnes, line 27: Since this application will be interacting with both our front-end and back-end Fastly services (for redirects & clearing image cache, respectively), I've stored these both so we can use the appropriate service ID per method.

app/Services/Fastly.php
```diff
@@ -0,0 +1,168 @@
...
+
+    /**
+     * Get all redirects.
+     *
+     * @return RedirectCollection
+     */
+    public function getAllRedirects()
+    {
+        $redirects = $this->get(
+            "service/$this->frontendServiceId/dictionary/$this->redirectsTableId/items",
```
DFurnes, line 41: I'd originally used `.` string concatenation here, but it made this hard to read since Prettier separated out the components of the URL on separate lines. PHP's double-quoted string interpolation makes this easy to read!

config/services.php
```diff
@@ -72,7 +72,12 @@
...
         'url' => 'https://api.fastly.com/',
-        'key' => env('FASTLY_API_TOKEN'),
-        'service_id' => env('FASTLY_SERVICE_ID'),
+        'api_key' => env('FASTLY_API_KEY'),
+        'redirects_table' => env('FASTLY_TABLE_REDIRECTS'),
+        'frontend_url' => env('FASTLY_FRONTEND_SERVICE_URL'),
+        'services' => [
+            'frontend' => env('FASTLY_FRONTEND_SERVICE_ID'),
+            'backend' => env('FASTLY_BACKEND_SERVICE_ID'),
+        ],
```
DFurnes, line 81: I'll make sure to set these on each environment before merging & deploying this.

resources/views/admin/redirects/create.blade.php
```diff
@@ -0,0 +1,31 @@
...
+                <h1>Create Redirect</h1>
+                <p>Any visitors to the address in "incoming path" will be instantly redirected to the target URL.</p>
+
+                @include('forms.errors')
+
+                <form action="{{ route('redirects.store') }}" method="POST">
+                    <div class="form-item -padded">
+                        <label for="path" class="field-label">Incoming Path</label>
+                        <input class="text-field" placeholder="e.g. /us/campaigns/old-path" name="path" type="text" id="path">
+                    </div>
```
DFurnes, line 19: I swapped these from the `{!! Form::` helpers like [this](https://github.com/DoSomething/aurora/blob/2bf89f3a9360d358dd4494ba150323c527869487/resources/views/redirects/create.blade.php#L16-L17) (from [this package](https://laravelcollective.com/docs/5.2/html)) to plain HTML, as we've generally found this package's abstraction to be less helpful than using markup as-is & I didn't want to import another dependency that we don't consistently use elsewhere.

app/Http/Controllers/Web/Admin/RedirectsController.php
```diff
@@ -0,0 +1,141 @@
...
+
+        if (!$successful) {
+            return redirect()
+                ->route('admin.redirects.index')
+                ->with('flash', 'Could not delete redirect.');
+        }
+
+        return redirect()
+            ->route('admin.redirects.index')
+            ->with('flash', 'BAM! Deleted.');
```
mendelB, line 139: We have fun 🎸

mendelB (approved):
👏

app/Providers/AppServiceProvider.php
```diff
@@ -26,6 +26,8 @@ public function boot()
      */
     public function register()
     {
-        // ...
+        $this->app->bind(Fastly::class, function ($app) {
+            return new Fastly(config('services.fastly'));
```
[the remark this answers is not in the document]
DFurnes: Oh, you're totally right – we don't! Removed in 6d3910e.

config/services.php
DFurnes: Done!

PR #1081, opened 2020-12-22
R&D - Magic links

aaronschachter:
### What's this PR do?

This pull request adds the [Laravel Passwordless Login](https://github.com/grosv/laravel-passwordless-login) contrib package, and uses it in a new `POST api/v2/users/{user}/magic-links` route to return a passwordless login link for a given user.

The end goal for using this endpoint would be for SMS -- sending an SMS to a user to something like a Phoenix `/blocks/:id` page, where they'd already be logged in (and could report back via the block permalink in their mobile web browser, instead using SMS messages to ask/receive each attribute of a photo post)

...

I had a tough time getting this package to install, first running into memory limit errors:
```
PHP Fatal error:  Allowed memory size of 1610612736 bytes exhausted (tried to allocate 4096 bytes) in phar:///usr/local/bin/composer/src/Composer/DependencyResolver/RuleWatchGraph.php on line 52
```
I did get it to install by overriding the memory limit in-line:

```
php -d memory_limit=-1 `which composer` require grosv/laravel-passwordless-login
```

What's then bizarre is that authentication requests aren't working properly. A `POST /auth/token` request errors with:
```
{
    "error": {
        "code": 500,
        "message": "Too few arguments to function App\\Auth\\Entities\\AccessTokenEntity::convertToJWT(), 0 passed in /home/vagrant/Code/northstar/vendor/league/oauth2-server/src/Entities/Traits/AccessTokenTrait.php on line 79 and exactly 1 expected"
    },
    "debug": {
        "file": "/home/vagrant/Code/northstar/app/Auth/Entities/AccessTokenEntity.php",
        "line": 52
    }
}
```

I was able to hack together a working `POST /users/{user}/magic-links` endpoint by installing the `master` branch, making the API request to get a valid token, then checking out this branch to test generating the link.

A Northstar route will work as advertised, e.g. `users/{userId}/edit`

> http://northstar.test/magic-login/5f3dc977ea73310d6443dfe3?expires=1608668636&redirect_to=users%2F5f3dc977ea73310d6443dfe3%2Fedit&user_type=app-models-user&signature=9a21fef1d3b05da376596260714587dde56ceb9cc4e8fc37c9514fcd187cc963"

But when I tried to pass a phoenix.test URL to redirect to, the user be redirected to Phoenix URL.. but not logged in on Phoenix. Taking a peek at the `AuthController` in the Phoenix repo, and thinking that it's Phoenix that probably needs to receive a request to generate the link (perhaps by then posting to this endpoint)

### Relevant tickets

References [Pivotal #]().

### Checklist

### DIFF
--- a/app/Http/Controllers/UserController.php
+++ b/app/Http/Controllers/UserController.php
@@ -236,14 +236,20 @@ class UserController extends Controller
      * @return \Illuminate\Http\Response
      * @throws NotFoundHttpException
      */
-    public function createMagicLink(User $user)
+    public function createMagicLink(User $user, Request $request)
     {
         // Require admin access to create a magic link for now.
         $this->authorize('delete', $user);
 
+        $this->validate($request, [
+            'redirect' => 'required',
+        ]);
+
         $generator = new LoginUrl($user);
 
-        $generator->setRedirectUrl('/somewhere/else');
+        //$generator->setRedirectUrl($request['redirect']);
+        // Try redirecting to a path in this project
+        $generator->setRedirectUrl('users/' . $user->id . '/edit');
 
         $url = $generator->generate();
 
--- a/app/Http/Controllers/UserController.php
+++ b/app/Http/Controllers/UserController.php
@@ -7,6 +7,7 @@ use App\Auth\Role;
 use App\Exceptions\NorthstarValidationException;
 use App\Http\Transformers\UserTransformer;
 use App\Models\User;
+use Grosv\LaravelPasswordlessLogin\LoginUrl;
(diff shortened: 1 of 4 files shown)


aaronschachter:
This package might work pretty well if our destination URL was served from this app, but it's served from Phoenix -- where the React frontend isn't aware that a user's logged into Northstar because the user doesn't click on any of the authentication links that redirect to Northstar for auth and back.

This PR describes some of the trouble I've had in getting tests to run properly and composer dependencies -- working on that separately in #1085

PR #1082, opened 2021-01-05
Adds no-referrer policy to Reset Password pages

aaronschachter:
### What's this PR do?

This pull request adds a `referrer` meta tag with `no-referrer` to prevent leaking password reset tokens to third-party services (e.g. ad trackers)

### How should this be reviewed?

Visit a valid password reset page, and verify the meta tag appears.

### Any background context you want to provide?

Initially I tried adding a new middleware file to use over in [this branch](https://github.com/DoSomething/northstar/compare/referrer-policy), but couldn't access the response headers within the Blade file (seems that'd be more useful for an API response vs adding to HTML anyway)

### Relevant tickets

References [Pivotal #175644976](https://www.pivotaltracker.com/story/show/175644976).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [ ] Documentation added for new features/changed endpoints.
- [x] Added appropriate feature/unit tests.
- [ ] If new attributes were added to users, there is a corresponding PR to surface these in Aurora.
- [ ] If new attributes were added to users, then the data team already knows about these changes.

PR #1083, opened 2021-01-05
Adds middleware for Referrer Policy header

aaronschachter:
### What's this PR do?

This pull request adds a  `Referrer-Policy` header set to a `no-referrer` on all password reset pages, to prevent leaking password reset tokens to third-party services (e.g. ad trackers).

👀 

I still get failed tests per the thread in https://github.com/DoSomething/northstar/pull/1082, so I'm testing in the blind a little bit 🐭 😢 . Does seem like an issue specific to my machine... 

### Relevant tickets

References [Pivotal #175644976](https://www.pivotaltracker.com/story/show/175644976).

### Checklist

### DIFF
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -54,7 +54,7 @@ class Kernel extends HttpKernel
         'auth' => \App\Http\Middleware\Authenticate::class,
         'can' => \Illuminate\Auth\Middleware\Authorize::class,
         'guest' => \App\Http\Middleware\RedirectIfAuthenticated::class,
-        'referrer' => \App\Http\Middleware\ReferrerPolicy::class,
+        'no-referrer' => \App\Http\Middleware\ReferrerPolicy::class,
         'scope' => \App\Http\Middleware\RequireScope::class,
         'role' => \App\Http\Middleware\RequireRole::class,
         'throttle' => \App\Http\Middleware\ThrottleRequests::class,
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -54,6 +54,7 @@ class Kernel extends HttpKernel
         'auth' => \App\Http\Middleware\Authenticate::class,
         'can' => \Illuminate\Auth\Middleware\Authorize::class,
         'guest' => \App\Http\Middleware\RedirectIfAuthenticated::class,
+        'referrer' => \App\Http\Middleware\ReferrerPolicy::class,
         'scope' => \App\Http\Middleware\RequireScope::class,
         'role' => \App\Http\Middleware\RequireRole::class,
         'throttle' => \App\Http\Middleware\ThrottleRequests::class,
--- a/app/Http/Controllers/Web/ResetPasswordController.php
+++ b/app/Http/Controllers/Web/ResetPasswordController.php
@@ -34,7 +34,7 @@ class ResetPasswordController extends Controller
     public function __construct()
     {
         $this->middleware('guest');
-        $this->middleware('referrer');
+        $this->middleware('no-referrer');
     }
 
     /**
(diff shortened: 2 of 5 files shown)


app/Http/Kernel.php
```diff
@@ -54,6 +54,7 @@ class Kernel extends HttpKernel
         'auth' => \App\Http\Middleware\Authenticate::class,
         'can' => \Illuminate\Auth\Middleware\Authorize::class,
         'guest' => \App\Http\Middleware\RedirectIfAuthenticated::class,
+        'referrer' => \App\Http\Middleware\ReferrerPolicy::class,
```
DFurnes, line 57: What do you think about naming this middleware `no-referrer` to make it clearer at-a-glance what it does?

DFurnes (approved):

DFurnes:
> ### Any background context you want to provide?
> I still get failed tests per the thread in #1082, so I'm testing in the blind a little bit 🐭 😢 . Does seem like an issue specific to my machine...

Gah, that's so frustrating! 😞 I've run into that when running a `composer update`, but not for normal installs (so far). I thiiink the issue is something that happens in a patch/minor update to the [oauth2-server](https://github.com/thephpleague/oauth2-server) package. Hoping it's an easy fix, but haven't had time to dig in too much myself!

aaronschachter:
@DFurnes Testing this on QA (with a dummy URL like https://identity-qa.dosomething.org/password/reset/forgot-password/askldfja) and Chrome inspector still shows a referrer-policy of `strict-origin-when-cross-origin`

Seems maybe we need a Fastly cache clear to resolve?

Screen Shot 2021-01-06 at 11 42 31 AM

DFurnes:
Oh snap, that's because we write this header [here](https://github.com/DoSomething/infrastructure/blob/8a8c7c234ce6de80ecaa1d52d79e2e6bed4582b9/components/fastly_backends/main.tf#L30-L33) – which actually makes me wonder... was this already working as intended?

I think the `strict-origin-when-cross-origin` should, as far as I can tell, omit the URL's path in the referrer header for cross-origin requests... which I think is roughly the same as what `no-referrer` is doing for us (assuming that we trust ourselves with this referrer header, which we probably do). If so, we might be able to revert this & just ask them to re-test as is...? 🤔

DFurnes:
I'm not sure if the `no-referrer` header, perhaps, signals additionally to Google Analytics et. al. not to track that particular URL, though? Or if there's something else we can do to keep these URLs out of third-party tracking/logs...

aaronschachter:
Maybe the fix could be to add a `rel="noreferrer"` into the Customer.io email template we use for sending the password reset emails? (and reverting this PR)

DFurnes:
I think that would then mean we'd not get the email provider's URL in the referrer header they send to us (say, `mail.google.com/mail/u/0` if I'd clicked this link from my work email). The more I think about it, the more I think we're already in good shape with referrer headers with the `strict-origin-when-cross-origin` header rule...

What we could consider doing is removing analytics from this particular page, so that these reset URLs don't end up in Google Analytics & Snowplow events, by maybe adding an optional `$skip_analytics => true` variable that excludes [these two includes](https://github.com/DoSomething/northstar/blob/b5d80d258cad19c3daac30d96fc64c78d5cc3baf/resources/views/layouts/app.blade.php#L15-L16) if set... the downside is that we then wouldn't have analytics here.

Not sure how big a trade-off that is, but might be worthwhile for the security peace of mind...

aaronschachter:
I'm happy to take that approach, maybe would be worth chatting with product/data about the analytics implications? It seems like for now, I should revert this PR.

DFurnes:
Yeah, that sounds like a solid plan. 👍

PR #1086, opened 2021-01-09
Updating config files.

weerd:
### What's this PR do?

This pull request updates a number of config files to bring them up-to-date with Laravel 6.x.

Careful 👀 s through the config settings. Some env variable names have changed or were added, but mostly concerned with the edited ones to ensure nothing breaks!

As we move towards consolidating apps and eventually upgrading this app to newer versions of Laravel for the additional features, we want to minimize the mental overhead when performing an upgrade. 

When an upgrade is completed, typically only the most necessary files are updated, and some that include minor changes are left untouched. As future upgrades happen, some of the minor changes that were not updated might cause confusion as to whether a particular line of code is still used or needed, etc. So to minimize confusion and smoother upgrades, it is ideal to bring these framework files up to par with the latest installed version (v6.x in this case).

### Relevant tickets

References [Pivotal #176369232](https://www.pivotaltracker.com/story/show/176369232).

### Checklist


config/auth.php
```diff
@@ -104,9 +101,23 @@
     'passwords' => [
         'users' => [
             'provider' => 'users',
-            'email' => 'auth.emails.password',
```
weerd, line 107: I think this is a vestige from Laravel v5.2; I don't think it is used.

config/auth.php
```diff
@@ -104,9 +101,23 @@
...
+    | Password Confirmation Timeout
+    |--------------------------------------------------------------------------
+    |
+    | Here you may define the amount of seconds before a password confirmation
+    | times out and the user is prompted to re-enter their password via the
+    | confirmation screen. By default, the timeout lasts for three hours.
+    |
+    */
+
+    'password_timeout' => 10800,
```
weerd, line 121: @DFurnes I think you would know better, but I suspect this `password_timeout` and `throttle` setting on line `106` are fine to add... 🤔

config/queue.php
```diff
@@ -96,8 +93,8 @@
     */
 
     'failed' => [
-        'driver' => 'mongodb',
-        'database' => 'mongodb',
+        'driver' => env('QUEUE_FAILED_DRIVER', 'database'),
```
weerd, line 96: I wasn't sure if the old setting of `mongodb` was correct. I assume `database` is correct here and then the `DB_CONNECTION` sets it to be `mongodb` 🤔

DFurnes: Not a typo! We need this (at least for the time being) to allow us to read/write failed jobs [to a MongoDB database](https://github.com/jenssegers/laravel-mongodb#queues).

config/auth.php
DFurnes: I think both fine to add! The password timeout setting is for [this feature](https://blog.laravel.com/password-confirmation), which we don't use yet (but I'm very excited to in our monolith-ish future)! The `throttle` setting also seems totally reasonable to me & I think can replace [this middleware](https://github.com/DoSomething/northstar/blob/611a21f44c00a3a3b6c63365c314eef104dbea91/app/Http/Controllers/Web/ForgotPasswordController.php#L29) that we'd added ourselves.

weerd: Nice! Good to know 👯‍♂️

config/queue.php
weerd: ‼️ Ok, I'll change it back 😅

weerd:
I will be checking to see if the https://github.com/DoSomething/infrastructure repo will require any updates to ENV variable names before this PR is merged.

config/app.php
```diff
@@ -36,7 +39,7 @@
     |
     */
 
-    'debug' => env('APP_DEBUG', false),
+    'debug' => (bool) env('APP_DEBUG', false),
```
lindsayrmaher, line 42: what does the `(bool)` addition add here?

config/filesystems.php
```diff
@@ -27,7 +26,7 @@
     |
     */
 
-    'cloud' => 's3',
+    'cloud' => env('FILESYSTEM_CLOUD', 's3'),
```
lindsayrmaher, line 30: noticed `FILESYSTEM_CLOUD` isn't in the .env example file, is it fine to be missing? Is the 's3' the default?

config/queue.php
```diff
@@ -50,38 +51,34 @@
             'driver' => 'database',
             'table' => 'jobs',
             'queue' => 'default',
-            'expire' => 60,
```
lindsayrmaher, line 53: do we want to completely get rid of expire vs retry_after?

lindsayrmaher (approved):
All the changes seem reasonable to me, left a few questions around changes that seemed like updates vs additions.

config/queue.php
weerd: `expire` was the setting in Laravel up to 5.2 and changed to `retry_after` in 5.3 to now! Just bringing us up to date.

PR #1087, opened 2021-01-11
Add user administration interface to Northstar.

DFurnes:
### What's this PR do?

This pull request moves Aurora's user administration interface into Northstar.

### How should this be reviewed?

This is a somewhat hefty PR, but it's largely "copy and paste" from Aurora.

First, a couple little touch-ups:

📝 Fixes a typo on the config for the "frontend" URL (e.g. `www.dosomething.org`). bceedd9

🍃 The logo now directly links you to the "admin home", rather than the user-facing "profile home". f417db6

Then, I copied over some configuration this work relies on:

🔽 I added a `forms.select` partial, borrowed [from Rogue](https://github.com/DoSomething/rogue/blob/0947a1114354060958cfcada80b5d7fbc571007d/resources/views/forms/select.blade.php). 4a345b7

📑 Configured Laravel to use Forge paginators & updated styles to match what we use in Aurora. d8df057

🤝 Add helpers & custom styles from Aurora that these pages rely on. ab673ff

📛 Extracts a `CauseInterests` type from the profile controller, so we can reference it in our admin controller too. c0ff5e9

Finally, the main event:

🏗️ Adds the user administration interface. This is less of a literal copy-paste because we're now _directly_ making these changes in the database (rather than via a `Northstar` API client) and because we'd previously relied on the [Forms](https://laravelcollective.com/docs/6.x/html) package, which is a dependency that I didn't want to pull along here. 6adbb9e

### Any background context you want to provide?

This, alongside other work in the [Unified Backend](https://docs.google.com/document/d/1zF9LuMviDPEbjhlcnE-YPct8vCRQSHmw-cDJ3UKQMi0/edit#) epic, will allow us to retire Aurora's dynos - saving us money & allowing related changes (new user profile fields & their form entries) to be made at the same time!

I have **not** included the password reset & merge user functionality in this PR since it was already getting really heavy! I also haven't yet tackled how we'll deal with logging access to sensitive profile fields since this no longer routes via our API. I'll be sure to have that in place before we toggle this on in QA or production.

### Relevant tickets

References [Pivotal #175943296](https://www.pivotaltracker.com/story/show/175943296).

### Checklist

- [ ] This PR has been added to the relevant Pivotal card.
- [ ] Documentation added for new features/changed endpoints.
- [ ] Added appropriate feature/unit tests.
- [ ] If new attributes were added to users, there is a corresponding PR to surface these in Aurora.
- [ ] If new attributes were added to users, then the data team already knows about these changes.

PR #1088, opened 2021-01-13
Fix hack that uses malformed user auth to hijack redirect.

DFurnes:
### What's this PR do?

This pull request fixes a bug in the `is_dosomething_domain` helper function under PHP 7.14.14. This was causing a spooky issue where this test would fail on CircleCI (which runs 7.4.14), but not on our local Homestead (which is on 7.4.12).

For full context, see the following two comments: https://github.com/DoSomething/northstar/pull/1087#issuecomment-759568830 and https://github.com/DoSomething/northstar/pull/1087#issuecomment-759574337.

In short, there was a bug in `parse_url` where a malformed username in a URL in a domain could be used to override the parsed host (e.g. `http://dosomething.org\@cobalt.io/xyz` would be parsed as a `cobalt.io` URL instead of a `dosomething.org` URL with a funky path). Chrome also falls for this, which could allow someone to use one of these funky URLs to "trick" someone into following a redirect through our domain to a third-party.

Since this PHP bug was fixed, we'll just check for the icky `\@` manually to cover this scenario & get the test passing.

It turns out parsing URLs is a lot harder than you'd think!

### Relevant tickets

N/A. Blocked #1087.

### Checklist

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -250,9 +250,9 @@ function is_dosomething_domain(string $url): bool
         return false;
     }
 
-    // Reject a host that includes an escaped '.', which could be used to
-    // bypass our domain check below (e.g. 'cobalt.io\.dosomething.org'):
-    if (str_contains($host, '\.')) {
+    // Reject a host that includes an escaped '.' or '@', which could be used to
+    // fool our domain check below (e.g. 'cobalt.io\.dosomething.org'):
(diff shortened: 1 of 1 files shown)


weerd (approved):
👀  ‼️

mendelB (approved):
Whoah 🤯

weerd:
Solid sleuthing @DFurnes! 🕵️‍♂️

PR #1091, opened 2021-01-15
BrowserKit Purge Part Uno (it begins...)

weerd:
### What's this PR do?

This pull request starts on some cleanup for tests in Northstar to remove the use of the older `BrowserKitTestCase`.

👀 🚥 

The `BrowserKitTestCase` class exists to provide compatibility between Laravel 5.3's testing API and the newer API introduced in Laravel 5.4 and currently in use. It'll make our lives a lot easier if we can switch to using the newer Laravel Test API so we stay up-to-date on current approaches to testing, unify all our tests using the same API and it would allow us to remove the  `BrowserKitTesting` support package.

### Relevant tickets

References [Pivotal #174652230](https://www.pivotaltracker.com/story/show/174652230).

### Checklist

### DIFF
--- a/tests/Http/Web/FacebookTest.php
+++ b/tests/Http/Web/FacebookTest.php
@@ -264,7 +264,9 @@ class FacebookTest extends TestCase
 
         $this->mockSocialiteFromUserToken($abstractUser);
 
-        $this->get('/facebook/verify');
+        $response = $this->followingRedirects()->get('/facebook/verify');
+
+        $response->assertStatus(200);
 
         $user = auth()->user();
 
@@ -289,7 +291,9 @@ class FacebookTest extends TestCase
 
         $this->mockSocialiteFromUserToken($abstractUser);
 
-        $this->get('/facebook/verify');
+        $response = $this->followingRedirects()->get('/facebook/verify');
+
+        $response->assertStatus(200);
 
         $user = auth()->user();
 
--- a/tests/Http/Web/FacebookTest.php
+++ b/tests/Http/Web/FacebookTest.php
@@ -261,7 +261,6 @@ class FacebookTest extends TestCase
         );
 
         $this->mockSocialiteFromUser($abstractUser);
-
         $this->mockSocialiteFromUserToken($abstractUser);
 
         $response = $this->followingRedirects()->get('/facebook/verify');
@@ -288,7 +287,6 @@ class FacebookTest extends TestCase
         );
 
         $this->mockSocialiteFromUser($abstractUser);
-
         $this->mockSocialiteFromUserToken($abstractUser);
 
         $response = $this->followingRedirects()->get('/facebook/verify');
@@ -315,7 +313,6 @@ class FacebookTest extends TestCase
         );
 
         $this->mockSocialiteFromUser($abstractUser);
-
         $this->mockSocialiteFromUserToken($abstractUser);
 
         $response = $this->get('/facebook/verify');
--- a/tests/Http/Web/FacebookTest.php
+++ b/tests/Http/Web/FacebookTest.php
@@ -2,8 +2,9 @@
 
 use App\Models\User;
 use Laravel\Socialite\AbstractUser;
+use Laravel\Socialite\Facades\Socialite;
 
-class FacebookTest extends BrowserKitTestCase
+class FacebookTest extends TestCase
 {
     /**
      * Mock a Socialite user for the given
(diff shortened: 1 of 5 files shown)


tests/Http/Web/PasswordResetTest.php
```diff
@@ -5,8 +5,13 @@
 use App\Models\User;
 use Illuminate\Support\Facades\Bus;
 
-class PasswordResetTest extends BrowserKitTestCase
+class PasswordResetTest extends TestCase
 {
+    private function getUserTokenFromResetUrl($url)
```
weerd, line 10: Since BrowserKit used to do some fancy, pseudo GUI-esk work where it could "submit" a form and the new approaches with just using `post()` request methods don't have this, I needed to nab the user token from the `$resetPasswordUrl` variable after parsing it to properly submit the form without getting a validation error. Came in handy! ✨

tests/Http/Web/FacebookTest.php
```diff
@@ -94,11 +98,14 @@ private function defaultMock()
      */
     public function testFacebookRedirect()
     {
-        $this->expectException(\Laravel\BrowserKitTesting\HttpException::class);
```
weerd, line 97: Specific exception that BrowserKit would throw; no longer needed.

tests/Http/Web/FacebookTest.php
```diff
@@ -232,12 +259,17 @@ public function testFacebookPartialBirthday()
...
         );
+
         $this->mockSocialiteFromUser($abstractUser);
+
         $this->mockSocialiteFromUserToken($abstractUser);
 
-        $this->visit('/facebook/verify');
+        $response = $this->followingRedirects()->get('/facebook/verify');
+
+        $response->assertStatus(200);
```
weerd, line 269: The BrowserKit `view()` helper method would do a status check when called, so now we do it explicitly.

tests/Http/Web/GoogleTest.php
```diff
@@ -119,9 +125,14 @@ private function defaultMock()
      */
     public function testGoogleRedirect()
     {
-        // $this->visit('/google/continue');
-        // @TODO: Test below results in a 404: "A request to [https://accounts.google.com/o/oauth2/auth?client_id=....M9] failed. Received status code [404]."
-        // $this->assertRedirectedTo('https://accounts.google.com');
```
weerd, line 124: Not entirely sure why the above test was failing... Might be best to not have these tests depend on getting a response from google, but that can be a fix for a different day 🤷‍♂️ 

It may also have been failing since the redirect to assertion method was checking against a simpler URL without the slew of auth query params... I switched it up to just check against the domain to simplify it.

tests/Http/Web/PasswordResetTest.php
```diff
@@ -26,41 +31,62 @@ public function testForgotPasswordResetFlow()
         $user = factory(User::class)->create([
             'email' => 'forgetful@example.com',
         ]);
+
         $resetPasswordUrl = '';
 
+        // User can initiate a password reset.
+        $stepOneResponse = $this->get('/password/reset');
```
weerd, line 38: The tests in this class have a number of steps. Not sure if maybe we want to separate them out into smaller, testable methods? Anyhoo, to make it easier to follow and for my sanity, I labeled each response in the process by a `$step{Number}Response`.

mendelB: Makes sense!

tests/Http/Web/PasswordResetTest.php
```diff
@@ -26,41 +31,62 @@ public function testForgotPasswordResetFlow()
...
+        // Assert that the event was created & take note of reset URL for the next step.
         Bus::assertDispatched(SendPasswordResetToCustomerIo::class, function (
             $job
         ) use (&$resetPasswordUrl) {
             $resetPasswordUrl = $job->getUrl();
 
             return true;
         });
+
+        // @TODO: Not sure why we are logging this, and if so should we do it for the test below with a $resetPasswordUrl?
```
mendelB, line 62: I wonder if this was initially added just to debug while constructing these tests 🤔

mendelB (approved):
🧪

tests/Http/Web/PasswordResetTest.php
weerd: Oh potentially.... good call @mendelB. @DFurnes do you happen to know? I can remove if deemed unnecessary.

DFurnes: Huh. Yeah, I think it must've been leftover from debugging!

PR #1092, opened 2021-01-15
Introduce Rogue-style types for causes, resets, and subscription topics

katiecrane:
### What's this PR do?

This pull request introduces "Types" like the ones we use in Rogue for the often repeated causes, reset types, and email subscription topics.

First, I stole [the `Type` class from Rogue](https://github.com/DoSomething/rogue/blob/master/app/Types/Type.php) and added the `get` method to allow grabbing one item at a time which was useful for the password reset work.

Then, for each of the three categories, I created the Type class and then tried to use that each place where the full list of items or even just some of them were being referenced. Goal was to not have the same lists repeated in different places and to minimize the number of places a change needs to be made if we change any of the items in a Type.

I think we have pretty good tests for this, and they are all passing. Is there anything I could do to make this cleaner? Ideas for other places to use these Types?

All in the ticket!

### Relevant tickets

References [Pivotal #174421934](https://www.pivotaltracker.com/n/projects/2401401/stories/174421934).

### Checklist


katiecrane:
hmmm I have circleci failing on code style stuff - I guess my VS code doesn't have the right settings?

katiecrane:
oh right PHP isn't gonna fix itself is it? I still wish it would yell at me though!

katiecrane:
Now failing several files on `ordered_imports` some of them were actually wrong and some of them were already correct

katiecrane:
alright well it's still failing on `ordered_imports` but all those files have imports [in line length order like they're supposed to be](https://github.com/DoSomething/communal-docs/tree/master/Contributing#php-code)?

katiecrane:
Got import orders working! this is ready for review now @Dinnall! cc: @aaronschachter you started the type work in Northstar so thought you might want to take a look but no pressure!

app/Http/Controllers/CauseUpdateController.php
```diff
@@ -31,20 +32,7 @@ public function store(User $user, string $cause)
...
-                'gender_rights_equality',
-                'homelessness_poverty',
-                'immigration_refugees',
-                'lgbtq_rights_equality',
-                'mental_health',
-                'physical_health',
-                'racial_justice_equity',
-                'sexual_harassment_assault',
-            ])
+            !in_array($cause, CauseInterestType::all())
```
Dinnall, line 35: Love this change!

Dinnall (approved):
Love the changes there.
Good use of types and updating the `PasswordResetType`
💸  💯

katiecrane:
thanks @Dinnall! I fixed the merge conflict so merging now

mendelB:
This is awesome! 🙌

PR #1093, opened 2021-01-21
Browser Kit Purge Part Dos

weerd:
### What's this PR do?

This pull request starts on some cleanup for tests in Northstar to remove the use of the older `BrowserKitTestCase`.

👀 🚥 

Related to the work in #1091 that began The Purge™.

### Relevant tickets

References [Pivotal #174652230](https://www.pivotaltracker.com/story/show/174652230).

### Checklist

### DIFF
--- a/tests/Http/Web/WebUserTest.php
+++ b/tests/Http/Web/WebUserTest.php
@@ -2,7 +2,7 @@
 
 use App\Models\User;
 
-class WebUserTest extends BrowserKitTestCase
+class WebUserTest extends TestCase
 {
     /**
      * Default headers for this test case.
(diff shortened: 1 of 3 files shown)


tests/Http/Web/WebUserTest.php
```diff
@@ -22,32 +22,44 @@ public function testProfileShowRedirect()
...
-        $this->visit('users/' . $user->id)->seePageIs('/');
+        $response = $this->get('users/' . $user->id);
+
+        $response->assertStatus(302);
+        $response->assertRedirect('/');
     }
 
     /**
-     * Test that users can click to edit their profile from the homepage.
+     * Test that users are provided a button to edit their profile from the homepage.
```
weerd, line 32: By removing BrowserKit and using the newer approach to testing in Laravel, a couple of these tests seemed silly to test the full flow as it was before.

We no longer load a form and submit clicks to test the functionality, instead we hit the form endpoint directly, etc. So I changed the description a bit on some of these tests to better reflect what it is now testing and for more of the "end-to-end" tests we should switch them to Cypress.

tests/TestCase.php
```diff
@@ -94,12 +94,11 @@ public function registerUpdated()
         // Make sure we're logged out before trying to register.
         auth('web')->logout();
 
-        $this->visit('register');
-        $this->submitForm('register-submit', [
+        return $this->post('/register', [
```
weerd, line 97: Returning the response helps to later call assertions on it in the test methods!

Dinnall (approved):

weerd:
Rebased the changes from the monster 👹  PR in #1097 and re-ran all tests! Looks like everything still works 🕺

PR #1096, opened 2021-01-21
Add support for MySQL connection.

DFurnes:
### What's this PR do?

This pull request adds support for a second MySQL database connection alongside our existing MongoDB connection. 

**BLOCKER:** We need to update environment variables on dev, QA, and production to read from `MONGODB_`!

Paired with Diego! 🤝 

We chose to default to the (soon to be added) MySQL connection to make these upcoming PRs simpler. This meant we had to go back and specify the MongoDB "secondary" connection manually in existing code.

This will allow us to read and write to Rogue's MySQL database so that we can migrate these resources into this one codebase.

### Relevant tickets

### Checklist


app/Auth/PasswordBrokerManager.php
```diff
@@ -0,0 +1,23 @@
...
+
+class PasswordBrokerManager extends BasePasswordBrokerManager
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function createTokenRepository(array $config)
+    {
+        return new DatabaseTokenRepository(
+            $this->app['db']->connection('mongodb'), // HACK: We swap to our non-default connection here.
```
DFurnes, line 16: This one was really tricky! Even though Laravel [supports custom connections here](https://git.io/vr2RQ), the extended class [in the MongoDB package does not](https://github.com/jenssegers/laravel-mongodb/blob/a362df4f7fec5b91fea92b64273a531db34adfca/src/Jenssegers/Mongodb/Auth/PasswordBrokerManager.php#L15)! We had to manually add the custom connection name here.

app/Auth/Registrar.php
```diff
@@ -63,11 +63,11 @@ public function validate(
         $rules = [
             'first_name' => 'max:50',
             'email' =>
-                'email|nullable|unique:users,email,' .
+                'email|nullable|unique:mongodb.users,email,' .
                 $existingId .
                 ',_id|required_without:mobile',
             'mobile' =>
-                'mobile|nullable|unique:users,mobile,' .
+                'mobile|nullable|unique:mongodb.users,mobile,' .
```
DFurnes, line 70: Same deal! These validators all use the "default" connection unless you manually prepend the connection name.

app/Models/AuthCode.php
```diff
@@ -19,6 +19,13 @@
  */
 class AuthCode extends Model
 {
+    /**
+     * The database connection that should be used by the model.
+     *
+     * @var string
+     */
+    protected $connection = 'mongodb';
```
DFurnes, line 27: This tells Laravel [which database connection to use](https://laravel.com/docs/8.x/eloquent#database-connections)!

tests/BrowserKitTestCase.php
```diff
@@ -66,6 +68,18 @@ public function getMongoDocument($collection, $id)
...
+     * Assert that a given where condition exists in the MongoDB database.
+     *
+     * @param  string  $table
+     * @param  array  $data
+     * @return $this
+     */
+    protected function seeInMongoDatabase($table, array $data)
+    {
+        return $this->seeInDatabase($table, $data, 'mongodb');
+    }
```
DFurnes, line 81: We added this because we thought it read a little cleaner than passing that third parameter!

config/queue.php
```diff
@@ -93,8 +93,8 @@
     */
 
     'failed' => [
-        'driver' => env('QUEUE_FAILED_DRIVER', 'mongodb'),
-        'database' => env('DB_CONNECTION', 'mongodb'),
+        'driver' => 'mongodb',
+        'database' => 'mongodb',
```
weerd, line 97: Just noting that we decided to hardcode these to ensure we do not accidentally override the values on one of our environments.

weerd (approved):
Looks good to me! 🎉 

Just the one error that seems to be associated with a clean install of NS, when the `northstar:setup` command runs and `migrates` the database.

config/app.php
```diff
@@ -154,7 +154,8 @@
         Illuminate\Pipeline\PipelineServiceProvider::class,
         Illuminate\Queue\QueueServiceProvider::class,
         Illuminate\Redis\RedisServiceProvider::class,
-        Illuminate\Auth\Passwords\PasswordResetServiceProvider::class,
+        // TODO: Restore once we use MySQL table for password resets:
+        // Illuminate\Auth\Passwords\PasswordResetServiceProvider::class,
```
mendelB, line 158: Is this replaced by the custom class?

tests/TestCase.php
```diff
@@ -65,6 +67,18 @@ public function getMongoDocument($collection, $id)
...
+     *
+     * @param  string  $table
+     * @param  array  $data
+     * @return $this
+     */
+    protected function assertMongoDatabaseHas($table, array $data)
+    {
+        return $this->assertDatabaseHas($table, $data, 'mongodb');
+    }
+
```
mendelB, line 81: I don't see this used anywhere but is this for future work since we're migrating off of `BrowserKit`?

mendelB (approved):
👏 👏 👏

DFurnes:
Fixed up the setup migration issue in 87d30d3 (and https://github.com/DFurnes/environmentalist/commit/3b6114b1125b89c393f390e525fdeb2d898a405a)! I also fixed up an annoying bug where re-running the `northstar:keys` command would error out if keys already existed.

tests/TestCase.php
DFurnes: Correct!!

config/app.php
DFurnes: Yep, this is replaced by the additions we made to `AppServiceProvider` for our patched MongoDB support. We're thinking this is a good candidate to move back to MySQL (and therefore back to the "first party" Laravel service provider) in the near future.

.circleci/config.yml
```diff
@@ -9,6 +9,11 @@ jobs:
   build:
     docker:
       - image: circleci/php:7.4-node-browsers
+      - image: circleci/mariadb:10.3
```
weerd, line 12: 🎉

DFurnes:
Alright, we're good to go _finally!_ We had to add MariaDB support to the CircleCI box and make sure that we were running MongoDB migrations wherever we ran `php artisan migrate`. We also ran into another incompatibility with a bumped package that had changed a class interface.

app/Console/Commands/KeysCommand.php
```diff
@@ -32,9 +33,13 @@ public function handle(RSA $rsa, KeyRepository $repository)
...
-        $repository->writePublicKey($keys['publickey']);
-        $repository->writePrivateKey($keys['privatekey']);
+        try {
+            $repository->writePublicKey($keys['publickey']);
+            $repository->writePrivateKey($keys['privatekey']);
 
-        $this->info('Encryption keys generated successfully.');
+            $this->info('Encryption keys generated successfully.');
+        } catch (FileExistsException $exception) {
+            $this->info('Keys already exist.');
```
weerd, line 42: Good catch!

DFurnes:
We've set the appropriate `MONGODB_` and `MYSQL_` environment variables on each app so this should be ready to deploy!!

weerd (approved):
🎉

PR #1099, opened 2021-02-01
Updating create customer io event job usage

weerd:
### What's this PR do?

This pull request updates instances of `CreateCustomerIoEvent()`, that came from imported Rogue code, to switch the arguments passed to match what Northstar uses. Northstar passes in a `$user` object instead of just a user id string. 

👀 

My code editor also highlighted that the `SendPostToCustomerIo()` and `SendSignupToCustomerIo()` methods where missing their `dispatch()` calls, so I also updated those respective classes to import and use the same traits other Northstar jobs use.

In Rogue, these classes [extended a `Jobs`]() abstract class, which didn't do much other than provide the list of traits and set a property on the class. I preferred to follow Northstar's approach and not use this convenience class (we didn't import it from Rogue), since it's not too hard to just import the traits needed and makes each class less opaque. We _may_ want to add the property it added to each class though... 🤔 

```php
    /**
     * Delete jobs if their model(s) no longer exist. This prevents things that
     * have been deleted (either as part of automated testing or account deletions)
     * from cluttering up our failed jobs table with false negatives.
     *
     * @var bool
     */
    public $deleteWhenMissingModels = true;
```

### Relevant tickets

### Checklist

### DIFF
--- a/app/Jobs/SendPostToCustomerIo.php
+++ b/app/Jobs/SendPostToCustomerIo.php
@@ -5,9 +5,16 @@ namespace App\Jobs;
 use App\Jobs\Middleware\CustomerIoRateLimit;
 use App\Models\Post;
 use App\Services\CustomerIo;
+use Illuminate\Bus\Queueable;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Foundation\Bus\Dispatchable;
+use Illuminate\Queue\InteractsWithQueue;
+use Illuminate\Queue\SerializesModels;
 
-class SendPostToCustomerIo extends Job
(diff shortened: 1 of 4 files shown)


lindsayrmaher (approved):
makes sense, is the musing about adding in `deleteWhenMissingModels` something that would be necessary for this PR or just something we might want down the line?

weerd:
> makes sense, is the musing about adding in deleteWhenMissingModels something that would be necessary for this PR or just something we might want down the line?

It can be a later PR. Just food for thought for now... none of these imported Rogue classes run just yet!

DFurnes (approved):
Woo!

PR #1100, opened 2021-02-01
Read user's display name directly from database!

DFurnes:
### What's this PR do?

This pull request replaces the `getUserById` GraphQL query with directly reading from the related model through Eloquent! This takes advantage of [jenssegers/laravel-mongodb](https://github.com/jenssegers/laravel-mongodb)'s [cross-database relationship](https://github.com/jenssegers/laravel-mongodb#cross-database-relationships) support.

I also added some missing model imports in factory classes.

It may be easiest to review this commit-by-commit.

🏭 I added missing `App\Models` imports to factories in ac8137f.

👨‍👩‍👧‍👦 I updated our seeders to create real MongoDB users for activity (rather than [seeded IDs](https://github.com/DoSomething/northstar/blob/37a13e7969e4a46ba43bfa4a5182d0f2f2fdcfcf/database/faker/FakerNorthstarId.php)) in a41c794. This gives proper relationships between users & their activity on local.

🥞 I directly reads the user model from Eloquent, rather than GraphQL, in 624d610.

🏷️ Finally, I configured model relationships (e.g. `$signup->user`) in 1f289e2.

🥞

### Relevant tickets

### Checklist

### DIFF
--- a/app/Console/Commands/SetupCommand.php
+++ b/app/Console/Commands/SetupCommand.php
@@ -2,6 +2,7 @@
 
 namespace App\Console\Commands;
 
+use Defuse\Crypto\Key;
 use DFurnes\Environmentalist\ConfiguresApplication;
 use Illuminate\Console\Command;
 
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -3,6 +3,7 @@
 namespace App\Models;
 
 use App\Models\Traits\HasCursor;
+use App\Models\User;
 use App\Notifications\PostTagged;
 use App\Services\GraphQL;
 use Hashids\Hashids;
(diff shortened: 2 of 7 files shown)


app/Console/Commands/SetupCommand.php
```diff
@@ -2,6 +2,7 @@
 
 namespace App\Console\Commands;
 
+use Defuse\Crypto\Key;
```
weerd, line 5: Did this somehow get removed again? Or did we never put it back?

weerd (approved):
🙌  🎉

app/Console/Commands/SetupCommand.php
DFurnes: Whoah, that's spooky. I swear we added that the other day (and I definitely didn't intentionally add it today)...

weerd: 👻

PR #1102, opened 2021-02-02
Switch to using abstract Job class.

weerd:
### What's this PR do?

This pull request imports the Rogue version of the `App/Jobs/Job` abstract class and then subsequently updates all the Job classes to extend this abstract class.

👀 

In my initial [description](https://github.com/DoSomething/northstar/pull/1099#issue-565478340) for #1099, I thought it'd be best to not use the abstract class and define the traits used in each Job class. However, after further deliberation with @DFurnes, we decided it would be best to use it and also extend the prior Northstar classes with it.

It helps clean up each Job class and allows adding comments in the abstract class.

### Relevant tickets

### Checklist

### DIFF
--- a/app/Jobs/SendUserToCustomerIo.php
+++ b/app/Jobs/SendUserToCustomerIo.php
@@ -2,19 +2,13 @@
 
 namespace App\Jobs;
 
+use App\Jobs\Job;
 use App\Models\User;
 use App\Services\CustomerIo;
-use Illuminate\Bus\Queueable;
-use Illuminate\Contracts\Queue\ShouldQueue;
-use Illuminate\Foundation\Bus\Dispatchable;
-use Illuminate\Queue\InteractsWithQueue;
-use Illuminate\Queue\SerializesModels;
 use Illuminate\Support\Facades\Redis;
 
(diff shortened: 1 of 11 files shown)


app/Jobs/SendUserToCustomerIo.php
```diff
@@ -2,19 +2,13 @@
 
 namespace App\Jobs;
 
+use App\Jobs\Job;
```
DFurnes, line 5: We can omit these `use App\Jobs\Job` imports if we want since we're inside the `App\Jobs` namespace here - it'll automatically pull the correct `Job` based on that!

DFurnes (approved):
Looks great!

app/Jobs/SendUserToCustomerIo.php
weerd: Oh ya, I was debating that and figured that if they weren't need needed `composer format` would remove them, but it didn't so thought it might be useful to leave them since it might be nice to know exactly where it lives by looking at the imports 🤷  I can remove if this feels useless though! haha 😅

PR #1103, opened 2021-02-02
Some prep to start enabling test suites...

DFurnes:
### What's this PR do?

This pull request adds some fixes that I needed to make while working on getting `SignupTest` working.

I'd recommend reviewing commit-by-commit for context!

I'll drop some comments into the diff!

### Relevant tickets

### Checklist

### DIFF
--- a/app/Models/Action.php
+++ b/app/Models/Action.php
@@ -7,6 +7,13 @@ use Illuminate\Database\Eloquent\Model;
 class Action extends Model
 {
     /**
+     * The database connection that should be used by the model.
+     *
+     * @var string
+     */
+    protected $connection = 'mysql';
+
+    /**
      * The attributes that should be cast to native types.
      *
      * @var array
--- a/app/Policies/CampaignPolicy.php
+++ b/app/Policies/CampaignPolicy.php
@@ -1,0 +1,23 @@
+<?php
+
+namespace App\Policies;
+
+use App\Models\Campaign;
+use Illuminate\Auth\Access\HandlesAuthorization;
+
+class CampaignPolicy
+{
+    use HandlesAuthorization;
+
+    /**
+     * Determine if the given post can be updated by the user.
+     *
+     * @param  Illuminate\Contracts\Auth\Authenticatable $user
+     * @param  App\Models\Campaign $campaign
+     * @return bool
+     */
+    public function update($user, Campaign $campaign)
+    {
+        return is_staff_user();
+    }
+}
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -1,5 +1,7 @@
 <?php
 
+use App\Auth\Role;
+use App\Auth\Scope;
 use App\Auth\Entities\ClientEntity;
 use App\Auth\Normalizer;
 use App\Auth\Repositories\AccessTokenRepository;
@@ -7,6 +9,7 @@ use App\Auth\Repositories\KeyRepository;
 use App\Auth\Repositories\ScopeRepository;
 use App\Models\Client;
 use Carbon\Carbon;
+use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\HtmlString;
 use Illuminate\Support\Str;
 use libphonenumber\PhoneNumber;
@@ -539,17 +542,11 @@ function getAgeFromBirthdate($birthdate)
  */
 function getNorthstarId($request)
 {
-    if (
-        token()->role() === 'admin' ||
-        token()->role() === 'staff' ||
-        in_array('admin', token()->scopes())
-    ) {
-        return isset($request['northstar_id'])
-            ? $request['northstar_id']
-            : auth()->id();
+    if (is_admin_user() && !empty($request['northstar_id'])) {
+        return $request['northstar_id'];
     }
 
-    return auth()->id();
+    return Auth::id();
 }
 
 /**
@@ -561,7 +558,7 @@ function getNorthstarId($request)
 function is_admin_user(): bool
 {
     // If this is a machine client, then it's de-facto an admin:
-    if (token()->exists() && !token()->id()) {
+    if (Scope::allows('admin') && !Auth::id()) {
         return true;
     }
 
@@ -576,7 +573,7 @@ function is_admin_user(): bool
  */
 function is_staff_user(): bool
 {
-    return is_admin_user() || optional(auth()->user())->role === 'staff';
+    return is_admin_user() || optional(Auth::user())->role === 'staff';
 }
 
 /**
@@ -587,7 +584,7 @@ function is_staff_user(): bool
  */
 function is_owner($resource): bool
 {
-    return auth()->id() === $resource->northstar_id;
+    return Auth::id() === $resource->northstar_id;
 }
 
 /**
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -1,12 +1,12 @@
 <?php
 
-use App\Auth\Role;
-use App\Auth\Scope;
 use App\Auth\Entities\ClientEntity;
 use App\Auth\Normalizer;
 use App\Auth\Repositories\AccessTokenRepository;
 use App\Auth\Repositories\KeyRepository;
 use App\Auth\Repositories\ScopeRepository;
+use App\Auth\Role;
+use App\Auth\Scope;
 use App\Models\Client;
 use Carbon\Carbon;
 use Illuminate\Support\Facades\Auth;
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -542,7 +542,7 @@ function getAgeFromBirthdate($birthdate)
  */
 function getNorthstarId($request)
 {
-    if (is_admin_user() && !empty($request['northstar_id'])) {
+    if (is_staff_user() && !empty($request['northstar_id'])) {
         return $request['northstar_id'];
     }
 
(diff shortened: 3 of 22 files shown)


app/Models/Action.php
```diff
@@ -6,6 +6,13 @@
 
 class Action extends Model
 {
+    /**
+     * The database connection that should be used by the model.
+     *
+     * @var string
+     */
+    protected $connection = 'mysql';
```
DFurnes, line 14: It seems like we weren't [defaulting to the MySQL connection](https://github.com/DoSomething/northstar/blob/dcb5e941ea2a60ea120894ea1be8a333754a7f89/config/database.php#L17) like I would have expected. Without this, we get some messy exceptions like this whenever interacting with a SQL model:

```
15) SignupTest::testSignupsIndexAsAdminWithOrderBy
Error: Call to a member function prepare() on null

/home/vagrant/Sites/dosomething/vendor/laravel/framework/src/Illuminate/Database/Connection.php:457
/home/vagrant/Sites/dosomething/vendor/laravel/framework/src/Illuminate/Database/Connection.php:662
/home/vagrant/Sites/dosomething/vendor/laravel/framework/src/Illuminate/Database/Connection.php:629
/home/vagrant/Sites/dosomething/vendor/laravel/framework/src/Illuminate/Database/Connection.php:464
/home/vagrant/Sites/dosomething/vendor/laravel/framework/src/Illuminate/Database/Connection.php:416
/home/vagrant/Sites/dosomething/vendor/laravel/framework/src/Illuminate/Database/Query/Processors/Processor.php:32
/home/vagrant/Sites/dosomething/vendor/laravel/framework/src/Illuminate/Database/Query/Builder.php:2705
/home/vagrant/Sites/dosomething/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Builder.php:1385
/home/vagrant/Sites/dosomething/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php:840
/home/vagrant/Sites/dosomething/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php:805
/home/vagrant/Sites/dosomething/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php:668
/home/vagrant/Sites/dosomething/vendor/laravel/framework/src/Illuminate/Database/Eloquent/FactoryBuilder.php:219
/home/vagrant/Sites/dosomething/vendor/laravel/framework/src/Illuminate/Support/Traits/EnumeratesValues.php:176
/home/vagrant/Sites/dosomething/vendor/laravel/framework/src/Illuminate/Database/Eloquent/FactoryBuilder.php:220
/home/vagrant/Sites/dosomething/vendor/laravel/framework/src/Illuminate/Database/Eloquent/FactoryBuilder.php:181
/home/vagrant/Sites/dosomething/database/factories/PostFactory.php:30
/home/vagrant/Sites/dosomething/vendor/laravel/framework/src/Illuminate/Database/Eloquent/FactoryBuilder.php:373
/home/vagrant/Sites/dosomething/vendor/laravel/framework/src/Illuminate/Database/Eloquent/FactoryBuilder.php:291
/home/vagrant/Sites/dosomething/vendor/laravel/framework/src/Illuminate/Database/Eloquent/FactoryBuilder.php:305
/home/vagrant/Sites/dosomething/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Concerns/GuardsAttributes.php:129
/home/vagrant/Sites/dosomething/vendor/laravel/framework/src/Illuminate/Database/Eloquent/FactoryBuilder.php:313
/home/vagrant/Sites/dosomething/vendor/laravel/framework/src/Illuminate/Database/Eloquent/FactoryBuilder.php:242
/home/vagrant/Sites/dosomething/vendor/laravel/framework/src/Illuminate/Database/Eloquent/FactoryBuilder.php:243
/home/vagrant/Sites/dosomething/vendor/laravel/framework/src/Illuminate/Database/Eloquent/FactoryBuilder.php:178
/home/vagrant/Sites/dosomething/tests/Http/SignupTest.php:899
```

app/Policies/CampaignPolicy.php
```diff
@@ -0,0 +1,23 @@
...
+     *
+     * @param  Illuminate\Contracts\Auth\Authenticatable $user
+     * @param  App\Models\Campaign $campaign
+     * @return bool
+     */
+    public function update($user, Campaign $campaign)
+    {
+        return is_staff_user();
+    }
+}
```
DFurnes, line 23: Whoops, we just forgot to copy these [from Rogue](https://github.com/DoSomething/rogue/tree/4d43ca518ab629af0dd0297a772615f8a824350a/app/Policies)!

app/helpers.php
```diff
@@ -539,17 +542,11 @@ function getAgeFromBirthdate($birthdate)
  */
 function getNorthstarId($request)
 {
-    if (
-        token()->role() === 'admin' ||
-        token()->role() === 'staff' ||
```
DFurnes, line 544: I'd like to use Northstar's built-in ([league/oauth2-server](https://github.com/thephpleague/oauth2-server)-based) authentication, rather than trying to use both that & Gateway's [`token()` helper & middleware](https://github.com/DoSomething/gateway/blob/master/docs/server/ResourceServer.md) side by side. For the most part, we'll just be replacing `token()` calls with the equivalent `Role`, `Scope`, or `Auth` static methods.

app/Policies/CampaignPolicy.php
weerd: 😅

database/factories/SignupFactory.php
```diff
@@ -11,7 +11,9 @@
     $faker->addProvider(new FakerNorthstarId($faker));
 
     return [
-        'northstar_id' => factory(User::class)->create(),
+        'northstar_id' => function () {
+            return factory(User::class)->create()->id;
+        },
```
DFurnes, line 16: Laravel will only run the function for an attribute if it's not provided as an override, so by placing this `factory(User::class)` call into a function we avoid creating additional users when we don't have to (e.g. for signups where we've provide the `northstar_id` for an existing user in the overrides array).

tests/CreatesApplication.php
```diff
@@ -55,6 +55,12 @@ public function createApplication()
             'name' => 'DoSomething Staffers Club',
             'leaderId' => new \MongoDB\BSON\ObjectId(),
         ]);
+        $this->graphqlMock
+            ->shouldReceive('getCampaignWebsiteByCampaignId')
+            ->andReturn([
+                'title' => 'Test Example Campaign',
+                'slug' => 'test-example-campaign',
+            ]);
```
DFurnes, line 63: This mock is just copy-pasted from Rogue's [`WithMocks`](https://github.com/DoSomething/rogue/blob/4d43ca518ab629af0dd0297a772615f8a824350a/tests/WithMocks.php#L60-L65) trait!

weerd: 🍝

tests/WithAuthentication.php
```diff
@@ -156,8 +156,10 @@ public function asMachine()
      * @param array $scopes
      * @return $this
      */
-    public function asUser($user, $scopes = [])
-    {
+    public function asUser(
+        $user,
+        $scopes = ['user', 'activity', 'client', 'role:admin', 'write']
+    ) {
```
DFurnes, line 162: We have a lot of tests in Rogue where we create a user & then compare things we do as them to the user (e.g. if I create a signup as this person, does it have their `northstar_id` in the response?). That made the `asNormalUser()` helper less handy, and I didn't want to have to define scopes in every single test.

This defaults to including all reasonable scopes on a request & we can override this argument if we write tests where we only want a subset of these.

app/helpers.php
weerd: So does that mean not checking for `staff` user? Not sure if it's just not necessary.

DFurnes: Oooh, nope that's a bad refactor! 🙈 Yeah, this should use `is_staff_user`.

DFurnes: Fixed in 5bdc4e4b!

database/factories/SignupFactory.php
weerd: Would using `make()` help here? The `make()` method makes the object in memory but never saves it to the database.

DFurnes: My gut says we want to `create` since it acts most similar to production (and if we had something that relied on the user _really_ existing, like a `JOIN` or whatnot it'd work). But for the moment it probably would work...

weerd: ah fair! ok, sounds good

DFurnes: Although these get used for `db:seed`, and I think we'd be bummed to have signups that errored out when calling `$signup->user` because we hadn't saved the user to the database outside of the seeder...

DFurnes: I am curious to think about where `make` could give us a performance boost though! Maybe in test methods?

weerd: Ya, I was thinking while we're moving tests over we should audit where tests use `create()` and see if we actually need the item in the database as part of the test, or we're testing some other functionality and just need a user object. Would definitely speed things up if we find some instances that we can switch to `make()`!

weerd (approved):
Looooooks good to me! 👍  💯

mendelB (approved):
+1 👍

PR #1105, opened 2021-02-03
Enable signup routes in Northstar.

DFurnes:
### What's this PR do?

This pull request enables `v3/signup` routes in Northstar! We already copied over all of the business logic as-is from Rogue, so this pull request was just concerned with fixing inconsistencies & updating test syntax.

I split the work into atomic commits based on each change:

🐎 Rogue allowed controllers to default to cursor pagination, and we should do that here too! 563d2b1

🧾 Rogue uses Laravel's default response format, whereas we'd used a custom format in Northstar. I updated Northstar to use the "out of the box" formatter for `ValidationException` in 64144f2, and fixed an issue where we weren't actually re-casting Northstar validation errors to `NorthstarValidationException` in 15a9df1.

🎛️ Updated `SignupController` to use Northstar's middleware & auth classes (rather than Gateway) in e8db625.

🚥 Finally, I added routes & tests in c2750ff. I had to update the tests to use `asUser` instead of `withAccessToken` and `assertMysqlDatabaseHas` instead of `assertDatabaseHas`. I cleaned up some factories & assertions as well.

It's happening! 🤩 

### Relevant tickets

### Checklist

### DIFF
--- a/app/Http/Controllers/ActivityApiController.php
+++ b/app/Http/Controllers/ActivityApiController.php
@@ -18,19 +18,4 @@ abstract class ActivityApiController extends BaseController
         ValidatesRequests,
         FiltersActivityRequests,
         TransformsResponses;
-
-    /**
-     * Throw the failed validation exception with our custom formatting. Overrides the
-     * `throwValidationException` method from the `ValidatesRequests` trait.
-     *
-     * @param  \Illuminate\Http\Request $request
-     * @param  \Illuminate\Contracts\Validation\Validator $validator
-     * @throws NorthstarValidationException
-     */
-    protected function throwValidationException(Request $request, $validator)
-    {
-        throw new NorthstarValidationException(
-            $this->formatValidationErrors($validator),
-        );
-    }
 }
--- a/app/Http/Controllers/Controller.php
+++ b/app/Http/Controllers/Controller.php
@@ -10,27 +10,45 @@ use Illuminate\Foundation\Bus\DispatchesJobs;
 use Illuminate\Foundation\Validation\ValidatesRequests;
 use Illuminate\Http\Request;
 use Illuminate\Routing\Controller as BaseController;
+use Illuminate\Validation\ValidationException;
 
 abstract class Controller extends BaseController
 {
     use DispatchesJobs,
         AuthorizesRequests,
-        ValidatesRequests,
         FiltersRequests,
         TransformsResponses;
+    use ValidatesRequests {
+        validate as traitValidate;
+    }
 
     /**
-     * Throw the failed validation exception with our custom formatting. Overrides the
-     * `throwValidationException` method from the `ValidatesRequests` trait.
+     * Validate the given request with the given rules.
      *
-     * @param  \Illuminate\Http\Request $request
-     * @param  \Illuminate\Contracts\Validation\Validator $validator
-     * @throws NorthstarValidationException
+     * @param  \Illuminate\Http\Request  $request
+     * @param  array  $rules
+     * @param  array  $messages
+     * @param  array  $customAttributes
+     * @return array
+     *
+     * @throws \Illuminate\Validation\ValidationException
      */
-    protected function throwValidationException(Request $request, $validator)
-    {
-        throw new NorthstarValidationException(
-            $this->formatValidationErrors($validator),
-        );
+    public function validate(
+        Request $request,
+        array $rules,
+        array $messages = [],
+        array $customAttributes = []
+    ) {
+        try {
+            return $this->traitValidate(
+                $request,
+                $rules,
+                $messages,
+                $customAttributes,
+            );
+        } catch (ValidationException $exception) {
+            // This tells our handler to use Northstar's custom response formatting.
+            throw new NorthstarValidationException($exception->errors());
+        }
     }
 }
--- a/app/Http/Controllers/SignupsController.php
+++ b/app/Http/Controllers/SignupsController.php
@@ -2,6 +2,7 @@
 
 namespace App\Http\Controllers;
 
+use App\Auth\Role;
 use App\Http\Transformers\SignupTransformer;
 use App\Managers\SignupManager;
 use App\Models\Campaign;
@@ -9,6 +10,7 @@ use App\Models\Signup;
 use Illuminate\Auth\Access\AuthorizationException;
 use Illuminate\Http\Request;
 use Illuminate\Support\Arr;
+use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Str;
 
 class SignupsController extends ActivityApiController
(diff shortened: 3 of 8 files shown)


app/Http/Controllers/ActivityApiController.php
```diff
@@ -18,19 +18,4 @@ abstract class ActivityApiController extends BaseController
...
-     * @param  \Illuminate\Http\Request $request
-     * @param  \Illuminate\Contracts\Validation\Validator $validator
-     * @throws NorthstarValidationException
-     */
-    protected function throwValidationException(Request $request, $validator)
-    {
-        throw new NorthstarValidationException(
-            $this->formatValidationErrors($validator),
-        );
-    }
```
DFurnes, line 35: Removing this "override" so that we throw normal `ValidationException` for these controllers.

app/Http/Controllers/Controller.php
```diff
@@ -10,27 +10,45 @@
...
+            return $this->traitValidate(
+                $request,
+                $rules,
+                $messages,
+                $customAttributes,
+            );
+        } catch (ValidationException $exception) {
+            // This tells our handler to use Northstar's custom response formatting.
+            throw new NorthstarValidationException($exception->errors());
+        }
```
DFurnes, line 52: Except, well, it turns out we weren't really overriding this method because `throwValidationException` was [removed in Laravel 5.5](https://github.com/laravel/framework/blob/5.5/src/Illuminate/Foundation/Validation/ValidatesRequests.php) – oy! I've updated this syntax to override `validate` instead, catching validation exceptions & re-throwing them as our "custom" exception class (so the `Handler` knows to format differently).

app/Http/Controllers/SignupsController.php
```diff
@@ -42,12 +44,12 @@ public function __construct(SignupManager $signups)
...
         $this->transformer = new SignupTransformer();
 
-        $this->middleware('scopes:activity');
+        $this->middleware('scope:activity');
         $this->middleware('auth:api', [
             'only' => ['store', 'update', 'destroy'],
         ]);
         $this->middleware('role:admin,staff', ['only' => ['destroy']]);
-        $this->middleware('scopes:write', [
+        $this->middleware('scope:write', [
```
DFurnes, line 52: This middleware was called `scopes:` in Rogue, but is called `scope:` here.

routes/api.php
```diff
@@ -9,6 +11,19 @@
  * @see \App\Providers\RouteServiceProvider
  */
 
+Route::group(
+    // TODO: Do we want to use 'api/' prefix for v1 & v2 routes too?
+    ['prefix' => 'api/v3', 'middleware' => ['guard:api']],
```
DFurnes, line 16: We didn't prefix API routes under `api/` in Northstar, but it's probably worth swapping this to be consistent. We can maybe make this change when we're updating all our other applications to use the new monolith URL...

routes/api.php
```diff
@@ -9,6 +11,19 @@
  * @see \App\Providers\RouteServiceProvider
  */
 
+Route::group(
```
DFurnes, line 14: I wonder if we should consider keeping these behind a feature flag at first? They _should_ work just fine as soon as they're merged (and truthfully, nobody should be hitting these yet), but you never know...

tests/Http/SignupTest.php
```diff
@@ -0,0 +1,1066 @@
...
+        // Make sure the signup is persisted.
+        $this->assertMysqlDatabaseHas(
+            'signups',
+            [
+                'northstar_id' => $user->id,
+                'campaign_id' => $campaignId,
+                'quantity' => null,
+                'details' => 'affiliate-messaging',
+            ],
+            'mysql',
```
weerd, line 53: You don't need to pass this third argument since using assertMysqlDatabaseHas() has you covered with that right?

DFurnes: Yes! Fixed in 276908ab.

tests/Http/SignupTest.php
```diff
@@ -0,0 +1,1062 @@
...
+
+        $response->assertOk();
+        $response->assertJsonCount(0, 'data.0.posts.data');
+
+        // $decodedResponse = $response->decodeResponseJson();
+
+        // $this->assertEquals(
+        //     true,
+        //     empty($decodedResponse['data'][0]['posts']['data']),
+        // );
```
weerd, line 356: Old Rogue code?

tests/Http/SignupTest.php
```diff
@@ -0,0 +1,1062 @@
...
+
+        // Test with admin that entire user is returned.
+        $response = $this->asAdminUser()->getJson(
+            'api/v3/signups?include=user',
+        );
+
+        $response->assertOk();
+        $response->assertJsonPath(
+            'data.0.user.data.first_name',
+            $user->first_name,
```
weerd, line 720: Curious if it would be better to test against the `last_name` since the `first_name` is part of the public data returned if not an admin 🤔

tests/Http/SignupTest.php
```diff
@@ -0,0 +1,1062 @@
...
+                'northstar_id',
+                'campaign_id',
+                'quantity',
+                'created_at',
+                'updated_at',
+                'why_participated',
+                'source',
+                'details',
+            ],
+        ]);
```
weerd, line 895: Maybe for a future refactor since this PR is just for getting these tests working as they did in Rogue... but I wonder if checking just the structure for the signup here would mean it's a "specific signup". I think maybe checking that the `$signup->id` and/or some other signup properties match what's returned in the JSON would be more ideal 🤔

tests/Http/SignupTest.php
```diff
@@ -0,0 +1,1062 @@
...
+                'id',
+                'northstar_id',
+                'campaign_id',
+                'quantity',
+                'created_at',
+                'updated_at',
+                'why_participated',
+                'source',
+                'details',
+            ],
```
weerd, line 925: Same as note above... ideally want to also confirm the same created signup is returned. Granted it's the only one in the database but that shouldn't be an assumption in the test, haha 😅

weerd (approved):
Looks great overall!! 🎉 

I had a few questions on some items but nothing really crucial to block merging, and some are likely just comments related to a future refactor to improve some test functionality but that wasn't the scope of this PR, so fine if we address in later update!

routes/api.php
weerd: Also wondering if we should bump versions... so they're all under the same version number moving forward in this merged universe?

tests/Http/SignupTest.php
DFurnes: Oh whoops!

DFurnes: Removed in 4559f29a. 👾

tests/Http/SignupTest.php
DFurnes: Good call! Updated to test against `addr_zip` in 1b158ff2 (since `last_name` is one of the fields that we require to be specifically "opted-in" in the response, and so isn't supported within a nested include like this).

tests/Http/SignupTest.php
DFurnes: Yeah, that's a good idea. 👍

routes/api.php
DFurnes: Hmm, maybe! I wonder how much versioning will make sense if we're mostly interacting with these APIs via GraphQL (which handles versioning in it's own way). May be something we can just remove altogether...

PR #1106, opened 2021-02-04
Enable Campaign routes in Northstar

weerd:
### What's this PR do?

This pull request enables `v3/campaigns` API routes, and `/campaign` web routes in Northstar 🎉

All the business logic was already copied over from Rogue, so this pull request just worked on getting the tests for these routes working again! ✅ 

👀 

### Relevant tickets

### Checklist

### DIFF
--- a/tests/Http/CampaignTest.php
+++ b/tests/Http/CampaignTest.php
@@ -1,0 +1,751 @@
+<?php
+
+use App\Models\Campaign;
+use App\Models\GroupType;
+use App\Models\Post;
+use App\Types\Cause;
+use Illuminate\Support\Facades\Artisan;
+
+class CampaignTest extends TestCase
+{
+    //     /**
+//      * Test that a GET request to /api/v3/campaigns returns an index of all campaigns.
+//      *
+//      * GET /api/v3/campaigns
+//      * @return void
+//      */
+//     public function testCampaignIndex()
+//     {
+//         factory(Campaign::class, 5)->create();
+//         $response = $this->getJson('api/v3/campaigns');
+//         $decodedResponse = $response->decodeResponseJson();
+//         $response->assertStatus(200);
+//         $this->assertEquals(5, $decodedResponse['meta']['pagination']['count']);
+//     }
+//     /**
+//      * Test that we can filter open or closed campaigns.
+//      *
+//      * GET /api/v3/campaigns
+//      * @return void
+//      */
+//     public function testFilteredCampaignIndex()
+//     {
+//         factory(Campaign::class, 5)->create();
+//         factory(Campaign::class, 'closed', 3)->create();
+//         $response = $this->getJson('api/v3/campaigns?filter[is_open]=true');
+//         $decodedResponse = $response->decodeResponseJson();
+//         $this->assertEquals(5, $decodedResponse['meta']['pagination']['count']);
+//         $response = $this->getJson('api/v3/campaigns?filter[is_open]=false');
+//         $decodedResponse = $response->decodeResponseJson();
+//         $this->assertEquals(3, $decodedResponse['meta']['pagination']['count']);
+//     }
+//     /**
+//      * Test that we can filter campaigns with an associated Contentful 'Website' entry.
+//      *
+//      * GET /api/v3/campaigns
+//      * @return void
+//      */
+//     public function testWebsiteFilteredCampaignIndex()
+//     {
+//         factory(Campaign::class, 5)->create([
+//             'contentful_campaign_id' => '123',
+//         ]);
+//         factory(Campaign::class, 3)->create();
+//         $response = $this->getJson('api/v3/campaigns?filter[has_website]=true');
+//         $decodedResponse = $response->decodeResponseJson();
+//         $this->assertEquals(5, $decodedResponse['meta']['pagination']['count']);
+//         $response = $this->getJson(
+//             'api/v3/campaigns?filter[has_website]=false',
+//         );
+//         $decodedResponse = $response->decodeResponseJson();
+//         $this->assertEquals(3, $decodedResponse['meta']['pagination']['count']);
+//     }
+//     /**
+//      * Test that we can filter campaigns by cause.
+//      *
+//      * GET /api/v3/campaigns
+//      * @return void
+//      */
+//     public function testCauseFilteredCampaignIndex()
+//     {
+//         $causes = Cause::all();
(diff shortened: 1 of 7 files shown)


tests/Http/CampaignTest.php
```diff
@@ -0,0 +1,360 @@
...
+    public function testCampaignIndex()
+    {
+        $this->withoutExceptionHandling();
+
+        factory(Campaign::class, 5)->create();
+
+        $response = $this->getJson('api/v3/campaigns');
+
+        $response->assertOk();
+        $response->assertJsonPath('meta.pagination.count', 5);
```
weerd, line 42: In the old tests it would decode the JSON response and then use array notation to `assertEquals()` with the number it was comparing against.

I preferred the use of `assertJsonPath()` since it allows you to use the easier to read dot-notation and essentially compares the value obtained from the dot-notation and compares against the number you provide to ensure the values are equal. Overall it made the test slightly shorter and a little cleaner :)

tests/Http/CampaignTest.php
```diff
@@ -0,0 +1,360 @@
...
+
+    /**
+     * Test that a GET request to /api/v3/campaigns returns an index of all campaigns.
+     *
+     * GET /api/v3/campaigns
+     * @return void
+     */
+    public function testCampaignIndex()
+    {
+        $this->withoutExceptionHandling();
```
DFurnes, line 35: 👾

tests/Http/CampaignTest.php
```diff
@@ -0,0 +1,360 @@
...
+     * @return void
+     */
+    public function testCauseFilteredCampaignIndex()
+    {
+        $causes = Cause::all();
+
+        // Let's test against pairs of three causes each so that we have a first, last, and middle cause
+        // (ensuring we're testing our filtering logic against surrounding commas).
+        $campaignWithFirstThreeCauses = factory(Campaign::class, 1)->create([
+            'cause' => array_slice($causes, 0, 3),
```
DFurnes, line 110: This makes me think it'd be nice to make a `$faker->cause(n)` generator that pulls a random `n` causes from the `Cause::all()` array... could be handy to have easily accessible in tests/seeders!

tests/Http/CampaignTest.php
```diff
@@ -0,0 +1,360 @@
...
+     * GET /api/v3/campaigns
+     * @return void
+     */
+    public function testCauseFilteredCampaignIndex()
+    {
+        $causes = Cause::all();
+
+        // Let's test against pairs of three causes each so that we have a first, last, and middle cause
+        // (ensuring we're testing our filtering logic against surrounding commas).
+        $campaignWithFirstThreeCauses = factory(Campaign::class, 1)->create([
```
DFurnes, line 109: If you remove the `, 1` here, it should return a single `Campaign` instance rather than a single-element collection (so we then don't have to call `->first()` all the time below).

tests/Http/CampaignTest.php
```diff
@@ -0,0 +1,360 @@
...
+
+        $responseOne = $this->asAdminUser()->getJson($endpoint);
+
+        $responseOne->assertOk();
+        $responseOne->assertJsonCount(3, 'data');
+
+        // Then, we'll use the last post's cursor to fetch the remaining two:
+        $json = $responseOne->json();
+
+        $lastCursor = $json['data'][2]['cursor'];
```
DFurnes, line 174: When refactoring the signup test suite, I noticed you can optionally pass a `$path` to `$response->json()` to directly reach in and grab a specific value. That could perhaps make this a little more concise!

```php
$lastCursor = $response->json('data.2.cursor');
$responseTwo = $this->asAdminUser()->getJson(
    $endpoint . '&cursor[after]=' . $lastCursor,
);
```

tests/Http/Web/WebCampaignTest.php
```diff
@@ -0,0 +1,118 @@
...
+
+        // Delete the campaign.
+        $this->actingAs($admin, 'web')->deleteJson(
+            'campaigns/' . $campaign->id,
+        );
+
+        // Make sure the campaign is deleted.
+        $response = $this->getJson('api/v3/campaigns/' . $campaign->id);
+
+        $response->dump();
```
DFurnes, line 110: 👾

tests/Http/Web/WebCampaignTest.php
```diff
@@ -0,0 +1,118 @@
...
+
+        $response->assertNotFound();
+        $response->assertJsonPath(
+            'error.message',
+            'That resource could not be found.',
+        );
```
DFurnes, line 116: I've been removing these assertions on specific error language as I go through tests, since I think it's covering something that machines shouldn't really be basing logic upon. (Fine to keep too, but wanted to share what I've been thinking about these!)

DFurnes (approved):
Looking great!

tests/Http/CampaignTest.php
weerd: gah! Good catch!

tests/Http/CampaignTest.php
weerd: Oh ya, I dig that and sound very helpful for future tests as well!

tests/Http/CampaignTest.php
weerd: Oh fair. I had left it how it was, but this would be a nice, quick improvement 👍

tests/Http/CampaignTest.php
weerd: Ah nice! I was wondering about that too. I wasn't a fan of the array notation, will update!

tests/Http/Web/WebCampaignTest.php
weerd: 😓

tests/Http/Web/WebCampaignTest.php
weerd: Ya I was wondering about this. I came across this as a thing when switching NS tests to remove BrowserKit. Another issue I found with this is that it only matches if the text is exact! So if you miss the period at the end, it'll fail, which seems a little brittle. Have you been switching to just confirming there's an `error` object?

DFurnes: Just the status code via `$response->assertNotFound()` (or whichever error type)!

weerd: Ah ok! It already checks for `assertNotFound()` so I'll just delete the `assertJsonPath()` assertion 👍

PR #1107, opened 2021-02-05
Send password updated emails via Customer.io App API

aaronschachter:
### What's this PR do?

This pull request refactors sending password updated confirmation emails via the new Customer.io App API instead of the existing Track API. This way, we won't need to maintain a Customer.io profile for a member to send the password updated confirmation email. The documentation updates detail this setup.

This required a bit of setup on the Customer.io side that I did for both staging and production Customer.io instances:

* [x] Create an App API key
* [x] Create transactional message IDs for the transactional emails to send:
    * One for password updated emails (used in this PR)
    * Another for forgot password emails (for a future PR)
* [x] Create a placeholder Customer.io profile to use for sending transactional emails to
 

I've been manually testing by enabling Customer.io locally and changing my password via profile to verify the transactional email is sent:

* This PR deprecates the `SendPasswordUpdatedToCustomerIo` job for the sake of using the re-usable `CreateCustomerIoEvent` - which was introduced later in the codebase, for simplicity.

* Holding off on adding tests, for now, to help keep this reviewable. The next PR will do some refactoring on how we generate password reset links, making it easier to write some integration tests to visit a password reset link and verify that either the Customer.io event was created or a Customer.io email was sent. 

### Relevant tickets

References [Pivotal #176873288](https://www.pivotaltracker.com/story/show/176873288).

### Checklist


app/Services/CustomerIo.php
```diff
@@ -8,22 +8,41 @@
 class CustomerIo
 {
     /**
-     * The Customer.io client.
+     * The Customer.io App API client.
```
mendelB, line 11: Would it be easier if these two were separate services or _more_ complicated?

app/Listeners/ReportPasswordUpdated.php
```diff
@@ -15,9 +17,31 @@ class ReportPasswordUpdated
...
+                 */
+                'password_updated',
+                [
+                    'updated_via' => $event->updatedVia,
+                ]
+            );
+        }
+
+        /*
+         * Send transactional emails for to send confirmation for password updates that don't need
```
mendelB, line 39: I think there's a typo here 👾

documentation/customer-io.md
```diff
@@ -0,0 +1,13 @@
...
+
+## Track API
+
+We use the [Track API](https://customer.io/docs/api/#tag/trackOverview) to track member activity in Customer.io, and send both promotional and transactional emails to subscribers. The documentation [here](http://docs.dosomething.org/customer-io) and [here](http://docs.dosomething.org/non-traditional-member-activation) provide more detail, and will soon be moved into this repo.
+
+## App API
+
+We use the [App API](https://customer.io/docs/api/#tag/appOverview) to send transactional emails for forgot password requests, and password updated events.
+
+We use the [send email](https://customer.io/docs/api/#operation/sendEmail) endpoint to send emails to members without requiring a Customer.io profile to exist. This is done by executing all send email requests for a specific ID of a placeholder Customer.io profile we've set up, set via the `CUSTOMER_IO_APP_IDENTIFIER_ID` config variable.
```
mendelB, line 13: 👏 📝

mendelB, line 13: Curious -- If a profile _does_ exist, does the transactional email get logged to their profile automatically?

mendelB (approved):
💥

app/Services/CustomerIo.php
aaronschachter: I think I prefer them in one service class, so when we're writing Northstar code - we don't need to worry about which function belongs to which class.

documentation/customer-io.md
aaronschachter: Nope, they will all go to the profile ID we set here.

PR #1113, opened 2021-02-10
Send forgot password emails via Customer.io App API

aaronschachter:
### What's this PR do?

This pull request continues work done in #1107 by refactoring forgot-password emails to be sent from the Customer.io App (Transactional API) instead of the Track API. This way, we won't require a Customer.io profile to exist in order to send a reset password emails to members who request one.

It also deprecates the `SendPasswordResetToCustomerIo` job in favor of the more general `CreateCustomerIoEvent` job, which was added to the codebase later.

👀 

We should eventually add the Send Reset Email form back into the admin Danger Zone section of a user's profile (this wasn't ported over in #1087 and makes it easier to manually test this functionality). Might save that for a future PR or add it into this one later (if it hasn't been reviewed yet).

### Relevant tickets

References [Pivotal #176873288](https://www.pivotaltracker.com/story/show/176873288).

### Checklist


app/Models/User.php
```diff
@@ -906,18 +919,26 @@ public function sendPasswordResetNotification($token)
...
     {
-        if (!$token) {
-            $tokenRepository = new DatabaseTokenRepository(
-                app('db')->connection('mongodb'),
-                app('hash'),
-                config('auth.passwords.users.table'),
-                config('app.key'),
-                config('auth.passwords.users.expire'),
-            );
-            $token = $tokenRepository->create($this);
```
DFurnes, line 917: Why was this moved into `getPasswordResetUrl`?

DFurnes (approved):
Looks good! 📧

app/Models/User.php
aaronschachter: Ah I suppose it doesn't need to be -- I meant to write a test to generate a URL, visit it, and then test which events were fired after updating the password, but realized later that the Web/PasswordResetTest generally does this -- but doesn't check for the events yet. I don't mind moving it back in

DFurnes: Ah gotcha! Yeah, I think I'd prefer to keep it here so we're not passing a `null` around for too long before defaulting it.

PR #1115, opened 2021-02-10
Enable Group Type routes in Northstar

weerd:
### What's this PR do?

This pull request enables `v3/group-types` API routes and `/group-types` Web routes in Northstar 🎉

All the business logic was already copied over from Rogue, so this pull request just worked on getting the tests for these routes working again! ✅ 

👀 

### Relevant tickets

References [Pivotal #176851325](https://www.pivotaltracker.com/story/show/176851325).

### Checklist

### DIFF
--- a/tests/Http/Web/WebGroupTest.php
+++ b/tests/Http/Web/WebGroupTest.php
@@ -9,7 +9,7 @@ class WebGroupTest extends TestCase
     /** @test */
     public function testAdminCanCreateGroup()
     {
-        $admin = factory(User::class, 'admin')->make();
+        $admin = factory(User::class, 'admin')->create();
 
         $groupType = factory(GroupType::class)->create();
 
@@ -31,7 +31,7 @@ class WebGroupTest extends TestCase
     /** @test */
     public function testAdminCannotCreateDuplicateGroup()
     {
-        $admin = factory(User::class, 'admin')->make();
+        $admin = factory(User::class, 'admin')->create();
 
         $group = factory(Group::class)->create();
 
@@ -49,7 +49,7 @@ class WebGroupTest extends TestCase
     /** @test */
     public function testStaffCanCreateGroup()
     {
-        $staff = factory(User::class, 'staff')->make();
+        $staff = factory(User::class, 'staff')->create();
 
         $groupType = factory(GroupType::class)->create();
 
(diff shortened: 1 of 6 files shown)


tests/Http/Web/WebGroupTest.php
```diff
@@ -9,7 +9,7 @@ class WebGroupTest extends TestCase
     /** @test */
     public function testAdminCanCreateGroup()
     {
-        $admin = factory(User::class, 'admin')->create();
+        $admin = factory(User::class, 'admin')->make();
```
weerd, line 12: @DFurnes just wanted to note, for these web tests where we need a User but it is not really necessary for them to be persisted in the DB, I've switched them from `create()` to `make()` so it's just in memory. It's a very slight bump in speed but every little bit counts!

tests/Http/Web/WebGroupTypeTest.php
```diff
@@ -0,0 +1,85 @@
...
+
+        $response->assertRedirect();
+
+        $this->assertDatabaseMissing(
+            'group_types',
+            [
+                'name' => $name,
+            ],
+            'mysql',
+        );
```
weerd, line 61: Another thing I've been doing is starting all http request action path names with the `/`. It feel it provide a nice quick visual clue that this is part of a request and it's hitting some endpoint, versus (as on line `56`) a string defining a database table name (which likely has a similar name as the endpoint).

Feels like a plus for readability!

cc: @DFurnes

DFurnes: I like this!

tests/Http/Web/WebGroupTest.php
DFurnes: I think I'd lean towards using `create()` since it's closer to how things work in production. I worry that `make()` might come back to bite us if we had application behavior that relied on the database like, for example, if we were to add an  `exists:users` validator to post requests.

weerd: Oh hmm... 🤔  That's a very good point. Ok! I'll switch it back!

I still think there are some areas where we can make use of `make()` 😏  but probs leave to only obviously safe resources! Glad I noted here for discussion 😅

weerd:
Ran into an odd error in CI:

```
General error: 1205 Lock wait timeout exceeded; try restarting transaction
```
Re-ran tests locally and they all passed so re-running the CI and seeing if just a fluke. But might be something to note 🤔

DFurnes:
What a spooky error! Glad it seems to have resolved itself with a re-run. 👻

DFurnes (approved):
Looks great!

PR #1118, opened 2021-02-11
Adds tests for transactional emails

aaronschachter:
### What's this PR do?

This pull request refactors the `SendCustomerIoEmail` jobs into two separate jobs, `SendForgotPasswordEmail` and `SendPasswordUpdatedEmail` to make it easier to test what types of emails are sent within the standard forgot password flow.

It also:
* Adds a `CustomerIoJob` class to DRY adding the middleware (a whole lot of other jobs could use this, but holding off on too many changes to make this easier to review)
* Refactors the `sendEmail` function in the Customer.io service to accept a User model instead of an email address.

👀 

I'm changing my mind about the `CreateCustomerIoEvent` class I added into this repo and Rogue as well -- I think it's cleaner to create separate jobs for each different event we want to track, as we do for post created, post reviewed, etc. Holding off on cleaning this up for a future PR.

### Relevant tickets

References [Pivotal #176873288](https://www.pivotaltracker.com/n/projects/2328687/stories/176873288).

### Checklist


mendelB (approved):
👍

tests/Http/Web/PasswordResetTest.php
```diff
@@ -184,6 +203,18 @@ public function testActivateAccountResetFlow()
...
+            $job
+        ) use ($user) {
+            $params = $job->getParams();
+
+            $this->assertEquals($params['user']->id, $user->id);
+            $this->assertEquals($params['eventData']['updated_via'], 'rock-the-vote-activate-account');
+
+            return true;
+        });
+
```
weerd, line 217: Not sure if this applies since I'm less familiar with the CustomerIO stuff, but @DFurnes added a nice helper for checking against CustomerIO events firing:

https://github.com/DoSomething/northstar/blob/c57493f22f161f1488405f4c5a91eab918558d14/tests/TestCase.php#L96-L118

config/services.php
```diff
@@ -24,8 +24,8 @@
             'url' => 'https://api.customer.io/v1/',
             'identifier_id' => env('CUSTOMER_IO_APP_IDENTIFIER_ID', 'app-api-user'),
             'transactional_message_ids' => [
-                'forgot_password' => env('CUSTOMER_IO_FORGOT_PASSWORD_TRANSACTIONAL_MESSAGE_ID', 2),
-                'password_updated' => env('CUSTOMER_IO_PASSWORD_UPDATED_TRANSACTIONAL_MESSAGE_ID', 3),
+                'FORGOT_PASSWORD' => env('CUSTOMER_IO_FORGOT_PASSWORD_TRANSACTIONAL_MESSAGE_ID', 2),
+                'PASSWORD_UPDATED' => env('CUSTOMER_IO_PASSWORD_UPDATED_TRANSACTIONAL_MESSAGE_ID', 3),
```
weerd, line 28: Not sure if these should be capitalized since it's not the standard and they aren't technically constants.

aaronschachter: Initially I had these as snake_case, but liked that the `getTransactionalMessageId` function would accept an all caps string to look similar to the `get` calls in the `Type` class. This is easy enough to swap out, so don't have a preference any way... just thought it'd keep the function parameters looking similar.

tests/Http/Web/PasswordResetTest.php
aaronschachter: Ooh, sweet! This will work for one test I've added here, but for some tests like this one you've commented -- we're testing individual properties saved to the job (e.g. the `updated_via` property sent along with the Customer.io event)

PR #1120, opened 2021-02-11
Add rotation, image, and original routes.

DFurnes:
### What's this PR do?

This pull request adds the `posts/{post}/rotate`, `images/{hash}` and `originals/{id}` routes.

I made one fix to how we generate image paths in 4464b5d, since we use the `local` driver in this application (not `public`, as we used in Rogue). Since we don't actually expose this images via the public symlink, this better matches real usage.

Then, each route & associated tests are added in an individual commit: 21ce15a, 001b34b, 36e0ab3

Finally, I tweaked `ReviewsController` to be [resourceful](https://laravel.com/docs/8.x/controllers#actions-handled-by-resource-controller) in 2b7c832 (rather than using a non-standard method name).

We didn't have any tests for `originals/{id}` or `posts/{post}/rotate`, so I whipped up some simple coverage.

### Relevant tickets

References [Pivotal #176851335](https://www.pivotaltracker.com/story/show/176851335).

### Checklist

### DIFF
--- a/app/Http/Controllers/RotationController.php
+++ b/app/Http/Controllers/RotationController.php
@@ -7,7 +7,6 @@ use App\Models\Post;
 use App\Services\Fastly;
 use App\Services\ImageStorage;
 use Illuminate\Http\Request;
-use Storage;
 
 class RotationController extends Controller
 {
@@ -35,17 +34,14 @@ class RotationController extends Controller
      *
      * @param Filesystem $filesystem
      */
-    public function __construct(
-        Fastly $fastly,
-        ImageStorage $storage,
-        PostTransformer $transformer
-    ) {
+    public function __construct(Fastly $fastly, ImageStorage $storage)
+    {
         $this->fastly = $fastly;
         $this->storage = $storage;
-        $this->transformer = $transformer;
+        $this->transformer = new PostTransformer();
 
-        $this->middleware('auth', ['only' => 'update']);
-        $this->middleware('role:staff,admin', ['only' => 'update']);
+        $this->middleware('auth:api');
+        $this->middleware('role:staff,admin');
     }
 
     /**
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -336,7 +336,7 @@ class Post extends Model
     {
         $path = parse_url($this->url, PHP_URL_PATH);
 
-        // We store local URLs prefixed with `/storage` when using the "public" adapter.
+        // We store local URLs prefixed with `/storage` when using the "local" adapter.
         if (config('filesystems.default') == 'local') {
             return str_replace('/storage/', '', $path);
         }
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -337,7 +337,7 @@ class Post extends Model
         $path = parse_url($this->url, PHP_URL_PATH);
 
         // We store local URLs prefixed with `/storage` when using the "public" adapter.
-        if (config('filesystems.default') == 'public') {
+        if (config('filesystems.default') == 'local') {
             return str_replace('/storage/', '', $path);
         }
 
--- a/tests/Http/RotationTest.php
+++ b/tests/Http/RotationTest.php
@@ -1,0 +1,67 @@
+<?php
+
+use App\Models\Post;
+use App\Models\User;
+
+class RotationTest extends TestCase
+{
+    /**
+     * Test that an admin can view originals.
+     *
+     * @return void
+     */
+    public function testRotateImage()
+    {
+        $post = factory(Post::class)
+            ->states('photo')
+            ->create();
+
+        $response = $this->asAdminUser()->postJson(
+            '/api/v3/posts/' . $post->id . '/rotate',
+            ['degrees' => 90],
+        );
+
+        $response->assertSuccessful();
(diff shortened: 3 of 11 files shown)


app/Http/Controllers/RotationController.php
```diff
@@ -35,17 +34,14 @@ class RotationController extends Controller
...
-        Fastly $fastly,
-        ImageStorage $storage,
-        PostTransformer $transformer
-    ) {
+    public function __construct(Fastly $fastly, ImageStorage $storage)
+    {
         $this->fastly = $fastly;
         $this->storage = $storage;
-        $this->transformer = $transformer;
+        $this->transformer = new PostTransformer();
```
DFurnes, line 41: I'll be honest – I swapped this to be instantiated inline because I didn't want Prettier to make the constructor's arguments span multiple lines. But this also follows the same pattern we've established elsewhere & since we're not doing any fancy dependency injection with transformers, works perfectly fine. 😊

tests/Http/RotationTest.php
```diff
@@ -0,0 +1,67 @@
...
+            ->create();
+
+        $response = $this->asAdminUser()->postJson(
+            '/api/v3/posts/' . $post->id . '/rotate',
+            ['degrees' => 90],
+        );
+
+        $response->assertSuccessful();
+
+        $this->fastlyMock->shouldHaveReceived('purge');
```
DFurnes, line 26: Ideally, I'd want to assert that we call [`purgeKey`](https://github.com/DoSomething/northstar/blob/c57493f22f161f1488405f4c5a91eab918558d14/app/Services/Fastly.php#L133-L149) with `'post-' . $post->id`, since that's most accurately asserting against the expected change that Fastly would see. But that's somewhat tricky right now, since we call that via `purge` and therefore would need this to be a partial mock.

Perhaps later we can extract `purge` into either the `ImageStorage` service class or `Post` model...

app/Models/Post.php
```diff
@@ -337,7 +337,7 @@ public function getMediaPath()
         $path = parse_url($this->url, PHP_URL_PATH);
 
         // We store local URLs prefixed with `/storage` when using the "public" adapter.
```
weerd, line 339: Should the comment be updated since not using "public"?

DFurnes: Whoops, yes it should! 🙈

DFurnes: Updated in 031385548e60beb92be3182a283272b7398dabe6.

tests/Http/Web/OriginalsTest.php
```diff
@@ -0,0 +1,45 @@
...
+    {
+        $admin = factory(User::class, 'admin')->create();
+
+        $post = factory(Post::class)
+            ->states('photo', 'accepted')
+            ->create();
+
+        $this->be($admin, 'web');
+
+        $response = $this->get('originals/' . $post->id);
```
weerd, line 23: I've been using `actingAs()` in the web tests I've done. Not sure if it matters, but reads a little nicer as part of the flow and reads more similarly to how the API tests are written :)

```php
$response = $this->actingAs($admin, 'web')->get('originals/' . $post->id);
```

tests/Http/Web/OriginalsTest.php
```diff
@@ -0,0 +1,45 @@
...
+     */
+    public function testGetOriginalImageAsNonOwner()
+    {
+        $somebodyElse = factory(User::class)->create();
+
+        $post = factory(Post::class)
+            ->states('photo', 'accepted')
+            ->create();
+
+        $this->be($somebodyElse, 'web');
```
weerd, line 40: Same as comment above regarding `actingAs()` if interested in changing!

tests/Http/Web/OriginalsTest.php
```diff
@@ -0,0 +1,45 @@
...
+
+        $response = $this->get('originals/' . $post->id);
+        $response->assertStatus(302); // Redirect to 'register'.
```
weerd, line 43: Thoughts on using `$response->assertRedirect('/register')` instead and can get rid of comment.

weerd (approved):
Just some small comments but overall looks great! 💯

tests/Http/Web/OriginalsTest.php
DFurnes: Mmm yeah, I agree that reads nicer!

tests/Http/Web/OriginalsTest.php
DFurnes: I think I'd like to update this middleware to show an "unauthorized" page like [this](https://github.com/DoSomething/aurora/blob/master/resources/views/auth/unauthorized.blade.php) if you try to visit the page when logged-in (since as of now it actually does a wonky double-redirect from here to `/register` and then, because you're already logged-in, from there to `/`). I'll make a cleanup ticket so we don't forget.

weerd: Ah ok! Sounds good 👍

tests/Http/Web/OriginalsTest.php
DFurnes: Updated in 68712ce!

PR #1121, opened 2021-02-11
Adds promotions_muted_at User field

aaronschachter:
### What's this PR do?

This pull request adds a new `promotions_muted_at` field to our User model and adds an admin form to set it to the current time (for testing). The next PR (#1124) will modify our User observer to delete the user's Customer.io profile if this field is set, but holding off that, for now, to hopefully keep this easy to review.

👀 

It'd be sweet to get the `data-confirm` code from Aurora working in this app to ask the staff member to confirm before muting (or sending a password reset), but out of scope for this PR.

### Relevant tickets

References [Pivotal #176771234](https://www.pivotaltracker.com/story/show/176771234).

### Checklist

### DIFF
--- a/app/Http/Controllers/Web/Admin/UserController.php
+++ b/app/Http/Controllers/Web/Admin/UserController.php
@@ -157,7 +157,7 @@ class UserController extends Controller
     }
 
     /**
-     * Mutes promotions for user
+     * Mutes promotions for user.
      *
      * @param User $user
      * @return \Illuminate\Http\Response
--- a/app/Http/Controllers/Web/Admin/UserController.php
+++ b/app/Http/Controllers/Web/Admin/UserController.php
@@ -8,6 +8,7 @@ use App\Http\Controllers\Controller;
 use App\Models\User;
 use App\Types\CauseInterestType;
 use App\Types\PasswordResetType;
+use Carbon\Carbon;
 use Illuminate\Http\Request;
 
 class UserController extends Controller
@@ -153,5 +154,25 @@ class UserController extends Controller
         return redirect()
             ->route('admin.users.show', $user->id)
             ->with('flash', 'Sent ' . $type . ' email to user.');
+    }
+
+    /**
+     * Mutes promotions for user
+     *
+     * @param User $user
+     * @return \Illuminate\Http\Response
+     */
+    public function mutePromotions(User $user)
+    {
+        $this->authorize('delete', $user);
+
+        logger('Muting promotions', ['user' => $user->id]);
+
+        $user->promotions_muted_at = Carbon::now();
+        $user->save();
+
+        return redirect()
+            ->route('admin.users.show', $user->id)
+            ->with('flash', 'Promotions muted for user. Shhh....');
     }
 }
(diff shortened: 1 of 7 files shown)


app/Http/Controllers/Web/Admin/UserController.php
```diff
@@ -154,4 +155,24 @@ public function sendPasswordReset(User $user, Request $request)
...
             ->with('flash', 'Sent ' . $type . ' email to user.');
     }
+
+    /**
+     * Mutes promotions for user.
+     *
+     * @param User $user
+     * @return \Illuminate\Http\Response
+     */
+    public function mutePromotions(User $user)
```
DFurnes, line 165: This is subjective, but I think I'd lean towards making a new `Admin\PromotionsController` for this, and putting this code under the `destroy` method (following [this old DHH advice](http://jeromedalbert.com/how-dhh-organizes-his-rails-controllers/)). It keeps our controllers from getting too long, and kinda elegantly mirrors the route structure. (And if we ever wanted  a resubscribe button, it'd be easy to put in `PromotionsController@store`!)

routes/web.php
```diff
@@ -102,6 +102,10 @@
                 'as' => 'users.resets.create',
                 'uses' => 'Admin\UserController@sendPasswordReset',
             ]);
+            Route::post('users/{user}/mute-promotions', [
+                'as' => 'users.mute-promotions',
+                'uses' => 'Admin\UserController@mutePromotions',
+            ]);
```
DFurnes, line 108: Also subjective, but I like the idea of keeping this RESTful & considering this a `promotions` sub-resource on the user that we can delete (and following the same convention for the route alias):

```suggestion
            Route::delete('users/{user}/promotions', [
                'as' => 'users.promotions.destroy',
                'uses' => 'Admin\PromotionsController@destroy',
            ]);
```

resources/views/admin/users/partials/danger-zone.blade.php
```diff
@@ -26,6 +26,25 @@
...
+            <div class="form-item">  
+                <label for="mute-promotions" class="field-label">Mute Promotions</label>
+
+                <p class="footnote">This will delete the user's Customer.io profile.</p>
+            </div>
+
+            <div class="form-actions">
+                <button type="submit" class="button -secondary">
+                    Mute promotions
+                </button>
```
DFurnes, line 43: Last thought – it'd be great to have a test for this form that asserts that the `muted_promotions_at` field is set & the Customer.io API request is fired! You can use [WebActionTest](https://github.com/DoSomething/northstar/blob/master/tests/Http/Web/WebActionTest.php) as an example for testing web routes.

app/Http/Controllers/Web/Admin/UserController.php
```diff
@@ -154,4 +155,24 @@ public function sendPasswordReset(User $user, Request $request)
...
+        $this->authorize('delete', $user);
+
+        logger('Muting promotions', ['user' => $user->id]);
+
+        $user->promotions_muted_at = Carbon::now();
+        $user->save();
+
+        return redirect()
+            ->route('admin.users.show', $user->id)
+            ->with('flash', 'Promotions muted for user. Shhh....');
```
DFurnes, line 176: This is great copy. 👏

DFurnes (approved):

resources/views/admin/users/partials/danger-zone.blade.php
aaronschachter: There are some tests in place for the Customer.io API request introduced in #1124 -- I can add a new test class for the admin user form over in that PR.

PR #1126, opened 2021-02-18
Admin view refactoring & cleanup

weerd:
### What's this PR do?

This pull request does some initial admin views refactoring of templates and setup for upcoming work to get the admin UI previously in Rogue working in Northstar.

Probably best to go commit-by-commit:

- 🚫   [#dffbb17](https://github.com/DoSomething/northstar/pull/1126/commits/dffbb1765be734111c49a4b7e5571e29b9cd8d95) Adding `.prettierignore` so that when delving into vendor code and editing it to troubleshoot, Prettier does not reformat the code on save.
- 📸   [#0d9ffcf](https://github.com/DoSomething/northstar/pull/1126/commits/0d9ffcf123c9f6e63083b3a29dc5bb59aeb2265a) Renaming `'status'` flash messages to match others that us `'flash'` so that all flash messages in the system use the same variable name (also the term `status` is used within some resources so best to go with `flash` and not confuse things!).
- 🌱   [#e5adb4b](https://github.com/DoSomething/northstar/pull/1126/commits/e5adb4bbd2814051b801da278f2b4b0553e7751c) Unifies the admin layouts (between Northstar & Rogue) to use a single main admin layout located in `views/admin/layouts/main`.
- 📣   [#42c3ae](https://github.com/DoSomething/northstar/pull/1126/commits/42c3ae1bd17cb801addc11f05f449a73920900a0) enables showing the `/admin/campaigns/create` view and ensuring it is grouped as an admin route.
- 🧼   [#b00515e](https://github.com/DoSomething/northstar/pull/1126/commits/b00515e6656849a106632888c352c26b51c9b4b6) & [#12f1650](https://github.com/DoSomething/northstar/pull/1126/commits/12f1650bf2613a2e3b7c32b1f34bc38a9063c436) General cleanup, bug fix and updates.

This PR contains some initial work towards setting things up for enabling the more complicated web UIs related to `post/put/patch/delete` actions.

It's a bit heft but most of the changes are pretty straightforward!

### Relevant tickets

### Checklist


resources/views/admin/layouts/main.blade.php
```diff
@@ -0,0 +1,54 @@
...
+                @yield('main_content')
+            </div>
+        </div>
+
+        {{-- {{ isset($state) ? scriptify($state) : scriptify() }} --}}
+        <script src="{{ elixir('app.js', 'dist') }}"></script>
+    </body>
+
+    {{-- {{ scriptify($auth, 'AUTH') }} --}}
+    {{-- {{ scriptify(['GRAPHQL_URL' => config('services.graphql.url')], 'ENV') }} --}}
```
weerd, line 53: Rogue had these commented out `scriptify` calls and seeing how they'll be incorporated into NS. Figured it would be fine to leave commented out for now.

weerd:
Yay! Finally green 🚥  @_@

phpunit.xml
```diff
@@ -21,6 +21,7 @@
         <env name="SESSION_DRIVER" value="array"/>
         <env name="QUEUE_CONNECTION" value="sync"/>
         <env name="SIXPACK_ENABLED" value="false"/>
+        <env name="DS_ENABLE_ADMIN_ROUTES" value="true"/>
```
weerd, line 24: Forgetting this item was giving me a world of trouble with the CircleCI tests! 😓

lindsayrmaher (approved):
it all makes sense to me after going through each commit! not sure if you want another set of eyes before merging.

weerd:
I think it should be ok! Nothing world-changing really happening in here 😄

resources/views/admin/redirects/create.blade.php
```diff
@@ -1,7 +1,10 @@
-@extends('layouts.admin')
+@extends('admin.layouts.main')
 
 @section('title', 'Redirects')
-@section('subtitle', 'Create & manage URL redirects.')
+
+@section('header_content')
+    @include('admin.layouts.header', ['subtitle' => 'Create & manage URL redirects.'])
+@endsection
```
DFurnes, line 7: I'm curious to hear more about this change! Why set `subtitle` via the embedded partial?

weerd: Ah yep! I meant to write something on this cause there was a bit extra thought placed on this! Thanks for the reminder and sorry if this is a long explanation! Just want to make sure what led me to this decision is clear! 😅 

I was trying to find a reasonable approach between combining what was imported from the Aurora admin layout and the Rogue admin layout to combine them. While similar in some aspects, they differed in certain approaches and wanted to update it while keeping it flexible.

The **Aurora** admin layout template hardcoded the header and would output the `title` and `subtitle` in specific `@yield` calls. The `@yield` for the `subtitle` defaulted to `...` and I got the impression that was just so that it wouldn't output an empty `<p>` tag if no `subtitle` variable was provided (you can't easily do `if` conditional with `@yield`-ed variables, so the default was necessary):

```php
// resources/views/layouts/admin.blade.php (old)
....
<header class="header" role="banner"> 
    <div class="wrapper"> 
        <h1 class="header__title">  
            @yield('title', 'Administration') 
        </h1> 
        <p class="header__subtitle">  
            @yield('subtitle', '…') 
        </p>  
    </div>  
</header> 

@yield('main_content')
...
```
The **Rogue** admin layout template used a partial layout file for the header but oddly included it _inside_ the `main_content` section. I find this makes it less flexible for future edits with the header tied directly to the output of the `main_content`:

```php
// resources/views/layouts/master.blade.php (old)
...
<div class="container">

    @yield('main_content')

</div>
...

// resources/views/campaigns/create.blade.php
...
@section('main_content')

    @include('layouts.header', ['header' => 'Campaign Creation'])
    ...
@stop
```

---

I didn't scour all the templates in Rogue, but got the impression that maybe there's a page that doesn't show the header or is slightly different which is why outputting the header wasn't controlled by the main layout file, but by the specific page template and thus the decision to place it inside the `main_content`.

With the approach I landed on, it uses a `@section('header_content') ... @show` section which essentially allows either no header, or defining a header, but letting the page decide if it needs one.

The decision to include the header as an embedded partial, apart from added benefits of a single file that defines the header and making it easier to maintain, it lets the page determine whether to include it or not, _and_ makes it easier to deal with variables passed to the partial. Where the `@yield` could only provide a default if no variable is provided, the template partial can actually handle conditional logic and just not output anything (no empty tags) if no variable is provided 😄

DFurnes: Got it! I'm curious if we would ever want an admin page without a subtitle, since it makes the design look a bit squished, but I appreciate the thought behind making sure we have that choice!

weerd: We defs had some coming from the Rogue side, so I figured it'd be helpful to accommodate 🤷  We can always add subtitles where missing in the future 😄

PR #1127, opened 2021-02-19
Tracks new Promotions Resubscribe event in Customer.io

aaronschachter:
### What's this PR do?

This pull request tracks a new `promotions_resubscribe` event if we recreate a muted member's Customer.io profile -- and introduces the path towards refactoring campaign activity events to recreate profiles for muted members. I moved deleting a user's Customer.io profile into the `updated` hook instead of the `updating` hook within the UserObserver, which seemed to DRY logic, but felt like the best place to make Customer.io updates since we already update profiles from that hook.

Unfortunately I still can't get the last unit test to provide coverage for a `promotions_resubscribe` event getting triggered from a call to the new `trackCustomerIoEvent` method in the User model that this PR introduces. Locally I was testing by using the admin UI when viewing user -- mute promotions for a subscribed user to verify their profile is deleted. Then send them a Rock the Vote Activate Account reset. This will trigger updating the customer.io profile, sending a `promotions_resubscribe` event, and then the `call_to_action_email` that's used to send an activate account email in the Customer.io side. 

Screen Shot 2021-02-19 at 4 58 41 PM

* This PR provides context for the conversation in [Slack](https://dosomething.slack.com/archives/C782XBKDM/p1613596884019300) about using the transactional API for campaign signup, post, and review transactional emails. If we're in the business of deleting profiles and re-creating them upon participating activity - I don't think there's a straight forward way to chain the campaign Customer.io events to occur after a user's profile is recreated -- which means we would potentially not be able to send the email because the campaign event could recreate the user's profile first. Using the transactional API seems like a good workaround for this. We could still track campaign activity events in order to segment by users' activity for promotional purposes (and where we wouldn't need the event to be created with a fully restored profile).

* The relevant jobs that create campaign signup, post, reviewed Customer.io events should be refactored to call the `$user->trackCustomerIoEvent` method instead of the `customerIo->trackEvent` method, to ensure we recreate a profile if a muted user happens to create new campaign activity. This is for a future PR (and wouldn't apply until we stop sending activity API requests to the Rogue app and to this app instead)

### Relevant tickets

References [Pivotal #176771234](https://www.pivotaltracker.com/story/show/176771234).

### Checklist

### DIFF
--- a/app/Models/MongoModel.php
+++ b/app/Models/MongoModel.php
@@ -79,9 +79,11 @@ class MongoModel extends BaseModel
      */
     public function removeNullAttributes()
     {
+        /*
         $this->attributes = array_filter($this->attributes, function ($value) {
             return !is_null($value);
         });
+        */
     }
 
     /**
--- a/app/Models/MongoModel.php
+++ b/app/Models/MongoModel.php
@@ -79,11 +79,15 @@ class MongoModel extends BaseModel
      */
     public function removeNullAttributes()
     {
-        /*
-        $this->attributes = array_filter($this->attributes, function ($value) {
-            return !is_null($value);
-        });
-        */
+        $this->attributes = array_filter($this->attributes, function ($value, $key) {
+            /*
+             * We need to save null values for promotions_muted_at to determine whether a user
+             * profile should be recreated in Customer.io.
+             * @see UserObserver@updated
+             * @see https://github.com/DoSomething/northstar/pull/1127#discussion_r579494892
+             */
+            return $key === 'promotions_muted_at' ? true : !is_null($value);
+        }, ARRAY_FILTER_USE_BOTH);
     }
 
     /**
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -1136,14 +1136,22 @@ class User extends MongoModel implements
     }
 
     /**
-     * Tracks a Customer.io event with given name and data.
+     * Tracks a Customer.io event with given name and data if user is a subscriber.
      * Unmutes promotions if muted.
      *
      * @param string $eventName
-     * @param array $eventDataa
+     * @param array $eventData
      */
     public function trackCustomerIoEvent($eventName, $eventData)
     {
+        // If user is not subscribed to any platform, do not send to Customer.io.
+        if (
+            !$this->email_subscription_status &&
+            !static::isSubscribedSmsStatus($this->sms_status)
+        ) {
+            return;
+        }
+
         if (!isset($user->promotions_muted_at)) {
             return CreateCustomerIoEvent::dispatch($this, $eventName, $eventData);
         }
@@ -1152,7 +1160,11 @@ class User extends MongoModel implements
         $this->promotions_muted_at = null;
         $this->save();
 
-        // But there's no guarantee this will happen after the profile update.
-        CreateCustomerIoEvent::dispatch($this, $eventName, $eventData);
+        /*
+         * Note: there's no guarantee this will happen after the profile update, so a user who has
+         * been unmuted may never receive a transactional email from campaigns triggered by this
+         * given eventName.
+         */
+        return CreateCustomerIoEvent::dispatch($this, $eventName, $eventData);
     }
 }
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -944,7 +944,7 @@ class User extends MongoModel implements
          * user's messaging journey per their source (e.g., Rock The Vote, newsletter subscription).
          */
         if (PasswordResetType::isActivateAccount($type)) {
-            return CreateCustomerIoEvent::dispatch($this, 'call_to_action_email', $data);
+            return $this->trackCustomerIoEvent('call_to_action_email', $data);
         }
 
         // Send transactional emails for forgot password requests that don't need to be tracked.
@@ -1152,7 +1152,8 @@ class User extends MongoModel implements
             return;
         }
 
-        if (!isset($user->promotions_muted_at)) {
+        // If promotions are not muted, send our event.
+        if (!isset($this->promotions_muted_at)) {
             return CreateCustomerIoEvent::dispatch($this, $eventName, $eventData);
         }
 
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -1134,4 +1134,25 @@ class User extends MongoModel implements
         $this->deletion_requested_at = now();
         $this->save();
     }
+
+    /**
+     * Tracks a Customer.io event with given name and data.
+     * Unmutes promotions if muted.
+     *
+     * @param string $eventName
+     * @param array $eventDataa
+     */
+    public function trackCustomerIoEvent($eventName, $eventData)
+    {
+        if (!isset($user->promotions_muted_at)) {
+            return CreateCustomerIoEvent::dispatch($this, $eventName, $eventData);
+        }
+
+        // Unmute promotions to recreate the profile in Customer.io.
+        $this->promotions_muted_at = null;
+        $this->save();
+
+        // But there's no guarantee this will happen after the profile update.
+        CreateCustomerIoEvent::dispatch($this, $eventName, $eventData);
+    }
 }
--- a/app/Observers/UserObserver.php
+++ b/app/Observers/UserObserver.php
@@ -65,7 +65,6 @@ class UserObserver
     public function updating(User $user)
     {
         $changed = $user->getDirty();
-
         $currentlySubscribed = [
             'email' => $user->email_subscription_status,
             'sms' => User::isSubscribedSmsStatus($user->sms_status),
@@ -136,6 +135,11 @@ class UserObserver
         ) {
             $user->promotions_muted_at = Carbon::now();
         }
+
+        /*
+         * Note: Is it cleaner to move the rest of this function code into our updated hook?
+         * We're no longer altering the values to save within this transaction here.
+         */
 
         // If we're updating a user's club, dispatch a Customer.io event.
         if (isset($changed['club_id'])) {
--- a/app/Observers/UserObserver.php
+++ b/app/Observers/UserObserver.php
@@ -170,10 +170,14 @@ class UserObserver
      */
     public function updated(User $user)
     {
+        logger('User updated hook', ['user_id' => $user->id]);
+
         $mutedPromotions = isset($user->promotions_muted_at);
 
         // If we just made a change to mute promotions:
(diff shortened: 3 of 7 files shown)


app/Models/User.php
```diff
@@ -1134,4 +1134,38 @@ public function requestDeletion(): void
...
+        }
+
+        // If promotions are not muted, send our event.
+        if (!isset($this->promotions_muted_at)) {
+            return CreateCustomerIoEvent::dispatch($this, $eventName, $eventData);
+        }
+
+        // Unmute promotions to recreate the profile in Customer.io.
+        $this->promotions_muted_at = null;
+        $this->save();
```
aaronschachter, line 1162: This ideally should be triggering a `promotions_resubscribe` event, but the logging added in https://github.com/DoSomething/northstar/pull/1127/commits/c08cb12c035f46a9d3c69c667c1ffad92486c707 indicates that `$user->wasChanged('promotions_muted_at')` is false in our UserObserver's `updated` hook, which is why the `promotions_resubscribe` event is not dispatched.

tests/Models/UserModelTest.php
```diff
@@ -466,5 +468,54 @@ public function testUnmutingPromotionsViaEmailTopicsChange()
...
+        //  Manually mute promotions.
+        $user->email_subscription_status = Carbon::now();
+        $user->save();
+
+        $user->trackCustomerIoEvent('test_event', ['foo' => 'bar']);
+
+        // Verify promotions are no longer muted.
+        $this->assertNull($user->promotions_muted_at);
+        // TODO: Why isn't this firing?
+        // $this->assertCustomerIoEvent($user, 'promotions_resubscribe');
```
aaronschachter, line 518: Debugging this in https://github.com/DoSomething/northstar/pull/1127/files#r579465129

app/Models/User.php
DFurnes: Hmm, that's strange! I'm wondering if there's a difference in `wasChanged` (which seems to read from `$model->changes`) vs. the `getDirty()` function that seems to manually compare the in-memory attributes to the `$this->original` that were first loaded from the DB)...

If you log or `dd()` the `$changed` array at the top fo that `updating` function, does it show that field?

aaronschachter: In the `updating` function, we do log ` $user->getChanged()` at the end. (https://github.com/DoSomething/northstar/blob/aa579060ba360e64e0595a0f18d298a9e0952066/app/Observers/UserObserver.php#L170-L173)

 It doesn't show the field :
```
[2021-02-19 20:40:37] local.DEBUG: updated user {"id":"60301696950bf34cae01468e","changed":{"audit":"*****"}} {"request_id":null,"user_id":"600201a023a8223a1e4575a3","client_id":"northstar"}
```
It will show up when I'm making updates in the profile (both whether I'm setting or unsetting the field) via the Mute Promotions button or editing the profile to unmute:
```
[2021-02-19 21:07:52] local.DEBUG: updated user {"id":"60301696950bf34cae01468e","changed":{"audit":"*****","sms_status":"active","promotions_muted_at":null}} {"request_id":null,"user_id":"600201a023a8223a1e4575a3","client_id":"northstar"}
```

aaronschachter: Debugging what `$user->getChanges()` displays in the `updated` function -- after the `trackCustomerIoEvent` it only has `audit` and `updated_at ` array key entries.

When making a profile update to unmute, the `promotions_muted_at` array key will appear with a `null` value (which is what I'd expect to see after making the `save` call from within `trackCustomerIoEvent`...

aaronschachter: If I set the user's `first_name` field in this code as a test, 

```
  $this->promotions_muted_at = null
  $this->first_name = 'WHY HAVE YOU FORSAKEN ME';
  $this->save();
```
It shows up but the `promotions_muted_at` doesn't I have a feeling it's because we're setting the value to null.

```
[2021-02-19 21:34:19] local.DEBUG: updated user {"id":"60301696950bf34cae01468e","changed":{"first_name":"WHY HAVE YOU FORSAKEN ME","audit":"*****"}} {"request_id":null,"user_id":"600201a023a8223a1e4575a3","client_id":"northstar"}
[2021-02-19 21:34:19] local.DEBUG: User updated hook {"user_id":"60301696950bf34cae01468e","first_name":"WHY HAVE YOU FORSAKEN ME","updated_at":{"MongoDB\\BSON\\UTCDateTime":{"$date":{"$numberLong":"1613770459032"}}}} {"request_id":null,"user_id":"600201a023a8223a1e4575a3","client_id":"northstar"}
```

Yes, it seems to be the case. If I set the `promotions_muted_at` to a date in the future here, 
```
$this->promotions_muted_at = Carbon::now()->addYears(25);
$this->save();
 ````

it shows up
```
[2021-02-19 21:37:38] local.DEBUG: updated user {"id":"60301696950bf34cae01468e","changed":{"audit":"*****","promotions_muted_at":{"MongoDB\\BSON\\UTCDateTime":{"$date":{"$numberLong":"2402689058718"}}}}} {"request_id":null,"user_id":"600201a023a8223a1e4575a3","client_id":"northstar"}
```

File: app/Models/MongoModel.php [hunk 3]
```diff
@@ -79,9 +79,11 @@ public function getClearable()
      */
     public function removeNullAttributes()
     {
+        /*
         $this->attributes = array_filter($this->attributes, function ($value) {
             return !is_null($value);
         });
+        */
```
aaronschachter, line 86: @DFurnes  - This is the culprit! If we allow null values for `promotions_muted_at` -- it would get things working properly. Would it be feasible to allow nulls just for this field?

app/Models/User.php
DFurnes: Ohhhhhh! 💡

app/Models/MongoModel.php
DFurnes: Ahh, tricky! I think the main reason we added this was that, at least at the time, MongoDB unique index constraints would not allow multiple records to have `null` (because, like, `null` and `null` would count as duplicates), and so we wanted to make sure that we _unset_ `email` or `mobile` when a user zeroed out those fields. (And then it was just a bonus that we also kept our Mongo documents looking a bit more tidy!)

I don't see a downside to exempting the `promotions_muted_at` field from this logic, although I'm curious if there's some other way we could better track whether there's a field that _was_ in the original but got unset...

aaronschachter: Thanks for the context. I think for now I'll work on exempting the `promotions_muted_at` from the logic -- not sure about a more general approach.  We can always change remove the exemption if we figure out a better way.

DFurnes: Works for me!

tests/Models/UserModelTest.php
aaronschachter: I see this happening when I test live now after committing https://github.com/DoSomething/northstar/pull/1127/commits/62cedc63683d8535425eab8f6728d5eb3cd3491c per [this thread](https://github.com/DoSomething/northstar/pull/1127#discussion_r579494892) -- but still stumped why this test is failing. When I try to add logging into the UserObserver `updated` and `updating` hooks, and run this specific test, I don't see the logging 🤔

app/Models/MongoModel.php
aaronschachter: https://github.com/DoSomething/northstar/pull/1127/commits/62cedc63683d8535425eab8f6728d5eb3cd3491c gets this working locally, but the test still fails for sending the `promotions_resubscribe`. I noticed that this commit still doesn't save the null `promotions_muted_at` field to the DB, so tried editing the `setClearable` to allow it to be saved. That still doesn't fix this test though https://github.com/DoSomething/northstar/pull/1127#discussion_r579466472.

tests/Models/UserModelTest.php
```diff
@@ -466,5 +468,54 @@ public function testUnmutingPromotionsViaEmailTopicsChange()
...
+        //  Manually mute promotions.
+        $user->email_subscription_status = Carbon::now();
+        $user->save();
+
+        $user->trackCustomerIoEvent('test_event', ['foo' => 'bar']);
+
+        // Verify promotions are no longer muted.
+        $this->assertNull($user->promotions_muted_at);
+        // TODO: Why isn't this firing? This is happening when I manually test.
+        // $this->assertCustomerIoEvent($user, 'promotions_resubscribe');
```
DFurnes, line 518: Weird! I wonder if this is somehow down to some implementation detail of how chained jobs are run?

DFurnes (approved):

app/Models/User.php
```diff
@@ -1134,4 +1134,38 @@ public function requestDeletion(): void
...
+
+        /*
+         * Note: there's no guarantee this will happen after the profile update, so a user who has
+         * been unmuted may never receive a transactional email from campaigns triggered by this
+         * given eventName.
+         */
```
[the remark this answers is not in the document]
aaronschachter: Sounds reasonable (and I think would allow us to avoid chaining jobs, but doesn't seem to leave this as-is). As far as scale/issues, thinking forward to when we ideally move Gambit messaging into SQS, could that potentially be a reason not to switch?

app/Observers/UserObserver.php
```diff
@@ -142,13 +133,14 @@ public function updating(User $user)
...
-
             $user->promotions_muted_at = Carbon::now();
-
-            DeleteCustomerIoProfile::dispatch($user);
         }
 
+        /*
+         * Note: Is it cleaner to move the rest of this function code into our updated hook?
+         * We're no longer altering the values to save within this transaction here.
+         */
```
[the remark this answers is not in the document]
aaronschachter: I'll make this change in a follow-up PR.

app/Models/User.php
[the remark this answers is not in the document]
DFurnes: Potentially! We may find we want to run certain queues in FIFO mode & others in "standard" mode too.

tests/Models/UserModelTest.php
aaronschachter: That'd make sense. Your comment got me searching for mocking chained events... but this still doesn't work -- it failed asserting that false is true --  `The expected chain was not pushed`

```
        Queue::assertPushedWithChain(UpsertCustomerIoProfile::class, [
            new CreateCustomerIoEvent($user, 'promotions_resubscribe', []),
        ]);
 ```
Technically I think we're covered for looking for the `promotions_resubscribe` event being fired via the observer update tests above...

PR #1130, opened 2021-02-22
Makes User.promotions_muted_at property fillable

aaronschachter:
### What's this PR do?

This pull request fixes a bug in #1121 -- when I added the `promotions_muted_at` field, it wasn't editable via API request.

### How should this be reviewed?

👀 

### Any background context you want to provide?

I wondered whether the additional test is kind of repetitive alongside a lot of the other tests, but at the same time -- I didn't catch this bug until trying to edit the field over in a new Chompy import.

### Relevant tickets

References [Pivotal #176771234](https://www.pivotaltracker.com/story/show/176771234).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [x] Documentation added for new features/changed endpoints.
- [x] Added appropriate feature/unit tests.
- [ ] If new attributes were added to users, there is a corresponding PR to surface these in Aurora.
- [ ] If new attributes were added to users, then the data team already knows about these changes.

PR #1131, opened 2021-02-23
Revert "Makes User.promotions_muted_at property fillable"

aaronschachter:
Reverts DoSomething/northstar#1130: When trying to make a `PUT /users/:id` request for users to mute promotions for users have been removed for GDPR, the [request fails with validation errors](https://dosomething.slack.com/archives/CJ340DEEL/p1614115585008100?thread_ts=1614106925.003200&cid=CJ340DEEL) because these users have had email and mobile fields cleared (e.g. the request fails with error - `The mobile field is required when email is not present.`).

To workaround this, we'll create a new `DELETE users/{user}/promotions` endpoint that the Mute Promotions import (https://github.com/DoSomething/chompy/pull/200) can call. It'll mock our admin PromotionsController added in https://github.com/DoSomething/northstar/pull/1121, which will allow us to set the `promotions_muted_at`  field to the current time the request is made without worrying about validating any existing fields on the user.

### DIFF
--- a/app/Auth/Registrar.php
+++ b/app/Auth/Registrar.php
@@ -79,7 +79,6 @@ class Registrar
             'sms_paused' => 'boolean',
             'sms_subscription_topics.*' => 'in:general,voting',
             'last_messaged_at' => 'date',
-            'promotions_muted_at' => 'date',
             'email_subscription_status' => 'boolean',
             'email_subscription_topics.*' => Rule::in(
                 EmailSubscriptionTopicType::all(),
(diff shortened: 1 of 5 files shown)


mendelB (approved):
👍 sounds good!

> It'll mock our admin PromotionsController added in #1121, which will allow us to set the promotions_muted_at field to the current time the request is made without worrying about validating any existing fields on the user.

Is there any reason these need to be separate or can we consolidate both requests into one general controller for this?

aaronschachter:
> Is there any reason these need to be separate or can we consolidate both requests into one general controller for this?

I think because the web controller redirects back to the user view with a flash message, we'd likely want separate controllers for API and web...

PR #1135, opened 2021-02-26
Admin Clubs CRUD UI

weerd:
### What's this PR do?

This pull request updates all the Clubs related admin functionality to make sure staff can create, edit and update clubs!

It also includes a bit of cleanup in a couple files 🧼 

👀 

Went through and ensured all the CRUD activities (aside from Delete which doesn't seem like it existed for Clubs on Rogue...?) are working as expected!

### Relevant tickets

References [Pivotal #]().

### Checklist


.env.example
```diff
@@ -59,6 +59,12 @@ FACEBOOK_REDIRECT_URL=https://northstar.dev/facebook/verify
 FACEBOOK_APP_SECRET=FacebookSecret
 FACEBOOK_APP_ID=FacebookAppId
 
+FASTLY_API_KEY=
+FASTLY_BACKEND_SERVICE_ID=
+FASTLY_FRONTEND_SERVICE_ID=
+FASTLY_FRONTEND_SERVICE_URL=https://dev.dosomething.org
+FASTLY_TABLE_REDIRECTS=
```
weerd, line 66: We were missing this set of ENV variables from the example file!

database/seeds/ClubTableSeeder.php
```diff
@@ -0,0 +1,18 @@
...
+    public function run()
+    {
+        // Clear the database.
+        DB::connection('mysql')
+            ->table('clubs')
+            ->delete();
+
+        factory(Club::class, 5)->create();
+    }
+}
```
weerd, line 18: Since the Admin front-end uses both Blade views and React it's still a bit confusing when the interface is requesting items from the local MySQL DB or via our GraphQL dev endpoint.

I was stuck on a 404 that was confusing me for a bit until I remembered that some views our output via Laravel and others for the same resource via React. There a plans to eventually unify this all and still deciding on an approach there, but just noting since it's caught me more than once now! 😓

resources/views/admin/clubs/edit.blade.php
```diff
@@ -1,44 +1,46 @@
...
                     <div class="form-item">
                         <label class="field-label">School ID</label>
-                        @include('forms.text', ['name' => 'school_id', 'placeholder' => 'The school universal ID associated with this club', 'value' => $club->school_id])
+                        @include('admin.forms.text', ['name' => 'school_id', 'placeholder' => 'The school universal ID associated with this club', 'value' => $club->school_id])
                     </div>
 
 
                     <ul class="form-actions -inline -padded">
                         <li><input type="submit" class="button" value="Update Club"></li>
+                        <li><a href="{{ url('/admin/clubs', [$club->id]) }}" class="button -tertiary">Cancel</a></li>
```
weerd, line 43: Figure it would be nice to have a "cancel" and just go back to the view for the single club :)

routes/web.php
```diff
@@ -102,6 +97,11 @@
                 'except' => ['index', 'show'],
             ]);
 
+            // Clubs
+            Route::resource('/clubs', 'Admin\ClubsController', [
+                'except' => ['index', 'show', 'destroy'],
+            ]);
+
```
weerd, line 104: Moving these down into the Admin area instead of sitting in the public set of routes (the rest are fine up there until updated since they controllers define auth middleware).

aaronschachter (approved):

PR #1138, opened 2021-03-01
Properly load user relationship for Customer.io activity events.

DFurnes:
### What's this PR do?

This pull request fixes an issue passing a user along to Customer.io when sending post, signup, or review events.

Gah, I should've caught this the first time around!! The reason that these mocks had been failing to register is because these jobs were loading the associated user via `->user()` (the method that configures the relationship), and not `->user` (the property that uses a magic method to load the related model).

With this fix, tests now pass properly without the hack! 😌

### Relevant tickets

References [Pivotal #176851319](https://www.pivotaltracker.com/story/show/176851319).

### Checklist


weerd (approved):
Nice find! 💯

PR #1142, opened 2021-03-04
Only set phone on Customer.io profile when member is SMS subscriber

aaronschachter:
### What's this PR do?

This pull request modifies our upsert of a member's Customer.io profile to only set their `phone` attribute when they are subscribed to SMS promotions.

👀 

This work is the first step towards only storing mobile numbers in Customer.io for members that are subscribed to SMS promotions -- that way we don't need to lookup each Northstar user when sending a Gambit broadcast to safety check that we're not accidentally sending a broadcast to a member who's unsubscribed.

The next step is removing phone numbers from all existing Customer.io profiles that are not subscribed to SMS promotions. This will likely be handled by running a Chompy import or by modifying the [BackfillCustomerIo]( https://github.com/DoSomething/northstar/blob/master/app/Console/Commands/BackfillCustomerIo.php) command to optionally accept a CSV of user ids (there's no need to backfill all users -- just ones who have a mobile set, are unsubscribed, and haven't had their promotions muted per our February Customer.io retirement import)

### Relevant tickets

References [Pivotal #177168236](https://www.pivotaltracker.com/story/show/177168236).

### Checklist

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -651,7 +651,7 @@ class User extends MongoModel implements
         $payload = [
             'id' => $this->id,
             'email' => $this->email,
-            'phone' => $this->mobile,
+            'phone' => $this->isSmsSubscribed() ? $this->mobile : null,
             'sms_status' => $this->sms_status,
             'sms_status_source' => data_get($this->audit, 'sms_status.source'),
             'sms_paused' => (bool) $this->sms_paused,
@@ -1057,13 +1057,15 @@ class User extends MongoModel implements
     }
 
     /**
-     * Whether user has a subscribed SMS status.
+     * Whether user is a SMS subscriber.
      *
      * @return bool
      */
     public function isSmsSubscribed()
     {
-        return isset($this->sms_status) &&
+        return
+            isset($this->mobile) &&
+            isset($this->sms_status) &&
             self::isSubscribedSmsStatus($this->sms_status);
     }
 
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -1057,7 +1057,7 @@ class User extends MongoModel implements
     }
 
     /**
-     * Whether user is a SMS subscriber.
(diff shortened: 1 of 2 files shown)


mendelB (approved):
👏

mendelB:
Curious -- why can't we just rely on the SMS status itself on the profile instead of needing to remove the phone numbers? Or is it that the less PII the better in c.io?

aaronschachter:
>  why can't we just rely on the SMS status itself on the profile instead of needing to remove the phone numbers

We do this currently -- but Gambit will still make a Northstar lookup for each broadcast request, just to make sure we didn't accidentally send to unsubscribes via the Customer.io filter (it's happened before, which is why we added the Northstar lookup).

The current Gambit broadcast endpoint expects a NS ID -- our end goal, once we've removed all mobile numbers for non-subscribes, is that the endpoint will instead accept a `mobile` field,. We can assume all requests with a `mobile` are indeed ones we should send, in order to remove the Northstar lookup (and simplify scaling/network requests). Does that make sense?

mendelB:
> > why can't we just rely on the SMS status itself on the profile instead of needing to remove the phone numbers
> 
> We do this currently -- but Gambit will still make a Northstar lookup for each broadcast request, just to make sure we didn't accidentally send to unsubscribes via the Customer.io filter (it's happened before, which is why we added the Northstar lookup).
> 
> The current Gambit broadcast endpoint expects a NS ID -- our end goal, once we've removed all mobile numbers for non-subscribes, is that the endpoint will instead accept a `mobile` field,. We can assume all requests with a `mobile` are indeed ones we should send, in order to remove the Northstar lookup (and simplify scaling/network requests). Does that make sense?

Aha! thanks for clarifying, that def makes sense. So moving the source of truth over to the mobile field itself for a more assured filter removing the need to confirm with Northstar since the subscription filter via c.io could be wonky.

PR #1143, opened 2021-03-04
Send tags to Customer.io and delay job from sending after review is completed

lindsayrmaher:
### What's this PR do?

This pull request adds `tags` as a field in the payload we send to Customer.Io. The field is set based on what tags have been added when the posts are reviewed, and we are sending it as a single comma separated string vs an array.

The PR also adds a 5 minute delay to when the payload is sent upon review of a post. This delay is to ensure that when a review is completed, it is not sent over unintentionally without tags. Anthony plans on utilizing the tags as liquid tags in c.io and we want to give campaign managers enough time to add them after choosing `accept` or `reject` on a particular post.

👀 

We are hoping that the addition of the tags will help users to understand why their posts may be getting rejected!

### Relevant tickets

References [Pivotal #176421256](https://www.pivotaltracker.com/story/show/176421256).

### Checklist

### DIFF
--- a/app/Managers/PostManager.php
+++ b/app/Managers/PostManager.php
@@ -129,7 +129,8 @@ class PostManager
     {
         $post = $this->repository->reviews($post, $data, $comment, $admin);
 
-        // sending review to customerIo is delayed to ensure tags have been added to the post before it triggers an email to users
+        // sending review to customerIo is delayed to ensure users can optionally add tags to rejected or accepted posts
+        // if they're added, we want them to be sent along in the payload!
         SendReviewedPostToCustomerIo::dispatch($post)->delay(
             now()->addMinutes(5),
         );
--- a/app/Managers/PostManager.php
+++ b/app/Managers/PostManager.php
@@ -129,7 +129,10 @@ class PostManager
     {
         $post = $this->repository->reviews($post, $data, $comment, $admin);
 
-        SendReviewedPostToCustomerIo::dispatch($post);
+        // sending review to customerIo is delayed to ensure tags have been added to the post before it triggers an email to users
+        SendReviewedPostToCustomerIo::dispatch($post)->delay(
+            now()->addMinutes(5),
+        );
 
         // Log that a post was reviewed.
         info('post_reviewed', [
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -381,6 +381,7 @@ class Post extends Model
         $action = $this->actionModel;
         $signup = optional($this->signup);
         $campaign = optional($signup->campaign);
+        $tags = $this->tagSlugs()->toArray();
 
         return array_merge(
             [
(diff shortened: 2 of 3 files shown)


app/Managers/PostManager.php
```diff
@@ -129,7 +129,10 @@ public function review($post, $data, $comment = null, $admin = null)
     {
         $post = $this->repository->reviews($post, $data, $comment, $admin);
 
-        SendReviewedPostToCustomerIo::dispatch($post);
+        // sending review to customerIo is delayed to ensure tags have been added to the post before it triggers an email to users
```
katiecrane, line 132: this comment will be super helpful when I look at this file in 4 months!

katiecrane (approved):
looks great! let me know if you need a hand testing on QA!

app/Managers/PostManager.php
```diff
@@ -129,7 +129,10 @@ public function review($post, $data, $comment = null, $admin = null)
...
+        SendReviewedPostToCustomerIo::dispatch($post)->delay(
+            now()->addMinutes(5),
```
weerd, line 134: Just making sure I understand this correctly, so it waits 5 minutes so that it gives the reviewers time to add tags to each post? Are tags required on the post? If they are, is there not a way to add a validation rule indicating that posts need to be tagged with at least 1 tag or something and thus wouldn't be considered "reviewed" unless it was tagged? 

*Disclaimer: I'm not completely familiar with the reviewing logic so my questions above might just not be feasible. 😅

lindsayrmaher: no it's a good q! I was thinking about this as well, but I think the reasoning was that tags aren't required and they also can be added after a post has been "reviewed" ie the accept or reject has been pressed. I don't know that it's a part of campaigns workflow right now to ensure a tag is added every review, but the hope is that it will be going forward. @katiecrane let me know if i didn't explain that correctly!

katiecrane: Totally agree @lindsayrmaher! Tags aren't required and aren't added until after "accept" or "reject" has been clicked. The idea is that a tag should always be added to a rejected post (so we can tell the member why it has been rejected) and will be optionally added on accepted posts.

weerd: Ah ok! Then I might recommend that the comment be reworded a little bit, because to me it implies that tags will always be added and it won't trigger an email to users without them 🤔

lindsayrmaher: fair point, @weerd ! i can wordsmith it a bit.

lindsayrmaher: updated the comment, lemme know what you think! @weerd

app/Managers/PostManager.php
```diff
@@ -129,7 +129,11 @@ public function review($post, $data, $comment = null, $admin = null)
     {
         $post = $this->repository->reviews($post, $data, $comment, $admin);
 
-        SendReviewedPostToCustomerIo::dispatch($post);
+        // sending review to customerIo is delayed to ensure users can optionally add tags to rejected or accepted posts
+        // if they're added, we want them to be sent along in the payload!
```
weerd, line 133: Nice! 👍

weerd (approved):

PR #1055, opened 2020-09-03
Extract "change password" fields into their own form.

DFurnes:
### What's this PR do?

This pull request extracts the "change password" fields into their own form, and adds a field to confirm the user's existing password before allowing a new password to be set. This makes it harder for an attacker to hijack a user's account.

Screen Shot 2020-09-03 at 11 57 52 AM

### How should this be reviewed?

Do these changes look good? Any edge cases I may have forgotten?

### Any background context you want to provide?

🔏

### Relevant tickets

References [Pivotal #173717920](https://www.pivotaltracker.com/story/show/173717920).

### Checklist

- [ ] This PR has been added to the relevant Pivotal card.
- [ ] Documentation added for new features/changed endpoints.
- [ ] Added appropriate feature/unit tests.
- [ ] If new attributes were added to users, there is a corresponding PR to surface these in Aurora.
- [ ] If new attributes were added to users, then the data team already knows about these changes.

PR #1144, opened 2021-03-05
Set SMS status via Profile Edit page

### LINKED ISSUE #855: Bug: Can't clear user profile fields
Introduced in #482-- if I try to unset my address or mobile number, I can't. We had a user change their mobile number to someone else's to avoid receiving texts.

Related: if someone does clear out their mobile number, should we unset their `sms_status`?

aaronschachter:
### What's this PR do?

This pull request moves the `mobile` and `sms_status` form fields from the Subscriptions Profile template (`/profile/subscriptions`) into a new partials template, and uses it within the Profile Edit page (`/user/:id/edit`) to require members to set their SMS status if they add a mobile number.

It also fixes #855 to allow members to clear non-required fields from their profile, such as their mobile number. This was made possible by moving changing passwords into a separate form over in https://github.com/DoSomething/northstar/pull/1055 (this PR also adds some cleanup we missed there) 

If a member does clear their mobile number, we set their `sms_status` to null -- as they're no longer subscribed if they don't have a mobile set.

👀 

I initially was going to make the validation rule that `sms_status` should be required if mobile in the Registrar globally -- but it resulted in quite a [bunch of tests that needed to be updated](https://github.com/DoSomething/northstar/compare/edit-sms-status).  I went with making the validation rule just in the profile edit form for now to keep this reviewable (and less potentially big breaking changes)

### Relevant tickets

References [Pivotal #176890667](https://www.pivotaltracker.com/story/show/176890667), [#169212591](https://www.pivotaltracker.com/n/projects/2475250/stories/169212591)

### Checklist


app/Http/Controllers/Web/UserController.php
```diff
@@ -112,20 +112,19 @@ public function update(Request $request, $id)
...
-        $values = array_diff($request->all(), ['']);
+        $values = $request->all();
 
-        $user->fill($values)->save();
-
-        if (isset($values['password'])) {
-            event(new PasswordUpdated($user, 'profile'));
+        if (!$values['mobile']) {
+            // HACK: Clear any existing SMS status values if we don't have the user's mobile number.
+            // This should likely happen in UserObserver -- but that logic is pretty complex.
```
DFurnes, line 122: Agreed! Seems reasonable to leave here for now, and appreciate the `// HACK:` documentation!

DFurnes (approved):
Looks great! ☎️

PR #1146, opened 2021-03-08
Add Badges field to User

lindsayrmaher:
### What's this PR do?

This pull request adds a `badges` field to a user as an array. We will be updating certain actions taken by users to add the badges users have earned here and reflect this in user profiles on phoenix! This PR also adds badges to documentation and tests for user creation as well as data payload that is sent to customer.io.

👀 Should I add anything additional as far as tests? I'm planning to add tests around updating when I add the badge checks to signups and posts.

Adding this field on the user so that we can permanently see their badges and display them in the user profile with a progress bar.

### Relevant tickets

References [Pivotal #175820443](https://www.pivotaltracker.com/story/show/175820443).

### Checklist


app/Models/User.php
```diff
@@ -783,6 +787,36 @@ public function toCustomerIoPayload()
...
+                $this->badges,
+            )
+                ? true
+                : false,
+            'three_staff_faves_badge' => in_array(
+                'three-staff-faves',
+                $this->badges,
+            )
+                ? true
+                : false,
```
weerd, line 819: I feel like there's a better/cleaner way to handle this... like mapping through the list of official badges and if you hit a match set it to true in a new array that you can then merge in. I get that this is copying the approach already established but feels like it's doing a lot of the same thing and can be improved. Not necessarily for this PR, but just thinking out loud! 📣

documentation/endpoints/users.md
```diff
@@ -163,6 +163,8 @@ Either a mobile number or email is required.
   source: String; // Immutable. Will only be set on new records.
   source_detail: String; // Only accepted alongside a valid 'source'.
   created_at: Number; // timestamp
+  causes: Array;
+  badges: Array;
```
weerd, line 167: Is this endpoint still active? Not actually sure... just confused since there's a `v2` as well! 😅

app/Types/BadgeType.php
```diff
@@ -0,0 +1,34 @@
...
+    {
+        return [
+            'signup' => 'Signup Badge',
+            'one_post' => 'One Post Badge',
+            'two_posts' => 'Two Posts Badge',
+            'three_posts' => 'Three Posts Badge',
+            'breakdown' => 'Breakdown Badge',
+            'one_staff_fave' => 'One Staff Fave Badge',
+            'two_staff_faves' => 'Two Staff Faves Badge',
+            'three_staff_faves' => 'Three Staff Faves Badge',
```
weerd, line 31: Can you use the constants above as the keys? I think most of the other type files use the constants that way :)

app/Models/User.php
```diff
@@ -1142,6 +1175,40 @@ public function getSchoolIdPreviewAttribute()
...
+    public function getBadgesAttribute($value)
+    {
+        //Ensure we always return an array value for the badges attribute.
+        return empty($value) ? [] : $value;
+    }
+
+    /**
+     * Add the given cause to the user's array of causes if it is not already there.
+     *
+     * @param string $cause
```
weerd, line 1193: DocBlock should mention badges and not causes.

weerd (dismissed):
Just a couple suggested updates but looking good otherwise!

weerd (commented):
Oop! Didn't mean to approve just yet since needs a couple updates 😬

documentation/endpoints/users.md
aaronschachter: Gambit still posts to the /v1/users endpoint FYI

lindsayrmaher: I added it to both based on the fact that club_id had been added in both doc files!

app/Types/BadgeType.php
lindsayrmaher: good q, i guess i'm wondering if that will effect how it gets used on admin pages. right now i'm hoping to use the `key` value as what gets added to the array and the `value` is what is used for the label of the checkbox. which is the setup we used for causes. if we use the const, will it have the same output essentially?

weerd: Yeah, I think it should have the same output, like using a variable. Defs used like that in other type files like the `ActionType`:

https://github.com/DoSomething/northstar/blob/e904d81b5d66d5b21aafc195133dc23df80c15d6/app/Types/ActionType.php#L26-L38

weerd: Oh wait, you already changed it! 👍

weerd (approved):
Huzzah! 💯

PR #1150, opened 2021-03-08
Fix issue with volunteer credits images.

DFurnes:
### What's this PR do?

This pull request should fix an issue where volunteer credits certificates were not including the user's images, because the images did not include the proper `Access-Control-Allow-Origin` header.

Mendel added a test for the expected headers in b0d4e21. I got that passing in ef08f14.

After that, I removed the now unused [laravel-cors](https://github.com/fruitcake/laravel-cors) dependency in b5ba89b. The "update dependencies" part of this step pulled in a newer version of [phpseclib](https://github.com/phpseclib/phpseclib) than our code expects (it was a sub-dependency pulled in from something else we'd relied on), so I added an explicit dependency on it in e737a11 to get everything passing again.

This applies the same solution that @mendelB and I applied in DoSomething/rogue#1016 early last year. I forgot this context when migrating these endpoints over into Northstar & we falsely assumed we could get away with the package's middleware!

### Relevant tickets

### Checklist

### DIFF
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -39,7 +39,7 @@ class Kernel extends HttpKernel
         'api' => [
             \App\Http\Middleware\ParseOAuthHeader::class,
             \Illuminate\Routing\Middleware\SubstituteBindings::class,
-            \Barryvdh\Cors\HandleCors::class,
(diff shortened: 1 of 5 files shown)


mendelB (approved):
🙌 🔐

PR #1124, opened 2021-02-12
Delete Customer.io profile upon muting promotions

aaronschachter:
### What's this PR do?

This pull request adds deleting a user's Customer.io profile upon setting their `promotions_muted_at` field (added in #1121), via a new `DeleteCustomerIoProfile` job. It also makes these changes:

* Adds a new `suppressCustomer` method to our Customer.io service to handle [176808789](https://www.pivotaltracker.com/n/projects/2328687/stories/176808789) -- we want to suppress a Customer.io customer when they delete their DS account, not delete a Customer.io customer (details in the Pivotal story)

* Don't send updates to the user's Customer.io profile if the user's promotions have been muted.

* Renames `SendUserToCustomerIo` job to `UpsertCustomerIoProfile` for naming consistency with `DeleteCustomerIoProfile`. Uses the `CustomerIoJob` introduced in #1118 to DRY.

### How should this be reviewed?

👀 

### Any background context you want to provide?

In a future PR, we'll tweak the User updating hook in the observer to check for specific changes to their subscription status fields to determine whether to recreate the muted user's Customer.io profile. 

### Relevant tickets

References [Pivotal #176771234](https://www.pivotaltracker.com/story/show/176771234), [#176808789](https://www.pivotaltracker.com/story/show/176808789).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [ ] Documentation added for new features/changed endpoints.
- [x] Added appropriate feature/unit tests.
- [ ] If new attributes were added to users, there is a corresponding PR to surface these in Aurora.
- [ ] If new attributes were added to users, then the data team already knows about these changes.

PR #1151, opened 2021-03-09
Adds console command to remove phone from Customer.io profiles if not subscribed to SMS

aaronschachter:
### What's this PR do?

This pull request adds a `northstar:cio-remove-mobile` Artisan command that finds all users that have not had promotions muted, have a mobile set, and a null or unsubscribed SMS status. Then, for each user found:

* if they are subscribed to email, update their Customer.io profile (which will remove their `phone` attribute per #1142)
* if they are not subscribed, set their `promotions_muted_at` attribute to trigger deleting their Customer.io profile (per #1124) 

This PR also adds a migration to create a compound index to the `mobile`, `promotions_muted_at`, and  `sms_status` fields on our users MongoDB collection to optimize finding the ~1M out of ~11-12M documents we want to update. Because of this migration -- **this should not be merged until we're in a good spot to put Northstar into maintenance mode for a deploy.**

👀 

* I left a TODO for fixing the test that asserts muting promotions for each still-subscribed-to-email user for now, not sure why this is falling. When I've tested locally, I've verified the `promotions_muted_at` field is set correctly.

* I think we may want to delete the `BackfillCustomerIo` job entirely -- as it makes it somewhat easy to accidentally re-add profiles we spent a weekend deleting if someone ran it on production. Left out that deletion for now, but any objections?
 
### Relevant tickets

References [Pivotal #177168318](https://www.pivotaltracker.com/story/show/177168318).

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,10 +12,9 @@ Northstar! :sparkles:
 
 Northstar currently uses two database connections: its original MongoDB connection, and the MySQL connection originally owned by [Rogue](https://github.com/DoSomething/rogue).
 
-To run a migration, specify the desired DB connection and path:
+To run a database migration, it's easiest to use our custom Artisan command:
 ```
-php artisan migrate --path="database/migrations-mongodb/" --database="mongodb"
-php artisan migrate --path="database/migrations/" --database="mysql"
+php artisan migrate:all
 ```
 
 ### Contributing
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Northstar [![wercker status](https://app.wercker.com/status/109bce734be9a06703562876265f5bd9/s/dev "wercker status")](https://app.wercker.com/project/byKey/109bce734be9a06703562876265f5bd9) [![StyleCI](https://styleci.io/repos/26884886/shield?style=flat-rounded)](https://styleci.io/repos/26884886)
+# Northstar
 
 This is **Northstar**, the DoSomething.org user & identity service. It's our single "source of truth" for member information.
 Northstar is built using [Laravel 6.0](https://laravel.com/docs/6.x), [OAuth 2.0 Server](https://oauth2.thephpleague.com), and [MongoDB](https://www.mongodb.com).
@@ -7,6 +7,16 @@ Northstar is built using [Laravel 6.0](https://laravel.com/docs/6.x), [OAuth 2.0
 
 Check out the [API Documentation](https://github.com/DoSomething/northstar/blob/master/documentation/README.md) to start using
 Northstar! :sparkles:
+
+### Migrations
+
+Northstar currently uses two database connections: its original MongoDB connection, and the MySQL connection originally owned by [Rogue](https://github.com/DoSomething/rogue).
+
+To run a migration, specify the desired DB connection and path:
+```
+php artisan migrate --path="database/migrations-mongodb/" --database="mongodb"
+php artisan migrate --path="database/migrations/" --database="mysql"
+```
 
 ### Contributing
 
--- a/app/Console/Commands/RemoveCustomerIoMobile.php
+++ b/app/Console/Commands/RemoveCustomerIoMobile.php
@@ -30,10 +30,9 @@ class RemoveCustomerIoMobile extends Command
      */
     public function handle()
     {
-        // Find all SMS unsubscribed users where promotions have not been muted.
-        $query = (new User())->newQuery()
-            ->whereNull('promotions_muted_at')
-            ->whereIn('sms_status', ['stop', 'undeliverable']);
+        $query = User::whereNull('promotions_muted_at')
+            ->whereNotNull('mobile')
+            ->whereIn('sms_status', [null, 'stop', 'undeliverable']);
 
         $progress = $this->output->createProgressBar($query->count());
 
--- a/app/Console/Commands/RemoveCustomerIoMobile.php
+++ b/app/Console/Commands/RemoveCustomerIoMobile.php
@@ -1,0 +1,58 @@
+<?php
+
+namespace App\Console\Commands;
+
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -8,6 +8,16 @@ Northstar is built using [Laravel 6.0](https://laravel.com/docs/6.x), [OAuth 2.0
...
 Northstar! :sparkles:
 
+### Migrations
+
+Northstar currently uses two database connections: its original MongoDB connection, and the MySQL connection originally owned by [Rogue](https://github.com/DoSomething/rogue).
+
+To run a migration, specify the desired DB connection and path:
+```
+php artisan migrate --path="database/migrations-mongodb/" --database="mongodb"
+php artisan migrate --path="database/migrations/" --database="mysql"
```
aaronschachter, line 18: I got the MongoDB migrations to run successfully -- but can't figure out how to get the MySQL migrations to run. This results in an error, trying to run the 1st Northstar migration. `php artisan migrate` produces these same error too... 
```
SQLSTATE[42S01]: Base table or view already exists: 1050 Table 'users' already exists (SQL: create table `users` (`id` int unsigned not null auto_increment primary key, `name` varchar(255) not null, `email` varchar(255) not null, `password` varchar(255) not null, `remember_token` varchar(100) null, `created_at` timestamp null, `updated_at` timestamp null) default character set utf8 collate 'utf8_unicode_ci')
```

DFurnes: I'm wondering if this could be from an inconsistent state in your local Rogue migrations table (since we had to rename this migration in a6222a8c4e2bc8c29d05ea5d933995fbe286643f to stop it from conflicting with the MongoDB migration of the same name).

Can you try running `migrate:fresh` to see if migrating a fresh database works?

```
php artisan migrate:fresh --path="database/migrations/" --database="mysql"
```

app/Console/Commands/RemoveCustomerIoMobile.php
```diff
@@ -0,0 +1,59 @@
...
+
+        $progress = $this->output->createProgressBar($query->count());
+
+        $query->chunkById(200, function (Collection $users) use ($progress) {
+            $users->each(function (User $user) use ($progress) {
+                // If the user is not subscribed to email:
+                if (!$user->email_subscription_status) {
+                    // Delete their profile entirely by muting promotions.
+                    $user->promotions_muted_at = now();
+                    $user->save();
```
DFurnes, line 46: Now that we're doing this in a few places, it might be nice to make a `$user->mute()` method that ran these two steps.

app/Console/Commands/RemoveCustomerIoMobile.php
```diff
@@ -0,0 +1,59 @@
...
+     * Execute the console command.
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        // Find all SMS unsubscribed users where promotions have not been muted.
+        $query = (new User())->newQuery()
+            ->whereNull('promotions_muted_at')
+            ->whereIn('sms_status', ['stop', 'undeliverable']);
```
aaronschachter, line 36: Looks like we may also need to run this script for users who have a mobile in Customer.io but a null `sms_status`. Anthony only wants to send a [re-permissioning broadcast to 3k users](https://dosomething.slack.com/archives/C03T8SDMS/p1615390194000200) but reports that there are about 38K profiles that have a mobile and no SMS status.

app/Observers/UserObserver.php
```diff
@@ -49,13 +49,17 @@ public function creating(User $user)
      */
     public function created(User $user)
     {
-        // Send payload to Blink for Customer.io profile.
+        // Send metrics to StatHat.
+        Log::info('user_created', ['source' => $user->source]);
+
+        // If user is not subscribed to any platform -- do not create a Customer.io profile.
+        if (!($user->email_subscription_status || $user->isSmsSubscribed())) {
```
DFurnes, line 56: Similarly, wondering if encapsulating this in a `$user->receivesMessaging` method might be handy?

tests/Console/RemoveCustomerIoMobileTest.php
```diff
@@ -0,0 +1,30 @@
...
+        $this->customerIoMock->shouldHaveReceived('updateCustomer')->times(11);
+        $this->customerIoMock->shouldHaveReceived('deleteCustomer')->times(7);
+
+        foreach ($removeMobiles as $user) {
+            $this->assertNull($user->promotions_muted_at);
+        }
+
+        foreach ($deleteProfiles as $user) {
+            // TODO: Why is this failling?
+          //$this->assertNotNull($user->promotions_muted_at);
```
DFurnes, line 27: You might have to use `$user->fresh()->promotions_muted_at` to load changes from the database.

DFurnes (approved):
Looking good! 🔇

README.md
aaronschachter: Yes, that worked!

app/Console/Commands/RemoveCustomerIoMobile.php
aaronschachter: Updated this PR to run the update or/delete additionally have a null `sms_status` field value set.

app/Observers/UserObserver.php
aaronschachter: I'd like to hold on this change for now-- there are a few spots where we check both `email_subscription_status` and `sms_status` field values -- but whether a user receives messaging also depends on the `promotions_muted_at` fields (and we inspect `email_subscription_status` and `sms_status` current and changed values to determine whether to set/unset in the `updating` hook)

aaronschachter:
I think this is ready for merge... but I want to hold off on doing so until we're ready to go in maintenance mode on production to run the MongoDB migration with its deployment.

PR #1152, opened 2021-03-11
Fix issue with post filters.

DFurnes:
### What's this PR do?

This pull request includes a fix for an issue filtering the `v3/posts` endpoint, detailed in [#177311555](https://www.pivotaltracker.com/story/show/177311555).

I was able to validate the problem is fixed by ff83b79 using my local Phoenix & by hitting the API via Paw, but the test in 3c95dc5 seems to be passing even without the fix applied!? I'd love a fresh set of eyes on what I could be missing here, so we can make sure we have good test coverage for this behavior!

~We might also need to apply this fix to other controllers!~ Nope, we don't!

We'd kept Rogue's `FiltersRequests` trait as the `FiltersActivityRequests` trait when merging these two applications (since it conflicted with Northstar's slightly different trait of the same name). Unfortunately, we didn't update our includes to require the renamed trait so we got the Northstar one in here by mistake!

### Relevant tickets

References [Pivotal #177311555](https://www.pivotaltracker.com/story/show/177311555).

### Checklist

### DIFF
--- a/app/Http/Controllers/PostsController.php
+++ b/app/Http/Controllers/PostsController.php
@@ -2,6 +2,7 @@
 
 namespace App\Http\Controllers;
 
+use App\Http\Controllers\Traits\FiltersActivityRequests;
 use App\Http\Controllers\Traits\FiltersRequests;
 use App\Http\Requests\PostRequest;
 use App\Http\Transformers\PostTransformer;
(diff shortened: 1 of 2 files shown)


DFurnes:
I'm still unable to get this test to fail with the `FiltersRequests` trait applied to this controller, but I did find an interesting clue – we're extending `ActivityApiController` here (which includes `FiltersActivityRequests`). I wonder if there's some funny-business going on with how the two traits are applied & override each other that differs in tests vs. real execution?

DFurnes:
Regardless, I've confirmed that this fix (just removing the unnecessary `FiltersRequests` trait override) does get this interface back to it's expected behavior, so let's ship it!

aaronschachter (approved):
🚢

PR #1153, opened 2021-03-12
Fix volunteer credit PDF & scholarship newsletter signup.

DFurnes:
### What's this PR do?

This pull request re-installs the [laravel-cors](https://github.com/fruitcake/laravel-cors) package for origin-specific `Access-Control-Allow-Origin` headers, but whitelists the `images/*` routes to use a `*` header. This gets us the security benefits of origin-specific headers on our API responses, while avoiding issues that had introduced with PDF generation.

This should fix issues with [volunteer credit PDF generation](https://dosomething.slack.com/archives/C09ANFQLA/p1615231090048800) and [scholarship newsletter signup](https://dosomething.slack.com/archives/CUQMTP0RM/p1615483885001600).

👀

So why was this not working, anyways? This package _does_ specify the correct `Vary` header on responses, so we're not seeing any issues (AFAIK) on cross-origin API requests with a domain-specific `Access-Control-Allow-Origin` header.

Normally, requests for images don't include an `Origin` header (or require `Access-Control-Allow-Origin`) unless the [`crossorigin`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-crossorigin) attribute is set on the `img` tag.

However, the [react-pdf](https://react-pdf.org) library loads these images via `XmlHttpRequest` which _does_ need the `Access-Control-Allow-Origin` header when loading requests across origins (but, I believe, is still not setting that `crossorigin` flag). To allow those requests through, we need to allow `*`.

I don't think we _really_ want anyone to be able to to load images from our site in this way, but it's an acceptable short-term fix. Longer term, it'd probably make sense to get these under the same `www.` (or `qa.`, etc.) domain so they don't have to be cross-domain requests in the first place...

### Relevant tickets

### Checklist


DFurnes:
I'm running into trouble getting the subscriptions endpoint to work as expected on my local in 29e1bd4, but I'm hoping it will work on QA (as this is effectively rolling back #1150 for that endpoint). If not, back to the drawing board!

config/cors.php
```diff
@@ -0,0 +1,59 @@
...
+    | If ['*'] is provided to allowed_methods, allowed_origins or allowed_headers
+    | all methods / origins / headers are allowed.
+    |
+    */
+
+    /*
+     * You can enable CORS for 1 or multiple paths.
+     * Example: ['api/*']
+     */
+    'paths' => ['api/*'],
```
mendelB, line 23: Will this cover our `/v1` & `/v2/` APIs?

mendelB (approved):
👏  This branch name shook me to the core

config/cors.php
DFurnes: Gah, I had that on my local but never committed it. 🙈

DFurnes: Pushed up in 00af2de!

DFurnes: Hmm, that seems to [cause issues](https://app.circleci.com/pipelines/github/DoSomething/northstar/967/workflows/446b2276-89ff-41bf-a3dc-4c8092c7ec6e/jobs/1003) with some of the OAuth grant tests. I've limited CORS support to `api/`, `v2/users` and `v2/subscriptions` to get us the minimum coverage we should need here for existing functionality.

PR #1154, opened 2021-03-12
Fix configuration for CORS middleware.

DFurnes:
### What's this PR do?

This pull request fixes some configuration issues with our CORS middleware:

🌐 According to [the documentation](https://github.com/fruitcake/laravel-cors#upgrading-from-0x--barryvdh-laravel-cors), this must be run as a global middleware! We were attaching it to a group. I've updated this in e8aab75.

📽️ It seems like the `allowed_origins` setting must include the protocol (e.g. `https://*.dosomething.org` instead of `*.dosomething.org`). This makes sense given that the protocol is included in the `Origin` header (but I wish it were documented more clearly)! Updated in 351e638.

🍨 Finally, I tested the image/certificate issue after applying these two fixes & it seems to work with the proper origin-specific `Access-Control-Allow-Origin` header! I'm going to try removing our custom override in 7c42bc3.

👀

Hoping this closes out this saga! 🙏 

### Relevant tickets

### Checklist


mendelB (approved):
🤯 🙏 🥇 

Iiiiinteresting! So the images requests _will_ have the `Origin` included despite the caching? (I think you've mentioned that the `Vary` _is_ working as it should)

DFurnes:
Yeah, I think so! I keep getting myself turned around on what is & isn't affecting this. It does seem like images are always getting `Vary: Origin` headers, even when `Origin` is not provided, and so this shouldn't cause issues with caching. We'll see!

PR #1155, opened 2021-03-15
Assign Badges for various User Actions

lindsayrmaher:
### What's this PR do?

This pull request adds some checks in the `SignupObserver`, `PostObserver`, and `UserObserver` as well as the `tag` method on the `Post Model` that will add the correct badges to a user when they've earned them. 

One big q I have, should I use the variables from the `BadgeType` vs explicitly writing them out? 

This is one of the last steps we need in order to start successfully awarding badges to users and saving them on the `User`

### Relevant tickets

References [Pivotal #175820943](https://www.pivotaltracker.com/story/show/175820943).

### Checklist

### DIFF
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -805,7 +805,6 @@ class Post extends Model
 
     /**
      * Checks whether a user should be given a badge based on their post.
-     *
      */
     public function calculatePostBadges()
     {
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -479,7 +479,7 @@ class Post extends Model
         }
 
         if ($this->user) {
-            $userPosts = $this->user->posts;
+            $userPosts = $this->user->posts->limit(3);
             foreach ($userPosts as $post) {
                 if ($post->tagSlugs()->contains('good-submission')) {
                     if (
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -6,6 +6,7 @@ use App\Models\Traits\HasCursor;
 use App\Models\User;
 use App\Notifications\PostTagged;
 use App\Services\GraphQL;
+use App\Types\BadgeType;
 use Hashids\Hashids;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\ModelNotFoundException;
@@ -481,19 +482,43 @@ class Post extends Model
             $userPosts = $this->user->posts;
             foreach ($userPosts as $post) {
                 if ($post->tagSlugs()->contains('good-submission')) {
-                    if (!in_array('one_staff_fave', $this->user->badges)) {
-                        $this->user->addBadge('one_staff_fave');
-                    } elseif (
-                        !in_array('two_staff_faves', $this->user->badges) &&
-                        in_array('one_staff_fave', $this->user->badges)
+                    if (
+                        !in_array(
+                            BadgeType::get('ONE_STAFF_FAVE'),
+                            $this->user->badges,
+                        )
                     ) {
-                        $this->user->addBadge('two_staff_faves');
+                        $this->user->addBadge(BadgeType::get('ONE_STAFF_FAVE'));
                     } elseif (
-                        !in_array('three_staff_faves', $this->user->badges) &&
-                        in_array('three_posts', $this->user->badges) &&
-                        in_array('two_staff_faves', $this->user->badges)
+                        !in_array(
+                            BadgeType::get('TWO_STAFF_FAVES'),
+                            $this->user->badges,
+                        ) &&
+                        in_array(
+                            BadgeType::get('ONE_STAFF_FAVE'),
+                            $this->user->badges,
+                        )
                     ) {
-                        $this->user->addBadge('three_staff_faves');
+                        $this->user->addBadge(
+                            BadgeType::get('TWO_STAFF_FAVES'),
+                        );
+                    } elseif (
+                        !in_array(
+                            BadgeType::get('THREE_STAFF_FAVES'),
+                            $this->user->badges,
+                        ) &&
+                        in_array(
+                            BadgeType::get('THREE_POSTS'),
+                            $this->user->badges,
+                        ) &&
+                        in_array(
+                            BadgeType::get('TWO_STAFF_FAVES'),
+                            $this->user->badges,
+                        )
+                    ) {
+                        $this->user->addBadge(
+                            BadgeType::get('THREE_STAFF_FAVES'),
+                        );
                     }
                 }
             }
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -802,4 +802,26 @@ class Post extends Model
     {
         return config('services.slack.url');
     }
+
+    /**
+     * Checks whether a user should be given a badge based on their post.
+     *
+     * @return void
+     */
+    public function calculatePostBadges()
+    {
+        $user = $this->user;
+        if ($user) {
+            $userPosts = $user->posts();
+            if ($userPosts->count() === 1) {
+                $user->addBadge(BadgeType::get('ONE_POST'));
+            } elseif ($userPosts->count() === 2) {
+                $user->addBadge(BadgeType::get('TWO_POSTS'));
+            } elseif ($userPosts->count() === 3) {
+                $user->addBadge(BadgeType::get('THREE_POSTS'));
+            }
+            $user->save();
+        }
+        return;
+    }
 }
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -813,11 +813,12 @@ class Post extends Model
         $user = $this->user;
         if ($user) {
             $userPosts = $user->posts();
-            if ($userPosts->count() === 1) {
+            $userPostsCount = $userPosts->count();
+            if ($userPostsCount === 1) {
                 $user->addBadge(BadgeType::get('ONE_POST'));
-            } elseif ($userPosts->count() === 2) {
+            } elseif ($userPostsCount === 2) {
                 $user->addBadge(BadgeType::get('TWO_POSTS'));
-            } elseif ($userPosts->count() === 3) {
+            } elseif ($userPostsCount === 3) {
                 $user->addBadge(BadgeType::get('THREE_POSTS'));
             }
             $user->save();
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -477,29 +477,27 @@ class Post extends Model
             }
         }
 
-        $userId = $this->northstar_id;
-        $user = User::findOrFail($userId);
-        if ($user) {
-            $userPosts = $user->posts;
+        if ($this->user) {
+            $userPosts = $this->user->posts;
             foreach ($userPosts as $post) {
                 if ($post->tagSlugs()->contains('good-submission')) {
-                    if (!in_array('one_staff_fave', $user->badges)) {
-                        $user->addBadge('one_staff_fave');
+                    if (!in_array('one_staff_fave', $this->user->badges)) {
+                        $this->user->addBadge('one_staff_fave');
                     } elseif (
-                        !in_array('two_staff_faves', $user->badges) &&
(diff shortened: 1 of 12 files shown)


app/Models/Post.php
```diff
@@ -477,6 +477,31 @@ public function tag($tag)
             }
         }
 
+        $userId = $this->northstar_id;
+        $user = User::findOrFail($userId);
+        if ($user) {
```
DFurnes, line 482: Since these now live in the same application, we can use Eloquent relationships to load this:

```suggestion
        if ($this->user) {
```

app/Models/Post.php
```diff
@@ -477,6 +477,31 @@ public function tag($tag)
...
+            $userPosts = $user->posts;
+            foreach ($userPosts as $post) {
```
DFurnes, line 484: How would this work for users with a lot of posts? Are there any ways we could do this more efficiently?

app/Observers/PostObserver.php
```diff
@@ -55,6 +56,20 @@ public function creating(Post $post)
...
+            $userPosts = $user->posts();
+            if ($userPosts->count() === 1) {
+                $user->addBadge('one_post');
+            } elseif ($userPosts->count() === 2) {
+                $user->addBadge('two_posts');
+            } elseif ($userPosts->count() === 3) {
+                $user->addBadge('three_posts');
+            }
+            $user->save();
+        }
```
DFurnes, line 72: It might be nice to encapsulate these calculations in a `$post->calculateBadges()` (and so on) method(s), like how we have a `updateOrCreateActionStats` above. This could come in handy for our backfill script too!

app/Observers/PostObserver.php
```diff
@@ -55,6 +56,20 @@ public function creating(Post $post)
...
     {
         $post->updateOrCreateActionStats();
+
+        $userId = $post->northstar_id;
+        $user = User::findOrFail($userId);
+        if ($user) {
+            $userPosts = $user->posts();
+            if ($userPosts->count() === 1) {
+                $user->addBadge('one_post');
+            } elseif ($userPosts->count() === 2) {
```
DFurnes, line 66: Will this recalculate the count each time we call `$userPosts->count()`, or does it keep a cached result? If not, we might want to move this up above or try using Eloquent's [relationship count](https://laravel.com/docs/8.x/eloquent-relationships#counting-related-models) functionality.

app/Observers/UserObserver.php
```diff
@@ -172,6 +186,14 @@ public function updating(User $user)
      */
     public function updated(User $user)
     {
+        if (
+            in_array('news', $user->email_subscription_topics) &&
+            !in_array('breakdown', $user->badges)
```
DFurnes, line 191: Looking at these side by side makes me wonder... would we want to name this badge `news_subscription`? I believe we went with more general "topic" naming for these subscription topics in case we ever re-branded a newsletter (and it also feels nice to keep these two variables consistent).

tests/Http/PostTest.php
```diff
@@ -1923,4 +1923,165 @@ public function testCreatingAPostWithGroupIdAndNoExistingSignup()
...
+            'quantity' => $quantity,
+            'details' => json_encode($details),
+        ]);
+
+        // Make sure the updated why_participated is updated on the signup.
+        $this->assertMysqlDatabaseHas('signups', [
+            'campaign_id' => $signup->campaign_id,
+            'northstar_id' => $signup->northstar_id,
+            'why_participated' => $why_participated,
+        ]);
```
DFurnes, line 1977: If this behavior (creating and storing a post/signup) is covered in another test, I think you can omit these assertions here to make it easier to follow the behavior that this specific test is checking against.

tests/Http/PostTest.php
```diff
@@ -1923,4 +1923,165 @@ public function testCreatingAPostWithGroupIdAndNoExistingSignup()
...
+    /**
+     * Test that after multiple posts have been successfully created, "one_post", "two_posts", and "three_posts" badges are added to the user.
+     *
+     * @return void
+     */
+    public function testMultiplePostsAddingABadges()
+    {
+        $photo_signup = factory(Signup::class)->create();
+        $quantity = $this->faker->numberBetween(10, 1000);
+        $hours_spent = $this->faker->randomFloat(2, 0.1, 999999.99);
```
DFurnes, line 1992: Nitpick: we use camel case for variable names, so these should be `$photoSignup` and `$hoursSpent`! 🐫

DFurnes:
> One big q I have, should I use the variables from the `BadgeType` vs explicitly writing them out?

I think that'd be a great idea!

app/Models/Post.php
lindsayrmaher: 🤯 whoa!

app/Models/Post.php
lindsayrmaher: I was wondering if maybe we could/should pull the first 3 or just limit to 3? thoughts?

lindsayrmaher: or add a break after the 3rd conditional?

app/Observers/PostObserver.php
lindsayrmaher: like this idea!

app/Models/Post.php
lindsayrmaher: when i made this update it throws a SQL error `SQLSTATE[HY000] [1040] Too many connections` maybe i missed something when writing? Seems like it's maybe pulling all users vs one?

app/Models/Post.php
DFurnes: Yeah, I like that! Limiting to 3 posts should get us everything we need here & avoid any worst-case results for power users with a ton of posts. You should be able to do that with a `limit(3)` [relationship constraint](https://laravel.com/docs/8.x/eloquent-relationships#querying-relations).

app/Models/Post.php
DFurnes: Huh, that's so strange! 🤔 I wonder if it's running into trouble from calling `$this->user` multiple times... what if you try holding `$user = $this->user` in memory and working with that `$user` variable?

DFurnes: We're currently using a [cross-database relationship](https://github.com/jenssegers/laravel-mongodb#cross-database-relationships) to power this particular relationship, so it might not be quite as efficient as Laravel's built-in relationship support for SQL models...

lindsayrmaher: that feels right to me 👍

app/Observers/UserObserver.php
lindsayrmaher: i was thinking this same thing. I think we would need to let Anthony know before launch, but other than that i think we could reasonably change the name before updating. @katiecrane am i missing anything here that would prevent us from doing this?

katiecrane: I don't think so! I like that idea and wish I had named it that from the start!

app/Models/Post.php
lindsayrmaher: so i was still running into issues with just $this->user `Undefined property: App\Observers\SignupObserver::$user` but i _think_ this is because it's in the observer rather than the model itself? So I just used `$signup->user` / `$post->user` since we are passing that in as a parameter. Seems to work as we expect.

DFurnes: OH! Yeah, good catch!

app/Observers/PostObserver.php
lindsayrmaher: just to make sure i follow, we want it to keep a cached result, yeah? i think it makes more sense to save it above as a var and reference vs recalculating in each conditional. that feels more readable as well.

app/Observers/UserObserver.php
lindsayrmaher: i'm going to add this in a follow up PR!!

app/Observers/PostObserver.php
DFurnes: Yep, that's what I was thinking. There's a chance Laravel already does this for us automatically, so it could be worth taking a peek with [Clockwork](http://northstar.test/__clockwork) to see if it's actually running this query three times or not. (Although, like you said, it might be more readable to save above regardless!)

tests/Http/SignupTest.php
```diff
@@ -50,6 +50,47 @@ public function testCreatingASignup()
...
 
+    /**
+     * Test that a POST request to /signups creates a new signup and adds a signup badge.
+     *
+     * POST /api/v3/signups
+     * @return void
+     */
+    public function testAddingFirstSignupBadge()
+    {
+        $this->withoutExceptionHandling();
```
DFurnes, line 61: Let's remove this before merge! 👾

tests/Http/SignupTest.php
```diff
@@ -50,6 +50,47 @@ public function testCreatingASignup()
...
+                'group_id' => null,
+            ],
+        ]);
+        // Make sure the signup is persisted.
+        $this->assertMysqlDatabaseHas('signups', [
+            'northstar_id' => $user->id,
+            'campaign_id' => $campaignId,
+            'quantity' => null,
+            'details' => 'badge-testing',
+        ]);
```
DFurnes, line 88: Like in the test above, I think we can omit these lines! We cover this behavior in [testCreatingASignup](https://github.com/DoSomething/northstar/blob/a1fe206763fe23525ce2570226df6ac442586e8a/tests/Http/SignupTest.php#L31-L50).

tests/Http/SignupTest.php
```diff
@@ -112,6 +153,7 @@ public function testCreatingASignupWithReferrerUserId()
      */
     public function testCreatingASignupForUserWithClubId()
     {
+        $this->withoutExceptionHandling();
```
DFurnes, line 156: 👾

tests/Http/TagsTest.php
```diff
@@ -13,6 +14,7 @@ class TagsTest extends TestCase
      */
     public function testTaggingAPost()
     {
+        $this->withoutExceptionHandling();
```
DFurnes, line 17: 👾

tests/Http/TagsTest.php
```diff
@@ -160,4 +162,310 @@ public function testWithoutTagScope()
...
+     * Test that a POST request to /tags with a good submission adds the staff fave badges.
+     *
+     * POST /v3/posts/:post_id/tag
+     * @return void
+     */
+    public function testGoodSubmissionEarnsABadge()
+    {
+        $post = factory(Post::class)->create();
+        $userId = $post->northstar_id;
+        $user = User::findOrFail($userId);
```
DFurnes, line 176: You could load these directly from the Eloquent relationship to save a little work:

```suggestion
        $user = $post->user;
```

tests/Http/UserTest.php
```diff
@@ -1215,4 +1215,55 @@ public function testSmsSubscriptionTopicsAreClearedWhenUnsubscribing()
...
+            ->states('email-subscribed-news')
+            ->create();
+
+        $this->assertMongoDatabaseHas('users', [
+            '_id' => $newsUser->id,
+            'email_subscription_status' => true,
+        ]);
+
+        $user = $newsUser->fresh();
+        $this->assertEquals(['breakdown'], $user->badges);
```
DFurnes, line 1236: What if we combined these into the `assertMongoDatabaseHas` above?

```suggestion
        $this->assertMongoDatabaseHas('users', [
            '_id' => $newsUser->id,
            'email_subscription_status' => true,
            'badges' => ['breakdown'],
        ]);
```

app/Models/Post.php
```diff
@@ -477,6 +478,53 @@ public function tag($tag)
             }
         }
 
+        if ($this->user) {
+            $userPosts = $this->user->posts->take(3);
```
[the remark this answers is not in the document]
lindsayrmaher: is that the `scopeWithTag` method? I'm having trouble wrapping my head around how to use it, there weren't examples that I could find in the codebase.

DFurnes (approved):
🎉 📛

PR #1157, opened 2021-03-16
Enable more admin functionality.

DFurnes:
### What's this PR do?

This pull request enables more admin views that we'd moved over from Rogue, and creates on-demand JWTs that can be used by these admin sessions to review, tag, and edit reportbacks. (These had previously been loaded from the `access_token` that we got in the authorization flow, but that's not a thing now that this all lives in one app!)

I'd recommend reviewing commit-by-commit. I'll add some notes in the diffs!

I think this should get us very close to being able to enable & ship these in production!

### Relevant tickets

### Checklist

### DIFF
--- a/app/Services/GraphQL.php
+++ b/app/Services/GraphQL.php
@@ -32,6 +32,12 @@ class GraphQL
      */
     public function query($query, $variables)
     {
+        // If we've disable server-side GraphQL requests, return an optional
+        // wrapper object (so reading keys from the query won't fail):
+        if (config('features.graphql_php') === false) {
+            return optional();
+        }
+
         $response = $this->client->query($query, $variables);
 
         return $response ? $response->getData() : [];
--- a/app/Services/GraphQL.php
+++ b/app/Services/GraphQL.php
@@ -32,7 +32,7 @@ class GraphQL
      */
     public function query($query, $variables)
     {
-        // If we've disable server-side GraphQL requests, return an optional
+        // If we've disabled server-side GraphQL requests, return an optional
         // wrapper object (so reading keys from the query won't fail):
         if (config('features.graphql_php') === false) {
             return optional();
--- a/config/features.php
+++ b/config/features.php
@@ -38,6 +38,9 @@ return [
     // If this is enabled, we'll expose the new "admin" routes from Aurora:
     'admin' => env('DS_ENABLE_ADMIN_ROUTES', false),
 
+    // Should we make server-side GraphQL requests?
+    'graphql_php' => env('DS_ENABLE_GRAPHQL_PHP', true),
+
     'gambit' => env('DS_ENABLE_GAMBIT_RELAY', false),
 
     'track_club_id' => env('DS_ENABLE_TRACK_CLUB_IDS', false),
--- a/resources/assets/admin/Administration.js
+++ b/resources/assets/admin/Administration.js
@@ -97,7 +97,7 @@ const Application = () => {
             <ShowUser selectedTab="referrals" />
           </Route>
 
-          <Route path="/admin/users/:id">
+          <Route path="/admin/activity/:id">
             <ShowUser />
           </Route>
 
--- a/resources/views/admin/app.blade.php
+++ b/resources/views/admin/app.blade.php
@@ -13,8 +13,6 @@
         <link rel="apple-touch-icon-precomposed" href="/assets/images/apple-touch-icon-precomposed.png">
         <link rel="stylesheet" href="{{ elixir('admin.css', 'dist') }}">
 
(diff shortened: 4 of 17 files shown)


DFurnes:
cc'ing @weerd since you've been working on a lot of this!

app/Services/GraphQL.php
```diff
@@ -32,6 +32,12 @@ public function __construct()
      */
     public function query($query, $variables)
     {
+        // If we've disable server-side GraphQL requests, return an optional
+        // wrapper object (so reading keys from the query won't fail):
+        if (config('features.graphql_php') === false) {
+            return optional();
+        }
```
DFurnes, line 39: I was using my own local GraphQL app while testing this out (so that I'd be able to see the same data as on server-rendered Blade views from my local database). However, the backend PHP process can't reach the `localhost:3000` running on my Mac & these queries error out! 😩

I'd like to find a better solution for this long-term, but this seemed like a nice quick fix to unblock testing any of the routes that make behind-the-scenes GraphQL requests.

resources/assets/admin/Administration.js
```diff
@@ -97,7 +97,7 @@ const Application = () => {
             <ShowUser selectedTab="referrals" />
           </Route>
 
-          <Route path="/admin/users/:id">
+          <Route path="/admin/activity/:id">
```
DFurnes, line 100: Rogue has a "user" page that shows a member's signups, posts, and summarized profile. But, of course, now we have the user's "real" profiles living at `/admin/users/:id`! I figured that, for now, it'd make sense to call this view "activity" (and then later we can try to merge these together).

resources/views/admin/app.blade.php
```diff
@@ -13,8 +13,6 @@
         <link rel="apple-touch-icon-precomposed" href="/assets/images/apple-touch-icon-precomposed.png">
         <link rel="stylesheet" href="{{ elixir('admin.css', 'dist') }}">
 
-        <script src="{{ asset('dist/modernizr.js') }}"></script>
-
```
DFurnes, line 17: We used to use Modernizr for progressive enhancement, but haven't built any features that don't have overwhelmingly wide browser support now in quite a while (and are even less likely to have this concern on staff-facing admin interfaces)!

Because Webpack clears out the build directory when watching for changes, this file almost always 404s on local machines. I figured it'd be best to remove for now so it doesn't appear as a "false positive" error in the console.

resources/views/admin/users/show.blade.php
```diff
@@ -75,7 +75,10 @@
 
     <div class="container">
         <div class="wrapper">
-            <div class="container__block -half">
+            <div class="container__block -narrow">
+                <a href="/admin/activity/{{ $user->id }}">View user's signups & posts</a>
+            </div>
```
DFurnes, line 80: I moved the link to the user's signup & posts up here (since it's not really an external service anymore)!

config/features.php
```diff
@@ -38,6 +38,9 @@
     // If this is enabled, we'll expose the new "admin" routes from Aurora:
     'admin' => env('DS_ENABLE_ADMIN_ROUTES', false),
 
+    // Should we make server-side GraphQL requests?
```
weerd, line 41: Hopefully can get some cleanup/experiment time in next week to play with a solution where we wouldn't need to!

config/features.php
```diff
@@ -38,6 +38,9 @@
...
+    'graphql_php' => env('DS_ENABLE_GRAPHQL_PHP', true),
```
weerd, line 42: Hopefully can get some cleanup/experiment time in next week to play with a solution where we wouldn't need to!

app/Services/GraphQL.php
```diff
@@ -32,6 +32,12 @@ public function __construct()
      */
     public function query($query, $variables)
     {
+        // If we've disable server-side GraphQL requests, return an optional
```
weerd, line 35: Small typo:

```suggestion
        // If we've disabled server-side GraphQL requests, return an optional
```

weerd, line 35: Small typo:

```suggestion
        // If we've disabled server-side GraphQL requests, return an optional
```

resources/assets/admin/Administration.js
weerd: Ah nice solution! 👍

weerd (approved):
Looks great to me! 🎉

config/features.php
DFurnes: Oooh excited to hear more! 💡

app/Services/GraphQL.php
DFurnes: Thanks for catching that!

PR #1158, opened 2021-03-16
Subscriptions endpoint updates

weerd:
### What's this PR do?

This pull request enables passing an array of Email Newsletter subscription topics to the `/v2/subscriptions` endpoint, along with the original function of accepting just a single topic as a string.

Updating the functionality was necessary for the new Newsletter Subscription component for unauthenticated members; it lets a guest user select multiple newsletters to sign up for once they submit their email.

For new users, this will activate a password reset email so they can create a new password and the email style will depend on the first topic selected.

👀 

I added some relevant tests to ensure that the prior functionality (a single string topic) and the new functionality (an array of topics) all works as intended! 🚥 

### Relevant tickets

References [Pivotal #177168963](https://www.pivotaltracker.com/story/show/177168963).

### Checklist

### DIFF
--- a/app/Http/Controllers/SubscriptionController.php
+++ b/app/Http/Controllers/SubscriptionController.php
@@ -47,40 +47,51 @@ class SubscriptionController extends Controller
      */
     public function create(Request $request)
     {
+        // If topics value is a string, change to an array.
+        if (is_string($request->get('email_subscription_topic'))) {
+            $request->merge([
+                'email_subscription_topic' => [
+                    $request->get('email_subscription_topic'),
+                ],
+            ]);
+        }
+
         $this->validate($request, [
             'email' => 'required|email',
-            'email_subscription_topic' => [
-                'required',
+            'email_subscription_topic' => 'required|array',
+            'email_subscription_topic.*' => [
+                'string',
                 Rule::in(EmailSubscriptionTopicType::all()),
             ],
-
             'source' => 'required',
             'source_detail' => 'required',
         ]);
 
-        $topic = $request->get('email_subscription_topic');
+        $topics = $request->get('email_subscription_topic');
 
         $existingUser = $this->registrar->resolve($request->only('email'));
 
-        // If the user already exists, only update the email topics
+        // If the user already exists, only update the email topics.
         if ($existingUser) {
-            $existingUser->addEmailSubscriptionTopic($topic);
+            $existingUser->addEmailSubscriptionTopics($topics);
 
             $existingUser->save();
 
             return $this->item($existingUser, 200);
         }
 
+        // Register new user.
         $newUser = $this->registrar->register($request->all());
 
-        $newUser->addEmailSubscriptionTopic($topic);
+        $newUser->addEmailSubscriptionTopics($topics);
+
         $newUser->source = $request->get('source');
         $newUser->source_detail = $request->get('source_detail');
 
         $newUser->save();
 
         // Send activate account email to new user
-        switch ($topic) {
+        switch (collect($topics)->first()) {
             case 'scholarships':
                 $newUser->sendPasswordReset(
                     PasswordResetType::get('PAYS_TO_DO_GOOD_ACTIVATE_ACCOUNT'),
--- a/app/Http/Controllers/SubscriptionController.php
+++ b/app/Http/Controllers/SubscriptionController.php
@@ -81,7 +81,14 @@ class SubscriptionController extends Controller
         }
 
         // Register new user.
-        $newUser = $this->registrar->register($request->all());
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/SubscriptionController.php
```diff
@@ -47,40 +47,51 @@ public function __construct(
...
+
         $this->validate($request, [
             'email' => 'required|email',
-            'email_subscription_topic' => [
-                'required',
+            'email_subscription_topic' => 'required|array',
+            'email_subscription_topic.*' => [
+                'string',
                 Rule::in(EmailSubscriptionTopicType::all()),
             ],
```
weerd, line 65: This will validate each string in the array to ensure it is a valid topic.

app/Models/User.php
```diff
@@ -1005,6 +1005,7 @@ public function sendPasswordReset($type, $token = null)
      * Add the given topic to the user's array of topics if it is not already there.
      *
      * @param string $topic
+     * @deprecated
```
weerd, line 1008: Future cleanup task! Deprecating this method in favor of the new one below that allows passing an array of one or more topics.

tests/Http/SubscriptionsTest.php
```diff
@@ -2,8 +2,33 @@
 
 use App\Models\User;
 
-class SubscriptionsTest extends BrowserKitTestCase
+class SubscriptionsTest extends TestCase
```
weerd, line 5: While I was adding tests for the subscription updates, I figured I'd update the file to remove the `BrowserTestCase` in favor of `TestCase` as well! 🧼

app/Http/Controllers/SubscriptionController.php
```diff
@@ -47,40 +47,51 @@ public function __construct(
...
-            $existingUser->addEmailSubscriptionTopic($topic);
+            $existingUser->addEmailSubscriptionTopics($topics);
 
             $existingUser->save();
 
             return $this->item($existingUser, 200);
         }
 
+        // Register new user.
         $newUser = $this->registrar->register($request->all());
```
DFurnes, line 84: Hmm, I think this should either explicitly pass along the two fields we want or `$request->validated()` so that somebody can't sneak in extra unexpected fields in the mass-assigned payload here.

app/Http/Controllers/SubscriptionController.php
```diff
@@ -47,40 +47,51 @@ public function __construct(
...
+                    $request->get('email_subscription_topic'),
+                ],
+            ]);
+        }
+
         $this->validate($request, [
             'email' => 'required|email',
-            'email_subscription_topic' => [
-                'required',
+            'email_subscription_topic' => 'required|array',
```
DFurnes, line 61: Should we add `min:1` so that at least one subscription topic is required here?

app/Models/User.php
```diff
@@ -1015,6 +1016,20 @@ public function addEmailSubscriptionTopic($topic)
...
     }
 
+    /**
+     * Add one or more provided topics to the user's array of topics if not already there.
+     *
+     * @param array $topics
+     */
+    public function addEmailSubscriptionTopics($topics)
+    {
+        // Add the new topic to the existing array of topics
```
DFurnes, line 1026: This comment could probably be omitted.

tests/Http/SubscriptionsTest.php
```diff
@@ -2,8 +2,33 @@
...
+     */
+    public function testInvalidSubscriptionNotAddedToUser()
+    {
+        $user = factory(User::class)->create([
+            'email_subscription_topics' => [],
+        ]);
+
+        $response = $this->json('POST', 'v2/subscriptions', [
+            'email' => $user->email,
+            'email_subscription_topic' => 'puppetslothtips',
```
DFurnes, line 18: How _isn't_ this an email newsletter we offer?!

DFurnes (approved):
Looks great! 📧

app/Http/Controllers/SubscriptionController.php
weerd: I tested this out and it will not validate with an empty array if a value is required. Seems like Laravel knows that an empty array is not valid!

app/Http/Controllers/SubscriptionController.php
weerd: Ah ya. This was existing code but defs a good fix!

app/Http/Controllers/SubscriptionController.php
weerd: Just to be safe and make sure this is always the case and doesn't break on a Laravel upgrade, I'm adding a test to ensure this is always the case!

PR #1160, opened 2021-03-17
Questionnaire Endpoint

mendelB:
### What's this PR do?

This pull request adds an `/api/v3/questionnaires` endpoint to enable POSTing user submissions for the new **Questionnaire Action** from Phoenix (https://github.com/DoSomething/phoenix-next/pull/2571).

commit-by-commit if preferred.

Do you think this implementation is sound?

A Questionnaire is composed of multiple questions, each tied to a separate Action ID. Our goal is to store individual Posts per question submission.

The Questionnaire endpoint enables us to accept specific questionnaire parameters which we use to create the individual Posts (& corresponding signups), and decorate the `details` with some questionnaire-specific information.

Additionally, we only want to track _one_ customer.io event per groups of Questionnaire posts, which is tackled in https://github.com/DoSomething/northstar/commit/28a6421ecf61d38a8ec619fb1ceb148b862bf5b8 via a new parameter to the `PostManager->create` method to specify tracking the customer.io event for created posts.

### Relevant tickets

References [Pivotal #177321999](https://www.pivotaltracker.com/story/show/177321999).

### Checklist

### DIFF
--- a/app/Http/Controllers/QuestionnaireController.php
+++ b/app/Http/Controllers/QuestionnaireController.php
@@ -75,7 +75,7 @@ class QuestionnaireController extends ActivityApiController
      */
     public function store(Request $request)
     {
-        $this->validate($request, $this->rules);
+        $validated = $this->validate($request, $this->rules);
 
         $northstarId = getNorthstarId($request);
 
@@ -94,7 +94,7 @@ class QuestionnaireController extends ActivityApiController
 
             if (!$signup) {
                 $signup = $this->signups->create(
-                    $request->all(),
+                    $validated,
                     $northstarId,
                     $campaignId,
                 );
@@ -108,7 +108,7 @@ class QuestionnaireController extends ActivityApiController
 
             $shouldTrackToCustomerIo = $index === 0;
 
-            $post = $this->posts->create(array_merge($request->all(), [
+            $post = $this->posts->create(array_merge($validated, [
                 'action_id' => $actionId,
                 'text' => $question['answer'],
                 'details' => json_encode($postDetails),
--- a/app/Http/Controllers/QuestionnaireController.php
+++ b/app/Http/Controllers/QuestionnaireController.php
@@ -1,0 +1,121 @@
+<?php
+
+namespace App\Http\Controllers;
+
+use App\Http\Requests\PostRequest;
+use App\Http\Transformers\PostTransformer;
+use App\Managers\PostManager;
+use App\Managers\SignupManager;
+use App\Models\Campaign;
+use App\Models\Post;
+use Illuminate\Http\Request;
+use Illuminate\Support\Arr;
+
+class QuestionnaireController extends ActivityApiController
+{
+    /**
+     * The post manager instance.
+     *
+     * @var \App\Managers\PostManager
+     */
+    protected $posts;
+
+    /**
+     * The SignupManager instance.
+     *
+     * @var \App\Managers\SignupManager
+     */
+    protected $signups;
+
+    /**
+     * @var \App\Http\Transformers\PostTransformer;
+     */
+    protected $transformer;
+
+    /**
+     * Create a controller instance.
+     *
+     * @param PostManager $posts
+     * @param SignupManager $signups
+     * @param PostTransformer $transformer
+     */
+    public function __construct(
+        PostManager $posts,
+        SignupManager $signups,
+        PostTransformer $transformer
+    ) {
+        $this->posts = $posts;
+        $this->signups = $signups;
+        $this->transformer = $transformer;
+
+        $this->middleware('auth:api');
+
+        $this->middleware('scope:activity');
+        $this->middleware('scope:write');
+
+        $this->rules = array_merge(
+            Arr::except((new PostRequest())->rules(), [
+                'action', 'action_id', 'campaign_id', 'type',
+            ]),
+            [
+                'questions' => 'required|array',
+                'questions.*.action_id' => 'required|integer|exists:mysql.actions,id',
+                'questions.*.title' => 'required|string',
+                'questions.*.answer' => 'required|string|max:500',
+                'contentful_id' => 'required|string',
+            ]
+        );
+    }
+
+    /**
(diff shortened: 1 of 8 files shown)


app/Http/Controllers/QuestionnaireController.php
```diff
@@ -0,0 +1,123 @@
...
+
+        $this->middleware('auth:api');
+
+        $this->middleware('scope:activity');
+        $this->middleware('scope:write');
+
+        $this->rules = array_merge(
+            Arr::except((new PostRequest())->rules(), [
+                'action', 'action_id', 'campaign_id', 'type',
+            ]),
```
mendelB, line 59: The `action_id`s are required as nested properties in each individual question. (Which means we don't need the `action` or `campaign_id`. Plus the `type` will be determined from the `Action`).

app/Http/Controllers/QuestionnaireController.php
```diff
@@ -0,0 +1,123 @@
...
+            $campaignId = Campaign::fromActionId($actionId)->id;
+            $signup = $this->signups->get($northstarId, $campaignId);
+
+            if (!$signup) {
+                $signup = $this->signups->create(
+                    $request->all(),
+                    $northstarId,
+                    $campaignId,
+                );
+            }
```
mendelB, line 101: We'll do this for each individual post to be safe, but realistically, they should all be associated with individual Actions of the same campaign which means we'll only need to store one signup per Questionnaire.

app/Http/Controllers/QuestionnaireController.php
```diff
@@ -0,0 +1,123 @@
...
+        foreach ($questions as $index => $question) {
+            $actionId = $question['action_id'];
+
+            // Get the campaign id from the request by action_id.
+            $campaignId = Campaign::fromActionId($actionId)->id;
+            $signup = $this->signups->get($northstarId, $campaignId);
+
+            if (!$signup) {
+                $signup = $this->signups->create(
+                    $request->all(),
```
weerd, line 97: Is there an alternative to `$request->all()`? It's just always a little scary to just throw all the fields into the DB. Would `$request->validated()` work if available?

weerd, line 97: Is there an alternative to `$request->all()`? It's just always a little scary to just throw all the fields into the DB. Would `$request->validated()` work if available?

app/Http/Controllers/QuestionnaireController.php
```diff
@@ -0,0 +1,123 @@
...
+
+            $postDetails = array_merge($requestDetails, [
+                'questionnaire' => true,
+                'question' => $question['title'],
+                'contentful_id' => $request->input('contentful_id'),
+            ]);
+
+            $shouldTrackToCustomerIo = $index === 0;
+
+            $post = $this->posts->create(array_merge($request->all(), [
```
weerd, line 111: Same with question above. Would `$request->validated()` work?

tests/Http/QuestionnaireTest.php
```diff
@@ -0,0 +1,200 @@
...
+     * @return void
+     */
+    public function testCreatingQuestionnairePostsAndSignups()
+    {
+        $user = factory(User::class)->create();
+        $referrerUser = factory(User::class)->create();
+
+        $campaignId = factory(Campaign::class)->create()->id;
+        $location = 'US-' . $this->faker->stateAbbr();
+        $school_id = $this->faker->word;
```
weerd, line 26: Not sure if it matters, but would it be better to use the `FakerSchoolId` class to pull from the the csv of fake schools?

app/Repositories/PostRepository.php
```diff
@@ -46,15 +46,11 @@ public function find($id)
...
-     * @param  string $authenticatedUserRole
      *
      * @return \App\Models\Post|null
      */
-    public function create(
-        array $data,
-        $signupId,
-        $authenticatedUserRole = null
-    ) {
+    public function create(array $data, $signupId)
```
weerd, line 52: On purpose removal of `$authenticatedUserRole`?

weerd: Oop! Just saw the changes in the `PostManager`! 😅

weerd (approved):
Looks good! Just a couple questions 👍

app/Http/Controllers/QuestionnaireController.php
mendelB: That seems way safer, and I feel like that should totally work here!

app/Repositories/PostRepository.php
mendelB: It seems to be a relic of way past and unused, so I just went ahead and removed it 🗑️

app/Http/Controllers/QuestionnaireController.php
mendelB: I ended up just using the return value of `$this->validate...` (for some reason `$request->validated()` was erroring out as an undefined method) in https://github.com/DoSomething/northstar/pull/1160/commits/fd45efb0c4ef2abb75eeb0487b33732dd2be42de -- and then tests failed since the `group_id` wasn't being validated! Updated in https://github.com/DoSomething/northstar/pull/1160/commits/c774d32c671ccbf96192d1bf76c2f9dfadb88377.

weerd: Ya, I ran into `$request->valdiated()` erroring out as well when updating the subscriptions endpoint, but I didn't really dig into it so not 100% sure. Still curious and I'll report back here 📸  if I end up finding out!

mendelB: I wonder if it only works on custom "form requests" per [the docs](https://laravel.com/docs/7.x/validation#form-request-validation) 🤔 (the Questionnaire is just using a standard `Request` -- was that the same in your case?)

weerd: Ah ya, I think that's the conclusion I was drawing as well! Glad to see you found info to confirm it! Good 🔍  @mendelB!

PR #1162, opened 2021-03-18
Adds Four Posts Badge!

lindsayrmaher:
### What's this PR do?

This pull request adds one additional badge for when a user takes a fourth action, since there should be 6 badges "visible" to users and we reference 6 badges in our communications about it.

👀 

This was meant to be added in to replace the voter registration badge but we forgot to prioritize!!

### Relevant tickets

References [Pivotal #177399018](https://www.pivotaltracker.com/story/show/177399018).

### Checklist

### DIFF
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -486,15 +486,12 @@ class Post extends Model
                 ->count();
 
             if ($userPostsWithFaveTagCount >= 3) {
-                $this->user->addBadge(BadgeType::get('ONE_STAFF_FAVE'));
-                $this->user->addBadge(BadgeType::get('TWO_STAFF_FAVES'));
                 $this->user->addBadge(BadgeType::get('THREE_STAFF_FAVES'));
             }
-            if ($userPostsWithFaveTagCount === 2) {
-                $this->user->addBadge(BadgeType::get('ONE_STAFF_FAVE'));
+            if ($userPostsWithFaveTagCount >= 2) {
                 $this->user->addBadge(BadgeType::get('TWO_STAFF_FAVES'));
             }
-            if ($userPostsWithFaveTagCount === 1) {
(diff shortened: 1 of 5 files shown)


app/Models/Post.php
```diff
@@ -787,7 +787,13 @@ public function calculatePostBadges()
...
+            if ($userPostsCount >= 4) {
+                $user->addBadge(BadgeType::get('ONE_POST'));
+                $user->addBadge(BadgeType::get('TWO_POSTS'));
+                $user->addBadge(BadgeType::get('THREE_POSTS'));
+                $user->addBadge(BadgeType::get('FOUR_POSTS'));
+            }
+            if ($userPostsCount === 3) {
                 $user->addBadge(BadgeType::get('ONE_POST'));
                 $user->addBadge(BadgeType::get('TWO_POSTS'));
                 $user->addBadge(BadgeType::get('THREE_POSTS'));
```
DFurnes, line 799: What do you think about changing these to `>=` so we can de-duplicate adding the 1-3 post badges?

```php
if ($userPostsCount >= 4) {
    $user->addBadge(BadgeType::get('FOUR_POSTS'));
}

if ($userPostsCount >= 3) {
    $user->addBadge(BadgeType::get('THREE_POSTS'));
}

if ($userPostsCount >= 2) {
    $user->addBadge(BadgeType::get('TWO_POSTS'));
}

if ($userPostsCount >= 1) {
    $user->addBadge(BadgeType::get('ONE_POST'));
}
```

DFurnes (approved):
🍀

app/Models/Post.php
lindsayrmaher: great idea ➕ I should be able to do for tag badges as well.

DFurnes (approved):
🎉

PR #1167, opened 2021-03-19
Add search functionality to user admin index.

DFurnes:
### What's this PR do?

This pull request adds support for searching for a user in "Roguestar" admin.

I extracted the logic we use to recognize if something is a phone number or email in ae31642 & ce86f54. Using those helpers, I created a `search` query scope that infers what term is given & queries the appropriate field in ba0acb9.

Finally, I added the search form & controller support (as a simple `?search=` query on the index) in ed14d86.

I hadn't moved the "search" interface from Aurora into this app in [#176850876](https://www.pivotaltracker.com/story/show/176850876). This is an important use-case for Aurora and Rogue, and something we'll need to support before we can turn off the old admin interfaces!

### Relevant tickets

### Checklist

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -214,7 +214,7 @@ function format_birthdate($birthdate)
 }
 
 /**
- * Is the given value an email address?
+ * Is the given value a phone number?
  *
  * @param string $value
  * @return bool
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -214,6 +214,43 @@ function format_birthdate($birthdate)
 }
 
 /**
+ * Is the given value an email address?
+ *
+ * @param string $value
+ * @return bool
+ */
+function is_phone_number(?string $value): bool
+{
+    if (!$value) {
+        return false;
+    }
+
+    $parser = PhoneNumberUtil::getInstance();
+
+    try {
+        // Make sure that libphonenumber can parse this phone.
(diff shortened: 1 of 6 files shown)


app/helpers.php
```diff
@@ -213,6 +213,58 @@ function format_birthdate($birthdate)
     return format_date($birthdate, 'Y-m-d');
 }
 
+/**
+ * Is the given value an email address?
```
mendelB, line 217:
```suggestion
 * Is the given value a phone number?
```

👾

app/helpers.php
```diff
@@ -213,6 +213,58 @@ function format_birthdate($birthdate)
...
+        );
+
+        return !empty($valid) && empty($repeat);
+    } catch (\libphonenumber\NumberParseException $e) {
+        return false;
+    }
+}
+
+/**
+ * Is the given value a phone number?
```
mendelB, line 254:
```suggestion
 * Is the given value an email address?
```

mendelB (approved):
💥

app/helpers.php
DFurnes: Oh my god, how did I get them swapped?! 😂

mendelB: Email is the new SMS! (so Anthony claims 🙃)

PR #1169, opened 2021-03-22
Fix issues with Slack tag notifications.

DFurnes:
### What's this PR do?

This pull request fixes an issue that was preventing us from sending Slack notifications when a post is tagged as "Good Submission", "Good Quote", et cetera. I've also updated this job to read user information directly from the database, rather than querying via GraphQL & done a little extra tidying up. 🍃 

I'd recommend reviewing commit-by-commit. The last commit is a grab-bag to reformat for readability!

💬

### Relevant tickets

References [Pivotal #177437948](https://www.pivotaltracker.com/story/show/177437948).

### Checklist

### DIFF
--- a/app/Notifications/PostTagged.php
+++ b/app/Notifications/PostTagged.php
@@ -35,7 +35,7 @@ class PostTagged extends Notification implements ShouldQueue
     public $admin;
 
     /*
-     * OLD: The ID of the amdin who tagged this post.
+     * OLD: The ID of the admin who tagged this post.
      *
      * @var string
      */
--- a/app/Notifications/PostTagged.php
+++ b/app/Notifications/PostTagged.php
@@ -106,11 +106,7 @@ class PostTagged extends Notification implements ShouldQueue
                     '":',
             )
             ->attachment(function ($attachment) use ($userName) {
-                $permalink = route(
-                    'signups.show',
-                    [$this->post->signup_id],
-                    true,
-                );
+                $permalink = url("/admin/posts/{$this->post->id}");
                 $image = $this->post->getMediaUrl();
 
                 $attachment
--- a/app/Notifications/PostTagged.php
+++ b/app/Notifications/PostTagged.php
@@ -30,7 +30,14 @@ class PostTagged extends Notification implements ShouldQueue
     /*
      * The admin who tagged this post.
      *
-     * @var string;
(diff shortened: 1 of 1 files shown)


app/Notifications/PostTagged.php
```diff
@@ -55,24 +62,13 @@ class PostTagged extends Notification implements ShouldQueue
      */
     public function __construct(Post $post, Tag $tag)
     {
-        $this->adminId = auth()->id();
+        $this->admin = auth()->user();
```
DFurnes, line 65: These previously just stored the ID, since we were living in another service where the user wasn't easily accessible. Now that this job lives in Northstar, we can just serialize the actual user model here!

app/Notifications/PostTagged.php
```diff
@@ -92,47 +88,32 @@ public function via($notifiable)
...
     {
-        $data = $this->queryForUserAndReviewer();
-
-        $userName = $data['user']['displayName'];
-        $adminName = $data['admin']['displayName'];
+        // TEMPORARY: If we'd previously serialized this job with a
+        // string 'adminId', turn this into a full admin user model:
+        if (isset($this->adminId)) {
+            $this->admin = User::find($this->adminId);
+        }
```
DFurnes, line 95: Laravel serializes these notifications as queued jobs, and so a bunch of these are sitting in `failed_jobs` to be retried (with an `adminId` string parameter). This allows those to be processed as if they'd be serialized with a user model.

app/Notifications/PostTagged.php
```diff
@@ -92,47 +88,32 @@ public function via($notifiable)
...
                         'Why Participated' => Str::limit(
                             $this->post->signup->why_participated,
                         ),
-                    ])
-                    ->color(Arr::random(['#fcd116', '#23b7fb', '#4e2b63']));
-
-                // Do not send images with local URL to Slack (ie: http://rogue.test/images/filename)
-                if ($image && config('app.env') !== 'local') {
-                    $attachment->image($image);
-                }
```
DFurnes, line 135: This feels like an edge-case (as we don't send these from our local environments usually). I've also tested this from my local without this condition and confirmed that it doesn't error (anymore) - it just doesn't show an image!

weerd (approved):
Awesome! 😄

PR #1171, opened 2021-03-24
Add Backfill Method for Badges

lindsayrmaher:
### What's this PR do?

This pull request creates a method for users that have recently been updated to double check they have all the correct badges assigned to them based on activity. It also adds a test to make sure this works as expected.

Any other cases that seem necessary for tests?

This is our solution to make sure all existing and active users have their accounts updated properly with badges they've earned for past activity.

### Relevant tickets

References [Pivotal #177327500](https://www.pivotaltracker.com/story/show/177327500).

### Checklist

### DIFF
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -478,25 +478,7 @@ class Post extends Model
             }
         }
 
-        if ($this->user) {
-            $userPostsWithFaveTagCount = $this->user
-                ->posts()
-                ->withTag('good-submission')
-                ->take(3)
-                ->count();
-
-            if ($userPostsWithFaveTagCount >= 3) {
-                $this->user->addBadge(BadgeType::get('THREE_STAFF_FAVES'));
-            }
-            if ($userPostsWithFaveTagCount >= 2) {
-                $this->user->addBadge(BadgeType::get('TWO_STAFF_FAVES'));
-            }
-            if ($userPostsWithFaveTagCount >= 1) {
-                $this->user->addBadge(BadgeType::get('ONE_STAFF_FAVE'));
-            }
-
-            $this->user->save();
-        }
+        $this->calculateStaffFaveTagBadges();
 
         return $this;
     }
@@ -796,6 +778,33 @@ class Post extends Model
             if ($userPostsCount >= 1) {
                 $user->addBadge(BadgeType::get('ONE_POST'));
             }
+            $user->save();
+        }
+    }
+
+    /**
+     * Checks whether a user should be given a badge based on their post.
+     */
+    public function calculateStaffFaveTagBadges()
+    {
+        $user = $this->user;
+        if ($user) {
+            $userPostsWithFaveTagCount = $user
+                ->posts()
+                ->withTag('good-submission')
+                ->take(3)
+                ->count();
+
+            if ($userPostsWithFaveTagCount >= 3) {
+                $user->addBadge(BadgeType::get('THREE_STAFF_FAVES'));
+            }
+            if ($userPostsWithFaveTagCount >= 2) {
+                $user->addBadge(BadgeType::get('TWO_STAFF_FAVES'));
+            }
+            if ($userPostsWithFaveTagCount >= 1) {
+                $user->addBadge(BadgeType::get('ONE_STAFF_FAVE'));
+            }
+
             $user->save();
         }
     }
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -783,7 +783,8 @@ class Post extends Model
     }
 
     /**
-     * Checks whether a user should be given a badge based on their post.
+     * Checks whether a user should be given a badge based on if their post has been tagged
+     * a good-submission by staff.
      */
     public function calculateStaffFaveTagBadges()
     {
--- a/app/Models/Signup.php
+++ b/app/Models/Signup.php
@@ -304,8 +304,8 @@ class Signup extends Model
     {
         $user = $this->user;
         if ($user) {
-            $userSignups = $user->signups();
-            if ($userSignups->count() === 1) {
+            $userSignups = $user->signups()->count();
+            if ($userSignups > 0) {
                 $user->addBadge(BadgeType::get('SIGNUP'));
                 $user->save();
             }
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -10,6 +10,7 @@ use App\Types\BadgeType;
 use App\Types\PasswordResetType;
 use Carbon\Carbon;
 use Email\Parse as EmailParser;
+use Facade\Ignition\DumpRecorder\Dump;
 use Illuminate\Auth\Authenticatable;
 use Illuminate\Auth\Passwords\CanResetPassword;
 use Illuminate\Contracts\Auth\Access\Authorizable as AuthorizableContract;
@@ -1320,6 +1321,30 @@ class User extends MongoModel implements
         ) {
             $this->addBadge(BadgeType::get('NEWS_SUBSCRIPTION'));
             $this->save();
+        }
+    }
+
+    /**
+     * Runs our badges calculators to be sure active users have the correct badges earned
+     */
+    public function backfillBadges()
+    {
+        //check for a news email subscription
+        $this->calculateUserSubscriptionBadges();
+
+        //check for a successful campaign signup
+        if ($this->signups()->count()) {
+            $this->signups()
+                ->first()
+                ->calculateSignupBadges();
+        }
+
+        //check for completed posts and good-submission tags
+        if ($this->posts()->count()) {
+            $firstUserPost = $this->posts()->first();
+
+            $firstUserPost->calculatePostBadges();
+            $firstUserPost->calculateStaffFaveTagBadges();
         }
     }
 }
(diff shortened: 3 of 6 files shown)


app/Models/Signup.php
```diff
@@ -304,8 +304,8 @@ public function calculateSignupBadges()
     {
         $user = $this->user;
         if ($user) {
-            $userSignups = $user->signups();
-            if ($userSignups->count() === 1) {
+            $userSignups = $user->signups()->count();
+            if ($userSignups > 0) {
```
katiecrane, line 308: I looked at the `addBadge` function a little more carefully and I actually don't think it handles de-duping the badges. I think we should handle that in `addBadge` so we don't have to worry about it elsewhere and that way this logic is still good!

katiecrane:
looks like tests are failing because there is no `rogue_test` database! however, that database does seem to get setup in the `config.yml` on circleci so I'm tagging in @DFurnes to see if he has any ideas since I see this is happening even when we run it again

app/Models/Signup.php
lindsayrmaher: Would the setter take care of it? We are deduping there. https://github.com/DoSomething/northstar/blob/309a1fa9d04a003bf8b5458b480c2b8e4b8b9d68/app/Models/User.php#L1253-L1262

katiecrane: Ah yeah should be good then! I was thrown by the comment with `addBadge` of
>Add the given badge to the user's array of badges if it is not already there

since there's nothing specific in that function to de-dupe, though I suppose that is not incorrect to say since that is what is happening behind the scenes!

app/Models/User.php
```diff
@@ -10,6 +10,7 @@
 use App\Types\PasswordResetType;
 use Carbon\Carbon;
 use Email\Parse as EmailParser;
+use Facade\Ignition\DumpRecorder\Dump;
```
aaronschachter, line 13: Is this used anywhere?

lindsayrmaher: it doesn't look like it no..

app/Models/Post.php
```diff
@@ -799,4 +781,31 @@ public function calculatePostBadges()
             $user->save();
         }
     }
+
+    /**
+     * Checks whether a user should be given a badge based on their post.
```
weerd, line 786: Not a deal breaker, but noting that the description for the method was confusing in relation to the method name and what it does. Maybe specify that it gives badge based on number of post tagged as staff favorite or something like that?

weerd:
> looks like tests are failing because there is no rogue_test database! however, that database does seem to get setup in the config.yml on circleci so I'm tagging in @DFurnes to see if he has any ideas since I see this is happening even when we run it again

So there's definitely something up because it should be using the Mongo test DB and not the Rogue one... I can try and help troubleshoot after I get my other PR submitted!

weerd:
I pulled down the branch and ran PHPUnit on my local and all green. Something's likely up w/ CircleCI or the CircleCI setup, but nothing that's immediately obvious to me. Still 👁️-ing...

katiecrane:
> > looks like tests are failing because there is no rogue_test database! however, that database does seem to get setup in the config.yml on circleci so I'm tagging in @DFurnes to see if he has any ideas since I see this is happening even when we run it again
> 
> So there's definitely something up because it should be using the Mongo test DB and not the Rogue one... I can try and help troubleshoot after I get my other PR submitted!

ah I think it _is_ supposed to hit the Rogue one in these cases because an update happens on the user, at which time we make sure their badges are up to date, which is why it's trying to get the signups:

> CommunityTopicBackfillTest::it_should_add_community_to_subscribed_users
Illuminate\Database\QueryException: SQLSTATE[42S02]: Base table or view not found: 1146 Table 'rogue_test.signups' doesn't exist (SQL: select count(*) as aggregate from `signups` where `signups`.`northstar_id` = 605b8bbdb685d72c727d309a and `signups`.`northstar_id` is not null and `signups`.`deleted_at` is null)

katiecrane:
But it's not failing on any of the actual Rogue tests so that's confusing!

weerd:
Ya, agreed. And it's also failing on tests with `signups` table aside from this particular test... I just don't get why the table would be considered missing.

weerd:
I might try digging in a bit more in a bit and SSH into the CircleCI box... 🤔

weerd:
Ok, still not really sure what's up, BUT I did SSH into the Circle CI box and ran the specific PHPUnit test from @lindsayrmaher's PR and got green... so the isolated test is fine, it's just when it runs in conjunction with rest of the suite, and possibly something else that just broke and is now just being highlighted in this code update! Lucky you Lindsay! 🍀

wtf?!

lindsayrmaher:
> Ok, still not really sure what's up, BUT I did SSH into the Circle CI box and ran the specific PHPUnit test from @lindsayrmaher's PR and got green... so the isolated test is fine, it's just when it runs in conjunction with rest of the suite, and possibly something else that just broke and is now just being highlighted in this code update! Lucky you Lindsay! 🍀

feeling blessed 😌 

well i guess more investigating needs to be done today 🕵️‍♀️

DFurnes:
Sorry, I've been swamped with meetings the past two days! I've blocked off time tomorrow to dig into this.

lindsayrmaher:
> Sorry, I've been swamped with meetings the past two days! I've blocked off time tomorrow to dig into this.

hooray! I would be down to pair if that works, or at the very least sort of hear a walk through if you pin point what's going on.

DFurnes:
This is monumentally spooky! It's interesting that the exact same tests have failed on each build [for this branch](https://app.circleci.com/pipelines/github/DoSomething/northstar?branch=add-backfill-method-for-badges) on CI, but everything looks fine when running tests locally or by SSH'ing into the CircleCI box! 🤔 Just to be sure, I also tried rerunning the builds for [the latest commit on `main`](https://app.circleci.com/pipelines/github/DoSomething/northstar/1058/workflows/ea8548b6-3c31-4c34-91a7-9f571f36fbd1/jobs/1103) and [the commit `a1fe2067` this PR was branched from](https://app.circleci.com/pipelines/github/DoSomething/northstar/1061/workflows/d42ec58c-7c65-49df-922d-17fadec6d39a/jobs/1104) & both pass. So, whatever's going on _somehow_ boils down to something in this PR (even if the underlying issue has been creeping about for longer).

It seems like each failure has this part of the stack trace in common, [here](https://github.com/DoSomething/northstar/blob/e3de6b60d74e9f2760978b8b9fb78468b50d9fbb/app/Models/User.php#L1335) and [here](https://github.com/DoSomething/northstar/blob/e3de6b60d74e9f2760978b8b9fb78468b50d9fbb/app/Observers/UserObserver.php#L190):

```
/home/circleci/project/app/Models/User.php:1335
/home/circleci/project/app/Observers/UserObserver.php:190
```

So, for some reason, it's not happy using that MySQL `signups` relationship query builder on the user...

DFurnes:
We certainly read/write to that signups table directly (or via post's relationship) elsewhere, and even read/write to it via a `$user->signups` relationship [here](https://github.com/DoSomething/northstar/blob/309a1fa9d04a003bf8b5458b480c2b8e4b8b9d68/app/Observers/UserObserver.php#L274) as part of our user deletion tests... I'm wondering if this is somehow unique to using the relationship `$user->signup()->…` builder?

DFurnes:
Sidenote: this makes me _really_ want to get these tests running faster! Four minutes feels like an _eternity_ when I'm running these tests back-to-back! 🐌 Re-evaluating how I spend my upcoming cleanup sprint time... 😂

DFurnes:
Alright, I was able to reproduce these failures on CircleCI by SSH'ing in, dropping the MySQL database, and re-running the test suite. I get the same `rogue.tests` failures & all subsequent runs pass (until I drop the database again).

That makes me think that there's some sort of order-of-operations issue with how we're migrating databases in tests.

DFurnes:
A ha!! We hadn't applied the `RefreshDatabase` test to our `BrowserKitTestCase` class (which is used by some of our older tests, like `DrupalPasswordHashTest`. Without this trait, we would only migrate the MongoDB database (since that's handled in `CreatesApplication@setUp` [here](https://github.com/DoSomething/northstar/blob/309a1fa9d04a003bf8b5458b480c2b8e4b8b9d68/tests/CreatesApplication.php#L103-L109)). 

Once we ran a "newer" `TestCase` test (where we _do_ use that trait), we'd migrate the database & any subsequent `BrowserKitTestCase` runs would be happy since they'd be able to use that previously migrated database.

DFurnes:
I think that 9324762 should get thing in order! Rogue's tests all use Laravel's newer built-in `TestCase` class, and so until this PR we'd never interacted with the MySQL database from any of the `BrowserKitTestCase` tests. Tricky!

weerd:
> Sidenote: this makes me really want to get these tests running faster! Four minutes feels like an eternity when I'm running these tests back-to-back! 🐌 Re-evaluating how I spend my upcoming cleanup sprint time... 😂

We can do parallel testing! Juuuuuuuust need to upgrade from v6.x to v8.x 😅

weerd (approved):
Woot woot! Get them badges! 🏅

weerd:
Nice work @DFurnes sleuthing what was going on! 👻 🕵️‍♂️

PR #1173, opened 2021-04-05
Add client administration interface.

DFurnes:
### What's this PR do?

This pull request moves the OAuth Clients administration interface from [Aurora](https://github.com/DoSomething/aurora) over into this application. This allows developers to create, edit, and view OAuth clients now that we've retired Aurora.

I've broken this work into separate commits:

💁‍♂️ I added some remaining helpers from Aurora that were used in this flow in 8397119

👀 I added views & assets in 80d73f4. These are slightly modified from [their original counterparts](https://github.com/DoSomething/aurora/blob/main/resources/views/clients/edit.blade.php) since we don't use the `{!! Form::` helpers in this application. Where it made sense, I used our existing form partials to build these forms.

🤖 I added the routes & controller in e5c7520. This involved merging logic from Aurora's [ClientController](https://github.com/DoSomething/aurora/blob/main/app/Http/Controllers/ClientController.php) and Northstar's API [ClientController](https://github.com/DoSomething/northstar/blob/a6b9ef4cdc1a24f2fd635813a0867fbb3bd9edcf/app/Http/Controllers/ClientController.php) to ensure we kept the same behavior.

🔗 I added links to this new section to the homepage & navigation bar in 30c4a24.

🥡 Finally, I updated the `Client` model to always return an array for `redirect_uri` in 6a29a5d. This is something we'd previously handled in the transformer, but now that this isn't hitting our API it makes sense to move into the model!

🔏

### Relevant tickets

References [Pivotal #176850872](https://www.pivotaltracker.com/story/show/176850872).

### Checklist


app/Http/Controllers/Web/Admin/ClientsController.php
```diff
@@ -0,0 +1,181 @@
+<?php
+
+namespace App\Http\Controllers\Web\Admin;
+
+use App\Auth\Scope;
+use App\Http\Controllers\Controller;
+use App\Models\Client;
+use Illuminate\Http\Request;
+
+class ClientsController extends Controller
```
DFurnes, line 10: I'd recommend looking at this side-by-side with Aurora's [ClientController](https://github.com/DoSomething/aurora/blob/main/app/Http/Controllers/ClientController.php) and Northstar's API [ClientController].

resources/views/admin/layouts/navigation.blade.php
```diff
@@ -49,16 +49,12 @@
                         </a>
                     </li>
                 @endif
-
-                <li>
-                    <a href="{{ url('/admin/faq') }}">
-                        <strong class="navigation__title">FAQ</strong>
-                        <span class="navigation__subtitle">How do I...</span>
-                    </a>
-                </li>
```
DFurnes, line 58: I moved this over to the "secondary" navigation since it was otherwise awkwardly wrapping to a second line.

lindsayrmaher (approved):
I'm not seeing anything jumping out as an issue! GTG

lindsayrmaher:
It does look like some tests need updating though 🤔

DFurnes:
Ooof, it sure does! 👀

DFurnes:
A ha! After using `$this->withoutExceptionHandling()` to print the 500 we were receiving on this test's HTTP request:

```
Symfony\Component\Debug\Exception\FatalThrowableError: Argument 1 passed to League\OAuth2\Server\Grant\AbstractGrant::validateRedirectUri() must be of the type string, array given, called in /home/vagrant/Sites/northstar/vendor/league/oauth2-server/src/Grant/AuthCodeGrant.php on line 263

/home/vagrant/Sites/northstar/vendor/league/oauth2-server/src/Grant/AbstractGrant.php:260
/home/vagrant/Sites/northstar/vendor/league/oauth2-server/src/Grant/AuthCodeGrant.php:263
/home/vagrant/Sites/northstar/vendor/league/oauth2-server/src/AuthorizationServer.php:157
/home/vagrant/Sites/northstar/app/Http/Controllers/Web/AuthController.php:65
```

Since the mutator now always returns an array for `$client->client_secret`, we have to pick the "first" one in this test. Otherwise we end up serializing an array (e.g. `&redirect_uri[0]=…`) into this URL and that's not what we want.

Fixed up in d7bd5cb & we're back to green! 🟢

PR #1174, opened 2021-04-06
Dimension Validation Message on the Post file field

mendelB:
### What's this PR do?

This pull request adds a custom validation message for `dimension` errors for the Post `file` field. 

👀 

We'd like to have clearer validation messaging for users reporting back on our website. We're overwriting the standard error message (which was missing from this config in the first place.)

(This is technically global, but we only enforce these dimension validations on the `file` field for Posts. If we add this for some other purpose we can configure this more locally.)

See https://github.com/DoSomething/phoenix-next/pull/2595 & https://github.com/DoSomething/phoenix-next/pull/2595#issuecomment-814218762

### Relevant tickets

References [Pivotal #177189106](https://www.pivotaltracker.com/story/show/177189106).

### Checklist

### DIFF
--- a/resources/lang/en/validation.php
+++ b/resources/lang/en/validation.php
@@ -34,8 +34,6 @@ return [
     'different' => 'The :attribute and :other must be different.',
     'digits' => 'The :attribute must be :digits digits.',
     'digits_between' => 'The :attribute must be between :min and :max digits.',
-    'dimensions' => "Photos must be no larger than 10MB, at least :min_width x :min_height, and no larger
-          than :max_width x :max_height. Try cropping your photo.",
     'email' => 'The :attribute must be a valid email address.',
     'filled' => 'The :attribute field is required.',
     'exists' => 'The selected :attribute is invalid.',
--- a/resources/lang/en/validation.php
+++ b/resources/lang/en/validation.php
@@ -1,8 +1,5 @@
 <?php
 
-$minImageSize = config('posts.image.min');
-$maxImageSize = config('posts.image.max');
-
 return [
     /*
     |--------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


resources/lang/en/validation.php
```diff
@@ -34,6 +37,8 @@
     'different' => 'The :attribute and :other must be different.',
     'digits' => 'The :attribute must be :digits digits.',
     'digits_between' => 'The :attribute must be between :min and :max digits.',
+    'dimensions' => "Photos must be no larger than 10MB, at least {$minImageSize['width']} x {$minImageSize['height']}, and no larger
+          than {$maxImageSize['width']} x {$maxImageSize['height']}. Try cropping your photo.",
```
DFurnes, line 41: Are you able to use a placeholder to load these from the validator, like `at least :min_width x :min_height`? It looks like that's supported for some other customizable validation messages, and might read a little more cleanly.

DFurnes: I'm also wondering, though, if it'd make more sense to define this in the `PostRequest`, like [this](https://laravel.com/docs/8.x/validation#customizing-the-error-messages). That way, if we ever defined a different `dimensions` validator (say, for uploading avatars or whatnot), we could apply different image validation rules & error messaging in each place...

DFurnes: And regardless, I think it's definitely smart to add language for this here! If we end up customizing this more specifically for this endpoint, we can still add [more general fallback language](https://github.com/laravel/laravel/blob/5808129a1f702f973c7c31203d16db2066bd9030/resources/lang/en/validation.php#L40) for this validator in here.

mendelB: Smarttttt! Updated in https://github.com/DoSomething/northstar/pull/1174/commits/c2deffb11d941e83aea8cba1ccc8e995908c5ad1, https://github.com/DoSomething/northstar/pull/1174/commits/e491bd9400f11e9dac856a713fe21772e3e5175e, and https://github.com/DoSomething/northstar/pull/1174/commits/777ed66f8207c8ae77902a46547dc42c0a473196! (Yay for the test making this easy to play with! 😌 )

DFurnes (approved):
Fantastic! 💯

PR #1175, opened 2021-04-06
Removes now-unused 'v2/clients' API.

DFurnes:
### What's this PR do?

This pull request removes the `v2/clients` API, which allowed applications to read and write OAuth clients. This was used by Aurora's "OAuth Clients" interface, but since that [now lives inside this application](https://github.com/DoSomething/northstar/pull/1173) it's no longer used!

🔥

This was always a somewhat scary API – if a token that had access to this endpoint was leaked, an attacker could use those credentials to create, view, or edit other tokens and really cause a ruckus! It's nice peace-of-mind to remove it. 😌

### Relevant tickets

References [Pivotal #176850872](https://www.pivotaltracker.com/story/show/176850872).

### Checklist


lindsayrmaher (approved):
😌

PR #1178, opened 2021-04-09
Fix bug where users who unselect "WYD" in registration flow are still subscribed to it

katiecrane:
### What's this PR do?

This pull request fixes a very specific bug! When a member went through the web registration form:
1. After first page, gets subscribed to WYD (correct!)
2. On third page, if they unchecked WYD and did not check any other newletters, they would still be subscribed to WYD. (incorrect!)

The issue was that Laravel does not include in the request fields that have not been filled out. A form with four checkboxes where none are checked appears to indicate no change, but it fact signifies clearing out all email topics since we know for sure that one of those boxes was checked by default. (If an existing user happens to hit this page, they will still get the correct experience too!) So, now we check the request for `email_subscription_topics`  and if none are found, we actively set that to `[]`.

See failing and then passing test as proof that this works!

### Relevant tickets

References [Pivotal #177290113](https://www.pivotaltracker.com/story/show/177290113).

### Checklist

### DIFF
--- a/app/Http/Controllers/Web/ProfileSubscriptionsController.php
+++ b/app/Http/Controllers/Web/ProfileSubscriptionsController.php
@@ -52,7 +52,13 @@ class ProfileSubscriptionsController extends Controller
 
         $currentMobile = $user->mobile;
 
-        $this->registrar->register($request->all(), $user, function (
+        // Make sure that we wipe out existing email topics if the form is submitted with none checked
+        $request = $request->all();
+        if (!array_key_exists('email_subscription_topics', $request)) {
+            $request['email_subscription_topics'] = [];
+        }
+
+        $this->registrar->register($request, $user, function (
             $user
         ) use ($currentMobile) {
             // Set the sms_status if we're adding or updating the user's mobile.
--- a/app/Http/Controllers/Web/ProfileSubscriptionsController.php
+++ b/app/Http/Controllers/Web/ProfileSubscriptionsController.php
@@ -53,12 +53,12 @@ class ProfileSubscriptionsController extends Controller
         $currentMobile = $user->mobile;
 
         // Make sure that we wipe out existing email topics if the form is submitted with none checked
-        $request = $request->all();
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/Web/ProfileSubscriptionsController.php
```diff
@@ -52,7 +52,13 @@ public function update(Request $request)
...
         $currentMobile = $user->mobile;
 
-        $this->registrar->register($request->all(), $user, function (
+        // Make sure that we wipe out existing email topics if the form is submitted with none checked
+        $request = $request->all();
+        if (!array_key_exists('email_subscription_topics', $request)) {
+            $request['email_subscription_topics'] = [];
+        }
+
+        $this->registrar->register($request, $user, function (
```
mendelB, line 61: Nit -- maybe worth naming this `$input` (like in the [Laravel docs](https://laravel.com/docs/7.x/requests#retrieving-input) example usage) to avoid overriding the `$request` itself?
```suggestion
        // Make sure that we wipe out existing email topics if the form is submitted with none checked
        $input = $request->all();
        if (!array_key_exists('email_subscription_topics', $input)) {
            $input['email_subscription_topics'] = [];
        }

        $this->registrar->register($input, $user, function (
```

mendelB (approved):
🙌 🕵️

app/Http/Controllers/Web/ProfileSubscriptionsController.php
katiecrane: LOVE that idea! I was actually concerned about overriding the `$request` too (even though we're not using it again as of now) for future confusion but wasn't sure about naming and this solves it!

katiecrane: Just pushed up this change! https://github.com/DoSomething/northstar/pull/1178/commits/cce3a3393c99d69bc46e8c95d2fa1a2d13d6ebde

PR #1182, opened 2021-04-16
Adding chompy models and moar

weerd:
### What's this PR do?

This pull request copies over Chompy models, types, helpers, facotries, seeders and a few other things into Northstar so we can begin the work of connecting the pieces to incorporate Chompy's functionality into Northstar.

👀 

Northstar is like Kirby absorbing all our service repositories.

### Relevant tickets

References [Pivotal #177733225](https://www.pivotaltracker.com/story/show/177733225).

### Checklist


app/Imports/RockTheVoteRecord.php
```diff
@@ -0,0 +1,250 @@
...
+                 */
+            } elseif (
+                ($key === 'referral' || $key === 'refferal') &&
+                str_to_boolean($value[1])
+            ) {
+                /**
+                 * Return result to force querying for existing user via this record email or mobile
+                 * upon import.
+                 */
+                $result['referrer_user_id'] = $userId; // @TODO: error! This would never be set... wtf?!
```
DFurnes, line 191: ⁉️

database/factories/RockTheVoteLogFactory.php
```diff
@@ -0,0 +1,20 @@
...
+
+$factory->define(RockTheVoteLog::class, function (Faker $faker) {
+    return [
+        'finish_with_state' => 'No',
+        'import_file_id' => $faker->randomDigitNotNull,
+        'pre_registered' => 'No',
+        'started_registration' => Carbon::now()->format('Y-m-d H:i:s O'),
+        'status' => 'Step 1',
+        'tracking_source' => 'ads',
+        'user_id' => factory(User::class)->create(),
```
DFurnes, line 18: This needs an `->id` at the end I think! And might be good to place in a closure in case we override:

```suggestion
        'user_id' => function () {
            return factory(User::class)->create()->id;
        },
```

DFurnes (approved):

database/factories/RockTheVoteLogFactory.php
DFurnes: Whoops, I forgot we were trying out auto-merge! 😂

DFurnes: We'll probably want to apply the same fix [here](https://github.com/DoSomething/northstar/blob/27740d4e9db172bb5bbd248badd9c35a2da4957e/database/factories/PostFactory.php#L33) as well!

PR #1183, opened 2021-04-16
More efficient user creation in factory setup.

weerd:
### What's this PR do?

This pull request updates how users are created in factories and improves it by wrapping it in a closure which is only executed if not value for the `user_id` field is provided.

👀 

Great suggestion by @DFurnes in [previous PR](https://github.com/DoSomething/northstar/pull/1182/files/2f33b64122c55a3ab20c54d08f72aaff6c69af88#diff-0f9d790228ea7bad6f8d105f9db55ecb83def4973ecf4cef290f996d1d37b1b2).

### Relevant tickets

References [Pivotal #177733225](https://www.pivotaltracker.com/story/show/177733225).

### Checklist

### DIFF
--- a/database/factories/PostFactory.php
+++ b/database/factories/PostFactory.php
@@ -15,6 +15,9 @@ $factory->define(Post::class, function (Faker $faker) {
     $faker->addProvider(new FakerSchoolId($faker));
 
     return [
+        'northstar_id' => function () {
+            return factory(User::class)->create()->id;
+        },
         'campaign_id' => function () {
             return factory(Campaign::class)->create()->id;
         },
@@ -29,9 +32,6 @@ $factory->define(Post::class, function (Faker $faker) {
             return factory(Action::class)->create([
                 'campaign_id' => $attributes['campaign_id'],
             ])->id;
-        },
-        'northstar_id' => function () {
-            factory(User::class)->create()->id;
         },
         'text' => $faker->sentence(),
         'location' => 'US-' . $faker->stateAbbr(),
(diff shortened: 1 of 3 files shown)


DFurnes (approved):
🆔

database/factories/PostFactory.php
```diff
@@ -15,6 +15,9 @@
     $faker->addProvider(new FakerSchoolId($faker));
 
     return [
+        'northstar_id' => function () {
+            return factory(User::class)->create()->id;
+        },
```
weerd, line 20: This stumped me for a bit and I was seeing failures in the `TagsTest.php`.

It turns out that since the `northstar_id` is a needed relationship to generate the `signup_id` attribute, it needs to be ready for when the `signup_id` attribute is being generated.

Prior to enclosing the `northstar_id` factory in a closure, it would be available regardless of the ordering in the array, but now that we're using a closure, the array ordering matters. If the `northstar_id` attribute is below the `signup_id` all that is returned for the `northstar_id` that the Signup factory uses is the closure. Thus, moving the `northstar_id` above the `signup_id` resolves the bug since the `signup_id` relationship depends on the Post factory having a `northstar_id` defined.

The [Laravel docs](https://laravel.com/docs/6.x/database-testing#relationships) showcase this, and I guess it's implied that the dependent attributes are defined and resolved ahead of the attribute with the relationship!

Twas a little tricky to figure out because it wasn't immediately obvious, and took a bit longer than ideal to figure out but it was a rather interesting little find (for me anyhoo 😊).

DFurnes: Whew, great sleuthing!! 🕵️

PR #1185, opened 2021-04-22
Adding import mute promotions job

weerd:
### What's this PR do?

This pull request starts to merge the Chompy queue jobs over into Northstar.

First up is the `ImportMutePromotions` job!

👀 

I wasn't sure whether to include tests for the existing Northstar API/Web endpoints for `/users/{id}/promotions`. Now that the job is in Northstar directly it does not need to use these endpoints. We will need the endpoints when we transition the web UI but I suspect that means getting rid of the API endpoint and keeping the web one. We can tackle that when we tackle the UI.

### Relevant tickets

References [Pivotal #177733312](https://www.pivotaltracker.com/story/show/177733312).

### Checklist

### DIFF
--- a/app/Http/Controllers/PromotionsController.php
+++ b/app/Http/Controllers/PromotionsController.php
@@ -30,6 +30,11 @@ class PromotionsController extends Controller
     /**
      * Mute promotions for the specified resource.
      *
+     * @TODO: this existing as a destroy method seems odd, since it is not directly
+     * deleting a resource but setting a property on the User model. Contextually,
+     * this will ultimately result in the deletion of the user from CustomerIO, but
+     * could be a bit clearer.
+     *
      * @param  string $id
      * @return \Illuminate\Http\Response
      */
--- a/app/Http/Controllers/Web/Admin/PromotionsController.php
+++ b/app/Http/Controllers/Web/Admin/PromotionsController.php
@@ -20,8 +20,11 @@ class PromotionsController extends Controller
 
     /**
      * Mutes promotions for user.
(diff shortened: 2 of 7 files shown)


app/Http/Controllers/PromotionsController.php
```diff
@@ -30,6 +30,11 @@ public function __construct(UserTransformer $transformer)
     /**
      * Mute promotions for the specified resource.
      *
+     * @TODO: this existing as a destroy method seems odd, since it is not directly
+     * deleting a resource but setting a property on the User model. Contextually,
+     * this will ultimately result in the deletion of the user from CustomerIO, but
+     * could be a bit clearer.
+     *
```
weerd, line 37: I could maybe see this as a renamed controller (that more explicitly declares what it does) and with just a single `__invoke()`-able method since there's not really anything else going on in the controller class.

app/Http/Controllers/Web/Admin/PromotionsController.php
```diff
@@ -21,6 +21,11 @@ public function __construct()
     /**
      * Mutes promotions for user.
      *
+     * @TODO: this existing as a destroy method seems odd, since it is not directly
+     * deleting a resource but setting a property on the User model. Contextually,
+     * this will ultimately result in the deletion of the user from CustomerIO, but
+     * could be a bit clearer.
+     *
```
weerd, line 28: Same same as [above](https://github.com/DoSomething/northstar/pull/1185/files#r618827419).

DFurnes (approved):
Looks great! 💯

PR #1187, opened 2021-04-27
Adding import email subscriptions job

weerd:
### What's this PR do?

This pull request pulls in the `ImportEmailSubscriptions` job from Chompy into Northstar. It also adds some new tests for the job which it previously did not have in Chompy-land.

👀 

### Any background ~context~ you want to provide?

Yes I do. Here's a nice one of the stars at night and a nice mountainous landscape.

image

### Relevant tickets

References [Pivotal #177733302](https://www.pivotaltracker.com/story/show/177733302).

### Checklist


app/Jobs/Imports/ImportEmailSubscriptions.php
```diff
@@ -0,0 +1,179 @@
...
+     * @param object $user
+     */
+    private function sendPasswordReset($user)
+    {
+        $logParams = ['user' => $user->id];
+
+        [
+            'enabled' => $topicResetEnabled,
+            'type' => $topicResetType,
+        ] = $this->getResetConfig();
```
weerd, line 148: This was fun! Associative Array Destructuring in PHP 🎉

app/Jobs/Imports/ImportEmailSubscriptions.php
```diff
@@ -0,0 +1,179 @@
...
+    /**
+     * Get the specified import config.
+     *
+     * @return array
+     */
+    private function getResetConfig()
+    {
+        $target = ImportType::getConfig(ImportType::$emailSubscription);
+
+        $key = "topics.$this->selected_subscription_topic.reset";
```
weerd, line 175: This felt pretty cool too; simple but I thought it helps make it a bit more readable! Variable interpolation to generate the dot-notation string that the `data_get()` method utilizes to grab the key from the provided array :)

database/factories/ImportFileFactory.php
```diff
@@ -10,7 +10,23 @@
...
+    'import_type' => ImportType::$emailSubscription,
+]);
+
+$factory->state(ImportFile::class, 'mute_promotions', [
+    'import_type' => ImportType::$emailSubscription,
+]);
+
+$factory->state(ImportFile::class, 'rock_the_vote', [
+    'import_type' => ImportType::$emailSubscription,
+]);
```
weerd, line 32: Now with states!™ Soon it will slice and dice and make julienne fries!

app/Jobs/Imports/ImportEmailSubscriptions.php
```diff
@@ -0,0 +1,179 @@
...
+
+        $this->importFile = $importFile;
+    }
+
+    /**
+     * Execute the job to create or update users by email and set subscription topics.
+     *
+     * @return array
+     */
+    public function handle()
```
weerd, line 75: I cleaned up this `handle()` function a tiny bit, but I can try and refactor it further if it still feels a little convoluted, which I think I eventually want to tackle, but didn't seem like a priority for the time being.

app/Jobs/Imports/ImportEmailSubscriptions.php
```diff
@@ -0,0 +1,179 @@
...
+     *
+     * @return array
+     */
+    private function getResetConfig()
+    {
+        $target = ImportType::getConfig(ImportType::$emailSubscription);
+
+        $key = "topics.$this->selected_subscription_topic.reset";
+
+        return data_get($target, $key);
```
weerd, line 177: love me some `data_get()` helper action!

app/Jobs/Imports/ImportEmailSubscriptions.php
```diff
@@ -0,0 +1,179 @@
...
+     * Execute the job to create or update users by email and set subscription topics.
+     *
+     * @return array
+     */
+    public function handle()
+    {
+        // @TODO: standardize logging for info messages; they seem all over the place!
+        info('progress_log: Processing: ' . $this->email);
+
+        $registrar = App::make(Registrar::class);
```
DFurnes, line 80: I believe you can inject this in the `handle` method, which might make the dependency read a little cleaner:

```php
public function handle (Registrar $registrar) {
 // ...

 $user = $registrar->resolve(['email' => $this->email]);
```

app/Jobs/Imports/ImportEmailSubscriptions.php
```diff
@@ -0,0 +1,179 @@
...
+
+        $user = $registrar->resolve(['email' => $this->email]);
+
+        $collatedSubscriptionTopics = $this->collateSubscriptionTopics($user);
+
+        $input = [
+            'first_name' => $this->first_name,
+            'email_subscription_status' => true, // @TODO: feels weird this is a boolean.
```
DFurnes, line 88: Is this `TODO` related to the field itself on the user or the usage here? If it's the former, I wonder if there's a better place to track this (since I'm not sure how likely we are to be able to act upon changing the structure of this field when we come upon this `TODO` if working in this file in the future).

app/Jobs/Imports/ImportEmailSubscriptions.php
```diff
@@ -0,0 +1,179 @@
...
+            'email_subscription_topics' => $collatedSubscriptionTopics,
+        ];
+
+        if ($user) {
+            // Update the user, filtering out null input values.
+            $registrar->register(array_filter($input), $user);
```
DFurnes, line 94: It could read a little more clearly to just use `$user->fill(…); $user->save();` here, since we're not really _registering_ a user in this case.

tests/Jobs/Imports/ImportEmailSubscriptionsTest.php
```diff
@@ -0,0 +1,138 @@
...
+            $importFile,
+            [
+                'source_detail' => 'phpunit',
+                'email_subscription_topic' => 'community',
+            ],
+        );
+
+        $user = App::make(Registrar::class)->resolve([
+            'email' => 'puppetsloth@dosomething.org',
+        ]);
```
DFurnes, line 134: Let's just use a plain `User::where('email', 'puppetsloth@dosomething.org')->first()` to limit how much we're relying on application code in our test assertions. (Since the "resolve" method here does some extra input normalization that could accidentally lead us to assert against the wrong user if there was a bug there!)

DFurnes (approved):
Looks great!

app/Jobs/Imports/ImportEmailSubscriptions.php
weerd: Oh right! Great suggestion! I'll try that.

app/Jobs/Imports/ImportEmailSubscriptions.php
weerd: Ya, I left it in moreso to prompt a discussion but I can remove. It was moreso the field... the name of the key makes me expect a status value of `subscribed` or `unsubscribed` but having the status be `true` or `false` makes it read oddly and seems confusing to me.

Mostly mentioned here in case it feels like something to address during the unified backend work and switch to PostgreSQL. Maybe a backlog ticket would make more sense?

tests/Jobs/Imports/ImportEmailSubscriptionsTest.php
weerd: Oooo, ya, good call!

app/Jobs/Imports/ImportEmailSubscriptions.php
DFurnes: Mmm, that's a good flag re: PostgreSQL work! Could you drop a comment on that field in the technical spec? That'll probably be the best way to remind us to look into reformatting that when it gets to migration time.

weerd: ok! I left a comment in the tech spec 📝

weerd:
This test is failing a lot, triggering a `General error: 1205 Lock wait timeout exceeded; try restarting transaction...` error so I'm digging into it because it will just happen more often. I think it's related to transactions with queue jobs.

PR #1189, opened 2021-05-05
Add a command to fix mismatched mobile/status fields.

DFurnes:
### What's this PR do?

This pull request adds an Artisan command to fix accounts which have no `mobile` but still have an "addressable" `sms_status` field. In cases where a user _had_ a phone number and removed it, this bug may be causing them to continue to receive SMS messaging & be unable to opt-out by texting STOP.

I added a command to fix potentially affected accounts & a test demonstrating expected behavior. This command also logs context about affected users (source, creation time, and whether or not they have a `phone` set in Customer.io).

I'll plan to run this command on QA & production – once as a "dry run", and once "for real".

You can read a full write-up on the issue that this addresses in [this Pivotal comment](https://www.pivotaltracker.com/story/show/177986264/comments/223965113).

### Relevant tickets

References [Pivotal #177986264](https://www.pivotaltracker.com/story/show/177986264).

### Checklist

### DIFF
--- a/app/Console/Commands/FixSmsStatusCommand.php
+++ b/app/Console/Commands/FixSmsStatusCommand.php
@@ -1,0 +1,66 @@
+<?php
+
+namespace App\Console\Commands;
+
+use App\Models\User;
+use App\Services\CustomerIo;
+use Illuminate\Console\Command;
+use Illuminate\Support\Collection;
+
+class FixSmsStatusCommand extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'northstar:fix-sms-status {--dry-run}';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Fix accounts with active sms_status but no mobile number.';
+
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle(CustomerIo $customerIo)
+    {
+        info('northstar:fix-sms-status - Starting up!');
+
+        // Get addressable users with a null mobile:
+        $query = User::whereIn('sms_status', [
+            'active',
+            'pending',
+            'less',
+        ])->where('mobile', 'exists', false);
+
+        $query->chunkById(200, function (Collection $users) use ($customerIo) {
+            foreach ($users as $user) {
+                $profile = $customerIo->getAttributes($user);
+
+                // We want to log a little context on each user that we're fixing
+                // to help identify patterns of potentially affected accounts:
+                info('Fixing user', [
+                    'id' => $user->id,
+                    'has_profile' => !empty($profile),
+                    'phone' => data_get($profile, 'phone'),
+                    'source' => $user->source,
+                    'created_at' => $user->created_at,
+                ]);
+
+                if ($this->option('dry-run')) {
+                    continue;
+                }
+
+                $user->unset('sms_status');
+            }
+        });
+
+        info('northstar:fix-sms-status - All done!');
+    }
+}
--- a/app/Services/CustomerIo.php
+++ b/app/Services/CustomerIo.php
@@ -101,6 +101,36 @@ class CustomerIo
     }
 
     /**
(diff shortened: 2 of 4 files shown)


app/Console/Commands/FixSmsStatusCommand.php
```diff
@@ -0,0 +1,66 @@
...
+                    'id' => $user->id,
+                    'has_profile' => !empty($profile),
+                    'phone' => data_get($profile, 'phone'),
+                    'source' => $user->source,
+                    'created_at' => $user->created_at,
+                ]);
+
+                if ($this->option('dry-run')) {
+                    continue;
+                }
```
mendelB, line 58: Neato!

app/Services/CustomerIo.php
```diff
@@ -100,6 +100,36 @@ public function trackEvent($user, $eventName, $eventData = [])
...
+        );
+
+        // For this endpoint, any status besides 200 means something is wrong:
+        if ($response->getStatusCode() !== 200) {
+            return null;
+        }
+
+        $json = json_decode((string) $response->getBody(), true);
+
+        return $json['customer']['attributes'];
```
weerd, line 130: Not vital, but maybe use the `data_get()` helper and can provide default `null` just in case?

app/Services/CustomerIo.php
```diff
@@ -100,6 +100,36 @@ public function trackEvent($user, $eventName, $eventData = [])
...
+        if (!$this->enabled()) {
+            info('Would have read attributes from Customer.io', [
+                'id' => $user->id,
+            ]);
+
+            return null;
+        }
+
+        $response = $this->appApiClient->get(
+            "https://beta-api.customer.io/v1/api/customers/$user->id/attributes",
```
mendelB, line 120: Ohhhh I didn't realize we ID folks in customer.io using their NS ID! 💡

tests/Console/FixSmsStatusCommandTest.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+use App\Models\User;
+use Illuminate\Support\Facades\Artisan;
+
+class FixSmsStatusCommandTest extends TestCase
+{
+    /** @test */
+    public function it_should_fix_birthdates()
```
mendelB, line 9: Should this be?
```suggestion
    public function it_should_fix_sms_statuses()
```

app/Services/CustomerIo.php
mendelB, line 130: Ooof, in the docs they don't seem to show a top-level `customer` key in the response payload 😦

mendelB (approved):
💥

app/Console/Commands/FixSmsStatusCommand.php
weerd, line 58: Nice! ↩️

tests/Console/FixSmsStatusCommandTest.php
DFurnes: Whoops, yes!

weerd (approved):
Small comment but otherwise looks great! 👍

app/Services/CustomerIo.php
DFurnes: Yeah, that tripped me up at first! I guess that's why it's a beta API. 😅

DFurnes: I think I'd prefer to keep this as-is, so that if we get an unexpected response structure we'll throw an exception. I'd be worried that if we silently turned this into `null`, we might not catch that something was amiss.

app/Services/CustomerIo.php
DFurnes: Yeah, it's very handy! They also seem to have their own "internal" IDs (which, for example, they now use for URLs), but it generally makes it nice and simple to find a corresponding profile based on our own user IDs which I really appreciate.

app/Services/CustomerIo.php
weerd: Ah, that's fair! Sounds good.

PR #1191, opened 2021-05-05
Impact goal for actions

lindsayrmaher:
### What's this PR do?

This pull request adds a new field to an action called the `impact_goal`. This field is for editors to define the impact goal for a campaign that involves collecting items etc. This goal amount will be used to populate the progress bar displayed on that campaign!

Let me know if any other test cases make sense here. I would like to add one to test an invalid data format when an action is created, but I was having trouble finding a good way to do that. Advice would be welcomed!

Screen Shot 2021-05-05 at 4 23 08 PM

I added the field to the create/edit forms for an action, but I can't yet add it to the react components that display actions because we are using a graphql query there to grab the data and I haven't added this field to graphql yet. Tricky, tricky!

### Relevant tickets

References [Pivotal #177933241](https://www.pivotaltracker.com/story/show/177933241).

### Checklist

### DIFF
--- a/database/migrations/2021_05_04_185150_add_impact_goal_to_actions.php
+++ b/database/migrations/2021_05_04_185150_add_impact_goal_to_actions.php
@@ -15,7 +15,7 @@ class AddImpactGoalToActions extends Migration
     {
         Schema::table('actions', function (Blueprint $table) {
             $table
-                ->string('impact_goal')
+                ->unsignedInteger('impact_goal')
                 ->after('quiz')
                 ->nullable()
                 ->comment('The impact goal associated with this action.');
--- a/database/migrations/2021_05_04_185150_add_impact_goal_to_actions.php
+++ b/database/migrations/2021_05_04_185150_add_impact_goal_to_actions.php
@@ -1,0 +1,36 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class AddImpactGoalToActions extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('actions', function (Blueprint $table) {
+            $table
+                ->string('impact_goal')
+                ->after('quiz')
+                ->nullable()
+                ->comment('The impact goal associated with this action.');
+        });
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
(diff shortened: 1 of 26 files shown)


database/migrations/2021_05_04_185150_add_impact_goal_to_actions.php
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('actions', function (Blueprint $table) {
+            $table
+                ->string('impact_goal')
```
DFurnes, line 18: Would it make sense to save this as an integer, in case we ever wanted to sort by this column?

tests/Http/ActionTest.php
```diff
@@ -38,4 +38,25 @@ public function testActionShow()
...
+    /**
+     * Test that when an action is created with an impact goal it saves properly
+     *
+     * @return void
+     */
+    public function testCreatingAnActionWithAnImpactGoal()
+    {
+        $action = factory(Action::class)->create(['impact_goal' => '3000']);
+
+        $response = $this->getJson('api/v3/actions/' . $action->id);
```
DFurnes, line 51: Let's rename this test to `testDisplayingActionWithAnImpactGoal` and reframe the assertions to ensure that the expected impact goal is found within the JSON response. 

We can then test creating an action via the [`WebActionTest`](https://github.com/DoSomething/northstar/blob/c3e6611ba9b7ca5a0cf11d616e40278a9b71262a/tests/Http/Web/WebActionTest.php#L9-L42) test case (since that's handled via Northstar's internal "web" API, and not the public API). I think that could make sense to bundle within the existing `testCreatingAnAction` test for the "success" case.

> Let me know if any other test cases make sense here. I would like to add one to test an invalid data format when an action is created, but I was having trouble finding a good way to do that. Advice would be welcomed!

One thing we're missing in `WebActionTest` is a test case for request validation, so this is a perfect opportunity to add it! You can assert validation errors occur using `assertSessionHasErrors`, like [this](https://github.com/DoSomething/northstar/blob/c3e6611ba9b7ca5a0cf11d616e40278a9b71262a/tests/Http/Web/WebAuthenticationTest.php#L82-L84).

DFurnes (approved):

database/migrations/2021_05_04_185150_add_impact_goal_to_actions.php
lindsayrmaher: yeah i was wondering about that as well, i think i'll make that update

PR #1192, opened 2021-05-05
Don't throw on 404s from the Customer.io API.

DFurnes:
### What's this PR do?

This pull request prevents the `northstar:fix-sms-status` command from throwing an exception if a request 404's.

We'll expect a subset of these users won't exist in Customer.io because their promotions have been muted & that shouldn't "fail" the script. For example, here's an error that halted the script when testing on QA earlier:

```
In RequestException.php line 113:

  Client error: `GET https://beta-api.customer.io/v1/api/customers/5d02cdd646e68f00e96de186/attributes` resulted in a `404 Not Found` response:
  {"errors":[{"detail":"not found (reference 01F4Z5B1MDK8FFTM2D2C5QM9GQ)","status":"404"}]}
```

Here's documentation on the Guzzle [http_errors](https://docs.guzzlephp.org/en/stable/request-options.html#http-errors) option.

This more closely mirrors how we handle requests in our Gateway library (where 404's also return `null`).

I always forget that this is the default behavior with Guzzle! 

### Relevant tickets

References [Pivotal #177986264](https://www.pivotaltracker.com/story/show/177986264).

### Checklist

### DIFF
--- a/app/Services/CustomerIo.php
+++ b/app/Services/CustomerIo.php
@@ -151,7 +151,12 @@ class CustomerIo
 
         $response = $this->trackApiClient->put('customers/' . $user->id, [
             'json' => $payload,
+            'http_errors' => false,
         ]);
+
+        if ($response->getStatusCode() === 404) {
+            return null;
+        }
 
         // For this endpoint, any status besides 200 means something is wrong:
         if ($response->getStatusCode() !== 200) {


app/Services/CustomerIo.php
```diff
@@ -151,8 +151,13 @@ public function updateCustomer(User $user)
...
             'json' => $payload,
+            'http_errors' => false,
         ]);
 
+        if ($response->getStatusCode() === 404) {
+            return null;
+        }
+
         // For this endpoint, any status besides 200 means something is wrong:
         if ($response->getStatusCode() !== 200) {
```
mendelB, line 162: Would we ever reach this point in the function if Guzzle is throwing their own exceptions for non-200s? (Or would this catch anything besides 4/500s?)

mendelB (approved):
⚡

app/Services/CustomerIo.php
DFurnes: It would not have! Now that we've set `http_errors = false`, it'll no longer throw these exceptions & so we need to catch and handle each potential status code ourselves (and this code block will actually be used).

mendelB: omg. literally this entire point of this PR 🤦 . coffee first.

PR #1194, opened 2021-05-06
Investigate flaky test failures.

DFurnes:
### What's this PR do?

We've been seeing a lot of flaky failures in this application's test suite lately, mostly (or all) contained to the new import job tests. It seems like these are caused by seeding a random integer into the `id` field in the ImportFile factory, which could be causing duplicates or otherwise conflicting with this column's auto-incremented setting.

I pushed up a change to this seeder in bae24a7 & then pushed 9 more empty commits to "stress test" this fix on CI.

I left notes as I investigated in [this Slack thread](https://dosomething.slack.com/archives/CUQMU4Q6B/p1620315425064900).

### Relevant tickets

References [Pivotal #178055976](https://www.pivotaltracker.com/story/show/178055976).

### Checklist

### DIFF
--- a/database/factories/ImportFileFactory.php
+++ b/database/factories/ImportFileFactory.php
@@ -8,7 +8,6 @@ use Faker\Generator as Faker;
 
 $factory->define(ImportFile::class, function (Faker $faker) {
     return [
-        'id' => $faker->randomDigitNotNull,
         'filepath' => $faker->imageUrl,
         'import_type' => $faker->randomElement([
             ImportType::$emailSubscription,


DFurnes:
One of these 10 test runs failed with the following error on `testUpdatedPostWithReaction`:

```
Failed asserting that a row in the table [posts] matches the attributes {
    "id": 164,
    "updated_at": "2021-05-06T16:38:21.000000Z"
}.

Found similar results: [
    {
        "id": 164,
        "updated_at": "2021-05-06 16:38:22",
    }
].
```

That seems to be a different (unrelated) flaky timing failure which I haven't seen before (or since).

weerd:
> That seems to be a different (unrelated) flaky timing failure which I haven't seen before (or since).

It fails because the time is off by a second (or also because the formats don't match)? I wonder if the formatting is playing a role in a rounding error or something 🤔

DFurnes:
Could be! I also wonder if it could have just happened to line up with when the second changed since we don't freeze the clock in this test.

DFurnes:
Alright, and that's ten successful builds without a `Lock wait timeout exceeded error`… I think this issue is solved! 🎉

weerd (approved):
All that trouble was caused from a single line! 🤯  Excellent find! 🕵️‍♂️

mendelB:
image

I'm printing & framing this.

PR #1197, opened 2021-05-10
Bring over Rogue API docs

katiecrane:
### What's this PR do?

This pull request brings over all the Rogue API docs which I noticed didn't live here yet! Not sure if maybe that was intentional in which case we can discuss!

- I brought over docs for all the endpoints that still exist in Northstar (so the [`/users` one](https://github.com/DoSomething/rogue/blob/master/docs/endpoints/users.md) that is not still a thing was not brought over). 
- I did not update any docs but I did update the table of contents to include a few routes that were missing. The rotation endpoint is missing docs and I did not add those but I did add it to the table.
- I put all of these into a `v3` folder because that seems to be how we generally structure our API docs but totally open to moving into the root `documentation` folder if we prefer that
- I moved the questionnaire docs into the `v3` folder and added that to the table as well since it is also v3

Does how I organized these make sense? Anything still obviously outdated?

Nope!

### Relevant tickets

🤠


DFurnes (approved):
Thank you!!!! 📚 This totally slipped my mind.

PR #1199, opened 2021-05-10
Add Accepted Quantity method to actions

lindsayrmaher:
### What's this PR do?

This pull request adds a method to the actions model that calculates how many posts associated with it have an `accepted` status. We then add that using `includes` on the `ActionTransformer` so that we can optionally query for it when pulling an action or actions. If we include it in the API request it should return as an `accepted_quantity` field.

I think I covered all the steps needed with the transformer but let me know if I missed something!

Need this so that we can pull this info and display on a campaign's progress bar!

### Relevant tickets

References [Pivotal #177933245](https://www.pivotaltracker.com/n/projects/2401401/stories/177933245).

### Checklist

### DIFF
--- a/app/Models/Action.php
+++ b/app/Models/Action.php
@@ -96,4 +96,16 @@ class Action extends Model
     {
         return $this->hasMany(Post::class);
     }
+
+    /**
+     * Get the quantity total associated with approved posts under this action.
+     *
+     * @return int
+     */
+    public function getAcceptedQuantity()
+    {
+        $accepted_posts = $this->posts->where('status', 'accepted');
+
+        return $accepted_posts->sum('quantity');
+    }
 }
--- a/documentation/endpoints/v3/actions.md
+++ b/documentation/endpoints/v3/actions.md
@@ -89,6 +89,12 @@ Example Response:
 
 ## Retrieve A Specific Action
 
+### Optional Query Parameters
+
+- **include** _(string)_
+  - Include additional related records in the response: `accepted_quantity`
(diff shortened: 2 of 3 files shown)


app/Models/Action.php
```diff
@@ -96,4 +96,16 @@ public function posts()
...
     }
+
+    /**
+     * Get the quantity total associated with approved posts under this action.
+     *
+     * @return int
+     */
+    public function getAcceptedQuantity()
+    {
+        $accepted_posts = $this->posts->where('status', 'accepted');
```
mendelB, line 107: Total nit, but I thiiiiiiiink our convention is to camel case variables in PHP? `$acceptedPosts`

documentation/endpoints/v3/actions.md
```diff
@@ -11,9 +11,14 @@ GET /api/v3/actions
 ### Optional Query Parameters
 
 - **filter[column]** _(string)_
+
   - Filter results by the given column: `id`, `campaign_id`, `callpower_campaign_id`
   - You can filter by more than one value for a column, e.g. `/actions?filter[id]=121,122`
 
+- **include** _(string)_
+  - Include additional related records in the response: `accepted_quantity`
```
mendelB, line 19: I know this is our typical usage for Includes, but in this case, could it make sense to add something along the lines of 
```suggestion
  - Include additional related records and/or computed attributes in the response: `accepted_quantity`
```

mendelB (approved):
This is neat! 👏

mendelB:
This could also be a good feature for testing, is that slated in for this work?

PR #1200, opened 2021-05-10
Add 'ImportCallPowerRecord' job & test.

DFurnes:
### What's this PR do?

This pull request adds the `ImportCallPowerRecord` job and test from Chompy into Northstar. 

I'd recommend reviewing commit-by-commit!

While we're not currently using CallPower, this will allow us to reactivate this partnership once Chompy is decommissioned.

### Relevant tickets

References [Pivotal #177733285](https://www.pivotaltracker.com/story/show/177733285).

### Checklist

### DIFF
--- a/app/Jobs/Imports/ImportCallPowerRecord.php
+++ b/app/Jobs/Imports/ImportCallPowerRecord.php
@@ -1,0 +1,146 @@
+<?php
+
+namespace App\Jobs\Imports;
+
+use App\Auth\Registrar;
+use App\Managers\PostManager;
+use App\Managers\SignupManager;
+use App\Models\Action;
+use App\Models\User;
+use Exception;
+use Chompy\Services\Rogue;
+use Illuminate\Bus\Queueable;
+use Illuminate\Queue\InteractsWithQueue;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Foundation\Bus\Dispatchable;
+
+class ImportCallPowerRecord implements ShouldQueue
+{
+    use Dispatchable, InteractsWithQueue, Queueable;
+
+    /**
+     * The call parameters sent from CallPower.
+     *
+     * @var array
+     */
+    protected $parameters;
+
+    /**
+     * Create a new job instance.
+     *
+     * @return void
+     */
+    public function __construct($parameters)
+    {
+        $this->parameters = $parameters;
+    }
+
+    /**
+     * Execute the job to create a CallPower post in Rogue.
+     *
+     * @return array
+     */
+    public function handle(
+        Registrar $registrar,
+        SignupManager $signups,
+        PostManager $posts
+    ) {
+        ['mobile' => $mobile, 'status' => $status] = $this->parameters;
+
+        $details = $this->extractDetails($this->parameters);
+
+        // If this call came from an anonymous mobile, we won't be able to
+        // tie it to a user account & so should just discard the record:
+        if (is_anonymous_mobile($mobile)) {
+            info('Cannot import phone call for anonymous mobile ' . $mobile);
+
+            return;
+        }
+
+        // Using the mobile number, get or create a user:
+        $user = $registrar->resolve(['mobile' => $mobile]);
+
+        if (!$user) {
+            $user = $registrar->register([
+                'mobile' => $mobile,
+                'sms_status' => 'stop',
+            ]);
+
+            info('ImportCallPowerRecord - created user', ['user' => $user->id]);
+        }
+
+        // Using the callpower_campaign_id, get the action & campaign:
+        $callPowerId = $this->parameters['callpower_campaign_id'];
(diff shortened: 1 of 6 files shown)


app/Jobs/Imports/ImportCallPowerRecord.php
```diff
@@ -0,0 +1,146 @@
...
+            'user' => $user->id,
+            'details' => $details,
+        ]);
+
+        $payload = [
+            'northstar_id' => $user->id,
+            'action_id' => $action->id,
+            'type' => 'phone-call',
+            'quantity' => 1,
+            'source' => 'importer-client', // @TODO: We should type this.
```
DFurnes, line 94: This will keep the same metadata that we're currently storing (and, I think, relying upon in Quasar) for signups & posts generated by Chompy imports (which use [this OAuth client](https://identity.dosomething.org/admin/clients/importer-client)).

DFurnes: For now, I hardcoded this in each job but I think it'd make sense to load via a type (in an OAuth-less future), or perhaps as a static property on the `Client` model...

app/Jobs/Imports/ImportCallPowerRecord.php
```diff
@@ -0,0 +1,146 @@
...
+
+        if (!$signup) {
+            $signup = $signups->create($payload, $user->id, $campaign->id);
+        }
+
+        $post = $posts->create($payload, $signup->id);
+
+        if ($post) {
+            $post->status = $status === 'completed' ? 'accepted' : 'incomplete';
+            $post->save();
```
DFurnes, line 109: This needs to be set here because the `status` field is (reasonably) not mass-assignable, and so this override would be discarded if we were to have provided it in the `$payload` variable above.

app/Jobs/Imports/ImportCallPowerRecord.php
```diff
@@ -0,0 +1,146 @@
...
+use App\Models\Action;
+use App\Models\User;
+use Exception;
+use Chompy\Services\Rogue;
+use Illuminate\Bus\Queueable;
+use Illuminate\Queue\InteractsWithQueue;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Foundation\Bus\Dispatchable;
+
+class ImportCallPowerRecord implements ShouldQueue
```
DFurnes, line 17: This is based upon the [ImportCallPowerRecord](https://github.com/DoSomething/chompy/blob/main/app/Jobs/CallPower/ImportCallPowerRecord.php) job from Chompy.

app/Jobs/Imports/ImportEmailSubscriptions.php
```diff
@@ -98,7 +98,7 @@ public function handle(Registrar $registrar)
             // We need to pass a "source" in order to save through the source_detail.
             $revisedInput = array_merge($input, [
                 'email' => $this->email,
-                'source' => 'northstar',
+                'source' => 'importer-client', // @TODO: We should type this.
```
katiecrane, line 101: Oooo I like that `@TODO`!

app/Jobs/Imports/ImportCallPowerRecord.php
```diff
@@ -0,0 +1,146 @@
...
+            'details' => $details,
+        ];
+
+        $signup = $signups->get($user->id, $campaign->id);
+
+        if (!$signup) {
+            $signup = $signups->create($payload, $user->id, $campaign->id);
+        }
+
+        $post = $posts->create($payload, $signup->id);
```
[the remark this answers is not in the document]
katiecrane: Ah yeah, since this logic currently lives [in the `store` method on the PostsController](https://github.com/DoSomething/northstar/blob/main/app/Http/Controllers/PostsController.php#L156-L162) that doesn't really help you here does it?

app/Jobs/Imports/ImportCallPowerRecord.php
katiecrane: OH I see, you pulled `$status` out of the parameters at the top

app/Jobs/Imports/ImportCallPowerRecord.php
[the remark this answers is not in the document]
DFurnes: Yep, exactly!

app/Jobs/Imports/ImportCallPowerRecord.php
```diff
@@ -0,0 +1,146 @@
...
+     *
+     * @return void
+     */
+    public function __construct($parameters)
+    {
+        $this->parameters = $parameters;
+    }
+
+    /**
+     * Execute the job to create a CallPower post in Rogue.
```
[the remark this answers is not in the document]
DFurnes: Oh it sure can! 🔥 Removed in dbb73a59.

PR #1205, opened 2021-05-13
Add 'ImportSoftEdgeRecord' job & test.

DFurnes:
### What's this PR do?

This pull request migrates the [`ImportSoftEdgeRecord`](https://github.com/DoSomething/chompy/blob/main/app/Jobs/SoftEdge/ImportSoftEdgeRecord.php) job from Chompy into Northstar.

I've written a simple test case to validate this job works as expected.

I don't believe we're actively importing these, but this isn't functionality we want to lose. We'll follow up with the API endpoints for these third-party integrations after this, so that it's easy to re-enable when we renew these partnerships.

### Relevant tickets

References [Pivotal #177733342](https://www.pivotaltracker.com/story/show/177733342).

### Checklist

### DIFF
--- a/app/Jobs/Imports/ImportSoftEdgeRecord.php
+++ b/app/Jobs/Imports/ImportSoftEdgeRecord.php
@@ -35,7 +35,7 @@ class ImportSoftEdgeRecord implements ShouldQueue
     }
 
     /**
-     * Execute the job to create a SoftEdge post in Rogue.
+     * Execute the job to create a SoftEdge post.
      *
      * @return array
      */
@@ -89,7 +89,7 @@ class ImportSoftEdgeRecord implements ShouldQueue
     }
 
     /**
-     * Parse the call and return details we want to store in Rogue as a JSON object.
+     * Parse the call and return details we want to store as a JSON object.
      *
      * @param array $call
      */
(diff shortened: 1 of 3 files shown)


mendelB (approved):
💥

app/Jobs/Imports/ImportSoftEdgeRecord.php
```diff
@@ -0,0 +1,115 @@
...
+class ImportSoftEdgeRecord implements ShouldQueue
+{
+    use Dispatchable, InteractsWithQueue, Queueable;
+
+    /**
+     * The call parameters sent from SoftEdge.
+     *
+     * @var array
+     */
+    protected array $parameters;
```
weerd, line 25: Is `array` supposed to be after the `protected` keyword? Maybe this is just a newer PHP thing?

DFurnes: Yep, this just indicates the [type of the property](https://www.php.net/manual/en/language.oop5.properties.php#language.oop5.properties.typed-properties) in a way that the compiler can enforce.

weerd: Ah nice! I only noticed it cause VSCode has it underlined in red! Thanks for the tip!

PR #1206, opened 2021-05-13
Add partner-specific API endpoints from Chompy.

DFurnes:
### What's this PR do?

This pull request adds partner API endpoints for CallPower & SoftEdge (which kick off each of these import jobs, added in #1200 and #1205 respectively). These are facades used by these third-party partners to interface with our internal APIs.

I separated out this work into individual commits & wrote test suites for each endpoint.

I also copied over Chompy's documentation & re-formatted to match Northstar's existing documentation (which use JS request parameter syntax & copy-pasteable cURL examples for request payload with appropriate headers).

We'll need to provide updated API endpoints to these partners when we decommission Chompy!

### Relevant tickets

References [Pivotal #177733275](https://www.pivotaltracker.com/story/show/177733275).

### Checklist

### DIFF
--- a/app/Http/Middleware/AuthenticateWithApiKey.php
+++ b/app/Http/Middleware/AuthenticateWithApiKey.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace App\Http\Middleware;
+
+use Closure;
+use Illuminate\Auth\AuthenticationException;
+
+class AuthenticateWithApiKey
+{
+    /**
+     * Valid API key headers & secrets.
+     *
+     * @var array
+     */
+    protected array $keys;
+
+    public function __construct()
+    {
+        $this->keys = [
+            'X-DS-CallPower-API-Key' => config('auth.partners.callpower'),
+            'X-DS-SoftEdge-API-Key' => config('auth.partners.softedge'),
+        ];
+    }
+
+    /**
+     * Handle an incoming request.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure  $next
+     * @param  $headers
+     * @return mixed
+     *
+     * @throws \Illuminate\Auth\AuthenticationException
+     */
+    public function handle($request, Closure $next, ...$headers)
+    {
+        // Check the header(s) for authorizing this route. If any match,
+        // we're good! If not, tell this request to hit the road.
+        foreach ($headers as $header) {
+            // Ensure that if we haven't set an environment variable for this
+            // environment that we throw an error rather than leave these ungated:
+            if (empty($this->keys[$header])) {
+                throw new \Exception(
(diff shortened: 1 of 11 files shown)


app/Http/Middleware/AuthenticateWithApiKey.php
```diff
@@ -0,0 +1,55 @@
...
+     *
+     * @throws \Illuminate\Auth\AuthenticationException
+     */
+    public function handle($request, Closure $next, ...$headers)
+    {
+        // Check the header(s) for authorizing this route. If any match,
+        // we're good! If not, tell this request to hit the road.
+        foreach ($headers as $header) {
+            // Ensure that if we haven't set an environment variable for this
+            // environment that we throw an error rather than leave these ungated:
```
katiecrane, line 41: Ah I like this addition!

tests/Http/Partners/CallPowerTest.php
```diff
@@ -0,0 +1,140 @@
...
+                'call_timestamp' => '2017-11-09 06:34:01.185035',
+                'call_duration' => 50,
+                'campaign_target_name' => 'Mickey Mouse',
+                'campaign_target_title' => 'Representative',
+                'campaign_target_district' => 'FL-7',
+                'callpower_campaign_name' => 'Test',
+                'number_dialed_into' => '+12028519273',
+            ],
+            [
+                'X-DS-CallPower-API-Key' => 'let-me-in!!!!', // <-- shockingly, not the token!
```
katiecrane, line 132: 😂

documentation/endpoints/partners/callpower.md
```diff
@@ -0,0 +1,85 @@
...
+  campaign_target_title: String,
+
+  // The district of the target the user called.
+  campaign_target_district: String,
+
+  // The CallPower campaign name given in CallPower. (required)
+  callpower_campaign_name: String,
+
+  // The number the user called to connect to the target. (required)
+  number_dialed_into: String,
```
katiecrane, line 43: We don't validate that this field is a string, should we?

katiecrane: Oh same with `mobile`! I'm thinking this is intentional but curious as to why!

katiecrane (approved):
Thanks for adding docs too! I had one non-blocking question!

documentation/endpoints/partners/callpower.md
DFurnes: Hmm, that's a really good catch! 👁️ And I'm not sure!

I'm thinking it might be helpful to log a few incoming request payloads from Chompy so that we can test them out in these new Northstar jobs & ensure they make the same posts in our system. That'd be a good chance to see if this is coming in as a string or not & potentially update our validation rules!

PR #1207, opened 2021-05-19
Adding Import RTV Record Job

weerd:
### What's this PR do?

This pull request provides an initial pass at adding the job and related classes from Chompy for Importing Rock The Vote records into the system.

It adds some initial tests that touch on most of the methods in the `ImportRockTheVote.php` job and that touch on methods in other classes.

👀 

While moving this job over from Chompy into Northstar, there were a number of approaches that could be refactored or simplified/improved now that it resides in Northstar and does not depend on Gateway to make API calls to obtain the data it then works with. I did manage to make a few improvements, but there are a number of other improvements I will be leaving for a cleanup sprint since the main objective is to get this all working in Northstar.

### Relevant tickets

### Checklist

### DIFF
--- a/app/Imports/RockTheVoteRecord.php
+++ b/app/Imports/RockTheVoteRecord.php
@@ -235,10 +235,10 @@ class RockTheVoteRecord
                 $userId = $value[1];
             } elseif ($key === 'group_id') {
                 $result['group_id'] = (int) $value[1];
+            } elseif (
                 //  If referral parameter is set to true, the user parameter belongs to the referring
                 //  user, not the user that should be associated with this voter registration record.
                 //  Expected key: "referral"
-            } elseif (
                 ($key === 'referral' || $key === 'refferal') &&
                 str_to_boolean($value[1])
             ) {
--- a/app/Imports/RockTheVoteRecord.php
+++ b/app/Imports/RockTheVoteRecord.php
@@ -27,67 +27,56 @@ class RockTheVoteRecord
      */
     public static $startedRegistrationFieldName = 'Started registration';
 
-    /**
-     * @var array
-     */
-    public $config;
+    public array $config;
+
+    public array $postData;
+
+    public array $trackingSource;
+
+    public array $userData;
 
     /**
-     * @var array
-     */
-    public $postData;
-
-    /**
-     * @var array
-     */
-    public $trackingSource;
-
-    /**
-     * @var array
-     */
-    public $userData;
-
-    /**
-     * Parses values to send to create a RTV Record from given CSV data, and provided config values.
+     * Parses values to send to create a RTV Record from given CSV payload data,
+     * and provided config values.
      *
-     * @param array $record
+     * @param array $payload
      * @param array $config
      */
-    public function __construct($record, $config = null)
+    public function __construct($payload, $config = null)
     {
-        $this->validate($record);
+        $this->validate($payload);
 
         $this->config =
             $config ?? ImportType::getConfig(ImportType::$rockTheVote);
 
         $rtvStatus = $this->parseVoterRegistrationStatus(
-            $record['Status'],
-            $record['Finish with State'],
+            $payload['Status'],
+            $payload['Finish with State'],
         );
 
         // Used in UserData and PostData.
         $this->trackingSource = $this->parseTrackingSource(
-            $record['Tracking Source'],
+            $payload['Tracking Source'],
         );
 
-        $this->setUserData($record, $rtvStatus);
+        $this->setUserData($payload, $rtvStatus);
 
-        $this->setPostData($record, $rtvStatus);
+        $this->setPostData($payload, $rtvStatus);
     }
 
     /**
-     * Set data that will be used to create a new post for record if post
+     * Set payload data that will be used to create a new post for record if post
      * does not already exist.
      *
-     * @var array
-     * @var string
+     * @param array $payload
+     * @param string $status
      */
-    private function setPostData($record, $status)
+    private function setPostData($payload, $status)
     {
         $this->postData = [
             'source' => $this->config['post']['source'],
             'source_details' => null,
-            'details' => $this->parsePostDetails($record),
+            'details' => $this->parsePostDetails($payload),
             'status' => $status,
             'type' => $this->config['post']['type'],
             'action_id' => $this->config['post']['action_id'],
@@ -100,16 +89,17 @@ class RockTheVoteRecord
     }
 
     /**
-     * Set data that will be used to find or create a new user for the record.
+     * Set payload data that will be used to find or create a new user for the record.
      *
-     * @var array
-     * @var string
+     * @param array $payload
+     * @param string $status
+     * @return void
      */
-    private function setUserData($record, $status)
+    private function setUserData($payload, $status)
     {
         $this->userData = [
-            'addr_zip' => $record['Home zip code'],
-            'email' => $record['Email address'],
+            'addr_zip' => $payload['Home zip code'],
+            'email' => $payload['Email address'],
             'mobile' => null,
             'sms_status' => null,
             // Source is required in order to set the source detail.
@@ -133,12 +123,12 @@ class RockTheVoteRecord
             return;
         }
 
-        $emailOptIn = str_to_boolean($record['Opt-in to Partner email?']);
+        $emailOptIn = str_to_boolean($payload['Opt-in to Partner email?']);
 
         $this->userData = array_merge($this->userData, [
-            'addr_street1' => $record['Home address'],
-            'addr_street2' => $record['Home unit'],
-            'addr_city' => $record['Home city'],
+            'addr_street1' => $payload['Home address'],
+            'addr_street2' => $payload['Home unit'],
+            'addr_city' => $payload['Home city'],
             'email_subscription_status' => $emailOptIn,
             'email_subscription_topics' => $emailOptIn
                 ? explode(
@@ -146,18 +136,18 @@ class RockTheVoteRecord
                     $this->config['user']['email_subscription_topics'],
                 )
                 : [],
-            'first_name' => $record['First name'],
-            'last_name' => $record['Last name'],
+            'first_name' => $payload['First name'],
+            'last_name' => $payload['Last name'],
             'mobile' =>
-                isset($record[static::$mobileFieldName]) &&
-                is_phone_number($record[static::$mobileFieldName])
-                    ? $record[static::$mobileFieldName]
+                isset($payload[static::$mobileFieldName]) &&
+                is_phone_number($payload[static::$mobileFieldName])
+                    ? $payload[static::$mobileFieldName]
                     : null,
         ]);
 
         // If a mobile was provided, set SMS subscription per opt-in value.
         if ($this->userData['mobile']) {
-            $smsOptIn = str_to_boolean($record[static::$smsOptInFieldName]);
+            $smsOptIn = str_to_boolean($payload[static::$smsOptInFieldName]);
 
             $this->userData['sms_status'] = $smsOptIn
                 ? SmsStatus::$active
@@ -245,7 +235,7 @@ class RockTheVoteRecord
                 $userId = $value[1];
             } elseif ($key === 'group_id') {
                 $result['group_id'] = (int) $value[1];
-            //  If referral parameter is set to true, the user parameter belongs to the referring
+                //  If referral parameter is set to true, the user parameter belongs to the referring
                 //  user, not the user that should be associated with this voter registration record.
                 //  Expected key: "referral"
             } elseif (
@@ -286,17 +276,17 @@ class RockTheVoteRecord
     }
 
     /**
-     * Parse the record for extra details and return them as a JSON object.
+     * Parse the record payload for extra details and return them as a JSON object.
      *
-     * @param  array $record
+     * @param  array $payload
      * @return string
      */
-    private function parsePostDetails($record)
+    private function parsePostDetails($payload)
     {
         $result = [];
 
         foreach (config('import.rock_the_vote.post.details') as $key) {
-            $result[$key] = $record[$key];
+            $result[$key] = $payload[$key];
         }
 
         return json_encode($result);
--- a/app/Imports/RockTheVoteRecord.php
+++ b/app/Imports/RockTheVoteRecord.php
@@ -79,8 +79,8 @@ class RockTheVoteRecord
      * Set data that will be used to create a new post for record if post
      * does not already exist.
      *
-     * @var array $record
-     * @var string $status
+     * @var array
+     * @var string
      */
     private function setPostData($record, $status)
     {
@@ -102,8 +102,8 @@ class RockTheVoteRecord
     /**
      * Set data that will be used to find or create a new user for the record.
      *
-     * @var array $record
-     * @var string $status
+     * @var array
+     * @var string
      */
     private function setUserData($record, $status)
     {
@@ -245,7 +245,7 @@ class RockTheVoteRecord
                 $userId = $value[1];
             } elseif ($key === 'group_id') {
                 $result['group_id'] = (int) $value[1];
-                //  If referral parameter is set to true, the user parameter belongs to the referring
+            //  If referral parameter is set to true, the user parameter belongs to the referring
                 //  user, not the user that should be associated with this voter registration record.
                 //  Expected key: "referral"
             } elseif (
--- a/app/Imports/RockTheVoteRecord.php
+++ b/app/Imports/RockTheVoteRecord.php
@@ -28,15 +28,165 @@ class RockTheVoteRecord
     public static $startedRegistrationFieldName = 'Started registration';
 
     /**
-     * Parses values to send to DS API from given CSV record, using given config.
+     *
+     *
+     * @var array
+     */
+    public $config;
+
+    /**
+     *
+     *
+     * @var array
+     */
+    public $postData;
+
+    /**
+     *
+     *
+     * @var array
+     */
+    public $trackingSource;
+
+    /**
+     *
+     *
+     * @var array
+     */
+    public $userData;
+
+    /**
+     * Parses values to send to create a RTV Record from given CSV data, and provided config values.
      *
      * @param array $record
      * @param array $config
      */
     public function __construct($record, $config = null)
     {
-        $validator = Validator::make($record, [
-            static::$startedRegistrationFieldName => 'required|date',
+        $this->validate($record);
+
+        $this->config =
+            $config ?? ImportType::getConfig(ImportType::$rockTheVote);
+
+        $rtvStatus = $this->parseVoterRegistrationStatus(
+            $record['Status'],
+            $record['Finish with State'],
+        );
+
+        // Used in UserData and PostData
+        $this->trackingSource = $this->parseTrackingSource(
+            $record['Tracking Source'],
+        );
+
+        $this->setUserData($record, $rtvStatus);
+
+        $this->setPostData($record, $rtvStatus);
+    }
+
+    /**
+     *
+     *
+     * @var array $record
+     * @var string $status
+     */
+    private function setPostData($record, $status)
+    {
+        $this->postData = [
+            'source' => $this->config['post']['source'],
+            'source_details' => null,
+            'details' => $this->parsePostDetails($record),
+            'status' => $status,
+            'type' => $this->config['post']['type'],
+            'action_id' => $this->config['post']['action_id'],
+        ];
+
+        $this->postData['group_id'] = $this->trackingSource['group_id'];
+
+        $this->postData['referrer_user_id'] =
+            $this->trackingSource['referrer_user_id'];
+    }
+
+    /**
+     *
+     *
+     * @var array $record
+     * @var string $status
+     */
+    private function setUserData($record, $status)
+    {
+        $this->userData = [
+            'addr_zip' => $record['Home zip code'],
+            'email' => $record['Email address'],
+            'mobile' => null,
+            'sms_status' => null,
+            // Source is required in order to set the source detail.
+            'source' => 'northstar', // @TODO: should this be something more specific? env?
+            'source_detail' => $this->config['user']['source_detail'],
+            'voter_registration_status' => Str::contains($status, 'register')
+                ? 'registration_complete'
+                : $status,
+        ];
+
+        $this->userData['id'] = $this->trackingSource['user_id'];
+
+        $this->userData['referrer_user_id'] =
+            $this->trackingSource['referrer_user_id'];
+
+        //  At step 1, a user has only provided their email and zip, but Rock The Vote will
+        //  sometimes mysteriously send through data for fields populated in later steps.
+        //  We do not want to save any other data until the status is at least step 2.
+        //  @see /docs/imports/README.md#status
+        if ($status === 'step-1') {
+            return;
+        }
+
+        $emailOptIn = str_to_boolean($record['Opt-in to Partner email?']);
+
+        $this->userData = array_merge($this->userData, [
+            'addr_street1' => $record['Home address'],
+            'addr_street2' => $record['Home unit'],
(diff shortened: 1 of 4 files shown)


app/Imports/RockTheVoteRecord.php
```diff
@@ -28,114 +28,184 @@ class RockTheVoteRecord
...
             'email' => $record['Email address'],
+            'mobile' => null,
+            'sms_status' => null,
             // Source is required in order to set the source detail.
-            'source' => config(
-                'services.northstar.client_credentials.client_id',
-            ),
-            'source_detail' => $config['user']['source_detail'],
-            'voter_registration_status' => Str::contains($rtvStatus, 'register')
+            'source' => 'northstar', // @TODO: should this be something more specific? env?
```
weerd, line 116: Not sure if this needs to be a different value or more specific? Indicate the environment?

app/Imports/RockTheVoteRecord.php
```diff
@@ -28,114 +28,184 @@ class RockTheVoteRecord
...
             'email' => $record['Email address'],
+            'mobile' => null,
+            'sms_status' => null,
             // Source is required in order to set the source detail.
-            'source' => config(
-                'services.northstar.client_credentials.client_id',
-            ),
-            'source_detail' => $config['user']['source_detail'],
-            'voter_registration_status' => Str::contains($rtvStatus, 'register')
+            'source' => 'northstar',
```
weerd, line 116: Not sure if this needs to be a different value or more specific? Indicate the environment?

app/Jobs/Imports/ImportRockTheVoteRecord.php
```diff
@@ -0,0 +1,679 @@
...
+        PostManager $posts,
+        Registrar $registrar,
+        SignupManager $signups
+    ) {
+        if (!$this->record && $this->exception) {
+            $this->importFile->incrementSkipCount();
+
+            return [
+                'response' => ['message' => $this->exception],
+            ];
```
weerd, line 94: Not sure what actually captures this to then utilize the message 🤔

app/Jobs/Imports/ImportRockTheVoteRecord.php
```diff
@@ -0,0 +1,679 @@
...
+            $user = $this->registrar->resolve([
+                'mobile' => $userData['mobile'],
+            ]);
+
+            if ($user) {
+                info('Found user by mobile: ', ['user' => $user->id]);
+            }
+
+            return $user;
+        }
```
weerd, line 195: Could these be consolidated with one `resolve()` call for multiple matching fields? 🤔

app/Jobs/Imports/ImportRockTheVoteRecord.php
```diff
@@ -0,0 +1,679 @@
...
+    {
+        $user = $this->registrar->register($this->record->userData);
+
+        $action = Action::find($this->record->postData['action_id']);
+
+        $post = $this->createPost($user, $action, $signups, $posts);
+
+        if (!$post) {
+            return [];
+        }
```
weerd, line 352: Is it fine to return empty array or maybe throw an error?

app/Jobs/Imports/ImportRockTheVoteRecord.php
```diff
@@ -0,0 +1,679 @@
...
+
+        $post = $this->getPost($user, $action);
+
+        $post = $post
+            ? $this->updatePost($post)
+            : $this->createPost($user, $action, $signups, $posts);
+
+        if (!$post) {
+            return [];
+        }
```
weerd, line 394: Same as https://github.com/DoSomething/northstar/pull/1207/files#r635633559

app/Jobs/Imports/ImportRockTheVoteRecord.php
```diff
@@ -0,0 +1,679 @@
...
+
+        if (!$signup) {
+            $signup = $signups->create(
+                $payload,
+                $user->id,
+                $action->campaign->id,
+            );
+        }
+
+        $post = $posts->create($payload, $signup->id);
```
weerd, line 296: After post is created should an `accepted` status be set? Not sure what that means for RTV records...

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
```diff
@@ -0,0 +1,158 @@
...
+    /**
+     * Test that user and post are created if user is not found.
+     */
+    public function testCreatesUserIfUserNotFound()
+    {
+        $record = $this->makeFakeRockTheVoteReportRecord();
+
+        $action = factory(Action::class)
+            ->states('voter-registration')
+            ->create();
```
weerd, line 86: Should the action type be different from what is defined in the factory state? Currently set to `attend-event`.

app/Jobs/Imports/ImportRockTheVoteRecord.php
```diff
@@ -0,0 +1,679 @@
...
+            if (!isset($post['details'])) {
+                continue;
+            }
+
+            $details = json_decode($post['details']);
+
+            if ($details->{$key} === $importRecordDate) {
+                return $post;
+            }
+        }
```
DFurnes, line 258: I'm curious if we could use [Eloquent's `array` cast](https://laravel.com/docs/8.x/eloquent-mutators#array-and-json-casting) and [`firstWhere`](https://laravel.com/docs/8.x/collections#method-first-where) to simplify this to something like:

```php
$importRecordDate = $this->record->getPostDetails()[$key];

return $posts->firstWhere("details.$key", $importRecordDate);
```

(This would probably require plenty of other refactoring, so more noting for later than anything else!)

app/Jobs/Imports/ImportRockTheVoteRecord.php
```diff
@@ -0,0 +1,679 @@
...
+        PostManager $posts
+    ) {
+        $user = $this->updateStatusIfChanged($user);
+
+        // If import record has a mobile number provided, update SMS subscription.
+        if (isset($this->record->userData['mobile'])) {
+            $user = $this->updateSmsSubscriptionIfChanged($user);
+        }
+
+        $action = Action::find($this->record->postData['action_id']);
```
DFurnes, line 384: Should this be a `findOrFail`?

DFurnes (commented):
This was an epic migration! 💯

app/Jobs/Imports/ImportRockTheVoteRecord.php
DFurnes: I believe there's a mapping of different statuses based on where in the voter-registration process the user is (e.g. Step 1 vs. Step 2, etc.). I _thiiiiiiink_ I remember seeing that handled elsewhere in here (and maybe it defaults to `pending` for the first step?)

app/Imports/RockTheVoteRecord.php
```diff
@@ -28,114 +28,184 @@ class RockTheVoteRecord
...
+
+    /**
+     * @var array
+     */
+    public $trackingSource;
+
+    /**
+     * @var array
+     */
+    public $userData;
```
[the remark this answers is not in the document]
weerd: Ya I dig it! I need to update my mac's PHP otherwise VSCode yells at me, but not a big deal!

app/Imports/RockTheVoteRecord.php
```diff
@@ -28,114 +28,184 @@ class RockTheVoteRecord
...
+
+    /**
+     * Parses values to send to create a RTV Record from given CSV data, and provided config values.
      *
      * @param array $record
      * @param array $config
      */
     public function __construct($record, $config = null)
```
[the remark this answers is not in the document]
weerd: Ya, makes sense to me! Seems like a good improvement. Might go with `$payload` since it's a bit more generic and used elsewhere.

app/Jobs/Imports/ImportRockTheVoteRecord.php
DFurnes: Hmm yeah, I thiiiink that should be possible but we'll want to make sure we test it carefully! I think the main potential gotcha we'd have to keep our eyes out for is making sure we resolve a user by email even if their mobile doesn't match (which if I remember correctly, was actually a relatively common scenario).

I wonder if we should make a `resolveByAny` (?) method that does this sort of "looser" match?

app/Jobs/Imports/ImportRockTheVoteRecord.php
DFurnes: I think this is used by the web-based test interface (which fires off this job synchronously and then returns the jobs outputted array to the user). It's somewhat awkward since we otherwise don't return from queued jobs (and otherwise run these all asynchronously on queue workers, rather than the "main" web process)...

I wouldn't mind thinking about if there's a cleaner way to expose this sort of debug info when we tackle that UI.

app/Jobs/Imports/ImportRockTheVoteRecord.php
weerd: I was wondering that too and in a few other places, but not sure yet the full scope of how this job is used, if used elsewhere (seems like with web view side, it ends up returning some response at a later point in the process), and if failing would kill things too early? 🤷

DFurnes: Hmm, yeah I'm not sure either! What if we just left a `TODO:` to investigate when we get there?

app/Jobs/Imports/ImportRockTheVoteRecord.php
weerd: Agreed. It was odd seeing this in the context of a queue job, and relates to a few other additions made to the job in service of when it's run via the web-based interface. Feels like there's a better way™ to handle all that. Makes sense to reassess once tackling the UI!

app/Jobs/Imports/ImportRockTheVoteRecord.php
```diff
@@ -0,0 +1,679 @@
...
+{
+    /**
+     * @var array
+     */
+    protected $config;
+
+    /**
+     * @var \Illuminate\Validation\ValidationException
+     */
+    protected $exception;
```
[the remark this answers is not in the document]
weerd: Yep! Good catch 🔍

app/Jobs/Imports/ImportRockTheVoteRecord.php
DFurnes: Wait actually, I think I was misremembering how this works! It looks like it does check for [any one of](https://github.com/DoSomething/northstar/blob/163b45e880438ab6740dc1ed2ef37bd3def756a8/app/Auth/Registrar.php#L129-L142) the given credentials, so it might be good as-is. (That raises whether _that_ is expected behavior or not haha, but could be helpful here!)

app/Jobs/Imports/ImportRockTheVoteRecord.php
weerd: Sounds good!

app/Jobs/Imports/ImportRockTheVoteRecord.php
weerd: Sound good! I'll add it as a `@TODO` and reference your comment here for cleanup sprint. I did a little bit of refactoring here to reduce the number of variables to just `$importRecordDate` but didn't touch further to reduce amount of refactoring work 😅

app/Jobs/Imports/ImportRockTheVoteRecord.php
weerd: 🤔 ... I have more tests that need to be added for the RTV Import from Chompy, so maybe I'll make this revision and add a test to ensure it's working as expected in an upcoming PR!

app/Jobs/Imports/ImportRockTheVoteRecord.php
katiecrane: I think I know about this! I believe the `parseVoterRegistrationStatus()` function determines what the status should be, but you can also poke around and see what statuses look like for VR posts in the Rogue database.

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
katiecrane: I think I might be able to help here - there is at least one action in the Rogue database with `post_type` of `voter-reg` that has `action_type` of `attend-event`, so I believe it should be okay to use that here. Not 100% sure if that's what you were getting at though!

weerd: @katiecrane ya I think so. Just seemed weird to me that `attend-event` would apply for a RTV import. I figured it might be something else but not sure what, but maybe whatever an online action of this kind would be... but only option that would seem to apply is `other` 🤷 

https://github.com/DoSomething/northstar/blob/main/app/Types/ActionType.php#L7-L17

app/Jobs/Imports/ImportRockTheVoteRecord.php
weerd: Thanks @katiecrane! I checked that method and it mostly handles to differentiate between `register-OVR` or `register-form` if status from RTV is `complete` along with if the member finished with the state form or not. I did find this helpful in the Chompy wiki though: 
https://github.com/DoSomething/chompy/tree/main/docs/imports#status-hierarchy

It was a bit confusing in code cause there's `status` but also `voter-registration-status` and there's overlap in the values but then very specific differences in certain values depending on context used. 🤪

katiecrane: Ah yeah I see what you're saying, `status` on the `post` but also `voter_registration_status` on the `user`. So the `parseVoterRegistrationStatus` method I believe determines what the `post` `status` will be - it's more specific. We use in most cases the status provided by RTV BUT for completed registrations we want to know if they completed it online or mailed in the form which is why that differentiation is added in.

 On the `user` for `voter_registration_status` we don't need it to be as detailed and that's where the confusing overlap but not full overlap comes in 😵 Writing this out partially to make sure I understand it!

app/Imports/RockTheVoteRecord.php
katiecrane: I don't know the answer here but folks on the data side might have thoughts - I forget if this was expected to be a particular value or not

app/Imports/RockTheVoteRecord.php
```diff
@@ -27,115 +27,175 @@ class RockTheVoteRecord
...
+
+        $this->userData['referrer_user_id'] =
+            $this->trackingSource['referrer_user_id'];
+
+        //  At step 1, a user has only provided their email and zip, but Rock The Vote will
+        //  sometimes mysteriously send through data for fields populated in later steps.
+        //  We do not want to save any other data until the status is at least step 2.
+        //  @see /docs/imports/README.md#status
+        if ($status === 'step-1') {
+            return;
```
katiecrane, line 123: I like how you re-arranged this part to return if on `step-1`!

app/Imports/RockTheVoteRecord.php
DFurnes: We chatted on this a bit in #1208! We're going to hardcode in the `importer-client` client name (that had previously been used by Chompy's queue workers) so this value stays consistent in Quasar, etc.

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
DFurnes: I wonder if we decided on that back when we thought a lot of voter registration was going to be in-person events. (A simpler time!) It does feel like it could be reasonable to have a distinct action type for this sort of thing...

DFurnes: Might be worth revisiting when we think about other taxonomy stuff with scholarships!

app/Jobs/Imports/ImportRockTheVoteRecord.php
```diff
@@ -0,0 +1,679 @@
...
+        if (!$user) {
+            return $this->importRecordAsNewUser($signups, $posts);
+        }
+
+        if (RockTheVoteLog::getByRecord($this->record, $user)) {
+            $this->skipImportingRecord($user);
+        }
+
+        $this->importRecordForExistingUser($user, $signups, $posts);
+    }
```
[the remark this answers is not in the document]
katiecrane: Huge +1 to this, makes it a lot easier to follow the high level flow!

app/Jobs/Imports/ImportRockTheVoteRecord.php
```diff
@@ -0,0 +1,682 @@
...
+     * @return \App\Models\User
+     */
+    public function updateStatusIfChanged(User $user)
+    {
+        $userData = $this->record->userData;
+
+        $shouldUpdateStatus = self::shouldUpdateStatus(
+            $user->voter_registration_status,
+            $userData['voter_registration_status'],
+        );
```
katiecrane, line 468: Pulling this out into a variable makes this easier to follow, good call!

app/Jobs/Imports/ImportRockTheVoteRecord.php
weerd: Yeah, it definitely took a bit of reading in the wiki and code and then re-reading to make sure I understood better, haha!

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
weerd: > (A simpler time!)

Indeed 😔

app/Jobs/Imports/ImportRockTheVoteRecord.php
weerd: Ya, I ❤️  prettier, but sometimes it makes some code less readable (if technically it considers it prettier, haha). Thus popping out a one-time variable that helps with readability!

app/Jobs/Imports/ImportRockTheVoteRecord.php
```diff
@@ -0,0 +1,682 @@
...
+                $this->getSmsSubscriptionPayload($user),
+            );
+        }
+
+        // Check if another user already owns the import mobile.
+        $this->mobileUser = $this->registrar->resolve([
+            'mobile' => $this->record->userData['mobile'],
+        ]);
+
+        dd(['poopie', $this->mobileUser]);
```
katiecrane, line 505: 😂 😂 😂  this seriously made my day!

weerd: 😂   whoops!

You win a prize for finding one of my "poopie" die dumps, lol!

katiecrane: Based on the key word here I think I might not want the prize... 😂

weerd: haha! 😅

app/Jobs/Imports/ImportRockTheVoteRecord.php
```diff
@@ -0,0 +1,680 @@
...
+                $user->id,
+                $action->campaign->id,
+            );
+        }
+
+        $post = $posts->create($payload, $signup->id);
+
+        if ($post) {
+            $post->source = 'importer-client';
+            $post->status = $this->record->postData['status'];
```
katiecrane, line 297: I was figuring these weren't `fillable` but they are - why update here rather than as part of `$payload` when creating the post, is there something else going on behind the scenes?

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
```diff
@@ -0,0 +1,158 @@
...
+    /**
+     * Test that record is skipped if the RTV data is invalid.
+     *
+     * @return void
+     */
+    public function testSkipsImportIfInvalidRecordData()
+    {
+        $record = $this->makeFakeRockTheVoteReportRecord([
+            'First name' => 'Poopie',
+            'Last name' => 'Doopie',
```
katiecrane, line 61: Possibly my favorite two lines of code to ever exist

app/Jobs/Imports/ImportRockTheVoteRecord.php
weerd: Yep yep,  @DFurnes took this approach with the [CallPower and SoftEdge import jobs PR](https://github.com/DoSomething/northstar/pull/1208) and I followed suit. He found that while they are mass-assignable in the `Post`, the `PostManager` (which we use for creating posts for these jobs) does a bunch of it's own reading and setting of values and sets `source` to a client ID thus overriding anything we push through during `create()`. He's got a bit more info in his description in that PR!

We chatted a bit during our 1:1's that during a cleanup sprint it might make sense to re-look at a lot of these classes and see where we can simplify/update approaches and reassess prior work.

katiecrane: Ah that makes sense, and a good catch too! Confusing!

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
weerd: LOL 😂  Totes forgot that I put those for the test user! I can change if we want or leave it to brighten someone else's day when they encounter it, haha

katiecrane: hahaha totally up to you, but I am here for it

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
```diff
@@ -0,0 +1,158 @@
...
+        config(['import.rock_the_vote.post.action_id' => $action->id]);
+
+        $importFile = factory(ImportFile::class)
+            ->states('rock_the_vote')
+            ->create();
+
+        ImportRockTheVoteRecord::dispatch($record, $importFile);
+
+        $this->assertMysqlDatabaseHas('posts', [
+            'northstar_id' => $user->id,
```
katiecrane, line 148: Is this how we know another `user` was not created? Because the `post` has the existing `northstar_id`?

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
```diff
@@ -0,0 +1,158 @@
+<?php
+
+use App\Imports\RockTheVoteRecord;
+use App\Jobs\Imports\ImportRockTheVoteRecord;
+use App\Models\Action;
+use App\Models\ImportFile;
+use App\Models\User;
+use Carbon\Carbon;
+
+class ImportRockTheVoteRecordTest extends TestCase
```
katiecrane, line 10: Chompy has loads of additional tests in the corresponding file - was wondering if the plan is to bring those over as well just perhaps in another PR?

weerd: Yep yep! This PR felt dense enough as it is, so I was doing just these initial tests that would force me to bring over most of the functionality to make them pass. More tests in upcoming PRs but hopefully they'll go a bit faster thanks to these initial ones that cover a large amount of what is in the import job!

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
weerd: Yep yep, the post has ID of the factory created user and the post type is `voter-reg` :)

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
weerd: Decided to change it to make it a more DS-style test name. Will help me in the future so when I use an impromptu "poopie" for debugging, I'm not running into old ones when doing a find in VSCode, haha!

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
weerd: Made a ticket for this: <https://www.pivotaltracker.com/story/show/178253251>

katiecrane (approved):
Great work on this enormous chunk of business logic! I don't think I have any other questions!

PR #1208, opened 2021-05-20
Ensure correct source is set for imported records.

DFurnes:
### What's this PR do?

This pull request ensures that imported records (via SoftEdge or CallPower) continue to be marked with the `importer-client` source when created via Northstar's local importer jobs. This ensures we can continue to attribute these correctly in our admin panels & data warehouse.

I added a (failing) test for this in 39c828f & fixed the issue in 835de21.

We actually _do_ allow mass-assigning `source` [on the Post](https://github.com/DoSomething/northstar/blob/b28242793bee2af1270557b5bc32de4c09813b07/app/Models/Post.php#L72), so you'd think this would have worked! I forgot, however, that we also only read/set specific fields [in the PostRepository](https://github.com/DoSomething/northstar/blob/b28242793bee2af1270557b5bc32de4c09813b07/app/Repositories/PostRepository.php#L91-L121) so this field is discarded before Eloquent ever gets to see it.

### Relevant tickets

References [Pivotal #177733275](https://www.pivotaltracker.com/story/show/177733275).

### Checklist

### DIFF
--- a/app/Jobs/Imports/ImportCallPowerRecord.php
+++ b/app/Jobs/Imports/ImportCallPowerRecord.php
@@ -91,7 +91,6 @@ class ImportCallPowerRecord implements ShouldQueue
             'action_id' => $action->id,
             'type' => 'phone-call',
             'quantity' => 1,
-            'source' => 'importer-client', // @TODO: We should type this.
             'source_details' => 'CallPower',
             'details' => $details,
         ];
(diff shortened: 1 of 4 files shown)


app/Jobs/Imports/ImportCallPowerRecord.php
```diff
@@ -105,6 +104,8 @@ public function handle(
         $post = $posts->create($payload, $signup->id);
 
         if ($post) {
+            // Set fields that repository does not allow mass-assigning:
+            $post->source = 'importer-client'; // @TODO: We should type this.
```
weerd, line 108: Hmmm, I take it I should use this same approach for the RTV Import job? https://github.com/DoSomething/northstar/pull/1207#discussion_r635553372

weerd (approved):
👍  Looks good! I'll make the same edits on my PR for RTV importer!

app/Jobs/Imports/ImportCallPowerRecord.php
DFurnes: Yeah, I think this is probably the easiest way to keep everything working consistently.

PR #1209, opened 2021-05-25
More ImportRockTheVoteRecord Tests 🚥

weerd:
### What's this PR do?

This pull request add more tests for the `ImportRockTheVoteRecord` job.

👀 

I capped it at this amount for now. There will be another PR with more tests but wanted to keep this one manageable!

### Relevant tickets

### Checklist

### DIFF
--- a/app/Jobs/Imports/ImportRockTheVoteRecord.php
+++ b/app/Jobs/Imports/ImportRockTheVoteRecord.php
@@ -97,7 +97,7 @@ class ImportRockTheVoteRecord extends Job
         }
 
         if (RockTheVoteLog::getByRecord($this->record, $user)) {
-            $this->skipImportingRecord($user);
+            return $this->skipImportingRecord($user);
         }
 
         $this->importRecordForExistingUser($user, $signups, $posts);
--- a/tests/CreatesApplication.php
+++ b/tests/CreatesApplication.php
@@ -56,9 +56,6 @@ trait CreatesApplication
         $this->gambitMock = $this->mock(\App\Services\Gambit::class);
         $this->gambitMock->shouldReceive('deleteUser');
 
-        $this->fastlyMock = $this->mock(\App\Services\Fastly::class);
-        $this->fastlyMock->shouldReceive('purge');
-
         // Configure a mock for GraphQL calls.
         $this->graphqlMock = $this->mock(\App\Services\GraphQL::class);
         $this->graphqlMock->shouldReceive('getSchoolById')->andReturn([
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -137,9 +137,11 @@ abstract class TestCase extends Illuminate\Foundation\Testing\TestCase
         });
 
         // We should not have tracked this event for the provided user:
-        return $this->customerIoMock
-            ->shouldNotHaveReceived('trackEvent')
-            ->with($userMatcher, $eventName, Mockery::any());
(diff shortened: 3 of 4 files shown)


tests/CreatesApplication.php
```diff
@@ -56,9 +56,6 @@ public function createApplication()
         $this->gambitMock = $this->mock(\App\Services\Gambit::class);
         $this->gambitMock->shouldReceive('deleteUser');
 
-        $this->fastlyMock = $this->mock(\App\Services\Fastly::class);
-        $this->fastlyMock->shouldReceive('purge');
-
```
weerd, line 61: Removed this because I noticed it was a duplicate of lines 53 & 54... 🤷‍♂️

tests/TestCase.php
```diff
@@ -124,6 +124,24 @@ public function assertCustomerIoEvent(User $user, string $eventName)
...
+    {
+        $userMatcher = Mockery::on(function ($argument) use ($user) {
+            return $user->is($argument);
+        });
+
+        // We should not have tracked this event for the provided user:
+        return $this->customerIoMock
+            ->shouldNotHaveReceived('trackEvent')
+            ->with($userMatcher, $eventName, Mockery::any());
+    }
```
weerd, line 143: Having some trouble with this and likely just because I'm bad at "mocks". Anyhoo, this is a copy of `assertCustomerIoEvent()` but tweaked to expect no event triggered. It keeps throwing an error in the test runs because the return from `$this->customerIoMock->shouldNotHaveReceived('trackEvent')` ends up being `null`, so when it chains the `with()` method, it throws an error trying to call `with()` on `null`.

It passes if I essentially chop the mock call and remove line 142, so it doesn't check against the user, etc, but not sure if that's just not a correct approach.

weerd: For additional context, within the importer code, the test gets to [this point](https://github.com/DoSomething/northstar/blob/c584e0bf60c8afbea7a2cf20bf8da143589f8af9/app/Jobs/Imports/ImportRockTheVoteRecord.php#L643-L651) to determine whether to send out the password reset email.

Since the user does not have `email_subscription_status` (because they have a `Step 1` status), it passes the condition and just returns out of the method without doing anything (thus no Customer IO event).

DFurnes: Huh that's so weird! According to the [Mockery docs](http://docs.mockery.io/en/latest/reference/spies.html), it looks like they don't support chaining arguments onto `shouldNotHaveReceived` "due to limitations". 🙃 

Looks like you can do it this way, though!

```php
$this->customerIoMock->shouldNotHaveReceived('trackEvent', [$userMatcher, $eventName, Mockery::any()]);
```

weerd: Ah will try that! Thanks for the help 🙌

weerd: Noice! All green! Tests pass 🎉

app/Jobs/Imports/ImportRockTheVoteRecord.php
```diff
@@ -97,7 +97,7 @@ public function handle(
         }
 
         if (RockTheVoteLog::getByRecord($this->record, $user)) {
-            $this->skipImportingRecord($user);
+            return $this->skipImportingRecord($user);
```
weerd, line 100: Realized we need a `return` here otherwise it continues to also import record below apart from marking a skip.

lindsayrmaher (approved):
looking reasonable to me!

PR #1210, opened 2021-05-26
Add 'CreateRockTheVoteReport' job & command.

DFurnes:
### What's this PR do?

This pull request migrates Chompy's `CreateRockTheVoteReport` queue job & Artisan command into Northstar. These allow us to request new "reports" from Rock The Vote on an hourly schedule. (We'll then check up on & parse those reports in the `ImportRockTheVoteReport`, to be added later!)

I added tests for the job & command in their respective commits. I'll drop some comments in the diff, as well.

🍽️

### Relevant tickets

References [Pivotal #178301632](https://www.pivotaltracker.com/story/show/178301632).

### Checklist

### DIFF
--- a/app/Console/Commands/ImportRockTheVoteCommand.php
+++ b/app/Console/Commands/ImportRockTheVoteCommand.php
@@ -1,0 +1,48 @@
+<?php
+
+namespace App\Console\Commands;
+
+use App\Jobs\Imports\CreateRockTheVoteReport;
+use Carbon\CarbonImmutable;
+use Illuminate\Console\Command;
+use Illuminate\Support\Facades\Log;
+
+class ImportRockTheVoteCommand extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'northstar:rock-the-vote';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Imports Rock The Vote registrations from the past hour.';
+
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        // We want all registrations in the past hour (including a 30 minute overlap to
+        // ensure that we don't lose any records to a gap between scheduled jobs):
+        $now = CarbonImmutable::now();
+        $since = $now->subHours(1)->subMinutes(30);
+
+        Log::debug('Executing import command', [
+            'since' => $since,
+            'before' => $now,
+        ]);
+
+        CreateRockTheVoteReport::dispatch(
+            $since->toDateTimeString(),
+            $now->toDateTimeString(),
+        );
+    }
+}
--- a/app/Models/RockTheVoteReport.php
+++ b/app/Models/RockTheVoteReport.php
@@ -2,6 +2,7 @@
 
 namespace App\Models;
 
+use Carbon\CarbonInterface;
 use App\Services\RockTheVote;
 use Illuminate\Database\Eloquent\Model;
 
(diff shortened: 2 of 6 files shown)


app/Console/Commands/ImportRockTheVoteCommand.php
```diff
@@ -0,0 +1,48 @@
...
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        // We want all registrations in the past hour (including a 30 minute overlap to
+        // ensure that we don't lose any records to a gap between scheduled jobs):
+        $now = CarbonImmutable::now();
+        $since = $now->subHours(1)->subMinutes(30);
```
DFurnes, line 36: In the original Chompy job, we created a traditional `Carbon` object & cloned it:

```php
// We want all registrations that have started before the current datetime...
$before = Carbon::now();
$since = clone $before;

// ... and after an hour ago (adding 30 overlapping minutes to ensure we don't miss any within the internal).        
$since->subHours(1)->subMinutes(30);
```

I swapped this to use `CarbonImmutable` instead, which always returns a new immutable copy of itself (rather than mutating the object that you call `subHours` or `subMinutes` on).

app/Models/RockTheVoteReport.php
```diff
@@ -55,7 +62,7 @@ class RockTheVoteReport extends Model
      */
     public static function createViaApi($since = null, $before = null)
     {
-        $userId = optional(auth()->user())->northstar_id;
+        $userId = auth()->id();
```
DFurnes, line 65: This used to read from Chompy's local `users` session table (which had a `northstar_id` column) – now it's as simple as calling Laravel's built-in `auth()->id()`! If nobody's logged in (because this is a queue job), it'll just return `null`.

app/Models/RockTheVoteReport.php
```diff
@@ -77,15 +84,16 @@ public static function createViaApi($since = null, $before = null)
         ]);
 
         // Parse response to find the new Rock The Vote Report ID.
-        $statusUrlParts = explode('/', $response->status_url);
+        // TODO: Can we read 'report_id' from the response?
+        $statusUrlParts = explode('/', $response['status_url']);
```
DFurnes, line 88: I'm wondering if perhaps the `report_id` field [on this response](https://rock-the-vote.github.io/Voter-Registration-Tool-API-Docs/#gregistrant_reports) didn't exist when we wrote this job [a year ago](https://github.com/DoSomething/chompy/pull/128)?

app/Services/RockTheVote.php
```diff
@@ -41,7 +41,7 @@ public function createReport($params)
             'json' => array_merge($params, $this->authQuery),
         ]);
 
-        return json_decode($response->getBody()->getContents());
+        return json_decode($response->getBody()->getContents(), true);
```
DFurnes, line 44: By convention, we always return associative arrays when parsing JSON – it makes it clearer that this isn't a "real" typed object/class, and makes it easier to mock responses in our test suites (without needing to cast `(object)`).

app/Console/Commands/ImportRockTheVoteCommand.php
mendelB: Nice! 🤩

mendelB (approved):
⚡

app/Models/RockTheVoteReport.php
mendelB: That would make sense!

DFurnes: Hmm, maybe not! I dumped the response when hitting this URL from my local and it indeed doesn't have `report_id`:

```json
{"status":"queued","record_count":null,"current_index":null,"status_url":"https://register.rockthevote.com/api/v4/registrant_reports/57568","download_url":null}
```

DFurnes: I did notice I was accidentally looking at the `gregistrant_reports` endpoint above, instead of [`registrant_reports`](https://rock-the-vote.github.io/Voter-Registration-Tool-API-Docs/#reports) (sneaky!), but that one is _also_ supposed to have a `report_id` field that's not showing up in practice. Alas!

I'll update this comment to explain why this hack seems to be necessary.

DFurnes: Added a `// HACK:` note in 98bb321.

PR #1213, opened 2021-06-02
ImportFileFactory updates

weerd:
### What's this PR do?

This pull request updates the `ImportFileFactory` to ensure all fields are included in the factory file to help with seeding records into the database. It also updates the prior tests that utilized the `ImportFileFactory` to ensure that expected values are used to ensure the tests pass.

👀 

While working on moving over the Chompy UI I needed to seed a bunch of import file records into the database but only a few of the fields were included in the factory. I added the rest of the fields and ensure they utilize random values that are appropriate for the fields.

However, after having done so, some tests broke, because they expected specific values (for example a `skip_count` of `1`, etc), so I added some helper methods to the test files to allow making an Import File with appropriate values for the tests.

### Relevant tickets

References [Pivotal #177733168](https://www.pivotaltracker.com/story/show/177733168).

### Checklist

### DIFF
--- a/database/factories/ImportFileFactory.php
+++ b/database/factories/ImportFileFactory.php
@@ -7,14 +7,21 @@ use App\Types\ImportType;
 use Faker\Generator as Faker;
 
 $factory->define(ImportFile::class, function (Faker $faker) {
+    $rowCount = $faker->numberBetween(10, 1250);
+    $importCount = $faker->numberBetween(1, $rowCount);
+
     return [
-        'filepath' => $faker->imageUrl,
+        'user_id' => null,
         'import_type' => $faker->randomElement([
             ImportType::$emailSubscription,
             ImportType::$mutePromotions,
             ImportType::$rockTheVote,
         ]),
-        'row_count' => $faker->numberBetween(10, 1250),
+        'options' => null,
+        'filepath' => $faker->imageUrl,
+        'row_count' => $rowCount,
+        'import_count' => $importCount,
+        'skip_count' => $rowCount - $importCount,
     ];
 });
 
@@ -23,9 +30,9 @@ $factory->state(ImportFile::class, 'email_subscription', [
 ]);
 
 $factory->state(ImportFile::class, 'mute_promotions', [
-    'import_type' => ImportType::$emailSubscription,
+    'import_type' => ImportType::$mutePromotions,
 ]);
 
 $factory->state(ImportFile::class, 'rock_the_vote', [
-    'import_type' => ImportType::$emailSubscription,
+    'import_type' => ImportType::$rockTheVote,
 ]);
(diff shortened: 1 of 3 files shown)


database/factories/ImportFileFactory.php
```diff
@@ -23,9 +30,9 @@
 ]);
 
 $factory->state(ImportFile::class, 'mute_promotions', [
-    'import_type' => ImportType::$emailSubscription,
+    'import_type' => ImportType::$mutePromotions,
 ]);
 
 $factory->state(ImportFile::class, 'rock_the_vote', [
-    'import_type' => ImportType::$emailSubscription,
+    'import_type' => ImportType::$rockTheVote,
```
weerd, line 37: These were an older copy-🍝 mistake!

tests/Jobs/Imports/ImportEmailSubscriptionsTest.php
```diff
@@ -6,6 +6,19 @@
 
 class ImportEmailSubscriptionsTest extends TestCase
 {
+    /**
+     * Make a fake unprocessed import file with no completed or skipped imports.
+     */
+    public function makeFakeUnprocessedImportFile()
```
katiecrane, line 12: Would it be helpful to do a factory state for this case as well? [We have ones elsewhere](https://github.com/DoSomething/northstar/blob/main/database/factories/UserFactory.php#L56-L66) that are more involved and similar to this!

katiecrane (approved):
Yay tests that tell the truth! I had a non-blocking question

tests/Jobs/Imports/ImportEmailSubscriptionsTest.php
weerd: Oh! That might actually be a little nicer too and remove need for helper method. I'll merge for now to unblock me from utilizing this in the Chompy UI move, but I'll try this out in upcoming PR and ping ya if it works out well!

katiecrane: Woohoo! I'd be interested to see how it goes!

weerd: Ok @katiecrane!

I looked into it and there's a small caveat. I'm already using a factory state in each test helper method to specify generating an import file as either an `email_subscription` or `rock_the_vote`.

So I added:

```php
$factory->state(ImportFile::class, 'unprocessed', [
    'import_count' => 0,
    'skip_count' => 0,
]);
```
But then when using the factory state I would need to pass the `import_type` each time and also import the `ImportType` class into each test to utilize the static properties of that class. Either that OR create states for an unprocessed import file for each type.

Debating how useful any of the aforementioned options would be 🤔

katiecrane: oooo hmmmmm I feel like the second option (create states for an unprocessed import file for each type) is most similar to how we use factories elsewhere but yeah not sure if at that point things become more confusing instead of simpler!

weerd: Ya, agreed. I'm gonna leave it for now and come back to it after all the Chompy tests are moved into NS and maybe reassess then if it's useful!

PR #1215, opened 2021-06-03
More ImportRockTheVoteRecord Tests (part deux) 🚥 

weerd:
### What's this PR do?

This pull request continues adding additional tests for the `ImportRockTheVoteRecord` job.

👀 

I capped it at this amount for now. There will be another PR with more tests but wanted to keep this one manageable!

### Relevant tickets

### Checklist

### DIFF
--- a/tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
+++ b/tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
@@ -364,6 +364,57 @@ class ImportRockTheVoteRecordTest extends TestCase
     }
 
     /**
+     * Test that existing user who started registration process is updated with import data if
+     * import record has higher voter registration status.
+     */
+    public function testUpdatesExistingUserIfImportDataHasHigherStatus()
+    {
+        $user = factory(User::class)->create([
+            'voter_registration_status' => 'step-1',
+        ]);
+
+        $action = $this->makeFakeVoterRegistrationPostAction();
+
+        $dateRegistrationStarted = $this->daysAgoInRockTheVoteFormat();
+
+        $post = factory(Post::class)
+            ->state('voter-reg')
+            ->create([
+                'action_id' => $action->id,
+                'northstar_id' => $user->id,
+                'status' => 'step-1',
+                'details' => json_encode([
+                    'Home zip code' => $user->addr_zip,
+                    'Finish with State' => 'No',
+                    'Pre-Registered' => 'No',
+                    'Started registration' => $dateRegistrationStarted,
+                    'Status' => 'Step 1',
+                    'Tracking Source' => 'ads',
+                ]),
+            ]);
+
+        $payload = $this->makeFakeReportPayloadForSpecificUser($user, [
+            'Finish with State' => 'No',
+            'Started registration' => $dateRegistrationStarted,
+            'Status' => 'Step 2',
+        ]);
+
+        $importFile = $this->makeFakeUnprocessedImportFile();
+
+        ImportRockTheVoteRecord::dispatch($payload, $importFile);
+
+        $this->assertMongoDatabaseHas('users', [
+            'voter_registration_status' => 'step-2',
+        ]);
+
+        $this->assertMysqlDatabaseHas('posts', [
+            'id' => $post->id,
+            'status' => 'step-2',
+            'type' => 'voter-reg',
+        ]);
+    }
+
+    /**
      * Test that a user is updated if their voter registration status changes.
      */
     public function testUpdatesVoterRegistrationStatusIfStatusChanged()
--- a/tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
+++ b/tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
@@ -236,6 +236,66 @@ class ImportRockTheVoteRecordTest extends TestCase
     }
 
     /**
+     * Test that existing user is not updated with import data if their voter registration status
+     * is already completed (at a higher status hierarchy).
+     */
+    public function testDoesNotUpdateExistingUserIfAlreadyAtHigherStatus()
+    {
+        $user = factory(User::class)->create([
+            'voter_registration_status' => 'registration_complete',
+        ]);
+
+        $action = $this->makeFakeVoterRegistrationPostAction();
+
+        $dateTwoDaysAgo = $this->daysAgoInRockTheVoteFormat(2);
+
+        $dateToday = $this->daysAgoInRockTheVoteFormat(0);
+
+        $postFromTwoDaysAgo = factory(Post::class)
+            ->state('voter-reg')
+            ->create([
+                'action_id' => $action->id,
+                'northstar_id' => $user->id,
+                'status' => 'register-OVR',
+                'details' => json_encode([
+                    'Home zip code' => $user->addr_zip,
+                    'Finish with State' => 'Yes',
+                    'Pre-Registered' => 'No',
+                    'Started registration' => $dateTwoDaysAgo,
+                    'Status' => 'Complete', // @Question: is this correct?
+                    'Tracking Source' => 'ads',
+                ]),
+            ]);
+
+        $payload = $this->makeFakeReportPayloadForSpecificUser($user, [
+            'Finish with State' => 'No',
+            'Started registration' => $dateToday,
+            'Status' => 'Step 1',
+        ]);
+
+        $importFile = $this->makeFakeUnprocessedImportFile();
+
+        ImportRockTheVoteRecord::dispatch($payload, $importFile);
+
+        $this->assertMongoDatabaseHas('users', [
+            'voter_registration_status' => $user->voter_registration_status,
+        ]);
+
+        $this->assertMysqlDatabaseHas('posts', [
+            'id' => $postFromTwoDaysAgo->id,
+            'status' => $postFromTwoDaysAgo->status,
+            'type' => 'voter-reg',
+        ]);
+
+        $this->assertMysqlDatabaseHas('rock_the_vote_logs', [
+            'import_file_id' => $importFile->id,
+            'started_registration' => $payload['Started registration'],
+            'status' => $payload['Status'],
+            'user_id' => $user->id,
+        ]);
+    }
+
+    /**
      * Test that record is skipped if the RTV data is invalid.
      *
      * @return void
--- a/tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
+++ b/tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
@@ -75,7 +75,7 @@ class ImportRockTheVoteRecordTest extends TestCase
     }
 
     /**
-     * Make a fake unprocessed import file with no completed or skipped imports.
+     * Make a fake unprocessed import file with no complete or skipped imports.
      *
      * @return \App\Models\ImportFile
      */
@@ -152,12 +152,12 @@ class ImportRockTheVoteRecordTest extends TestCase
     }
 
     /**
-     * That that if an existing completed post is found for the specified user,
+     * Test that if an existing complete post is found for the specified user,
      * the import does not create or update the post.
      *
      * @return void
      */
-    public function testDoesNotCreateOrUpdatePostIfExistingCompletedPostFound()
+    public function testDoesNotCreateOrUpdatePostIfExistingCompletePostFound()
     {
         $user = factory(User::class)->create();
 
@@ -176,7 +176,7 @@ class ImportRockTheVoteRecordTest extends TestCase
                     'Finish with State' => $payload['Finish with State'],
                     'Pre-Registered' => $payload['Pre-Registered'],
                     'Started registration' => $payload['Started registration'],
-                    'Status' => $payload['Status'],
+                    'Status' => 'Complete',
                     'Tracking Source' => $payload['Tracking Source'],
                 ]),
             ]);
@@ -194,8 +194,11 @@ class ImportRockTheVoteRecordTest extends TestCase
 
         $this->assertMysqlDatabaseHas('posts', [
             'id' => $post->id,
-            'status' => $post->status,
+            'status' => 'register-form', // unchanged!
         ]);
+
+        // Only existing completed post should exist in the posts table.
+        $this->assertCount(1, Post::all());
     }
 
     /**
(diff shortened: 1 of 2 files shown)


tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
```diff
@@ -376,4 +505,156 @@ public function testUserMobileAddedIfImportMobileProvided()
...
+            'sms_subscription_topics' => ['general', 'voting'],
+        ]);
+
+        $this->assertMongoDatabaseHas('users', [
+            '_id' => $mobileUser->id,
+            'sms_subscription_topics' => ['general'],
+        ]);
+
+        // @Question: this test in Chompy had something about the mobile user's sms_status
+        // being set to "stop", but doesn't seem like it is coded that way in the job.
```
weerd, line 658: @see [Chompy's `testMobileOwnerIsUpdatedIfImportUserHasNoMobileAndMobileIsTaken()`](https://github.com/DoSomething/chompy/blob/main/tests/Jobs/RockTheVote/ImportRockTheVoteRecordTest.php#L555)

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
```diff
@@ -147,6 +151,53 @@ public function testCreatesUserIfUserNotFound()
...
+            'started_registration' => $payload['Started registration'],
+            'status' => $payload['Status'],
+            'user_id' => $user->id,
+        ]);
+
+        $this->assertMysqlDatabaseHas('posts', [
+            'id' => $post->id,
+            'status' => $post->status,
+        ]);
+    }
```
katiecrane, line 199: Sorry I'm a bit confused on this test! Is the assertion on the `rock_the_vote_logs` table making sure we actually processed the import that you dispatched and can therefore perform our later checks? I see that there's a check for the existing post not being updated, but how do we know another one wasn't created?

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
```diff
@@ -235,6 +290,68 @@ public function testDoesNotImportRecordIfLogExists()
...
+            ->create([
+                'action_id' => $action->id,
+                'northstar_id' => $user->id,
+                'status' => 'register-OVR',
+                'details' => json_encode([
+                    'Home zip code' => $user->addr_zip,
+                    'Finish with State' => 'Yes',
+                    'Pre-Registered' => 'No',
+                    'Started registration' => $dateTwoDaysAgo,
+                    'Status' => 'Complete', // @Question: is this correct?
```
katiecrane, line 322: Based on [this](https://github.com/DoSomething/chompy/wiki/Rock-The-Vote#status-translation-rules) I believe this is correct!

DFurnes (approved):
Looking good!! 🚥 🎉

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
DFurnes: Confusing!! Is this based on whether or not they choose to opt-in to messaging in the RTV form? 🤔

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
DFurnes: Oh that's also a good catch! This makes me wish there was a `assertDatabaseHasCount()` assertion we could use to ensure that there's exactly one post for the user after this test completes... 🤔

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
weerd: Thanks for the additional 👀 . All the statuses plus how we convert statuses and then different formatting in statuses has been a real 🤯  so it's nice to get additional eyes confirming this is correct! 😅

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
katiecrane: That'd be awesome! Could we instead grab all posts and make sure there is only one?

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
katiecrane: Totally agreed, it's confusing!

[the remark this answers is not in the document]
katiecrane: Hmm @DFurnes I think that doc is indicating that TurboVote had a `confirmed` status but RTV does not?

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
```diff
@@ -235,6 +290,68 @@ public function testDoesNotImportRecordIfLogExists()
...
+            'Started registration' => $dateToday,
+            'Status' => 'Step 1',
+        ]);
+
+        $importFile = $this->makeFakeUnprocessedImportFile();
+
+        ImportRockTheVoteRecord::dispatch($payload, $importFile);
+
+        $this->assertMongoDatabaseHas('users', [
+            'voter_registration_status' => $user->voter_registration_status,
```
katiecrane, line 338: Should this be hardcoded to check for `registration_complete`?

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
```diff
@@ -235,6 +290,68 @@ public function testDoesNotImportRecordIfLogExists()
...
+        ]);
+
+        $this->assertMysqlDatabaseHas('posts', [
+            'id' => $postFromTwoDaysAgo->id,
+            'status' => $postFromTwoDaysAgo->status,
+            'type' => 'voter-reg',
+        ]);
```
katiecrane, line 345: I'm a little confused what we're checking for here - do we want to make sure there's NOT a post with a different status? I'm not super familiar with how the post stuff works so might have to bear with me a bit here!

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
```diff
@@ -235,6 +290,68 @@ public function testDoesNotImportRecordIfLogExists()
         );
     }
 
+    /**
+     * Test that existing user is not updated with import data if their voter registration status
+     * is already completed (at a higher status hierarchy).
```
katiecrane, line 295: Would it make sense to say "user and post" here since there are checks for both?

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
```diff
@@ -376,4 +505,156 @@ public function testUserMobileAddedIfImportMobileProvided()
...
+        ]);
+
+        $importFile = $this->makeFakeUnprocessedImportFile();
+
+        ImportRockTheVoteRecord::dispatch($payload, $importFile);
+
+        $this->assertMongoDatabaseHas('users', [
+            '_id' => $user->id,
+            'mobile' => $user->mobile,
+            'sms_status' => $user->sms_status,
```
katiecrane, line 550: I think this one would be clearer to me if when the user was created an `sms_status` was explicitly set and then we check for that same string here so we know for sure it hasn't changed (same with `sms_subscription_topics` below)

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
```diff
@@ -376,4 +505,156 @@ public function testUserMobileAddedIfImportMobileProvided()
...
+        $this->assertMongoDatabaseHas('users', [
+            '_id' => $user->id,
+            'mobile' => $phoneNumber,
+        ]);
+    }
+
+    /**
+     * Test that the owner of the mobile number has their sms subscription topics updated
+     * if the import data user does not have a mobile set and the provided number in import
+     * data points to a different user owning that nubmer.
```
katiecrane, line 621: Say that 5 times fast 🤯

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
katiecrane: I'm also confused - why is the mobile user getting their `voting` topic removed here?

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
```diff
@@ -376,4 +505,156 @@ public function testUserMobileAddedIfImportMobileProvided()
...
+            'Started registration' => $dateRegistrationStarted,
+        ]);
+
+        $importFile = $this->makeFakeUnprocessedImportFile();
+
+        ImportRockTheVoteRecord::dispatch($payload, $importFile);
+
+        $this->assertMongoDatabaseHas('users', [
+            '_id' => $user->id,
+            'sms_subscription_topics' => $user->sms_subscription_topics,
```
katiecrane, line 586: I think this is another spot where it might make sense to hardcode!

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
weerd: So from what I've concluded, User01 fills out the RTV form and adds a phone number, and this is all part of the import record. When we go to import the record, we look to see if the user already exists (typically by email also provided in the import form). Once User01 is found, before updating their record, there's a check to see if a different user has the same phone number (I guess this would be a thing if someone changed their phone number and both old and now new numbers are going into our database, or maybe the user signed up with one account via web and another via mobile number and haven't been merged 🤷‍♂️ ).

So if a different member, User02 has the same mobile number as User01, we remove the `voting` subscription from their sms subscriptions since User02 didn't fill out the RTV form, User01 had done that and would get subscribed to sms subscription related to voting moving forward.

I was suspecting we'd also want to `null` the mobile number from User02 since now it seems User01 has it... but I also don't know the full nuance of edge cases here. Could I fill out the RTV form and just add Dave's phone number to get sms messages related to voting? How does that all work?

Ok, my 🧠  hurst again...

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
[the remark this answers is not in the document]
weerd: Ya, the link @katiecrane added in [earlier comment](https://github.com/DoSomething/northstar/pull/1215#discussion_r645685357) is the one I initially went off of and seemed correct and indicating that TurboVote used `confirmed` but not RTV which uses `complete` instead.

And in the docs link you added it shows `confirmed` in the hierarchy list but also specifically says:

>The TurboVote data (which we imported before Rock The Vote - see Notes), would supply a confirmed status - similar to the Rock the Vote Completed status.

Which adds to the confusion. There's also competing docs, between the `docs/` directory and the GitHub wiki @_@

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
weerd: > Sorry I'm a bit confused on this test! Is the assertion on the rock_the_vote_logs table making sure we actually processed the import that you dispatched and can therefore perform our later checks? I see that there's a check for the existing post not being updated, but how do we know another one wasn't created?

Ya, the initial assertion on the `rock_the_vote_logs` checks to make sure the import file was actually processed. The following assertion on the `posts` table confirms that the status on the post was not changed. (Although I hadn't realized that the `status` in the created Post `details` should also have been set to `Complete` instead of the `$payload` value, so I made that update).

I also added an additional assertion to grab all posts and assert the count is only 1! Good catch @katiecrane 🕵️‍♀️

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
weerd: Ya... not really keen how much this test is doing... A lot of this stuff could use some love during a cleanup/refactor sprint! 😅

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
weerd: Essentially; we want to confirm that the post which was already completed is not overwritten with uncompleted-related status, etc from the import record (which just might be older data). 

So if the post voter registration status is `register-OVR` that means the registration was completed and we don't want to import a record where the same post ID has a status of `step-2`.

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
```diff
@@ -376,4 +505,156 @@ public function testUserMobileAddedIfImportMobileProvided()
...
+            'started_registration' => $dateRegistrationStarted,
+            'user_id' => $user->id,
+        ]);
+
+        $payload = $this->makeFakeReportPayloadForSpecificUser($user, [
+            'Phone' => $user->mobile,
+            'Started registration' => $dateRegistrationStarted,
+        ]);
+
+        $importFile = $this->makeFakeUnprocessedImportFile();
```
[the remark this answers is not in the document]
weerd: Agreed! Definitely something to come back and review during a cleanup/refactor sprint!

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
katiecrane: Woohoo!!

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
katiecrane: @weerd oooo so would User02 have been given the `voting` sms topic as part of the import job? I was thinking it already existed on the user prior to that, thus my confusion!

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
katiecrane: Ah got it! Thanks!

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
weerd: Nope, not as part of import but in case they have it for some other reason. I think it just aims to remove the `voting` topic from User02 just in case if they happen to have one, given that they are no longer the expected owner of the phone number they shouldn't be receiving voting related sms!

katiecrane: Hmmm I don't think I really understand this but given that we're sending User02 fewer messages and not more, seems okay to me!

weerd: Yeah, it's all a bit confusing and 🍝 -esk... I'm not sure I get it 100% either 😅

PR #1219, opened 2021-06-07
More ImportRockTheVoteRecord Tests (part trois) 🚥 

weerd:
### What's this PR do?

This pull request adds a third grouping of tests for the `ImportRockTheVoteRecord` job.

👀

This should be the last of the tests to add for merging this job from Chompy to Northstar! 😪 

### Relevant tickets

### Checklist

### DIFF
--- a/tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
+++ b/tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
@@ -943,4 +943,66 @@ class ImportRockTheVoteRecordTest extends TestCase
             'sms_subscription_topics' => ['general'],
         ]);
     }
+
+    /**
+     * Test that existing user with completed voter registration and stop sms status will not
+     * have SMS subscription topics changed with SMS opt-in via imported data.
+     */
+    public function testUserWithStopSmsStatusRemainsUnchangedWhenOptingIn()
+    {
+        $user = factory(User::class)->create([
+            'sms_status' => 'stop',
+            'sms_subscription_topics' => [],
+            'voter_registration_status' => 'registration_complete',
+        ]);
+
+        $this->makeFakeVoterRegistrationPostAction();
+
+        $payload = $this->makeFakeReportPayloadForSpecificUser($user, [
+            'Opt-in to Partner SMS/robocall' => 'Yes',
+            'Phone' => '+12345678910',
+            'Status' => 'Complete',
+        ]);
+
+        $importFile = $this->makeFakeUnprocessedImportFile();
+
+        ImportRockTheVoteRecord::dispatch($payload, $importFile);
+
+        $this->assertMongoDatabaseHas('users', [
+            '_id' => $user->id,
+            'sms_status' => 'active',
+            'sms_subscription_topics' => ['voting'],
+        ]);
+    }
+
+    /**
+     * Test that existing user with completed voter registration and stop sms status
+     * will have SMS subscription topics updated if SMS opt-out via imported data.
+     */
+    public function testUserWithStopSmsStatusUpdatesSmsSubscriptionsWhenOptingOut()
+    {
+        $user = factory(User::class)->create([
+            'sms_status' => 'stop',
+            'sms_subscription_topics' => [],
+            'voter_registration_status' => 'registration_complete',
+        ]);
+
+        $this->makeFakeVoterRegistrationPostAction();
+
+        $payload = $this->makeFakeReportPayloadForSpecificUser($user, [
+            'Opt-in to Partner SMS/robocall' => 'No',
+            'Phone' => '+12345678910',
+            'Status' => 'Complete',
+        ]);
+
+        $importFile = $this->makeFakeUnprocessedImportFile();
+
+        ImportRockTheVoteRecord::dispatch($payload, $importFile);
+
+        $this->assertMongoDatabaseHas('users', [
+            '_id' => $user->id,
+            'sms_status' => 'stop', // unchanged!
+            'sms_subscription_topics' => null,
+        ]);
+    }
 }
--- a/tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
+++ b/tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
@@ -700,7 +700,7 @@ class ImportRockTheVoteRecordTest extends TestCase
      * Test that existing user with completed voter registration can opt-in to
      * SMS via imported data.
      */
-    public function testUpdateUserSmsWhenAlreadyCompletedVoterRegistration()
+    public function testUserWithCompletedVoterRegistrationCanOptInToSms()
     {
         $user = factory(User::class)->create([
             'sms_status' => null,
@@ -731,7 +731,7 @@ class ImportRockTheVoteRecordTest extends TestCase
      * Test that existing user with completed voter registration can opt-out of
      * SMS via imported data.
      */
-    public function testSomething()
+    public function testUserWithCompletedVoterRegistrationCanOptOutOfSms()
     {
         $user = factory(User::class)->create([
             'sms_status' => null,
@@ -755,6 +755,37 @@ class ImportRockTheVoteRecordTest extends TestCase
             '_id' => $user->id,
             'sms_status' => 'stop',
             'sms_subscription_topics' => null,
+        ]);
+    }
+
+    /**
+     * Test that existing user with completed voter registration and active sms status
+     * will remain unchanged with SMS opt-in via imported data.
+     */
+    public function testUserWithActiveSmsStatusRemainsUnchangedWhenOptingIn()
+    {
+        $user = factory(User::class)->create([
+            'sms_status' => 'active',
+            'sms_subscription_topics' => ['general', 'voting'],
+            'voter_registration_status' => 'registration_complete',
+        ]);
+
+        $this->makeFakeVoterRegistrationPostAction();
+
+        $payload = $this->makeFakeReportPayloadForSpecificUser($user, [
+            'Opt-in to Partner SMS/robocall' => 'Yes',
+            'Phone' => '+12345678910',
+            'Status' => 'Complete',
+        ]);
+
+        $importFile = $this->makeFakeUnprocessedImportFile();
+
+        ImportRockTheVoteRecord::dispatch($payload, $importFile);
+
+        $this->assertMongoDatabaseHas('users', [
+            '_id' => $user->id,
+            'sms_status' => 'active', // unchanged!
+            'sms_subscription_topics' => ['general', 'voting'], // unchanged!
         ]);
     }
 }
--- a/tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
+++ b/tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
@@ -722,6 +722,7 @@ class ImportRockTheVoteRecordTest extends TestCase
 
         $this->assertMongoDatabaseHas('users', [
             '_id' => $user->id,
+            'mobile' => $user->mobile,
             'sms_status' => 'active',
             'sms_subscription_topics' => ['voting'],
         ]);
@@ -753,6 +754,7 @@ class ImportRockTheVoteRecordTest extends TestCase
 
         $this->assertMongoDatabaseHas('users', [
             '_id' => $user->id,
+            'mobile' => $user->mobile,
             'sms_status' => 'stop',
             'sms_subscription_topics' => null,
         ]);
--- a/tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
+++ b/tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
@@ -759,8 +759,8 @@ class ImportRockTheVoteRecordTest extends TestCase
     }
 
     /**
-     * Test that existing user with completed voter registration and active sms status
-     * will remain unchanged with SMS opt-in via imported data.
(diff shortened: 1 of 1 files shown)


tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
```diff
@@ -664,4 +665,498 @@ public function testMobileOwnerIsUpdatedIfImportUserHasNoMobileAndMobileIsTaken(
         // @Question: this test in Chompy had something about the mobile user's sms_status
         // being set to "stop", but doesn't seem like it is coded that way in the job.
     }
+
+    /**
+     * Test that existing user with mobile number retains mobile number and updates sms status
+     * and subscriptions even if the import data mobile number is different.
```
DFurnes, line 671: Woah, this might be behavior we want to re-consider in the future... I'm not sure that'd be how I'd expect this to work (if I opted-in to messaging with a different phone number) and not sure how this plays with CTIA rules. cc: @ngjo

(I know this is just existing behavior from Chompy & not something we should change now!)

DFurnes: Just to spell it out – this test is asserting (existing Chompy behavior) that if a user _has_ a phone number A that's texted `STOP` & they opt-in to messaging in Rock The Vote flow with a _different_ phone number B, then that _existing_ phone number A would be re-enrolled in messaging.

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
```diff
@@ -664,4 +665,498 @@ public function testMobileOwnerIsUpdatedIfImportUserHasNoMobileAndMobileIsTaken(
...
+            '_id' => $user->id,
+            'sms_status' => 'active', // unchanged!
+            'sms_subscription_topics' => ['general', 'voting'], // unchanged!
+        ]);
+    }
+
+    /**
+     * Test that existing user with completed voter registration and active sms status
+     * will have SMS subscription topics updated if SMS opt-out via imported data.
+     */
```
DFurnes, line 795: Nice! 💯

DFurnes (approved):
Great work! These tests were all really pleasant to read. 💯

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
weerd: Yep yep. It seemed wonky to me but that's essentially what the original test did as I translated it over, and it passed when I completed writing it. Seemed weird but I also lack a lot of the SMS related context/rules that were decided over time.

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
weerd: I re-wrote a lot of the descriptions (to make them easier to hopefully grok on first read) and test titles (for readability and make it easier to group related tests alphabetically)! Took some time but I think it was worth it! 😅

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
```diff
@@ -664,4 +665,498 @@ public function testMobileOwnerIsUpdatedIfImportUserHasNoMobileAndMobileIsTaken(
...
+
+        $importFile = $this->makeFakeUnprocessedImportFile();
+
+        ImportRockTheVoteRecord::dispatch($payload, $importFile);
+
+        $this->assertMongoDatabaseHas('users', [
+            '_id' => $user->id,
+            'sms_status' => 'active',
+            'sms_subscription_topics' => ['voting'],
+        ]);
```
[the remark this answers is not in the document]
weerd: Ah ya, I can add that. Just tried it and apparently it retains the old number...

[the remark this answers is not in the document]
weerd: @see [prior comment](https://github.com/DoSomething/northstar/pull/1219/files#r648583206)

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
DFurnes: Yeah totally! And I agree that we should keep existing behavior for now – just flagging in case we want to change.

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
DFurnes: It absolutely was! All these tests were super clear to read. 👏

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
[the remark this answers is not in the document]
weerd: @DFurnes I think I originally left it off on subsequent tests because there is a [test](https://github.com/DoSomething/northstar/pull/1219/files#diff-57a9819259125bfc4b0d86ee070fe556af537f9bcb8f23c060184e8d7364e548R673) specifically checking to ensure the original phone number is retained. I can update a few of these if it'll be clearer though!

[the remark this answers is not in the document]
DFurnes: I don't have very strong feelings on that, though!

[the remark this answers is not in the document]
weerd: Ya I think we're in the same boat; my comment snuck in before your response to my prior question, haha 😅

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
```diff
@@ -664,4 +665,500 @@ public function testMobileOwnerIsUpdatedIfImportUserHasNoMobileAndMobileIsTaken(
...
+            'Status' => 'Complete',
+        ]);
+
+        $importFile = $this->makeFakeUnprocessedImportFile();
+
+        ImportRockTheVoteRecord::dispatch($payload, $importFile);
+
+        $this->assertMongoDatabaseHas('users', [
+            '_id' => $user->id,
+            'mobile' => $user->mobile,
```
katiecrane, line 757: Is this more making sure that the mobile number was not wiped out rather than making sure it still has a particular value?

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
```diff
@@ -664,4 +665,500 @@ public function testMobileOwnerIsUpdatedIfImportUserHasNoMobileAndMobileIsTaken(
...
+            'Status' => 'Complete',
+        ]);
+
+        $importFile = $this->makeFakeUnprocessedImportFile();
+
+        ImportRockTheVoteRecord::dispatch($payload, $importFile);
+
+        $this->assertMongoDatabaseHas('users', [
+            '_id' => $user->id,
+            'sms_status' => 'active', // unchanged!
```
katiecrane, line 820: I'm kind of confused about why we change the topics but not the status, but I'm guessing this is how it currently works?

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
```diff
@@ -664,4 +665,500 @@ public function testMobileOwnerIsUpdatedIfImportUserHasNoMobileAndMobileIsTaken(
...
+        ]);
+
+        $importFile = $this->makeFakeUnprocessedImportFile();
+
+        ImportRockTheVoteRecord::dispatch($payload, $importFile);
+
+        $this->assertMongoDatabaseHas('users', [
+            '_id' => $user->id,
+            'sms_status' => 'active',
+            'sms_subscription_topics' => ['general', 'voting'], // unchanged!
```
katiecrane, line 852: Would Chompy add any other topics via RTV import other than `voting`? I'm wondering if it might make sense to make sure that `voting` is NOT added here, since maybe it did try to alter it but doesn't appear to since the user already had the `voting` topic? Or maybe we're more making sure no topics are removed?

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
```diff
@@ -664,4 +665,500 @@ public function testMobileOwnerIsUpdatedIfImportUserHasNoMobileAndMobileIsTaken(
...
+        $this->assertMongoDatabaseHas('users', [
+            '_id' => $user->id,
+            'sms_status' => 'stop',
+            'sms_subscription_topics' => null,
+        ]);
+    }
+
+    /**
+     * Test that existing user will have SMS subscription topics updated if SMS opt-out via
+     * imported data.
```
katiecrane, line 1075: This and the test below confused me a bit at first since I was having trouble spotting the difference. Would it be worth calling out that what we want to see is the `voting` topic removed?

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
weerd: Maybe a little of both. It was [a bit weird to Dave and me and we chatted about it a bit](https://github.com/DoSomething/northstar/pull/1219#discussion_r648603827) in the test before this one which is similar. But since the default behavior is a little counter-intuitive we agreed best to just ensure phone number stays the same as original and not overridden by the import data one.

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
weerd: I think because they are technically just opting out of `voting` related sms texts, but that's my best guess. And yep, all this is based on current Chompy functionality!

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
weerd: Ya, I think that's a good recommendation! Especially when coming back to these test at a later date and forgetting a bit.

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
weerd: I think it's more confirming that even with a `less` SMS status and a completed registration (so they have a `voting` topic already), when their record from RTV is imported it doesn't remove the `voting` topic. I tried it by initially not having `voting` in the original topic list and it _does_ get added to the final array. 🤷  In the original test in Chompy, there's a [specific comment](https://github.com/DoSomething/chompy/blob/35332a9a0aa9a305d939dacfc15d0b565968279c/tests/Jobs/RockTheVote/ImportRockTheVoteRecordTest.php#L706) indicating that the user already had the `voting` topic so I went with that in this test.

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
katiecrane: Gotcha! I'm not sure that this is making sure the phone number isn't overridden, because if it was overridden it would still be set to `$user->mobile` wouldn't it?

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
katiecrane: AH thanks for the background here!

tests/Jobs/Imports/ImportRockTheVoteRecordTest.php
weerd: Might be misunderstanding your question, but when the User is created via the factory they get assigned a phone number (maybe adding it as part of the array passed to `create()` would make that clearer?). So when the payload is imported, the test is confirming that they've retained that original number and not the one supplied in the `$payload['Phone']` which is a different number.

katiecrane: Right! But if the phone number on `$user` had been updated, when you pull `$user->mobile` won't that be the `mobile` currently on the user whether or not it has been updated, and not necessarily the value originally set by the factory?

weerd: 🤔  I wouldn't think so since it's the in memory saved record of the `$user` when created pre-database update via import logic. Wouldn't that essentially mean that the ID could be changed too so why compare against the `$user->id`... but now you have me doubting all the things lol 😅 

I ran a test on a different test in that file (`testUserWithCompletedVoterRegistrationCanOptInToSms()`) where a user value is changed for the test and dumped the `$user` after creation and at the end of the test and the properties all remained from when the user was initial saved to `$user`.

I also tried it by amending the `testUserWithCompletedVoterRegistrationCanOptInToSms()` test (on line 703); instead of asserting an `sms_status` value of `active` as a string, I changed it to `$user->sms_status` and it failed indicating that `$user` variable contained the original value when first created and saved to variable.

I _think_ that because `mobile` is a direct property on the user object it wouldn't be something that needs to hit the database again when retrieving and thus retains its original value.

katiecrane: Ah so since we never did a `fresh()` here it will always still have that same value? I guess it works then but a bit confusing read-ablity wise to me since it appears that we are just asserting that the user's mobile is set as the user's mobile which will always be true. I think hardcoding a starting mobile on the user and then checking that that specific value is still the one in the database at the end would be easier for me to wrap my head around but we may be looking at it in different ways!

weerd: Yep yep! Agreed 👍 

I'll aim to get that update in a separate PR and ping ya 😄  Defs a good thing to question and always better to strive for readability!

PR #1221, opened 2021-06-11
Importer UI part deux

weerd:
### What's this PR do?

This pull request continues adding and updating the Chompy UI into Northstar's admin UI.

👀 

...

### Relevant tickets

References [Pivotal #177733168](https://www.pivotaltracker.com/story/show/177733168).

### Checklist


resources/views/admin/imports/mute-promotions/show.blade.php
```diff
@@ -0,0 +1,43 @@
...
+        <table class="table">
+            <thead>
+                <tr class="row">
+                    <th>User</th>
+                </tr>
+            </thead>
+
+            @foreach($importedItems as $importedItem)
+                <tr class="row">
+                    {{-- @TODO: Change the link to view list of import files that include specified user ID. --}}
```
weerd, line 35: I hadn't realized entirely since the prior Chompy view files were a bit confusing, but the resource list view would allow showing either the import item list for a specific import file OR switching to show a list of the import files that contained a record for a specified user. I think in this new approach it would be ideal to enable the latter view via a `ImportUsersController`, so for now I just linked to view the individual user in NS Admin and we can add that controller/route/view(s) during a cleanup sprint since it doesn't seem vital for now!

app/Http/Controllers/Web/Admin/MutePromotionImportController.php
```diff
@@ -0,0 +1,190 @@
...
+        $records = $csv->getRecords();
+
+        foreach ($records as $record) {
+            ImportMutePromotions::dispatch(
+                ['northstar_id' => $record['northstar_id']],
+                $importFile,
+            )
+                ->delay(now()->addSeconds(3))
+                ->onQueue($queue);
+        }
```
DFurnes, line 131: I believe we can replace this with dispatching the [ParseMutePromotions](https://github.com/DoSomething/northstar/blob/main/app/Jobs/Imports/ParseMutePromotions.php) job (which takes some of this heavier processing off the web thread & gives it a bit more memory headroom in case of a really big CSV!)

app/Http/Controllers/Web/Admin/MutePromotionImportController.php
```diff
@@ -0,0 +1,190 @@
...
+     * Display the specified resource.
+     *
+     * @param  int  $id
+     * @return \Illuminate\Http\Response
+     */
+    public function show($id)
+    {
+        $importFile = ImportFile::findOrFail($id);
+
+        $importedItems = MutePromotionsLog::where(
```
DFurnes, line 146: Let's use route model binding here!

```suggestion
     * @param  ImportFile  $importFile
     * @return \Illuminate\Http\Response
     */
    public function show(ImportFile $importFile)
    {
        $importedItems = MutePromotionsLog::where(
```

app/Http/Controllers/Web/Admin/RockTheVoteImportController.php
```diff
@@ -0,0 +1,158 @@
...
+     *
+     * @return \Illuminate\Http\Response
+     */
+    public function create()
+    {
+        $config = ImportType::getConfig($this->importType);
+
+        if (request('source') === 'test') {
+            return $this->renderTestView($config);
+        }
```
DFurnes, line 61: For later – let's pull this out into a separate `TestRockTheVoteImportController` so we don't have this awkward "two routes in one" flow, since they're really doing very different things.

```suggestion
        // TODO: Extract this into a separate controller!
        if (request('source') === 'test') {
            return $this->renderTestView($config);
        }
```

app/Services/CustomerIo.php
```diff
@@ -184,7 +190,7 @@ public function deleteCustomer(string $id)
      */
     public function suppressCustomer(string $id)
     {
-        if (!config('features.delete-api')) {
+        if (!config('features.customer_io') && !config('features.delete-api')) {
```
DFurnes, line 193: This should be an `||` so that if either of these are disabled, we skip running this code!

```suggestion
        if (!config('features.customer_io') || !config('features.delete-api')) {
```

resources/views/admin/imports/index.blade.php
```diff
@@ -4,53 +4,53 @@
...
     <span class="caret"></span>
   </button>
 
   <ul class="dropdown-menu" aria-labelledby="dropdown-filter">
-    <li><a href="/admin/imports?type=email-subscription">Email Subscription</a></li>
-    <li><a href="/admin/imports?type=mute-promotions">Mute Promotions</a></li>
-    <li><a href="/admin/imports?type=rock-the-vote">Rock The Vote</a></li>
+    <li><a href="/admin/imports/email-subscriptions">Email Subscription</a></li>
+    <li><a href="/admin/imports/mute-promotions">Mute Promotions</a></li>
+    <li><a href="/admin/imports/rock-the-vote">Rock The Vote</a></li>
```
DFurnes, line 20: 🎉

routes/web.php
```diff
@@ -114,8 +115,55 @@
...
+            'Admin\RockTheVoteImportController@index',
+        )->name('imports.rock-the-vote');
+        Route::get(
+            'imports/rock-the-vote/create',
+            'Admin\RockTheVoteImportController@create',
+        )->name('imports.rock-the-vote.create');
+        Route::get(
+            'imports/rock-the-vote/{id}',
+            'Admin\RockTheVoteImportController@show',
+        )->name('imports.rock-the-vote.show');
```
DFurnes, line 132: What would you think about using [`Route::resource`](https://laravel.com/docs/8.x/controllers#resource-controllers) here to make this more concise?

DFurnes (approved):
It's all coming together! 🎉

app/Http/Controllers/Web/Admin/EmailSubscriptionImportController.php
```diff
@@ -0,0 +1,121 @@
...
+     * Create a new controller instance.
+     *
+     * @return void
+     */
+    public function __construct()
+    {
+        $this->middleware('auth:web');
+        $this->middleware('role:admin,staff');
+
+        $this->importType = ImportType::$emailSubscription;
```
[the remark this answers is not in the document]
weerd: Ya, I went back and forth on it. This just felt a little cleaner and less confusing than seeing a slew of `ImportType::$emailSubscription` sprinkled throughout the controller. It made sense before to use the static property everywhere since the one import controller handled all the different types, but now each type has its own controller.

app/Http/Controllers/Web/Admin/RockTheVoteImportController.php
weerd: Ya, I like that idea 👍

app/Services/CustomerIo.php
weerd: Ah, right. Then same for the `deleteCustomer()` above?

routes/web.php
weerd: Ya, I initially did that but it would bungle up the route name removing the `imports` prefix which I'd prefer to have included. I was going to ask about this and see if you had other ideas that would allow for using a resource. Alternatively it could mean defining these outside of the current group that sets the `admin.` prefix and creating a separate one below that whose prefix is `admin.imports.` to then have useful naming.

routes/web.php
```diff
@@ -114,8 +115,59 @@
...
+        Route::get(
+            'imports/mute-promotions',
+            'Admin\MutePromotionImportController@index',
+        )->name('imports.mute-promotions');
+        Route::get(
+            'imports/mute-promotions/create',
+            'Admin\MutePromotionImportController@create',
+        )->name('imports.mute-promotions.create');
+        Route::get(
+            'imports/mute-promotions/{importFile}',
```
weerd, line 162: @DFurnes After merging in your model binding suggestion in the `MutePromotionImportController@show` action, it wasn't working correctly and it's cause it was missing having the URI segment match the name of the variable. Updated it and works fine 😄  Buuuut not sure if we switch this to a `Route::resource` is auto-magically knows to do that and not use `{id}`.... 🤔  Don't remember offhand.

app/Services/CustomerIo.php
DFurnes: Oh, yes! 👍

routes/web.php
DFurnes: I think `Route::resource` should handle this out of the box!

app/Jobs/Imports/ParseEmailSubscriptions.php
```diff
@@ -66,6 +67,7 @@ public function handle()
             'import_type' => ImportType::$emailSubscription,
             'row_count' => iterator_count($records),
             'user_id' => optional($this->user)->id,
+            'options' => $this->options ? json_encode($this->options) : null,
```
weerd, line 70: Added the `option` property for the `ImportFile` since it's used on the front-end.

app/Jobs/Imports/ParseMutePromotions.php
```diff
@@ -51,6 +66,7 @@ public function handle()
             'import_type' => ImportType::$mutePromotions,
             'row_count' => iterator_count($records),
             'user_id' => optional($this->user)->id,
+            'options' => $this->options ? json_encode($this->options) : null,
```
weerd, line 69: Same here, added the `option` property for the `ImportFile` since it's used on the front-end.

routes/web.php
weerd: Ah cool! Good to know! 

It still leaves the issue with how it formats the named routes which I'd like to figure out a solution for before changing to a resource route 🤔

PR #1223, opened 2021-06-18
🦖 Roar!  Docusaurus is here!

weerd:
### What's this PR do?

This pull request implements the use of [Docusaurus](https://docusaurus.io/) for our DoSomething Admin documentation.

https://user-images.githubusercontent.com/105849/122840226-b2802700-d2c7-11eb-8391-c6d0cb2ca74c.mp4

👀   But may also help to review the [Docusaurus docs](https://docusaurus.io/docs) a bit as well for how it works, etc.

A bunch of background can be found in the [RFC for better documentation tools](https://github.com/DoSomething/rfcs/pull/13/files#diff-f904db6847fee28a707dd73f4c959ec283c2dfe8dc7d265a5f2f48cd4bd94b57R83).

### Relevant tickets

References [Pivotal #178614470](https://www.pivotaltracker.com/story/show/178614470).

### Checklist


docs/src/pages/index.js
```diff
@@ -0,0 +1,39 @@
...
+          <Link className="button button--secondary button--lg" to="/">
+            Docusaurus Tutorial - 5min ⏱️
+          </Link>
+        </div>
+      </div>
+    </header>
+  );
+}
+
+export default function Home() {
```
weerd, line 26: Left in the `index.js` page for now. We could decide to not have a landing home page and end up pointing directly to the docs intro, but figured I would leave it for now and we could decide later. Maybe the home page could direct you to other DS docs or something 🤔

docs/src/components/HomepageFeatures.js
```diff
@@ -0,0 +1,64 @@
...
+    description: (
+      <>
+        Docusaurus was designed from the ground up to be easily installed and
+        used to get your website up and running quickly.
+      </>
+    ),
+  },
+  {
+    title: 'Focus on What Matters',
+    Svg: require('../../static/img/undraw_docusaurus_tree.svg').default,
```
weerd, line 18: I thought this was kinda neat how they referred to SVGs in the content list array of objects instead of importing them all at the top.

.gitignore
```diff
@@ -1,20 +1,16 @@
...
-.env
 
-# NPM
-/node_modules
 npm-debug.log
-
-# macOS & editor
-.DS_Store
-.idea/
+stats.json
```
weerd, line 16: I removed the commented categories because it just made it a bit more confusing where to place new ignored files if they didn't fit into a category. Instead I went with hierarchy list as such:

- directories
- dotfiles
- files

weerd: We sorta did this in [Phoenix](https://github.com/DoSomething/phoenix-next/blob/main/.gitignore) but it needs a little cleaning sort.

docs/src/css/custom.css
```diff
@@ -0,0 +1,25 @@
...
+/* You can override the default Infima variables here. */
+:root {
+  --ifm-color-primary: #615bc4;
+  --ifm-color-primary-dark: rgb(73, 67, 183);
+  --ifm-color-primary-darker: rgb(50, 43, 170);
+  --ifm-color-primary-darkest: rgb(28, 30, 132);
+  --ifm-color-primary-light: rgb(122, 117, 208);
+  --ifm-color-primary-lighter: rgb(147, 143, 219);
+  --ifm-color-primary-lightest: rgb(199, 197, 237);
+  --ifm-code-font-size: 95%;
```
weerd, line 17: I went with some DS blurple action pulled from Tailwind... except for the `*-primary-lightest` value, which I just made up.

weerd:
This only adds a bit of the documentation for the Importer work. More will be added over time, but at least it gets the ball rolling. Didn't have as much time after setting up and configuring Docusaurus to get much farther on more importer docs 😅

weerd:
When Docusaurus builds the static site for the docs via the `npm run docs` command, it builds and exports the static site into a directory called `build/`. I debated how to best set this up, but I figured it would be easiest to have it place the `build/` directory in Laravel's public directory, so the docs could be easily accessible. One caveat is that for this to work as I have it set up, the URL for the docs would be `identity.dosomething.org/build/docs` which is not ideal (I think `/docs` would be preferred), but I wasn't sure how to best achieve that.

Alternatively, we could maybe setup a URL to point outside of the Laravel `public/` directory and keep the build in the `docs/` directory, but not sure if this is ideal either.

Thoughts?

weerd:
Docusaurus has some scripts already setup to help with deploying the docs as a GitHub Pages website. If we end up deciding that we really like Docusaurus and that all our technical documentation should live in the same place, I could maybe see the docs living within their own repository, combining the Northstar and Phoenix docs together and deploying as a GH Pages site, which could resolve the Laravel `public/` directory URL issue mentioned earlier 🤔 

Although maybe this will eventually happen anyways, as we move towards the idea of a monorepo™.... as all the Furnes of the world rejoice in unison 🎺

weerd:
A nice thing about Docusaurus is that there are some recognizable open-source projects using it for their docs, so it makes for some great examples on how to structure and set up docs.

DFurnes (approved):
Very cool!! 🦖

docs/README.md
```diff
@@ -0,0 +1,33 @@
+# Website
+
+This website is built using [Docusaurus 2](https://docusaurus.io/), a modern static website generator.
+
+## Installation
+
+```console
+yarn install
+```
```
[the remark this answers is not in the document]
weerd: Oh yes, definitely. I think that's just what the docs say, but when I was running it locally and when building the static site for production, it was all with NPM.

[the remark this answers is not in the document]
weerd: There's still a few files to edit and or remove and this read me they auto-add inside the docusaurus init directory is likely one of them.

weerd:
Also @DFurnes lemme know if you have thoughts on [this comment](https://github.com/DoSomething/northstar/pull/1223#issuecomment-865408101).

docs/src/components/HomepageFeatures.js
DFurnes: Hmm, yeah. I do like keeping this closer to where it's used!

On the other hand, it feels somewhat awkward to remember to grab the default export (via the `.default`) at the end, and I'm curious if this would make it harder for Webpack to tree-shake out unused code in some circumstances...

weerd: Ah ya... also good point. 🤔

DFurnes:
> One caveat is that for this to work as I have it set up, the URL for the docs would be `identity.dosomething.org/build/docs` which is not ideal (I think `/docs` would be preferred), but I wasn't sure how to best achieve that.

Can we set the output directory to `public/docs`, or does it always need a `build` in there for other things to live in? There's something really appealing about always having the accurate documentation for whichever commit is deployed to each instance! (And worst case `/build/docs` isn't all that terrible a URL to have this at.)

docs/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "docs",
+  "version": "0.0.0",
+  "private": true,
+  "scripts": {
+    "docusaurus": "docusaurus",
+    "start": "docusaurus start",
+    "clean": "rm -rf ../public/build",
+    "build": "docusaurus build && mv build ../public",
```
DFurnes, line 9: It looks like we might be able to use the [`--out-dir` option](https://docusaurus.io/docs/cli#docusaurus-build-sitedir) for `docusaurus build` to place this at a custom path.

weerd: Oh nice find! I missed that option thus the fun `mv` in the script 😅

weerd:
> Can we set the output directory to public/docs, or does it always need a build in there for other things to live in?

I believe so, but I'm going to play around with that setting you found and see what that reveals!

weerd:
Great find @DFurnes! Using that `--out-dir` flag now only lets us specify a path to set for where the script builds the static site to, but it also lets us name the directory. We ended up going with `/dev/docs` since these are development docs!

PR #1227, opened 2021-06-24
Enable serving our dev docs.

weerd:
### What's this PR do?

This pull request attempts to update the Nginx config to see if it will resolve the issues serving the static files for the docs via Heroku.

👀 

While the docs are working on QA after deploying the PR to enable Docusaurus, hitting the `/dev/docs` URL results in a `403 Forbidden` error, but if you add `/dev/docs/index.html` to the URL they work. We believe this is just an issue with how Nginx is set up to serve files and just need to indicate that we want to serve `index.html` files.

### Relevant tickets

References [Pivotal #178614470](https://www.pivotaltracker.com/story/show/178614470).

### Checklist

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -5,6 +5,11 @@ location / {
     try_files $uri $uri/ /index.php?$query_string;
 }
 
+# Allow our application to serve our development docs static site.
+location /dev {
+  index index.html;
+}
+
 # Allow our application to serve '.well-known' routes.
 location ^~ /.well-known/ {
     allow all;


nginx.conf
```diff
@@ -5,6 +5,11 @@ location / {
     try_files $uri $uri/ /index.php?$query_string;
 }
 
+# Allow our application to serve our development docs static site.
+location /dev {
```
weerd, line 9: I figured it'd be best to scope it directly to the `/dev` directory.

DFurnes: 100%! I think otherwise we'd run into issues with Laravel's `index.php` not being used.

DFurnes (approved):

PR #1230, opened 2021-06-28
Upgrade to Laravel 8.

weerd:
### What's this PR do?

This pull request updates Northstar to Laravel 8 following the upgrade guide.

👀 

...

### Relevant tickets

References [Pivotal #178201843](https://www.pivotaltracker.com/story/show/178201843).

### Checklist


app/Console/Commands/SetupCommand.php
```diff
@@ -15,7 +15,7 @@ class SetupCommand extends Command
      *
      * @var string
      */
-    protected $signature = 'northstar:setup {--reset}';
+    protected $signature = 'northstar:setup {--reset : Delete existing env file and recreate it.}';
```
weerd, line 18: Just added a helpful description :)

.env.example
```diff
@@ -35,7 +35,7 @@ QUEUE_CONNECTION=sync
 SESSION_DRIVER=file
 SESSION_LIFETIME=120
 
-MAIL_DRIVER=log
+MAIL_MAILER=log
```
DFurnes, line 38: We'll have to remember to [update this environment variable](https://github.com/DoSomething/infrastructure/blob/83741e7d41017d91c946c3b599b0bc6895b078c4/applications/northstar/main.tf#L97) before we deploy this!

app/Auth/Normalizer.php
```diff
@@ -84,10 +84,12 @@ public function mobile($mobile)
     }
 
     /**
-     * Parse a string into a Carbon date.
+     * Parse an array of string into Carbon dates.
+     * @see https://laravel.com/docs/7.x/upgrade#date-serialization dates
+     * using toArray are returned differently.
```
DFurnes, line 89: It sounds like this might only be when calling `toArray()` on Eloquent models (rather than `Collections`, which we're just using here for the more "friendly" `map` syntax). We should double-check that before merging though.

app/Auth/Normalizer.php
```diff
@@ -84,10 +84,12 @@ public function mobile($mobile)
...
      *
-     * @param string[] $strings
-     * @return Carbon[]
+     * @param array $strings
+     * @return Carbon
```
DFurnes, line 92: I think this still takes an array of strings & returns an array of `Carbon` instances, right?

tests/Http/Web/WebUserTest.php
```diff
@@ -59,7 +63,10 @@ public function testCancelProfileEdit()
         $response = $this->get($userUrl . '/edit');
 
         $response->assertStatus(200);
-        $response->assertSee('<a href="' . url($userUrl) . '">Cancel</a>');
+        $response->assertSee(
+            '<a href="' . url($userUrl) . '">Cancel</a>',
+            false,
+        );
```
DFurnes, line 69: This `false` tripped me up at first glance! I wonder if we should just check for `$response->assertSee('Cancel')` so we don't have to disable escaping (or perhaps add an inline comment explaining the `false`?

DFurnes (approved):
Epic job on the Laravel 7.x upgrade (especially with all the test updates)!!

app/Exceptions/Handler.php
```diff
@@ -40,65 +41,65 @@ class Handler extends ExceptionHandler
      *
      * This is a great spot to send exceptions to Sentry, Bugsnag, etc.
      *
-     * @param  \Exception $e
+     * @param  \Throwable $exception
```
weerd, line 44: I switched this variable name to be more explicit; I usually prefer not abbreviating it because my brain sometimes thinks `error` instead of `exception` when it's just depicted as `$e`. The Laravel codebase also uses the full word, so another reason to follow that as well :)

config/queue.php
```diff
@@ -93,8 +97,8 @@
     */
 
     'failed' => [
-        'driver' => 'mongodb',
-        'database' => 'mongodb',
+        'driver' => env('QUEUE_FAILED_DRIVER', 'database-uuids'),
+        'database' => env('DB_CONNECTION', 'mysql'),
```
weerd, line 101: @DFurnes I was a bit confused on this... not sure if this was just missing being updated when we merged in Rogue, but figured this should be set to the mysql connection since the `failed_jobs` table lives there and not in mongodb 🤔  I could be missing something 🤷‍♂️

app/Observers/UserObserver.php
```diff
@@ -224,8 +221,8 @@ public function updated(User $user)
                 // TODO: Refactor this to be a new TrackPromotionsResubscribeCustomerIoEvent job.
                 new CreateCustomerIoEvent($user, 'promotions_resubscribe', []),
             ])
-                ->dispatch($user)
-                ->onQueue($queue);
+                ->onQueue($queue)
+                ->dispatch($user);
```
weerd, line 225: This just needed the ole switcheroo!

app/Providers/RouteServiceProvider.php
```diff
@@ -14,6 +14,8 @@ class RouteServiceProvider extends ServiceProvider
      * This namespace is applied to your controller routes.
      *
      * In addition, it is set as the URL generator's root namespace.
+     * @see https://laravel.com/docs/8.x/upgrade#automatic-controller-namespace-prefixing consider
+     * proceeding with above upgrade to make development easier in IDEs.
```
weerd, line 18: I thought this was super cool and would be awesome for our IDEs to be able to jump to corresponding controller class etc, but felt like a cleanup task for a later date!

phpunit.xml
```diff
@@ -1,32 +1,33 @@
...
-         bootstrap="bootstrap/autoload.php"
+<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:noNamespaceSchemaLocation="./vendor/phpunit/phpunit/phpunit.xsd"
+         bootstrap="vendor/autoload.php"
          colors="true"
-         convertErrorsToExceptions="true"
-         convertNoticesToExceptions="true"
-         convertWarningsToExceptions="true"
-         processIsolation="false"
-         stopOnFailure="false">
```
weerd, line 10: All the configs that were deleted are already set as the default values for PHPUnit.

artisan
```diff
@@ -9,11 +10,11 @@
 | Composer provides a convenient, automatically generated class loader
 | for our application. We just need to utilize it! We'll require it
 | into the script here so that we do not have to worry about the
-| loading of any our classes "manually". Feels great to relax.
+| loading of any of our classes manually. It's great to relax.
```
DFurnes, line 13: Changes like this in the Laravel diffs always make me chuckle. It _is_ great to relax. 😌

config/queue.php
```diff
@@ -50,6 +50,7 @@
             'table' => 'jobs',
             'queue' => 'default',
             'retry_after' => 90,
+            'after_commit' => false,
```
DFurnes, line 53: I was curious what this setting was for & looked it up in [the docs](https://laravel.com/docs/8.x/queues#jobs-and-database-transactions). It solves a pretty sneaky edge case! I don't think we use transactions enough to run into this sort of issue ourselves, but it might not be a bad idea to set this to `true` anyways. (Perhaps during cleanup sprint!)

DFurnes (approved):
Looks great!! 🆙

tests/LegacyHttp/LegacyUserTest.php
```diff
@@ -1391,19 +1421,19 @@ public function testFilterBySearchFieldParam()
             'GET',
             'v1/users?search[email]=' . $user->email,
         );
-        $this->assertCount(1, $this->decodeResponseJson()['data']);
+        $this->assertCount(1, $this->response->json('data'));
         $this->assertEquals(
-            $this->decodeResponseJson()['data'][0]['email'],
+            $this->response->json('data.0.email'),
```
katiecrane, line 1426: ooooo I love this!

app/Auth/Normalizer.php
weerd: right. So it should be `@return array`? Or better to use the `Carbon[]`? Just not familiar with this syntax!

DFurnes: Yeah, I think either would be correct but `Carbon[]` is a bit more specific (and likewise `string[]` for the param).

weerd: gotcha! Good to know!

app/Auth/Normalizer.php
weerd: I added a `testNormalizeDates()` method because I couldn't find anywhere we use the `dates()` method on this class and helped to confirm it was working as expected.

tests/LegacyHttp/LegacyUserTest.php
DFurnes: +1! This is a really nice shortcut. 🤩

routes/channels.php
```diff
@@ -0,0 +1,18 @@
...
+|
+| Here you may register all of the event broadcasting channels that your
+| application supports. The given channel authorization callbacks are
+| used to check if an authenticated user can listen to the channel.
+|
+*/
+
+Broadcast::channel('App.Models.User.{id}', function ($user, $id) {
+    return (int) $user->id === (int) $id;
+});
```
DFurnes, line 18: I think we might need to modify our [`BroadcastServiceProvider`](https://github.com/DoSomething/northstar/blob/758a741a3032c4bf042655f3e20577e677a5595f/app/Providers/BroadcastServiceProvider.php) to match [the framework example](https://github.com/laravel/laravel/blob/8e5510458e1a4c0bf4b78024d9b0cf56102c8207/app/Providers/BroadcastServiceProvider.php#L10-L20) in order for this `routes/channels.php` to be sourced.

routes/console.php
```diff
@@ -0,0 +1,19 @@
...
+|
+| This file is where you may define all of your Closure based console
+| commands. Each Closure is bound to a command instance allowing a
+| simple approach to interacting with each command's IO methods.
+|
+*/
+
+Artisan::command('inspire', function () {
+    $this->comment(Inspiring::quote());
+})->purpose('Display an inspiring quote');
```
DFurnes, line 19: Likewise, I think we might need to source this file in our [`Console/Kernel.php`](https://github.com/DoSomething/northstar/blob/758a741a3032c4bf042655f3e20577e677a5595f/app/Console/Kernel.php#L21-L29) (like [this](https://github.com/laravel/laravel/blob/8e5510458e1a4c0bf4b78024d9b0cf56102c8207/app/Console/Kernel.php#L30-L40))!

DFurnes (approved):

DFurnes:
Fixed up the failing `northstar:setup` step on CircleCI builds by installing [`dfurnes/environmentalist@^2.0`](https://github.com/DFurnes/environmentalist/pull/3) in 758a741 & then fixed a [failing test](https://app.circleci.com/pipelines/github/DoSomething/northstar/1371/workflows/b92853de-4fdd-4ff3-a333-54ff30e7f9fc/jobs/1439) from the [tweaked language files](https://github.com/DoSomething/northstar/pull/1230/commits/583a5a284f5387343ce4380dac27ca9adb0fa0ae) in 703d92e. Builds are now passing! 💚 

Last step – wrap up & publish next Gateway release.

routes/channels.php
weerd: Yep yep! I was slowly making my way through upgrading the framework files and hand't gotten to that part yet. I was doing what I could before merging and that shouldn't affect functionality. I can do this and continue with framework file updates in following PR so we can get this one merged!

routes/console.php
weerd: [Yep!](https://github.com/DoSomething/northstar/pull/1230#discussion_r670671417)

routes/channels.php
weerd: But I can do it in this one and get it in before merge if preferred! Just lemme know.

DFurnes: Yeah, let's add those two changes into this PR to complete this task (just in case we don't get around to the other half of the job & then find ourselves tripped up by this file not working as expected down the line).

PR #1232, opened 2021-06-29
Extract recounts into a queued job.

DFurnes:
### What's this PR do?

This pull request extracts the database query to recount accepted/pending posts into a queue job. This takes this hefty query off of the main web thread & attempts to reduce "over-calculating" this when we don't need to.

I'll add some notes in the diff!

Ideally, we'll also want to find how we can make this query less intensive (likely by adding a more appropriate index than the one the query planner is currently going with). However, that's proving to be tricker than I'd thought it would be & this gets us a quick fix! We'll revisit the query itself once we've migrated over to PostgreSQL.

### Relevant tickets

References [Pivotal #178612113](https://www.pivotaltracker.com/story/show/178612113).

### Checklist

### DIFF
--- a/app/Jobs/RefreshCampaignPostCounts.php
+++ b/app/Jobs/RefreshCampaignPostCounts.php
@@ -1,0 +1,71 @@
+<?php
+
+namespace App\Jobs;
+
+use App\Models\Campaign;
+use App\Models\Post;
+use Illuminate\Bus\Queueable;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Foundation\Bus\Dispatchable;
+use Illuminate\Queue\InteractsWithQueue;
+use Illuminate\Queue\SerializesModels;
+
+class RefreshCampaignPostCounts implements ShouldQueue
+{
+    use Dispatchable, InteractsWithQueue, Queueable, SerializesModels;
+
+    /**
+     * The campaign we're refreshing.
+     *
+     * @var Campaign
+     */
+    protected $campaign;
+
+    /**
+     * When was this job dispatched?
+     *
+     * @var \Carbon\Carbon
+     */
+    protected $dispatchedAt;
+
+    /**
+     * Create a new job instance.
+     *
+     * @return void
+     */
+    public function __construct(Campaign $campaign)
+    {
+        $this->campaign = $campaign;
+        $this->dispatchedAt = now();
+
+        $this->onQueue(config('queue.names.low'));
+    }
+
+    /**
+     * Execute the job.
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        // If we've already updated this campaign after this job was dispatched,
+        // we can probably skip recounting to reduce unnecessary database load:
+        if ($this->campaign->updated_at > $this->dispatchedAt) {
+            return;
+        }
+
+        $accepted_count = Post::getPostCount($this->campaign, 'accepted');
+        $pending_count = Post::getPostCount($this->campaign, 'pending');
+
+        $this->campaign->pending_count = $pending_count;
+        $this->campaign->accepted_count = $accepted_count;
+
+        info('Recalculated post counts', [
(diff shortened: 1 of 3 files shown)


app/Jobs/RefreshCampaignPostCounts.php
```diff
@@ -0,0 +1,71 @@
...
+     * Create a new job instance.
+     *
+     * @return void
+     */
+    public function __construct(Campaign $campaign)
+    {
+        $this->campaign = $campaign;
+        $this->dispatchedAt = now();
+
+        $this->onQueue(config('queue.names.low'));
```
DFurnes, line 41: Since this is a "heftier" job, we'll place it on our low-priority queue so more time-sensitive jobs (like sending profile updates & events to Customer.io) don't get blocked behind this task.

app/Jobs/RefreshCampaignPostCounts.php
```diff
@@ -0,0 +1,71 @@
...
+use App\Models\Post;
+use Illuminate\Bus\Queueable;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Foundation\Bus\Dispatchable;
+use Illuminate\Queue\InteractsWithQueue;
+use Illuminate\Queue\SerializesModels;
+
+class RefreshCampaignPostCounts implements ShouldQueue
+{
+    use Dispatchable, InteractsWithQueue, Queueable, SerializesModels;
```
weerd, line 15: Should we just extend the `app/Jobs/Job.php` class that we brought in from Rogue that includes the above?

app/Jobs/RefreshCampaignPostCounts.php
```diff
@@ -0,0 +1,71 @@
...
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        // If we've already updated this campaign after this job was dispatched,
+        // we can probably skip recounting to reduce unnecessary database load:
+        if ($this->campaign->updated_at > $this->dispatchedAt) {
+            return;
+        }
```
DFurnes, line 55: Because these are placed onto our low-priority queue, I think there'll be a decent chance that they'll pile up before we get around to calculating them. If that's the case, we can "batch up" a bunch of recalculations for the same campaign into one call (and discard all the rest).

This isn't a perfect way of doing this (since any changes to a campaign, such as renaming it) could also cause us to skip recounting, but I think that's probably unlikely to happen in practice. We can keep an eye on logs to see how many of these jobs end up being skipped in practice!

weerd (approved):
Small comment! Otherwise looks great and nice to have this moved over into a queue! 👏

app/Models/Campaign.php
```diff
@@ -100,17 +100,6 @@ public function posts()
...
-    /**
-     * Run a quick 'COUNT(*)' query to get the count of pending and accepted
-     * posts for this campaign (so we can use this efficiently later).
-     */
-    public function refreshCounts()
-    {
-        $this->pending_count = Post::getPostCount($this, 'pending');
-        $this->accepted_count = Post::getPostCount($this, 'accepted');
-        $this->save();
-    }
```
DFurnes, line 112: I removed this method from the `Campaign` model to discourage using this synchronously, since we really want to push all this work to the queued job. (Also chuckling to myself at the "quick `COUNT(*)` query" from the docblock… it was once… 😂)

app/Jobs/RefreshCampaignPostCounts.php
DFurnes: Oh yes, definitely! I just used the `make:job` generator… wonder if we can customize those to use our preferred templates.

weerd: Ya! That'd be cool. Otherwise adding a new command based off original 🤷‍♂️

DFurnes: Updated in 39357bf! Good catch! 🙇

PR #1234, opened 2021-06-30
Remove events logic.

DFurnes:
### What's this PR do?

This pull request removes model events, relationships, and controller/transformer for our old events logic. This is no longer being used in any application logic & the table has gotten far rather unwieldy so we're going to clear it out in preparation for our upcoming PostgreSQL migration.

🔥🔥🔥

Once this is deployed, I'll follow up with another pull request to drop the actual table.

### Relevant tickets

References [Pivotal #178201917](https://www.pivotaltracker.com/story/show/178201917).

### Checklist


app/Providers/ModelServiceProvider.php
```diff
@@ -18,34 +18,13 @@ class ModelServiceProvider extends ServiceProvider
...
-                    // Use the authenticated user if coming from the web,
-                    // otherwise use the id of the user in the request.
-                    'user' => auth()->id()
-                        ? auth()->id()
-                        : $signup->northstar_id,
-                ]);
-            }
-        });
-
-        // When Posts are created reject test events.
```
katiecrane, line 42: Not sure if you meant to remove this comment or not!

app/Models/Post.php
```diff
@@ -495,8 +487,6 @@ public function untag($tagName)
         // Update timestamps on the Post when removing a tag
         $this->touch();
 
-        // @TODO: create an event here when we refactor events system.
```
katiecrane, line 498: Joke's on you TODO!

app/Providers/ModelServiceProvider.php
DFurnes: Yeah, it felt somewhat confusing to refer to "test events" here (since it's really test posts).

I figured there was probably enough context from the conditional & job below.

app/Models/Post.php
DFurnes: 😂

app/Providers/ModelServiceProvider.php
katiecrane: Oh you know what, I think there is just a mis-wording in this comment and it should say "When test Posts are created, reject them" since nothing with events is actually happening here is it?

DFurnes: Yeah, that'd make much more sense!

DFurnes: Cleaned this up in a48312c!

katiecrane (approved):
SO SQUEAKY CLEAN!!!

https://www.youtube.com/watch?v=ABYnqp-bxvg

PR #1237, opened 2021-07-16
Fix issue with 'status' route parameter clobbering Laravel internals.

DFurnes:
### What's this PR do?

This pull request fixes a _gnarly_ bug that was causing the reportback index to error out when loading via direct link.

This bug was caused by a change in Laravel 8 that made `status` a "reserved word" for [view routes](https://laravel.com/docs/8.x/routing#view-routes).

If you added a `status` parameter, it would "clobber" Laravel's internal status variable provided to the [`ViewController`](https://github.com/laravel/framework/blob/83b70546fc43fa3278d5d101141033fd37edb4c6/src/Illuminate/Routing/ViewController.php#L35). Since our "status" route parameter would overwrite that internal property, Laravel could no longer rely on grabbing the last four elements of this array to get the expected values… and it would instead try to load the `pending` view!

This was a tricky one! Great pairing session with @weerd! 🕵️ 🕵️ 

### Relevant tickets

References [Pivotal #178913541](https://www.pivotaltracker.com/story/show/178913541).

### Checklist

Co-Authored-By: Diego Lorenzo <diego@diegolorenzo.com>


katiecrane (approved):
oh this is tricky!! good find!

PR #1238, opened 2021-07-16
Drop more unused tables.

DFurnes:
### What's this PR do?

This pull request drops unused MariaDB tables & MongoDB collections, in preparation for our upcoming PostgreSQL move!

I'll drop some comments with context into the diff!

🔥 🗑️

### Relevant tickets

References [Pivotal #178201917](https://www.pivotaltracker.com/story/show/178201917).

### Checklist


database/migrations-mongodb/2021_07_16_161844_drop_api_keys_table.php
```diff
@@ -0,0 +1,28 @@
...
+class DropApiKeysTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::connection('mongodb')->dropIfExists('api_keys');
```
DFurnes, line 16: Huh – I'd included this on my list of tables to drop, but now I'm not seeing it when I check any of Northstar's databases. This table existed a long time back but was [renamed to `clients`](https://github.com/DoSomething/northstar/blob/4dbda054e21cffcce1e43cd882f2f647a2356607/database/migrations-mongodb/2016_04_15_192326_RenameClientTable.php) back in 2016.

This migration will just be a no-op since the table doesn't, in fact, exist! 🙃

database/migrations-mongodb/2021_07_16_161901_drop_tokens_table.php
```diff
@@ -0,0 +1,28 @@
...
+class DropTokensTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::connection('mongodb')->dropIfExists('tokens');
```
DFurnes, line 16: This table _does_ currently exist & is no longer used! It was created for our pre-OAuth homegrown authentication tokens, and has not had any new data written to it since October 2017.

database/migrations/2021_07_16_160354_drop_clients_table.php
```diff
@@ -0,0 +1,33 @@
...
+class DropClientsTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::drop('clients');
```
DFurnes, line 16: This table was used by Rogue to store it's own access tokens (for example, used when queued jobs would then make requests to Northstar or other internal OAuth-secured APIs). Now that this is just one application, it no longer has a need to fetch or store any access tokens!

database/migrations/2021_07_16_160642_drop_jobs_table.php
```diff
@@ -0,0 +1,36 @@
...
+class DropJobsTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::drop('jobs');
```
DFurnes, line 16: This table is used to power Laravel's [database queue driver](https://laravel.com/docs/8.x/queues#database) (which stores queued jobs in a SQL table, rather than a more traditional queue backend like Redis or SQS). I don't believe we ever actually used this, and so it's probably best to remove to avoid any confusion.

database/migrations/2021_07_16_161442_drop_cache_table.php
```diff
@@ -0,0 +1,28 @@
...
+class DropCacheTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::dropIfExists('cache');
```
DFurnes, line 16: I have no clue what's up with this one! Our production Rogue database has a `cache` table which looks like this:

```sql
CREATE TABLE `cache` (
  `key` varchar(255) COLLATE utf8_unicode_ci NOT NULL,
  `value` text COLLATE utf8_unicode_ci NOT NULL,
  `expiration` int(11) NOT NULL,
  UNIQUE KEY `cache_key_unique` (`key`)
)
```

We don't have a migration for it, so I'm honestly not sure where it came from. The table is empty. 👻

DFurnes: My best guess is this is somehow left over from super early Rogue development, or perhaps somehow accidentally got copied over from our old Drupal database during an attempt at a data migration? Unclear!

weerd (approved):
Huzzah! 🗑️ 🎉

PR #1239, opened 2021-07-16
Laravel 8 cleanup 🧽

weerd:
### What's this PR do?

This pull request includes additional cleanups and updates to get the codebase closer to the latest version of the Laravel framework version 8.

👀 

Getting these core framework files updated will help with making future updates easier and less confusing in cases where a specific file hasn't been updated for a number of versions and the differences can seem confusing.

### Relevant tickets

References [Pivotal #178938855](https://www.pivotaltracker.com/story/show/178938855).

### Checklist


config/session.php
```diff
@@ -187,11 +189,11 @@
...
-    | do not enable this as other CSRF protection services are in place.
+    | will set this value to "lax" since this is a secure default value.
     |
-    | Supported: "lax", "strict", "none"
+    | Supported: "lax", "strict", "none", null
     |
     */
 
-    'same_site' => null,
+    'same_site' => 'lax',
```
weerd, line 198: Just want to confirm this is ok!

weerd:
There were 4 files that I did not fully address because the changes made throughout Northstar's history seemed purposeful and I did not want to mess with their configuration. I'll note them in the Pivotal ticket for future reference!

app/Http/Middleware/TrustProxies.php
```diff
@@ -10,14 +10,19 @@ class TrustProxies extends Middleware
     /**
      * The trusted proxies for this application.
      *
-     * @var array|string
+     * @var array|string|null
      */
-    protected $proxies = '*';
+    protected $proxies;
```
mendelB, line 15: I noticed [this](https://laravel.com/docs/8.x/requests#trusting-all-proxies) in the documentation, are we using AWS Elastic Load Balancing? (I see below that we assign it as one of the `$headers`)

app/Http/Middleware/RedirectIfAuthenticated.php
```diff
@@ -31,13 +32,17 @@ public function __construct(Auth $auth)
...
-     * @param  string|null  $guard
+     * @param  string|null  ...$guards
      * @return mixed
      */
-    public function handle($request, Closure $next, $guard = 'api')
+    public function handle(Request $request, Closure $next, ...$guards)
     {
-        if ($this->auth->guard($guard)->check()) {
-            return $this->handleAuthenticated($request);
+        $guards = empty($guards) ? [null] : $guards;
```
mendelB, line 40: Is there a specific reason to populate the default array with `null` instead of leaving it empty?

mendelB (approved):
LGTM! 👍 Just a few Qs

config/session.php
DFurnes: Agreed, I think this should be good!

config/cache.php
```diff
@@ -79,6 +82,10 @@
             'table' => env('DYNAMODB_CACHE_TABLE', 'cache'),
             'endpoint' => env('DYNAMODB_ENDPOINT'),
         ],
+
+        'octane' => [
+            'driver' => 'octane',
+        ],
```
DFurnes, line 88: Ooh, this is another one of the things I'd love to set aside some time to play with now that we're on Laravel 8!! 🏎️

config/logging.php
```diff
@@ -44,14 +44,14 @@
...
-            'level' => 'debug',
+            'level' => env('LOG_LEVEL', 'debug'),
         ],
 
         'daily' => [
             'driver' => 'daily',
             'path' => storage_path('logs/laravel.log'),
             'tap' => [App\Logging\ContextFormatter::class],
-            'level' => 'debug',
+            'level' => env('LOG_LEVEL', 'debug'),
```
DFurnes, line 54: Oh, excellent! We should set this `LOG_LEVEL=info` in our base [`heroku_app`](https://github.com/DoSomething/infrastructure/blob/main/components/heroku_app/main.tf) Terraform module.

app/Http/Middleware/TrustProxies.php
DFurnes: Our applications sit behind [Heroku's router](https://devcenter.heroku.com/articles/http-routing#routing) (which handles load balancing and, like ELB, has dynamic IP addresses that we can't specifically whitelist). Let's keep this as `*` so that we continue trusting these in production.

app/Http/Middleware/TrustProxies.php
```diff
@@ -10,14 +10,19 @@ class TrustProxies extends Middleware
...
      *
      * @var int
      */
-    protected $headers = Request::HEADER_X_FORWARDED_AWS_ELB;
+    protected $headers =
+        Request::HEADER_X_FORWARDED_FOR |
+        Request::HEADER_X_FORWARDED_HOST |
+        Request::HEADER_X_FORWARDED_PORT |
+        Request::HEADER_X_FORWARDED_PROTO |
+        Request::HEADER_X_FORWARDED_AWS_ELB;
```
DFurnes, line 27: Likewise, I think we only want to allow `HEADER_X_FORWARDED_AWS_ELB` headers since some of these (like `HEADER_X_FORWARDED_HOST` _aren't_ sent by Heroku and so we could get ourselves in trouble by trusting them).

DFurnes: Here's some [documentation from Heroku](https://devcenter.heroku.com/articles/getting-started-with-laravel#trusting-the-load-balancer) that explains how this works & why we use these settings.

app/Providers/EventServiceProvider.php
```diff
@@ -30,12 +30,12 @@ class EventServiceProvider extends ServiceProvider
...
     /**
-     * Register any other events for your application.
+     * Register any events for your application.
      *
      * @return void
      */
     public function boot()
     {
-        parent::boot();
+        //
```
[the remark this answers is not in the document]
weerd: Ya, I just followed along with what the update was, but I was curious and looked at the change and seems it's just [not necessary](https://github.com/laravel/laravel/commit/d53c56e4e0bdc2b9a3c14511bdcbe48c4914f084).

config/session.php
```diff
@@ -165,7 +167,7 @@
     |
     */
 
-    'secure' => env('SESSION_SECURE_COOKIE', null),
+    'secure' => env('SESSION_SECURE_COOKIE'),
```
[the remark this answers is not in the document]
weerd: Not 100% sure but it's this way in the core framework files. I suspect maybe to ensure that if it _is_ used somewhere and nothing is set for it it throws an exception instead of failing silently 🤷‍♂️

app/Http/Middleware/RedirectIfAuthenticated.php
weerd: 🤔  Ya, I was a bit curious about that but didn't get a chance to dig into it. Just a change that was made and shows up as such in Laravel 8. [This is when the change was made](https://github.com/laravel/laravel/pull/5329/files).

app/Http/Middleware/TrustProxies.php
weerd: Ah, yep. I think I got confused here and thought this was a change because the extended from class defaulted to `*` but I checked and was wrong. Will change this back!

config/cache.php
weerd: Ya definitely!

PR #1240, opened 2021-07-19
Fix issue submitting Change Password form.

DFurnes:
### What's this PR do?

This pull request fixes an issue submitting the "Change Password" form. Because we were using an `{id}` route parameter, but expecting a `{user}` route model binding, this form would not correctly resolve the intended user. (It would, instead, inject an uninitialized `User` instance.)

I added a test to ensure this form works as expected.

This was a (unrelated) bug that I noticed while testing our Laravel 8 upgrade! 🐛

### Relevant tickets

References [Pivotal #178903994](https://www.pivotaltracker.com/story/show/178903994).

### Checklist


routes/web.php
```diff
@@ -53,7 +53,7 @@
 Route::patch('profile/subscriptions', 'ProfileSubscriptionsController@update');
 
 // Change Password
-Route::patch('users/{id}/password', 'PasswordController@update')->name(
+Route::patch('users/{user}/password', 'PasswordController@update')->name(
```
weerd, line 56: 🔍

tests/Http/Web/ChangePasswordTest.php
```diff
@@ -0,0 +1,73 @@
...
+    public function testCannotChangeOtherPasswords()
+    {
+        $attacker = factory(User::class)->create();
+        $rando = factory(User::class)->create([
+            'password' => 'old-password',
+        ]);
+
+        $this->be($attacker, 'web');
+
+        $newPassword = 'hackingggggg1231231232';
```
weerd, line 62: 😱

weerd (approved):
Awesome addition of the test for this! 👏

tests/Http/Web/ChangePasswordTest.php
weerd: LOL 😹

PR #962, opened 2019-11-18
Update dependencies, including PHP 7.3, and switch to CircleCI.

DFurnes:
#### What's this PR do?
This pull request updates this application's dependencies & switches our PHP version requirement to `7.3.*`. Since our homegrown Wercker box was stuck on PHP 7.2, I've taken this opportunity to move this application over to CircleCI for continuous integration.

#### How should this be reviewed?

This is very similar to our [Phoenix Next CircleCI config](https://github.com/DoSomething/phoenix-next/blob/master/.circleci/config.yml), except we swap out MariaDB for MongoDB.

#### Relevant Tickets

[#169214201](https://www.pivotaltracker.com/story/show/169214201)

#### Checklist

- [ ] Documentation added for new or changed endpoints.
- [ ] Tests added for new features/bug fixes/API updates.
- [ ] Post a message in #dev-northstar if this includes something that causes a rebuild!
- [ ] If new attributes were added to users, there is a corresponding PR to surface these in Aurora.
- [ ] If new attributes were added to users, then the data team already knows about these changes.
- [ ] This PR has been added to the relevant Pivotal card.

PR #1241, opened 2021-07-19
Experimental: Reduce per-request memory limit.

DFurnes:
### What's this PR do?

We currently allocate 256MB per request due to Northstar's Image API, which needs a lot of memory in order to format & resize images. However, the [overwhelming majority](https://gist.github.com/DFurnes/6119edab9a8fbecc4d5e18a7e6fb77a5) of requests use less than 5MB of memory!

This pull request experiments with setting a lower `memory_limit` per request (from 256MB to 32MB per request).

I've set the [global memory limit](https://devcenter.heroku.com/articles/php-concurrency#configuring-the-memory_limit-for-php-fpm) to `32M` (which determines how much memory to allocate per request, and thus how many concurrent requests Heroku's router sends to each dyno before they start queueing up). This should still give most requests decent headroom, while 8x'ing the amount of normal traffic each dynos can handle.

Since the Images API requires more memory, I used `ini_set` to [increase `memory_limit` at runtime](https://devcenter.heroku.com/articles/php-concurrency#runtime-changes-of-memory_limit) for this endpoint.

Since we get so few hits to the Images API due to caching those responses in Fastly, it should be unlikely that we get too many requests there at once. If we do, I think the worst-case scenario is that we [page to disk](https://devcenter.heroku.com/articles/error-codes#r14-memory-quota-exceeded) (slowing performance, but not immediately failing the request). If that happens, we'll see it in our logs.

This brings us back to Northstar's (pre-Rogue) memory limit, and should make it easier for us to scale up to smaller spikes in traffic. I'm planning to deploy this first to QA & test a bit by hand, and then send it up to production & keep an eye on logs/metrics.

### Relevant tickets

### Checklist


.circleci/config.yml
```diff
@@ -63,7 +63,7 @@ jobs:
           name: run phpunit tests
           command: |
             mkdir -p ~/phpunit
-            vendor/bin/phpunit -d memory_limit=512M --log-junit ~/phpunit/junit.xml
+            vendor/bin/phpunit --log-junit ~/phpunit/junit.xml
```
DFurnes, line 66: Since all of the PHPUnit tests run within the same process, I had to move this to an `ini_set` within PHPUnit's `setUp` method so that it didn't get permanently overwritten to `256M` by the `ImagesController`.

DFurnes: For context, I'd originally set this to `512M` in #962 due to [out-of-memory issues](https://github.com/DoSomething/northstar/pull/962#discussion_r347632916) caused by Faker. I think that, since all these tests run in the same process, they're more susceptible to memory leaks & so need a bit more headroom than a "single-use" web request.

tests/CreatesApplication.php
```diff
@@ -86,6 +86,9 @@ public function createApplication()
      */
     public function setUp(): void
     {
+        // Increase memory limit for test runner:
+        ini_set('memory_limit', '512M');
```
DFurnes, line 90: See above change to `.circleci/config.yml` for context on why we set this memory limit for tests.

katiecrane (approved):
ooo this is really interesting and also looks good to me! excited to see how this goes

PR #1242, opened 2021-07-20
Signups Index - Campaign Website Filter

mendelB:
### What's this PR do?

This pull request adds the `has_website` filter (similarly to the [`/campaigns` endpoint](https://github.com/DoSomething/northstar/blob/3c6040d66d560c4d6e977b212a45c2eb5e23a216/app/Http/Controllers/CampaignsController.php#L54-L61)) to the `/signups` index.

👀 

For the Account Campaigns page on Phoenix, we're running a GraphQL query to fetch member's signups & associated Northstar Campaigns, and in turn, associated Website (Contentful) Campaigns. 

This explodes when the Signup's Northstar Campaign has no associated Website (`contentful_campaign_id` attribute).

We can use this filter to avoid this issue.

While this should only affect older accounts / staff accounts. But peering into the database, I found a huge quantity of such signups. This feels like a good preemptive move which will also make QA-ing the new User Campaigns page easier.
 
### Relevant tickets

References [Pivotal #178248537](https://www.pivotaltracker.com/story/show/178248537/comments/225561969).
[Slack](https://dosomething.slack.com/archives/CUQMU4Q6B/p1626377621005000?thread_ts=1626375918.004000&cid=CUQMU4Q6B)
https://www.pivotaltracker.com/story/show/178248559/comments/225583872

### Checklist


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -19,5 +19,5 @@ References [Pivotal #]().
 - [ ] This PR has been added to the relevant Pivotal card.
 - [ ] Documentation added for new features/changed endpoints.
 - [ ] Added appropriate feature/unit tests.
-- [ ] If new attributes were added to users, there is a corresponding PR to surface these in Aurora.
+- [ ] If new attributes were added to users, these changes are also surfaced in the admin interface for user profiles.
```
mendelB, line 22: Just happened to notice this one! Does the new copy make sense?

app/Http/Controllers/SignupsController.php
```diff
@@ -117,6 +117,15 @@ public function index(Request $request)
             $query = $query->withVisiblePosts($types);
         }
 
+        // Apply scope for the has_website filter:
+        if (isset($filters['has_website'])) {
```
mendelB, line 121: I went with `has_website` to keep it the same as the `/campaigns` filter, but I could hear the argument for e.g. `campaign_has_website`.... 🤔

DFurnes (approved):

app/Http/Controllers/SignupsController.php
DFurnes: Hmm, yeah this is a tricky one to name! I feel like `has_website` is probably fine for now, since I can't imagine a future where we have websites for signups that this could conflict with....

app/Http/Controllers/SignupsController.php
```diff
@@ -117,6 +117,15 @@ public function index(Request $request)
...
+            if (filter_var($filters['has_website'], FILTER_VALIDATE_BOOLEAN)) {
+                $query->WhereCampaignHasWebsite();
```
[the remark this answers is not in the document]
mendelB: Oops! Copy pasta will get ya 🤦 fixed in https://github.com/DoSomething/northstar/pull/1242/commits/1d07da2d4cd6f8612c2de24df2b61b703e2e429c!

PR #1244, opened 2021-07-20
Revert "Experimental: Reduce per-request memory limit."

DFurnes:
This pull request reverts DoSomething/northstar#1241, as I continue to investigate some issues with image uploads.


katiecrane (approved):
can confirm this is the opposite of yesterday's!

PR #1245, opened 2021-07-22
Profile Subscriptions Form - Ensure no SMS Status is Assigned Without a Mobile

mendelB:
### What's this PR do?

This pull request ensures that we don't assign an SMS status to users when they haven't provided a mobile number. It clears the SMS status entirely ensuring that members with no mobile assigned don't have a misleading SMS status.

👀 

I copied over the logic we use for this for the profile edit form in the web [UserController#update](https://github.com/DoSomething/northstar/blob/28dda3419e4abd34995431ae059c9f6792fea7ce/app/Http/Controllers/Web/UserController.php#L120-L124). 

I wonder if we might extract this logic into the [Registrar#register](https://github.com/DoSomething/northstar/blob/28dda3419e4abd34995431ae059c9f6792fea7ce/app/Auth/Registrar.php#L224-L248) method as more of a catch-all? (I neglected to do this since it seems like we'd rather carefully place this in the `UserObserver` at some point per the comment).

### Relevant tickets

References [Pivotal #177567262](https://www.pivotaltracker.com/story/show/177567262).

### Checklist


app/Http/Controllers/Web/ProfileSubscriptionsController.php
```diff
@@ -58,6 +58,12 @@ public function update(Request $request)
             $input['email_subscription_topics'] = [];
         }
 
+        // HACK: Clear any existing SMS status values if we don't have the user's mobile number.
+        // This should likely happen in UserObserver -- but that logic is pretty complex.
```
DFurnes, line 62: Yeah, I agree re: `UserObserver`! It could be worth taking some time to untangle all that stuff so it's not to intimidating to try to fit more things in there... maybe something we can backlog into a sprint next quarter.

DFurnes (approved):

DFurnes:
> I wonder if we might extract this logic into the [Registrar#register](https://github.com/DoSomething/northstar/blob/28dda3419e4abd34995431ae059c9f6792fea7ce/app/Auth/Registrar.php#L224-L248) method as more of a catch-all? (I neglected to do this since it seems like we'd rather carefully place this in the `UserObserver` at some point per the comment).

Yeah, I think this was the right trade-off for now! I'd love to set some time aside in the future & think about the different ways we set things like this (via `Registrar`, `UserObserver`, mutators on the model...) and standardize/clean up.

mendelB:
> > I wonder if we might extract this logic into the [Registrar#register](https://github.com/DoSomething/northstar/blob/28dda3419e4abd34995431ae059c9f6792fea7ce/app/Auth/Registrar.php#L224-L248) method as more of a catch-all? (I neglected to do this since it seems like we'd rather carefully place this in the `UserObserver` at some point per the comment).
> 
> Yeah, I think this was the right trade-off for now! I'd love to set some time aside in the future & think about the different ways we set things like this (via `Registrar`, `UserObserver`, mutators on the model...) and standardize/clean up.

Oh heck yes, that would be awesome! (And heavy +1 to https://github.com/DoSomething/northstar/pull/1245#discussion_r674854964). Especially with this User model where we've got so much complexity across different flows, it'd be so neat to have this all clean & standardized 😌

PR #1246, opened 2021-07-22
Use ImageMagick instead of GD for image processing.

DFurnes:
### What's this PR do?

This pull request swaps [Intervention](http://image.intervention.io) and [Glide](https://glide.thephpleague.com) to use [ImageMagick](https://imagemagick.org/index.php) to process (crop, resize, reformat, etc) images, rather than [GD](https://libgd.github.io). This significantly reduces memory usage on the `images/{id}` endpoint (like from ~100MB down to 20MB!!)

I installed the ImageMagick PHP extension (and uninstalled the GD extension) in our `composer.json` ([docs](https://devcenter.heroku.com/articles/php-support#extensions)).

I've [published Intervention's configuration file](http://image.intervention.io/getting_started/configuration), which allows us to swap which driver this library uses for image processing. Although Glide uses Intervention behind the scenes, it requires it's [own configuration](https://glide.thephpleague.com/2.0/config/image-driver/) (which I've decided to read from that same `config/image.php` setting).

I left a lot of notes in [this ticket](https://www.pivotaltracker.com/story/show/178903960) while trying to figure out the root cause of this issue if you want to follow along on the journey!

### Relevant tickets

### Checklist


.circleci/config.yml
```diff
@@ -26,7 +26,7 @@ jobs:
           name: install php extensions
           command: |
             sudo apt-get install libmagickwand-dev --no-install-recommends
-            yes '' | sudo pecl install mongodb redis imagick
+            sudo docker-php-pecl-install install mongodb redis imagick
```
DFurnes: Alas! It turns out this [really useful helper](https://github.com/OpenMage/docker-php-base/blob/master/src/root/usr/local/bin/docker-php-pecl-install) was _not_ in the official base PHP box.

.circleci/config.yml
```diff
@@ -25,10 +25,12 @@ jobs:
       - run:
           name: install php extensions
           command: |
-            sudo pecl install mongodb redis
-            sudo apt-get install libpng-dev libjpeg-dev
+            sudo apt-get install libpng-dev libjpeg-dev libmagickwand-dev --no-install-recommends
+            (yes '' || true) | sudo pecl install mongodb
+            (yes '' || true) | sudo pecl install redis
+            (yes '' || true) | sudo pecl install imagick
```
DFurnes, line 31: The `yes ''` command repeatedly pipes a newline character, which auto-selects any prompts from `pecl install` (which is necessary there's no `--non-interactive` option to avoid prompts, and so those would otherwise just wait forever and time out).

The `yes` command just keeps printing the same "enter key" to the console indefinitely, so by default this fails with an error code 147 due to `pecl install` not consuming all of the provided input. We can use the `|| true` on `yes` to swallow up that error code from `yes` while still allowing errors from `pecl install` to fail the build.

.circleci/config.yml
```diff
@@ -25,10 +25,12 @@ jobs:
...
             sudo docker-php-ext-configure gd --with-jpeg
-            sudo docker-php-ext-enable mongodb redis
+            sudo docker-php-ext-enable mongodb redis imagick
```
DFurnes, line 33: This (alongside the `libmagickwand-dev` above) installs ImageMagick on this Docker box. I've kept `gd` installed because Laravel uses it to generate fake images for the `UploadedFile` helper.

.circleci/config.yml
DFurnes: I've also split out the `pecl install` steps into separate lines based on [the documentation for this Docker box](https://hub.docker.com/_/php). Otherwise, any errors installing any but the last extension are just ignored! 😯

DFurnes: (They also suggest pinning exact extension versions, which I don't think we want to do here since that means more overhead for us to make sure we update these over time. If we ever end up installing a version that isn't compatible with our PHP version, we can pin then!)

app/Http/Controllers/Web/ImagesController.php
```diff
@@ -52,6 +52,7 @@ public function show(string $hash, Request $request)
             'response' => new LaravelResponseFactory($request),
             'cache' => new Flysystem(new MemoryAdapter()),
             'source' => $this->filesystem->getDriver(),
+            'driver' => config('image.driver'),
```
DFurnes, line 55: Even though Glide uses Intervention behind the scenes, it _doesn't_ use Intervention's default driver. To make sure this also uses ImageMagick, I've just referenced the `image.driver` config here in the Glide constructor.

.circleci/config.yml
mendelB: Whoa. But who doesn't appreciate an agreeable config ☺️

mendelB (approved):
> This significantly reduces memory usage on the images/{id} endpoint (like from ~100MB down to 20MB!!)

😲 👏 🏅 

This is an epic adventure. Major kudos on this spectacular investigation, and making this so easy to follow along here and in the ticket! 🙌

PR #1248, opened 2021-07-26
Reduce memory usage when uploading/resizing images.

DFurnes:
### What's this PR do?

This pull request reduces memory usage when uploading new photos (to `v3/posts`) or rotating them (via `v3/posts/{id}/rotate`). In both cases, we needed to ensure that we were operating upon a temporary file on the disk rather than loading the whole image as a string in the PHP request. 🏋️ 

This one is probably easiest to review commit-by-commit. I'll add some comments with context.

As in #1246, I left [some comments](https://www.pivotaltracker.com/story/show/178903960/comments/225760808) on my process debugging this in the ticket!

### Relevant tickets

### Checklist


app/Http/Controllers/RotationController.php
```diff
@@ -57,15 +57,24 @@ public function update(Post $post, Request $request)
...
+            // Write edited image to S3 & purge cache:
+            $this->storage->edit($post, $rotatedImage);
+            $this->fastly->purge($post);
+        } finally {
+            // If we wrote a temporary file to disk, always clean up after
+            // ourselves so the disk doesn't get cluttered up:
+            if ($image) {
+                unlink($image->basePath());
+            }
+        }
```
DFurnes, line 77: I moved all this inside a `try / finally` block so that in case any exceptions occur while fetching or rotating the image, we'll still delete this temporary file. (Since there's no `catch`, those exceptions will still bubble up to the handler and fail the request!)

app/Services/ImageStorage.php
```diff
@@ -9,27 +9,31 @@
...
     /**
-     * Get image for the given post.
+     * Get image for the given post. Be sure to delete this file with 'unlink'
+     * when you're done with it, so we don't clutter up temporary directory!
      *
      * @param Post $post
      * @return Image
      */
-    public function get(Post $post)
+    public function download(Post $post)
```
DFurnes, line 28: I renamed this method from `get` to `download` to make it more clear that it's now creating a temporary resource on the local disk (and added a note in the docblock reminding anyone who uses it to clean up after themselves). If anyone can think of a better API to make this more obvious, I'm down to refactor!

app/Services/ImageStorage.php
```diff
@@ -40,30 +44,15 @@ public function get(Post $post)
...
     {
-        $extension = $file->guessExtension();
-        $contents = file_get_contents($file->getPathname());
+        $path = Storage::putFile($this->base, $file);
 
-        // Make sure we're only uploading valid image types
-        if (!in_array($extension, ['jpeg', 'jpg', 'png', 'gif'])) {
-            throw new UnprocessableEntityHttpException(
-                'Invalid file type. Upload a JPEG, PNG or GIF.',
-            );
```
DFurnes, line 52: Since we already validate mime types in [`PostRequest`](https://github.com/DoSomething/northstar/blob/de2246275aa5a3016c84961d2bfc982d3f31506d/app/Http/Requests/PostRequest.php#L53), this seemed redundant to me.

app/Services/ImageStorage.php
```diff
@@ -40,30 +44,15 @@ public function get(Post $post)
...
-            $this->base .
-            $signupId .
-            '-' .
-            md5($contents) .
-            '-' .
-            time() .
-            '.' .
-            $extension;
-
-        return $this->write($path, $contents);
```
DFurnes, line 66: Laravel's [`putFile`](https://laravel.com/docs/8.x/filesystem#automatic-streaming) automatically generates a unique filename so we no longer need to do this!

DFurnes: Also, since we have to write uploads & edits slightly differently (passing `UploadedFile` to `putFile`, or passing Intervention's stream to `putStream`), I inlined this logic into each method rather than abstracting in `$this->write`.

app/Services/ImageStorage.php
```diff
@@ -83,22 +72,14 @@ public function edit(Post $post, Image $image)
         }
 
         $path = $post->getMediaPath();
-        $contents = (string) $image->encode();
 
-        return $this->write($path, $contents);
-    }
+        // Write modified image to temporary file on the local filesystem. We'll store
+        // originals as full-quality JPEGs so multiple edits don't degrade quality:
+        $image->save(null, 100, 'jpg');
```
DFurnes, line 78: This is a little unclear now that I look at it, but setting the first parameter to `null` just allows us to customize quality & format while still writing to the existing filename that the image was loaded from.

database/faker/FakerPostUrl.php
```diff
@@ -2,7 +2,7 @@
 
 use App\Services\ImageStorage;
 use Faker\Provider\Base;
-use Symfony\Component\HttpFoundation\File\UploadedFile;
+use Illuminate\Http\UploadedFile;
```
DFurnes, line 5: We need this to be a _Laravel_ `UploadedFile` in order for `putFile` to work when using faked images in our test suite (since `putFile` relies on Laravel-specific methods on this class)!

app/Services/ImageStorage.php
weerd: Makes sense! Just maybe good to confirm `PostRequest` is the only avenue where all image uploads pass through before storing (mostly thinking about SMS which I'm not 100% sure of off-hand).

app/Http/Controllers/RotationController.php
```diff
@@ -57,15 +57,24 @@ public function update(Post $post, Request $request)
...
 
-        // Overwrite with edited image & purge cache:
-        $this->storage->edit($post, $rotatedImage);
-        $this->fastly->purge($post);
+            // Write edited image to S3 & purge cache:
+            $this->storage->edit($post, $rotatedImage);
+            $this->fastly->purge($post);
+        } finally {
+            // If we wrote a temporary file to disk, always clean up after
+            // ourselves so the disk doesn't get cluttered up:
```
weerd, line 73: 👏  🧼

app/Services/ImageStorage.php
weerd: I like it! Makes it much clearer 👍

app/Services/ImageStorage.php
weerd: Ya, the `null` through me off a little but got the general gist. IDE should help clarify the `save()` method arguments if hovered over the method too.

app/Services/ImageStorage.php
```diff
@@ -83,22 +72,14 @@ public function edit(Post $post, Image $image)
...
 
-    /**
-     * Write the image contents to the storage backend.
-     *
-     * @param string $extension
-     * @param string $contents
-     *
-     * @return string - URL of stored image
-     */
-    protected function write(string $path, string $contents)
```
weerd, line 99: So just confirming, no more `write()` method since essentially the `put()` performs this function?

weerd (commented):
Looks good to me! Nice work 💯

app/Services/ImageStorage.php
DFurnes: Yeah, good call! Luckily, we use Northstar's `v3/posts` API for both Phoenix and Gambit. We would defs have to think about validating file-types if we ever added other ways to submit photos (like a new API, console command, etc).

PR #1249, opened 2021-07-26
Getting Inertia-ed! 🎉 

weerd:
### What's this PR do?

This pull request begins an initial implementation of [InertiaJS](https://inertiajs.com/) for rebuilding the DoSomething administrative user interface.

It only tackles the Imports index list view for now, but as I'm tackling building out different pages, it will allow me to build a number of reusable React components that will be utilized on other pages (like the `Pagination` and `Table` components, etc).

image

I included a video to better illustrate how Inertia is helping update the page without a page refresh as I move through page links in the pagination:

https://user-images.githubusercontent.com/105849/127076789-fe2dd359-2d10-469f-bf56-d0d729f8752b.mp4

### Relevant tickets

### Checklist


composer.json
```diff
@@ -51,7 +52,8 @@
     "softonic/graphql-client": "^1.2",
     "sokil/php-isocodes": "^3.0",
     "spomky-labs/otphp": "^10.0",
-    "symfony/psr-http-message-bridge": "^1.0.0"
+    "symfony/psr-http-message-bridge": "^1.0.0",
+    "tightenco/ziggy": "^1.3"
```
weerd, line 56: Ziggy is great for providing access to route names from the backend to the frontend via a nice helpful `routes()` function.

resources/views/admin/app.blade.php
```diff
@@ -11,28 +11,15 @@
...
-        <meta name="csrf-token" content="{{ csrf_token() }}">
-    </head>
-
-    <body class="modernizr-no-js">
+        <link rel="stylesheet" href="{{ webpack_asset('inertia.css') }}">
 
-        <div class="chrome">
-            <div class="wrapper">
-                @include('admin.layouts.navigation')
+        @routes
```
weerd, line 17: This directive tells Ziggy to make the Laravel route list available to the client-side via the `route()` helper.

app/Http/Middleware/HandleInertiaRequests.php
```diff
@@ -0,0 +1,44 @@
...
+class HandleInertiaRequests extends Middleware
+{
+    /**
+     * The root template that's loaded on the first page visit.
+     *
+     * @see https://inertiajs.com/server-side-setup#root-template
+     * @var string
+     */
+    protected $rootView = 'admin/app';
+    // @TODO: maybe have both admin/app and client/app?
```
weerd, line 17: Not sure how best to handle different root views... maybe utilize a ternary based on whether intending to view admin or not? Otherwise, maybe a base root view of just `/app` and that decides which other view to load? 🤔

resources/js/Shared/components/Table/Table.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+const Table = ({ children, className }) => (
+  <table className={classnames('rounded shadow', className)}>{children}</table>
+);
```
weerd, line 7: This is kind of bare-bones for now, but aiming not to complicate it. This, the `TableHeader` and `TableBody` components are meant to be used as needed for some basic tables. Sometimes you use all three, other times (like in the `Imports/Index.js` file, you use `Table` and one of the others. In the case of `Imports/Index.js` I wanted to do some additional styling to the table body row items so instead I used HTML/JSX there for the `<tbody>` and components for `<Table>` and `<TableHeader>`. Feels like a nice yet flexible balance!

resources/js/Shared/components/Pagination/Pagination.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+import PageLink from './PageLink';
+
+const Pagination = ({ className, links }) => {
+  // If 3 links then only 1 page of results (plus previous & next links),
+  // so no need for pagination.
+  return links.length === 3 ? null : (
```
DFurnes, line 10: I'm having trouble visualizing this. Can you drop an example `links` payload in a comment here?

app/Models/ImportFile.php
```diff
@@ -59,4 +67,24 @@ public function incrementSkipCount()
...
+
+    /**
+     * Get the created at time.
+     *
+     * @return bool
+     */
+    public function getCreatedTimeAttribute()
+    {
+        return Carbon::create($this->attributes['created_at'])->format('h:i:s');
+    }
```
DFurnes, line 89: This is probably something we could derive pretty easily in the user interface using [date-fns](https://date-fns.org)!

DFurnes (approved):
Excited to see this taking shape! 🎉

app/Http/Middleware/HandleInertiaRequests.php
DFurnes: Hmm, yeah. I wonder if we'd want to have two separate middleware (one that we apply to the user-facing route group, and a separate `HandleAdminInertiaRequests` that we use for the administrative route group)?

That could also be handy if there are other defaults we'd want to apply to one but not the other.

app/Http/Controllers/Web/Admin/ImportController.php
```diff
@@ -25,10 +24,11 @@ public function __construct()
      */
     public function __invoke()
     {
-        $imports = ImportFile::paginate(15);
+        $imports = ImportFile::orderBy('created_at', 'DESC')->paginate(3);
```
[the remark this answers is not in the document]
weerd: Eeep! Yep, I switched it to make it easier for the pagination component to show up while working on this locally 👍  Good catch!

app/Models/ImportFile.php
weerd: Ya, I was debating this. I was going back and forth between just having the server add the items in the response so client doesn't have to worry about it. Initially I was going to reach for `date-fns` but didn't see it in the NS package.json, but also found it interesting to play around with appending formatted items in the JSON response via the backend. 🤷‍♂️  I can add `date-fns` and switch it in upcoming PR if that's preferred!

app/Http/Middleware/HandleInertiaRequests.php
weerd: Oh ya, I hadn't considered possibly splitting it into two middleware. That definitely seems plausible and worth testing out! 

> That could also be handy if there are other defaults we'd want to apply to one but not the other.

And yep, totally!

resources/js/Shared/components/Pagination/Pagination.js
weerd: image

I should probably update the comment to be clearer. Also maybe have a tiny helper function in the component like `hasMultiplePages()`, to make it clearer not to render pagination if there aren't multiple pages 🤷‍♂️

resources/views/admin/app.blade.php
```diff
@@ -11,28 +11,15 @@
         </title>
 
         <link rel="apple-touch-icon-precomposed" href="/assets/images/apple-touch-icon-precomposed.png">
-        <link rel="stylesheet" href="{{ webpack_asset('admin.css') }}">
 
-        <meta name="csrf-token" content="{{ csrf_token() }}">
```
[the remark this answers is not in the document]
weerd: ooopa! ya 😅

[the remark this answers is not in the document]
weerd: Also restored the `AUTH` and `ENV` items via `scriptify()`.

resources/js/Shared/components/Pagination/Pagination.js
weerd: @DFurnes I [updated the logic a bit](https://github.com/DoSomething/northstar/pull/1249/commits/9750f840814a3b024f2fc0a28953658462940211) and I think it's a bit more readable and clearer!

DFurnes: Ahh, that makes a lot more sense seeing the example payload. And nice refactoring! 🎉

app/Models/ImportFile.php
DFurnes: Yeah totes! Definitely nice to see how we'd go about doing something like this & makes sense to limit how many dependencies we add at once. Longer term, I think we might find we need more different date formats that become tedious to add one-by-one to every model (and so being able to do this in views with date-fns could be handy).

PR #1250, opened 2021-07-27
Reduce memory usage when downloading originals.

DFurnes:
### What's this PR do?

This pull request significantly reduces memory usage when downloading originals from S3 (following in the footsteps of #1246 and #1248) by directly streaming the response, rather than loading the full image into memory and re-encoding it.

The [`Storage::download`](https://laravel.com/docs/8.x/filesystem#downloading-files) method is documented here.

I've included [more detailed performance tracings](https://www.pivotaltracker.com/story/show/178903960/comments/225792271) on the ticket!

Like in the other linked PRs, our goal is to make sure we never need the full images in memory! Since S3 already has perfectly valid image files, it's enough for us to just stream that multi-part response directly to the authorized user!

### Relevant tickets

### Checklist


mendelB (approved):
👏

PR #1251, opened 2021-07-27
Inertia cleanup 🧼

weerd:
### What's this PR do?

This pull request does a bit of cleanup based on some [suggestions from prior PR](https://github.com/DoSomething/northstar/pull/1249#discussion_r677539132) (notably, removing the appended JSON data for formatted dates and switching to just using `date-fns` on the client side), along with some additional updates!

It adds a new `helpers/` directory to store helper global helper functions and split out into category files like we do in [Phoenix](https://github.com/DoSomething/phoenix-next/tree/main/resources/assets/helpers).

👀 

Code tweaks and some visual tweaks!

image

### Relevant tickets

### Checklist


resources/js/Admin/Shared/MainNav/menu.js
```diff
@@ -6,61 +6,51 @@ module.exports = {
...
       title: 'Clubs',
-      href: '#',
-      routeName: 'admin.clubs.*',
+      href: '/admin/clubs',
     },
     {
       title: 'Imports',
-      href: '#',
+      href: '/admin/imports',
       routeName: 'admin.imports.*',
```
weerd, line 21: This was fun... I realized I could resort to using the `routeName` item on this menu object to decide whether to render an `<InertiaLink />` or a regular `<a />` link that goes to non-Inertia admin page. 😀

resources/js/helpers/url.js
```diff
@@ -0,0 +1,32 @@
...
+  };
+};
+
+/**
+ * Determine if specified route name is the current active route.
+ *
+ * @param  {String} name
+ * @return {Boolean}
+ */
+export const isActiveRoute = name => (name ? route().current(name) : false);
```
DFurnes, line 22: Hmm, where is this `route` global coming from? Inertia?

Is there a way to `import` it as an ES module to make the dependency clearer?

DFurnes (approved):
Looks great!

resources/js/helpers/url.js
weerd: It's the `route()` helper from Ziggy that helps expose all the routes. I had added a eslint `global` up top to help indicate its a global helper; is that suitable?

DFurnes: Gotcha! Yeah, I saw the ESLint tag but was still wondering where exactly the global came from.

It looks like we might be able to import it as a module, which would help make this a bit clearer to anyone who's not familiar with Ziggy (but hopefully still let us use some of Ziggy's auto-configuration from that `@routes` directive):

```
import route from 'ziggy';
```

Want to give that a try? Another potential solution is to make our own `route` helper in `helpers/url.js` that references the global and explains where the function and configuration come from.

weerd: oh cool! Ya, I'll try importing first and see if that works fine. Thanks for the suggestion!

weerd: So it looks like if we want to `import` it, the setup requires adding an alias to webpack to the Ziggy `dist` but also when using `route()`, we'd need to import a `Ziggy` configuration and pass it to the function:

Example from Ziggy doc:
```
// app.js

import route from 'ziggy';
import { Ziggy } from './ziggy';

// ...

route('home', undefined, undefined, Ziggy);
```

Not keen on having to do that every time we call `route()`.... but we could combine both your suggestions and add all this to the `helpers/url.js`, wrapping it in our own helper function.

I might hold off on that for next PR, since it's more involved.

One caveat is that we can't really use the `@routes` directive. Without it we would need to use an artisan command to generate the list of routes as a config file (the aforementioned Ziggy config above) that is dumped in a directory we specify and import it and pass it to the `routes()` command:

```
php artisan ziggy:generate 
```

Anyhoo, if we go this route (pun-intended 😈 ), thoughts on a name for our wrapping helper function?

DFurnes: Hmm, yeah that's definitely a lot more involved! I do really like what the `@routes` directive does for us, even if it's a little "magical". I agree that the wrapper helper function makes the most sense, given the complications you've noted.

Since it's scoped to the module, we could probably actually just call it `route` if we wanted (and then reference `window.route` to get the Ziggy global). Not sure if that's too clever, but saves us from having to think up another name!

weerd: Ah, ya I can give that a shot :D 

Magic ✨

PR #1254, opened 2021-07-28
Fix memory issues on action's accepted_quantity include.

DFurnes:
### What's this PR do?

This pull request fixes an issue where we'd load _all_ of an action's posts into memory to sum their impact. 😱 By swapping this to a relationship method, we were able to offload this work into a `SUM()` query that the database can handle easily.

Here's some documentation on the [difference between relationship methods & properties](https://laravel.com/docs/8.x/eloquent-relationships#relationship-methods-vs-dynamic-properties).

When looking at the Papertrail logs, I noticed that requests to get these totals for actions with the most posts were running out of memory with `Allowed memory size of 268435456 bytes exhausted` errors from within `Illuminate\Database`. This was confirmed when I profiled this endpoint with [Blackfire](https://blackfire.io) (based on my learnings in [#178903960](https://www.pivotaltracker.com/story/show/178903960)! As you can see, there's _toooons_ of memory being used by `PDOStatement::fetchAll`:

Screen Shot 2021-07-28 at 2 09 15 PM

That doesn't seem right, since we should be able to just use a `SELECT SUM('quantity')` here and keep all this work in the database! It turns out, that _very sneakily_, we have to make sure to use the query builder relationship `$this->posts()` and not the relationship itself `$this->posts`. Oof!

### Relevant tickets

References [Pivotal #178663566](https://www.pivotaltracker.com/story/show/178663566).

### Checklist


weerd (approved):
Sneaky!! Excellent sleuthing 🕵️‍♂️  and sweet that you used Blackfire to help figure it out!

PR #1255, opened 2021-07-29
Reduce memory_limit for better concurrency.

DFurnes:
### What's this PR do?

This pull request re-applies the reduced memory limit from #1241.

Thanks to memory optimizations in #1246, #1248, #1250, and #1254, now _all_ of this applications' requests should fit well below this 32MB threshold so we don't need to increase this memory limit for any individual routes like we did before! 📊 

These PHP configuration settings are documented [here](https://www.php.net/manual/en/ini.core.php).

I'd originally applied this in #1241 and then rolled back in #1244 after noticing that it was causing image uploads to fail. Now that we've fixed those underlying issues, it's safe to re-apply again!

Since Heroku determines how many concurrent requests a dyno can handle by dividing total RAM by `memory_usage`, this should allow 8x more throughput per dyno and therefore a lot less requests getting queued in the router!

### Relevant tickets

### Checklist


DFurnes:
I just pulled down logs of production memory usage after deploying #1254 and we're looking really good!!

All of this application's requests over the past day use less than ~25MB (so way below this 32MB cap).

|        | min     | mean    | p90     | p95      | max       |
|--------|---------|---------|---------|----------|-----------|
| before optimizations | 2.99 MB | 4.69 MB | 5.75 MB | 22.39 MB | 242.54 MB |
| after optimizations  | 3.61 MB | 4.52 MB | 6.31 MB | 7.27 MB  | 25.08 MB  |

weerd (approved):
Such great work tackling and solving this memory issue! 👏

PR #1256, opened 2021-08-04
Correctly mute users who do not exist in Northstar, but are in Customer.io.

DFurnes:
### What's this PR do?

This pull request updates the `ImportMutePromotions` job to handle cases where a user doesn't exist in Northstar, but might still exist in Customer.io. This is an edge case that's affecting old Runscope test users from a couple years back, when we were automatically (hard) deleting these users from Northstar's database but _not_ deleting them from Customer.io.

I updated the test to reflect this new behavior in 9cad221 & adjusted the job in 58e418a.

This should clear out a large amount of failed jobs that are otherwise retrying every morning & unable to complete.

### Relevant tickets

References [Pivotal #179119168](https://www.pivotaltracker.com/story/show/179119168).

### Checklist


app/Jobs/Imports/ImportMutePromotions.php
```diff
@@ -39,13 +40,17 @@ public function __construct($record, ImportFile $importFile)
     /**
      * Execute the job to mute user promotions.
      */
-    public function handle()
+    public function handle(CustomerIo $customerIo)
     {
-        $user = User::withTrashed()->findOrFail($this->userId);
+        $user = User::find($this->userId);
```
DFurnes, line 45: I removed the `withTrashed` and `orFail` here, since we now have logic to handle scenarios where an ID does not exist in Northstar. I don't think it's particularly helpful to write `muted_promotions_at` to records that have already been soft-deleted, so we'll treat those the same way as a record that is entirely absent.

app/Jobs/Imports/ImportMutePromotions.php
```diff
@@ -54,7 +59,6 @@ public function handle()
 
         info('import.mute-promotions', [
             'user_id' => $this->userId,
-            'promotions_muted_at' => $user->promotions_muted_at->toDateTimeString(),
```
DFurnes, line 57: This line caused me some trouble since we don't always have a `$user` at this point. It also seemed somewhat redundant – we _know_ when promotions were muted based on when this log was written!

app/Jobs/Imports/ImportMutePromotions.php
```diff
@@ -39,13 +40,17 @@ public function __construct($record, ImportFile $importFile)
...
 
-        logger('Import job handling muting promotion', ['user' => $user->id]);
-
-        $user->mutePromotions();
+        if ($user) {
+            logger('Muting user', ['user' => $this->userId]);
+            $user->mutePromotions();
+        } else {
+            logger('Muting missing user', ['id' => $this->userId]);
+            $customerIo->deleteCustomer($this->userId);
```
weerd, line 52: Nice!

weerd (approved):
Looks good to me! 👍

PR #1257, opened 2021-08-04
Correctly resolve account if email provided but doesn't match.

DFurnes:
### What's this PR do?

This pull request should fix up some import jobs that are currently stuck in a failure loop.

Each of these Rock The Vote records has an email & mobile provided, but the email does not match an account in our system. Since we currently return `null` after trying to resolve an email (whether or not it exists), we immediately move on to the "create new account" flow. But had we checked mobile, we'd find that account already exists!!

I added a test case that replicates this bug (as [seen in our logs](https://my.papertrailapp.com/groups/10447062/events?q=sender%3Adosomething-northstar+program%3Aapp%2Fqueue+%22production.ERROR%22+E11000)) in e960f5a:

```
1) Tests\Jobs\Imports\ImportRockTheVoteRecordTest::testDoesNotCreateUserIfUserFoundByMobile
MongoDB\Driver\Exception\BulkWriteException: E11000 duplicate key error collection: northstar-test.users index: mobile_1 dup key: { : "+17704546680" }
```

I then applied a fix in ddb6eae to return everything to green. ✅ 

This addresses some jobs that were repeatedly failing (and retrying).

### Relevant tickets

References [Pivotal #179119168](https://www.pivotaltracker.com/story/show/179119168).

### Checklist


weerd (approved):
Nice catch! 👍

PR #1258, opened 2021-08-05
Updating Email Subscriptions Imports index with Inertia

weerd:
### What's this PR do?

This pull request exposes the `/admin/imports/email-subscriptions` route to utilize Inertia JS implementing the use of some of the components created in prior updates along with some new ones.

image

👀 

Here's a quick clip showing the page in action with the new `KeyValueCard` component in the sidebar:

https://user-images.githubusercontent.com/105849/128288107-32271b0b-0644-4afc-9abe-0a0fc4302b8a.mp4

### Relevant tickets

### Checklist


weerd:
Just to note, there may be a more ideal table column depiction, but I went with showing the columns that seemed most useful to me, and tossing rest in metadata sidebar to declutter the table. It's something that can easily be modified if staff prefer a different showcase of columns! This PR is more about getting more pages converted and UI components developed!

resources/js/Shared/components/Card/KeyValueCard.js
```diff
@@ -0,0 +1,51 @@
+import React, { Fragment } from 'react';
+import PropTypes from 'prop-types';
+import { startCase } from 'lodash';
+import classnames from 'classnames';
+
+const KeyValueCard = ({ className, data, placeholder, style, title }) => (
```
weerd, line 6: Does `KeyValueCard` make sense for the name of the component? Alternatives could be `DefinitionListCard` or `DescriptionListCard`... thoughts? 🤷‍♂️

app/Models/ImportFile.php
```diff
@@ -29,6 +29,13 @@ class ImportFile extends Model
         'user_id',
     ];
 
+    /**
+     * The accessors to append to the model's array form.
+     *
+     * @var array
+     */
+    protected $appends = ['metadata'];
```
weerd, line 37: I'm appending this `metadata` attribute which is based on the JSON string from the `options` attribute. It helps so that we don't need to decode to JSON on the frontend. Also, renaming it to metadata made more sense since the data doesn't really have anything to do with "options", and kept sounding confusing to me! 😅

resources/js/helpers/url.js
```diff
@@ -30,3 +19,21 @@ export const isActiveRoute = name => (name ? route().current(name) : false);
 export const findUrlSegmentByKey = key => {
   return get(urlSegmentMap(), key, key);
 };
+
+/**
+ * Return the Ziggy global route helper.
+ *
+ * @returns window.route()
+ */
+export const route = () => window.route();
```
weerd, line 28: Addressing [comment](https://github.com/DoSomething/northstar/pull/1251#discussion_r678339863) from prior PR!

resources/js/Shared/components/Card/KeyValueCard.js
```diff
@@ -0,0 +1,51 @@
...
+    <h1 className="border-b border-white p-4 text-gray-500 text-sm uppercase">
+      {title}
+    </h1>
+
+    <div className="px-4 py-8">
+      {data ? (
+        <dl>
+          {Object.keys(data).map((key, index) => {
+            return (
+              <Fragment key={index}>
```
weerd, line 17: Something new I learned, you can't add a `key` to a React fragment if using the shorthand `<>`... need to use the component!

resources/js/Admin/Imports/EmailSubscriptions/Index.js
```diff
@@ -0,0 +1,89 @@
...
+import Layout from '../../Shared/Layout';
+import Table from '../../../Shared/components/Table/Table';
+import TableHeader from '../../../Shared/components/Table/TableHeader';
+import KeyValueCard from '../../../Shared/components/Card/KeyValueCard';
+import Pagination from '../../../Shared/components/Pagination/Pagination';
+
+const Index = ({ imports }) => {
+  const { data, links } = imports;
+
+  const [rowMetadata, setRowMetadata] = useState(null);
```
DFurnes, line 14: Instead of saving the metadata itself into state here, what if we saved the ID? This would allow us to, for both the table & the sidebar, read the data directly from the "source of truth" props provided to the component. It would also mean that if we ever needed things other than metadata in this column, they'd be easy to access!

```js
const [selectedRow, setSelectedRow] = useState(null);
```

```jsx
onMouseEnter={() => setSelectedRow(id)}
```

```jsx
  <KeyValueCard
           className="sticky"
           data={data.find(item => item.id == selectedRow).metadata}
```

DFurnes (approved):
Looks fantastic!!! 🤩 I love the screencast demo in the PR.

resources/js/Shared/components/Card/KeyValueCard.js
DFurnes: 💡

app/Models/ImportFile.php
DFurnes: Agreed that this is more clear! Might be nice to rename the column in a future cleanup sprint.

resources/js/Shared/components/Card/KeyValueCard.js
DFurnes: Naming! 🙈 I think I'd lean towards `DescriptionListCard` since it makes the "intent" of the component (describing something) clearer, whereas "key value" and "definition list" both feel a bit more abstract.

app/Models/ImportFile.php
weerd: Yep yep! Just figured renaming it was a bit out of scope for now :)

resources/js/Admin/Imports/EmailSubscriptions/Index.js
weerd: Ah! Nice, I dig that 💡  Great suggestion @DFurnes 🙌  Will update!

app/Models/ImportFile.php
DFurnes: Yeah, definitely! I think that was the right call. 👍

resources/js/Shared/components/Card/KeyValueCard.js
weerd: Haha! Ya, I was kinda leaning towards `DescriptionListCard` as well, but realized that name after building it all 😂 

I came upon that name when reviewing the React docs to figure out how they recommended adding a `key` to a fragment: 
image

haha!

resources/js/Admin/Imports/EmailSubscriptions/Index.js
weerd: @DFurnes I ended up modifying the approach you suggested since the `selectedRow` on initial load is the default of `null` so when it tries to find an item with `null` as the ID it results in `undefined` and it errors when trying to get the `metadata` of `undefined`.

So I separated it out into a variable `rowMetadata` and then use `get()` to grab it via the prop set in the component.

Alternative, which maybe felt like overkill, is to set the default value for the state to be the first item's metadata 🤷‍♂️ , but figured the approach I took was less brittle.

DFurnes: Makes sense! 👍

PR #1259, opened 2021-08-05
Reducing payloads with resources

weerd:
### What's this PR do?

This pull request takes advantage of Laravel API resources to help reduce the payload size of the JSON that is provided to the frontend by Inertia. Resources will allow us to define what model data we want to send over so we can only send what the UI actually needs.

👀 

I wanted to come back to address reducing the Inertia JSON payload size, because I imagine for some other pages, the payload could end up being pretty hefty. So I wanted to try out Laravel API resource to help address this early on, so it would be something we utilize moving forward!

### Relevant tickets

### Checklist


resources/js/Admin/Imports/EmailSubscriptions/Index.js
```diff
@@ -9,7 +9,7 @@ import Pagination from '../../../Shared/components/Pagination/Pagination';
 import DescriptionListCard from '../../../Shared/components/Card/DescriptionListCard';
 
 const Index = ({ imports }) => {
-  const { data, links } = imports;
+  const { data, meta } = imports;
```
weerd, line 12: With resource collections, the `links` property is now an object with 4 links `prev`, `next`, `first`, `last`, and the full meta with links and pagination items is now in `meta.links`. Just noting why it changed!

app/Http/Resources/ImportFileResource.php
```diff
@@ -0,0 +1,29 @@
...
+     * @param  \Illuminate\Http\Request  $request
+     * @return array
+     */
+    public function toArray($request)
+    {
+        return [
+            'id' => $this->id,
+            'import_count' => $this->import_count,
+            'import_type' => $this->import_type,
+            'metadata' => $this->metadata,
```
weerd, line 21: @DFurnes so this could be an alternative to handling the `$options` in the `ImportFile` model, and appending a new `metadata` attribute in that class. So I think it could be possible to do:

```php
...
'metadata' => json_decode($this->options),
...
```
in this resource file, and remove the appended property that adds it to the model. Thoughts?

weerd: I think either works, but might be nice to do the manipulation in this resource file so it's all contained and understood it's explicitly for the JSON response. 🤔

DFurnes: Makes sense, and mirrors how we've handled transformation for other fields (in Fractal transformers) in the past!

weerd: Ya, definitely! Ok, I'm gonna try changing it to what I have in the example :) Thanks for taking a peek!

DFurnes:
This looks very similar to what we accomplish for other routes using Fractal transformers? Do you think this would be a new approach that would replace those eventually? What benefits (and potential downsides) are there from switching?

resources/js/Admin/Imports/EmailSubscriptions/Index.js
DFurnes: Got it! Is this something that we could customize on the `JsonResponse` if we wanted to keep the same structure?

weerd: I believe so. One approach is to specify a [resource collection class](https://laravel.com/docs/8.x/eloquent-resources#writing-resource-collections) for the same model and you can customize the metadata.

I will say, I dig the pagination structure from the resource class a lot better than the default JSON response version. Nicer to have all the pagination properties in compartmentalized properties:

image

weerd:
> This looks very similar to what we accomplish for other routes using Fractal transformers. Do you think this would be a new approach that would replace those eventually? What benefits (and potential downsides) are there from switching?

I do think it could be a new approach that would eventually replace those, and figured trying it out on data that isn't available via the NS API would be alright.

It feels like a nice benefit that API resources are natively supported out of the box by Laravel, and generally less setup/pieces to need to manually define like we do in Fractal with `items` and `collections` there. I think the downside would be that API Resources may not be as customizable as Fractal Transformers, but some might debate that's a good thing, haha!

Given that API Resources are natively supported, there's no additional package to install and if we decide to switch this to using Fractal, we totally could and I think the lift would be minimal. Just curious to see how well API Resources could work for our needs and simplify our setup!

weerd:
This PR is also evaluating the use of Eloquent API Resources for how we transform our model data into JSON responses. We want to determine and discuss in an [upcoming RFC](https://www.pivotaltracker.com/story/show/179136779), whether we want to switch from our existing approach using Fractal, to using Laravel's native approach that was introduced in Laravel v5.5.

resources/js/Admin/Imports/EmailSubscriptions/Index.js
DFurnes: Ok cool! Yeah, new structure looks good but wouldn't be looking forward to having to refactor everything to use it if we were to switch from Fractal to resource collections down the road.

DFurnes (approved):

resources/js/Admin/Imports/EmailSubscriptions/Index.js
weerd: Haha, ya. Might be something specifically for Admin resources utilizing Inertia and not ones for the NS API. We can see. Worth exploring!

PR #1260, opened 2021-08-06
Upgrading Mute Promotions & Rock The Vote index pages to Inertia.

weerd:
### What's this PR do?

This pull request continues moving more Admin import-related pages to utilizing InertiaJS. Specifically, it switches the Mute Promotions imports index page and Rock The Vote imports index page to use the new UI.

It also makes a couple updates to components and fixes a bug when trying to run RTV test import.

image

👀 

I added a `+ New Import File` on each index page so it makes it easier to get to the old create interface for each respective import type:

https://user-images.githubusercontent.com/105849/128557673-54bc4bad-be17-4da6-9e67-b334ae46ed8f.mp4

### Relevant tickets

### Checklist


app/Managers/PostManager.php
```diff
@@ -51,21 +51,23 @@ public function create($data, $signupId, $shouldSendToCustomerIo = true)
...
-            optional(User::find($post->referrer_user_id), function (
-                $referrerUser
-            ) use ($post) {
-                CreateCustomerIoEvent::dispatch(
-                    $referrerUser,
-                    'referral_post_created',
-                    $post->getReferralPostEventPayload(),
-                );
-            });
+        if (config('features.customer_io')) {
```
weerd, line 54: @DFurnes this should be cool right? I wrapped these in conditionals based on the feature flag so that when running RTV import tests, they don't try to send out requests to CustomerIO.

Same same in the `SignupManager.php`

DFurnes: Recapping a Slack conversation – it looks like these these eventually end up kicking off [CreateCustomerIoEvent](https://github.com/DoSomething/northstar/blob/main/app/Jobs/CreateCustomerIoEvent.php) (and thus calling [CustomerIo@trackEvent](https://github.com/DoSomething/northstar/blob/c636d04d51a1fdd820e731b2047b7fade3180eae/app/Services/CustomerIo.php#L59-L77) which _should_ get properly feature-flagged by this same config value. I'd love if we could sort out how this is sneaking through so that we don't have other cases where this feature flag is letting us down.

weerd: Will investigate and document in [Pivotal Ticket #179149134](https://www.pivotaltracker.com/story/show/179149134).

weerd:
In the future, it'd be nice to take a bunch of the button component-izing work done in Phoenix, and bring it in here (or even better, make Forge a component library that the two repos share 🔮 )

resources/js/Admin/Shared/MainNav/menu.js
```diff
@@ -53,5 +55,9 @@ module.exports = {
       title: 'OAuth Clients',
       href: '/admin/clients',
     },
+    {
+      title: 'Documentation',
+      href: '/dev',
+    },
```
weerd, line 61: Added a quick link to the Dev Documentation 🕺  So much easier to expand the menu when it's vertical!

resources/js/Shared/components/Card/DescriptionListCard.js
```diff
@@ -48,7 +48,7 @@ DescriptionListCard.propTypes = {
 
 DescriptionListCard.defaultProps = {
   className: null,
-  data: {},
+  data: null,
```
weerd, line 51: Ensures placeholder is rendered on initial load.

resources/views/admin/imports/rock-the-vote/create.blade.php
```diff
@@ -8,7 +8,9 @@
 
 <div style="display: block; margin-bottom: 15px; margin-top: 15px; overflow: auto;">
 
-    <a href="/admin/imports/rock-the-vote/test" class="pull-left">Test Import</a>
+    @if(config('import.import_test_form_enabled'))
+        <a href="/admin/imports/rock-the-vote-test" class="pull-left">Test Import</a>
```
weerd, line 12: This link was broken and now it's fixed!

app/Http/Controllers/Web/Admin/RockTheVoteTestController.php
```diff
@@ -89,6 +90,13 @@ public function store(Request $request)
             'import_type' => ImportType::$rockTheVote,
         ]);
 
+        // Test import needs an existing action to work with.
+        $action = factory(Action::class)
+            ->states('voter-registration')
+            ->create();
+
+        config(['import.rock_the_vote.post.action_id' => $action->id]);
```
DFurnes, line 98: I think we'd, once upon a time, used this form to test imports on dev/QA. Do you think that'd still be useful? If so, is there a way we could create this action in our seeders so it doesn't have to make it on-demand? If not, should we remove the "Action ID" field from the form, since that action won't actually be used here?

DFurnes (approved):
Just one question!

app/Http/Controllers/Web/Admin/RockTheVoteTestController.php
weerd: I've found it useful for generating records while working on the new UI or when I was merging Chompy into Northstar, but admittedly have not used it on Dev/QA. We could just remove all this and create a legitimate seeder to seed RTV import examples, not just an Action ID and maybe that would be much more ideal? Thoughts?

weerd: Ok! I went with the basic seeder approach to ensure local/dev/QA have an action record with ID specified from the import config settings. This should be helpful for now, but will also create a pivotal tracker issue to discuss whether we still need this RTV test import via UI functionality!

weerd: ^ [Pivotal ticket](https://www.pivotaltracker.com/story/show/179186867) related to above.

PR #1263, opened 2021-08-13
Command to remove 'news' email subscription topic from users and replace with 'lifestyle'

katiecrane:
### What's this PR do?

This pull request is part of the work to retire the Breakdown and re-name the Boost. For all users currently subscribed to the Breakdown, we want to remove that subscription topic and add one for the boost instead.

I haven't written one of these in Northstar in a while, any best practices I missed? How is my logic?

All above and in ticket!

### Relevant tickets

References [Pivotal #178968220](https://www.pivotaltracker.com/story/show/178968220).

### Checklist


app/Console/Commands/LifestyleNewsSubscriptionSwap.php
```diff
@@ -0,0 +1,96 @@
...
+use Illuminate\Support\Collection;
+
+class LifestyleNewsSubscriptionSwap extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'northstar:boostdown';
```
weerd, line 16: LOL, "boostdown"

app/Console/Commands/LifestyleNewsSubscriptionSwap.php
```diff
@@ -0,0 +1,96 @@
...
+    public function handle()
+    {
+        // Use low priority queue for these updates
+        config(['queue.jobs.users' => 'low']);
+
+        // Grab users who have 'news' in their email_subscription_topics
+        $query = (new User())->newQuery();
+        $query = $query->whereIn('email_subscription_topics', ['news']);
+
+        $totalCount = $query->count();
```
weerd, line 56: Is this a different syntax? Just a little thrown off by it, and others might be as well when reviewing at a later date 😅 

I might be misremembering, but I believe if you don't call `get()` or `count()` at the end of the query, it returns a query object, so then you can reuse for the `count()` and the `chunk()` afterwards:

```php
$query = User::whereIn('email_subscription_topics', ['news']);

$totalCount = $query->count();

// ...

$query->chunkById(...
```

But I could be mistaken! Worth a try 😄

app/Console/Commands/LifestyleNewsSubscriptionSwap.php
```diff
@@ -0,0 +1,96 @@
...
+                $user->save();
+
+                $this->line(
+                    'northstar:boostdown - User ' .
+                        $user->id .
+                        ' replaced news with lifestyle',
+                );
+            });
+
+            // Logging to track progress (may read over 100% at the end when there are less than 200 users in the last chunk)
```
weerd, line 78: Not a big deal but not sure I get why this would be if it's counting up to match the total?

weerd (commented):
Looks good to me! 🎉  Just a couple questions. Also, very much appreciated the tests! 🚥

app/Console/Commands/LifestyleNewsSubscriptionSwap.php
katiecrane: @weerd ah yeah - it adds 200 to the current total at the end of each chunk rather than checking how many users were in the chunk (since we expect it to be 200 for all but the final one) and the final chunk may have less than 200 users

app/Models/User.php
```diff
@@ -1043,6 +1043,21 @@ public function addEmailSubscriptionTopic($topic)
         );
     }
 
+    /**
+     * Remove the given topic from the user's array of topics if it is there.
+     *
+     * @param string $topic
+     * @deprecated
```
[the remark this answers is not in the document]
katiecrane: Whoops that is a copy paste mistake! good catch!

app/Console/Commands/LifestyleNewsSubscriptionSwap.php
weerd: Oh right, duh... since adding hardcoded value of `200`. So I take it there isn't a way to get the chunk size then to go off of that?

katiecrane: I forget why exactly we started doing it that way but I think it was to avoid extra computations since we've had memory issues with scripts like this before and for the progress stuff we really only care about how far we are until we reach the end and at the end we just want to know that we're done if that makes sense? I could move up the `currentCount` counter into the above block though and increment for each user if that'd make things clearer though!

weerd: Nah, I think not a big deal since in a script anyhoo. I was just curious... was pondering if maybe the chunk would have a property or something on it that would allow it to know it's own size that was retrieved or something, but not crucial!

app/Console/Commands/LifestyleNewsSubscriptionSwap.php
katiecrane: This is a great question too and I don't remember why we ever had it written this way but I see it in a few other commands too - I'm onboard to switch it to the more usual way you have written here and see what happens! My only guess is we tried this before for memory reasons?

app/Console/Commands/LifestyleNewsSubscriptionSwap.php
katiecrane: Ah yeah, that'd be cool! Will leave this as is for now given context and our discussion!

app/Console/Commands/LifestyleNewsSubscriptionSwap.php
weerd: Oh hmmm... it may have been due to not knowing about how the Eloquent query would return a query object if not finalized with a `get()` type call (or other call that ends the query and executes the database search).

But ya, give it a whirl and see if it works!

app/Models/User.php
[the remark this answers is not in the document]
katiecrane: Hmm the one I copied it from isn't actually deprecated either as far as I can tell so I'm removing it from both!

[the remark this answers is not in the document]
katiecrane: I dug into it and you're actually the person to talk to about it haha - the other one was marked as deprecated in favor of the new on you made that accepts an array of one or more topics! I guess I should add that back in!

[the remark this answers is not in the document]
katiecrane: I'm updating the spots I used the old one to use the new one instead!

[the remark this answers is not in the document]
katiecrane: All updated!

weerd (approved):
🎉

PR #1264, opened 2021-08-16
Adding Show pages for some import types.

weerd:
### What's this PR do?

This pull request adds a "show" page for individual Email Subscription and Mute Promotion import files:

image

It also does a little bit of cleanup on some styles, some directory naming and updates the `ImportFileFactory` to add in fake `options` data into the seeded records so they more closely match what a real record looks like and makes it easier to work on development of the new Admin UI.

👀 

...

### Relevant tickets

### Checklist


resources/js/Admin/Imports/RockTheVote/Index.js
```diff
@@ -2,7 +2,7 @@ import React, { useState } from 'react';
 import { get } from 'lodash';
 import { format, parseISO } from 'date-fns';
 
-import Layout from '../../Shared/Layout';
+import Layout from '../../shared/Layout';
```
weerd, line 5: Renamed the directory to lowercase since it shouldn't be capitalized.

resources/js/Admin/shared/MainNav/menu.js
```diff
@@ -3,14 +3,14 @@
...
-    {
-      title: 'FAQ',
-      href: '/admin/faq',
-    },
     {
       title: 'Campaigns',
       href: '/admin/campaigns',
     },
+    {
+      title: 'Members',
```
DFurnes, line 11: How do we define members nowadays? Historically, I think we had folks who were _users_ (had an account) but not _members_ (who had to be addressable via email or SMS). Is that still a distinction, and if so would we want to keep "Users" as the label here?

DFurnes (approved):
Looks good! One question on naming.

resources/js/Admin/shared/MainNav/menu.js
weerd: 🤔  not 100% sure. This was a suggestion from Leah after showing a bit of the new Admin UI, but maybe we dig into it a bit more before officially changing. I hand't fully considered the differences in the terms. Good call out!

resources/js/Admin/Imports/RockTheVote/Index.js
DFurnes: What's the style rule on when these directions should & shouldn't be capitalized?

weerd:
@DFurnes made a quick update in [this commit](https://github.com/DoSomething/northstar/pull/1264/commits/b89649c3d658d4e597253b188845b9e6ba2fd4d2) to add a new `FakerImportOptions` class to handle what I previously had as a function in the `ImportFilesFactory`. I believe since I had defined it as a standalone function there, every time the factory ran it was recalling it and throwing an error since the function already existed. So I popped it out into it's own class, like some of those other faker providers we include :)

resources/js/Admin/Imports/RockTheVote/Index.js
weerd: I think essentially capitalizing the Inertia related page directories/components that match the routing structure like `Admin/Imports/Show.js`, but for shared components and other partials, utilities, etc, following what we do in Phoenix and Northstar with container directories lowercased and component directory/files capitalized.

I'm planning a bit of a cleanup PR to fix up a few things with the directories, etc and once settled I can make sure to add to the docs regarding the directory structure for the React components and any naming rules if that sounds good?

PR #1266, opened 2021-08-17
Removing RTV test env var, config and routes.

weerd:
### What's this PR do?

This pull request removes the Rock The Vote test import Admin UI related code.

It didn't seem like we were really using this much, and not worth supporting this code if it is not utilized. There are also probably better alternatives to test how RTV records files are imported.

👀 

We discussed this a bit in the linked Pivotal Tracker ticket below to confirm we were ok to remove this code.

### Relevant tickets

References [Pivotal #179186867](https://www.pivotaltracker.com/story/show/179186867).

### Checklist


weerd:
I believe this covers all code items related to the RTV test import.

katiecrane (approved):
Looks good! I agree with the discussion in the ticket that we will likely want something like this when we get back to VR (I believe Aaron built it out of necessity for testing) but we can always bring it back if that's the case!

PR #1270, opened 2021-08-28
Admin RTV import show view ✨

weerd:
### What's this PR do?

This pull request updates the RTV Imports individual show page to use InertiaJS and React for the Admin UI.

https://user-images.githubusercontent.com/105849/131199706-117740a7-8306-4bc0-94ac-7b62ca8a93ee.mp4

👀 

Here's a more helpful commit specific list:

- Fixing routing and front-end nav issues: [2fffeec](https://github.com/DoSomething/northstar/pull/1270/commits/2fffeecc3464ecb08b0f4e71d5f647fb002df38e) & [56a39ad](https://github.com/DoSomething/northstar/pull/1270/commits/56a39adeda32528ad7d339ea7c6251a8231a57c8)
- Updating RTV Controller for Inertia along with resource for data sent to frontend: [511b349](https://github.com/DoSomething/northstar/pull/1270/commits/511b3491acfcfaa7376c0c6503c7314225123146)
- Fixing date issue: [d06a973](https://github.com/DoSomething/northstar/pull/1270/commits/d06a97303011ee879a09ae864ca4a6b849650965)
- Adding RTV import individual Show view: [628515](https://github.com/DoSomething/northstar/pull/1270/commits/6285155662ed23b651a5de4827d1743debae9916)

The rest are just some quick style updates and cleanup!

### Relevant tickets

References [Pivotal #178699626](https://www.pivotaltracker.com/story/show/178699626).

### Checklist


app/Http/Resources/ImportFileResource.php
```diff
@@ -15,14 +16,16 @@ class ImportFileResource extends JsonResource
...
     {
         return [
+            'created_at' => $this->created_at,
             'id' => $this->id,
             'import_count' => $this->import_count,
             'import_type' => $this->import_type,
-            'metadata' => json_decode($this->options),
+            'metadata' => $this->options
+                ? json_decode($this->options)
+                : new stdClass(),
```
weerd, line 25: If this ternary defaults to an empty array if `$this->options` is `false`, the empty array on the frontend is converted to an empty array, but to help make the JS cleaner and allow adding in `metadata` properties in JS, it's better for this to default to an empty `{}`. But to do this, we need to use `new stdClass()` for it to be converted to an empty `{}`.

I had this comment in the code:
```
// stdClass forces conversion to empty object in JS instead of array.
```
Not sure if best to leave a comment in code or just mention here (or maybe a comment in the docblock?)... This also applies to the `RockTheVoteLogResource.php` which also has a `metadata` item.

app/Http/Resources/RockTheVoteLogResource.php
```diff
@@ -0,0 +1,46 @@
...
+     * Format the tracking source data into an associative array.
+     *
+     * @return array|null
+     */
+    public function formatMetadata()
+    {
+        $stringsArray = explode(',', $this->tracking_source);
+
+        return strings_array_to_assoc_array($stringsArray, ':', 'source');
+    }
```
weerd, line 45: Needed to add this because the `tracking_source` data is not a JSON encoded string, but a string segment from data in a URL that is separated by `,` and each key/value pair is separated by `:`.

app/Models/RockTheVoteLog.php
```diff
@@ -105,4 +106,15 @@ public static function hasUpdatedSmsSubscription(
...
+    /**
+     * Get the started registration date.
+     *
+     * @param  string  $value
+     * @return string
+     */
+    public function getStartedRegistrationAttribute($value)
+    {
+        return Carbon::parse($value)->toDateTimeString();
+    }
```
weerd, line 119: The way the date for started registration is saved in the database is not a format that `date-fns` could deal with properly to parse it, so we convert it to a datetime string on retrieval so `date-fns` can format it to its heart's desire 💓 .

resources/js/Admin/Imports/EmailSubscriptions/Index.js
```diff
@@ -44,9 +44,7 @@ const Index = ({ imports }) => {
               user_id,
             } = dataItem;
 
-            if (metadata) {
-              metadata['created_by'] = user_id || '🤖 Console Robot';
-            }
+            metadata.created_by = user_id || '🤖 Console Robot';
```
weerd, line 47: With `metadata` always coming is as an object (empty or with data) from the backend, we can clean this up and just add properties to it.

database/migrations/2021_04_15_000002_create_rock_the_vote_logs_table.php
```diff
@@ -20,7 +20,7 @@ public function up()
             $table->string('tracking_source');
             $table->string('status');
             $table->string('pre_registered');
-            $table->string('started_registration');
+            $table->string('started_registration'); // @TODO: should this have been a datetime?
```
weerd, line 23: Related to [this comment](https://github.com/DoSomething/northstar/pull/1270/files#r698660430).

app/Http/Resources/ImportFileResource.php
DFurnes: Makes sense! What if we used `(object) []` (which does the same thing, AFAIK, but reads closer to the JS intent).

app/helpers.php
```diff
@@ -902,3 +902,38 @@ function webpack_asset($file, $buildDirectory = 'dist')
...
+    $defaultKey = 'key'
+) {
+    if (!is_array($strings)) {
+        return null;
+    }
+
+    // If single item, then use default key for associative array.
+    if (count($strings) === 1) {
+        return [$defaultKey => $strings[0]];
+    }
```
DFurnes, line 927: What's an example of when this conditional would run?

DFurnes (approved):
Looks great! 🚀

app/Http/Resources/ImportFileResource.php
weerd: Ah! I did not try that, but seems like a great idea and would be more straightforward indicating intent! Will give it a go 🐎

app/helpers.php
weerd: I confirmed we have entries on QA where the `tracking_source` is just `ads` or `ds-wl-verify`... with no `:` or anything else which is why I ended up having to account for it 🤷‍♂️

PR #1272, opened 2021-09-08
Admin Import Mute Promotions Create

weerd:
### What's this PR do?

This pull request adds the Mute Promotions import create page to initiate adding the list of users who have requested to mute their promotional messaging, queueing it up to be imported and processed by the system.

image

👀 

Here's a clip of the page functionality in action!

https://user-images.githubusercontent.com/105849/132598034-8b1cd21e-f1f0-490d-a403-83af9747059f.mp4

This PR also does a bit of heading style updates across some prior Inertia pages to bring them all in line with current design.

### Relevant tickets

References [Pivotal #179271621](https://www.pivotaltracker.com/story/show/179271621).

### Checklist


app/Http/Controllers/Web/Admin/MutePromotionImportController.php
```diff
@@ -58,11 +58,7 @@ public function index()
      */
     public function create()
     {
-        $config = ImportType::getConfig($this->importType);
```
weerd, line 61: I got rid of this because we didn't really need it since no config options for mute promotions to present on create form.

resources/js/Admin/Imports/RockTheVote/Create.js
```diff
@@ -29,10 +29,10 @@ const Create = ({ importConfig }) => {
     },
   );
 
-  function handleChange(e) {
-    const key = e.target.id;
+  function handleChange(event) {
```
weerd, line 32: I know it's more verbose but I like to be more explicit since `e` is sometimes also used for `error`... 🤷‍♂️

resources/js/Admin/Imports/RockTheVote/Create.js
```diff
@@ -79,7 +79,7 @@ const Create = ({ importConfig }) => {
           Create Import
         </h2>
 
-        <div className="col-span-full mb-10">
+        <div className="col-span-full mb-10" style={{ maxWidth: '75ch' }}>
```
weerd, line 82: This was a fun way to help control readable line-length!

resources/js/Admin/Imports/MutePromotions/Create.js
```diff
@@ -0,0 +1,170 @@
...
+
+    post(route('admin.imports.mute-promotions.store'), {
+      onSuccess: () => reset(),
+    });
+  }
+
+  return (
+    <div className="base-12-grid mx-auto pt-8 pb-12 relative">
+      {/* @TODO: Extract FlashMessage as a component! */}
+      {flash.info && withFlashMessage && (
```
katiecrane, line 55: Not sure I've seen this pattern before - how does this second `&&` here work?

resources/js/Admin/Imports/RockTheVote/Create.js
katiecrane: Ooooo this is cool!

resources/js/Admin/Imports/MutePromotions/Create.js
weerd: Ya, this is an approach I saw in the InertiaJS docs and dig it. Essentially, it's kind of a shortcut from using a ternary and needing to output `null` if the condition doesn't pass. Feels a little cleaner when it's a simple check to output an element.

In this particular case it's using a double check, but the last check is always going to pass if the prior ones did and thus it outputs the markup.

katiecrane: OOOO because the markup evaluates to `true` yeah? This is definitely cleaner than having to have the `else null` situation!

katiecrane (approved):
Looks awesome! The video was super helpful and I learned a new cool shortcut!

PR #1276, opened 2021-09-13
Calculate campaign pending/accepted counts more reliably.

DFurnes:
### What's this PR do?

This pull request updates Northstar to increment/decrement the campaign's counters whenever a post's status is updated, rather than running a massive `COUNT(*)` query to tally these up from scratch. This should fix up issues where the old approach (which only ran after reviews & on a cron timer) would end up with inconsistent counts.

I wrote up some better documentation on when each event in the `PostObserver` runs in 77dc3ca, and then added the logic to increment/decrement these counters and a test demonstrating the new behavior in fdc3306.

As you can imagine, it's way easier to simply increment or decrement a number rather than try to tally up all the posts from scratch! This allows us to make these updates instantly whenever a change is made, rather than queueing them up for later.

### Relevant tickets

References [Pivotal #170111820](https://www.pivotaltracker.com/story/show/170111820).

### Checklist


app/Observers/PostObserver.php
```diff
@@ -49,41 +51,58 @@ public function creating(Post $post)
...
+     * Handle the model "created" event. This fires after
+     * the record is created for the first time.
      *
      * @param  \App\Models\Post  $post
      * @return void
      */
     public function created(Post $post)
     {
-        $post->updateOrCreateActionStats();
-
```
DFurnes, line 60: I moved this down into the `saved` event (which fires on both creation & updates) to de-dupe.

app/Observers/PostObserver.php
```diff
@@ -49,41 +51,58 @@ public function creating(Post $post)
...
      */
     public function deleting(Post $post)
     {
+        // Delete associated files & cached responses, and anonymize
+        // this database record before we soft-delete it:
         $this->storage->delete($post);
         $this->fastly->purge($post);
 
         $post->update([
+            'status' => 'rejected', // Ensure we've removed this from pending/accepted counts.
```
DFurnes, line 105: Rather than have a separate check to decrement counters in the delete listener, I figured it was simpler to mark any deleted posts as rejected & therefore we'd get to rely upon the logic in `saved` above for this case as well.

app/Observers/PostObserver.php
```diff
@@ -49,41 +51,58 @@ public function creating(Post $post)
...
+    public function saved(Post $post)
     {
         $post->updateOrCreateActionStats();
+
+        // If we've changed the status of this post, increment or decrement
+        // the appropriate counts on the associated campaign model:
+        $originalStatus = $post->getOriginal('status');
+        if ($post->campaign && $post->status !== $originalStatus) {
+            if (in_array($post->status, ['accepted', 'pending'])) {
+                $post->campaign->increment($post->status . '_count');
```
DFurnes, line 81: This makes a SQL query that looks like `SET accepted_status = accepted_status + 1`, which means that we don't need to worry about race conditions if we create/approve/reject multiple posts rapid-fire. (If we'd done `$campaign->accepted_status += 1` in code, we'd potentially run into race conditions if multiple requests were handled at once!)

app/Observers/PostObserver.php
mendelB: 💡 Smart!

mendelB (approved):
This is awesome, thanks for the helpful comments!

PR #1277, opened 2021-09-15
Add redirect URLs to seeder.

DFurnes:
### What's this PR do?

This pull request fixes an issue where the seeded `dev-oauth` client would give an "Invalid Client" error because it didn't have any redirect URIs set. We had previously used this to allow _any_ redirect URI to be provided for easier local development, but that now seems to (somewhat reasonably) be disallowed by the package.

Any other URIs we might commonly use for local development? We've got a lot less of these to think about now that we merged in a lot of our previously decoupled administrative frontends into Northstar!

This just fixes a little issue I ran into when testing my other work.

### Relevant tickets

References [Pivotal #179594139](https://www.pivotaltracker.com/story/show/179594139).

### Checklist


katiecrane (approved):
ah makes sense to me and no other ones off the top of my head!

PR #1278, opened 2021-09-15
Fixes issue where we would include soft-deleted posts in aggregate counts.

DFurnes:
### What's this PR do?

This pull request fixes an issue where the `rogue:recount` command would include soft-deleted posts in the aggregate post totals, which could lead to "ghost" counts that would appear on the inbox but not when clicking into a campaign.

I checked this by adding a `->toSql()` on this query builder before the change:

```sql
select campaign_id, count(*) as count from `posts` where `type` in (?, ?) and `status` = ? group by `campaign_id`
```

…and after:

```sql
select campaign_id, count(*) as count from `posts` where `type` in (?, ?) and `status` = ? and `deleted_at` is null group by `campaign_id`
```

This command is necessary to "initialize" our counter caches based on the current number of rows in the database. Once the counts are accurate, we should be able to rely on our model observers (added in #1276) to increment/decrement these correctly and efficiently whenever we create/review/delete posts.

### Relevant tickets

References [Pivotal #170111820](https://www.pivotaltracker.com/story/show/170111820).

### Checklist


mendelB (approved):
🦅 👁️

PR #1271, opened 2021-09-02
Admin import RTV create

weerd:
### What's this PR do?

This pull request adds the RTV import create page to initiate generating a report from RTV that will later be queued for importing into the system.

image

### How should this be reviewed?

👀 

Here's a clip of the page functionality in action!

https://user-images.githubusercontent.com/105849/132067536-06cfb73d-3093-4175-9cd1-b75095cbbac7.mp4

### Any background context you want to provide?

...

### Relevant tickets

References [Pivotal #]().

### Checklist

- [x] This PR has been added to the relevant Pivotal card.

PR #1279, opened 2021-09-15
Form components and Email Subscriptions Import Create Admin UI ✨

weerd:
### What's this PR do?

This pull request updates the Admin UI to add an Email Subscription Create page. But _more importantly_, it begins the process of breaking out form fields/inputs into reusable and functional components that will be able to be used throughout the Admin UI to speed up building interfaces! ⚡ 

image

Here's a quick clip showcasing the new Email Subscriptions create page and its functionality:

https://user-images.githubusercontent.com/105849/134060547-8cdde98b-d687-4382-8ba9-a9a8a87afda9.mp4

This create page included similar fields/inputs as other forms, so it made sense to begin considering how to componentize the form elements. Since the application is technically divided between a client app that our members see (login/profile) and an admin app that our staff sees, it made sense to consider how to componentize form elements in a way that the core functionality of the inputs could be reused on both client & admin interfaces, but the styling/visuals could be customized differently for the needs of the client or admin interface.

The solution I came up with was to break form "inputs" into components shared by both client AND admin, but form "field" components (defined for the app as input plus any labels or other elements + styling) are distinct to client OR admin. Fields always include inputs, but inputs can be used standalone.

So for example, a [`TextInput`](https://github.com/DoSomething/northstar/pull/1279/files#diff-689d63f6002fc73508ad42548a3129c154e72e6a74d12f92cf94ed9c69baa54d) is a standalone `input` element with expected functionality for adding to a React-based form. It can be used in either the Client UI or the Admin UI since it inherently has no styling but can be styled as desired.

Then we have a [`TextField`](https://github.com/DoSomething/northstar/pull/1279/files#diff-cdbe29a2b64022e5df59ac54da2fda4cc412ac67afe2ed811ac9bb056883db48) for the Admin UI which imports the `TextInput` but also contains optional `label` element, optional description text and will showcase errors if there is an issue with the field when the form is submitted. This is all styled specifically for the Admin UI.

Not created in this PR, we would also have a `TextField` for the Client UI that also imports the `TextInput` but will add its own additional elements and styling as necessary for that interface's design.

The hope is that this pattern will make it easier to develop pages with forms faster over time as more of these components are developed; unifying functionality while maintaining customizability.

_Note: I'll be updating the Docs in an upcoming PR to describe this pattern along with providing examples, etc._

### Relevant tickets

References [Pivotal #179271621](https://www.pivotaltracker.com/story/show/179271621).

### Checklist


tailwind.config.js
```diff
@@ -181,6 +182,19 @@ module.exports = {
...
-  plugins: [],
+  variants: {
+    backgroundColor: ['responsive', 'active', 'hover', 'focus', 'checked'],
+    borderColor: ['responsive', 'active', 'hover', 'focus', 'checked'],
+  },
+  plugins: [
+    plugin(function({ addUtilities }) {
+      const newUtilities = {
+        '.max-chars': {
+          maxWidth: '75ch',
```
weerd, line 193: New utility to help wrap text at a max 75 characters. Helps with readability!

tailwind.config.js
```diff
@@ -181,6 +182,19 @@ module.exports = {
     // Avoid conflict with Forge's 'container' pattern:
     container: false,
   },
-  variants: {},
-  plugins: [],
+  variants: {
+    backgroundColor: ['responsive', 'active', 'hover', 'focus', 'checked'],
+    borderColor: ['responsive', 'active', 'hover', 'focus', 'checked'],
```
weerd, line 187: Specifically added for the `checked` variant so I can override default styles on radio buttons (and eventually checkboxes) using Tailwind classes.

resources/js/Admin/Imports/MutePromotions/Create.js
```diff
@@ -51,20 +26,7 @@ const Create = () => {
...
-
-            <CloseButton
-              className="absolute p-3 right-0 top-0 text-white"
-              color="#fff"
-              callback={() => setWithFlashMessage(false)}
-            />
-          </div>
-        </div>
-      )}
+      <FlashMessage messages={flash} />
```
weerd, line 29: Now a component! 🎉

resources/js/Admin/Imports/MutePromotions/Create.js
```diff
@@ -83,53 +45,16 @@ const Create = () => {
...
+        <FileUploadField
+          accept=".csv,.txt"
+          className="max-w-2xl"
+          clearErrors={clearErrors}
+          file={data.file}
+          error={errors.file}
+          id="file"
+          progress={progress}
+          setData={setData}
+        />
```
weerd, line 57: Booya! All that boilerplate reduced to a reusable file upload field component 🕺

resources/js/Admin/Imports/MutePromotions/Create.js
```diff
@@ -74,7 +36,7 @@ const Create = () => {
         Create Import
       </h2>
 
-      <div className="col-span-full mb-10" style={{ maxWidth: '75ch' }}>
+      <div className="col-span-full max-chars mb-10">
```
weerd, line 39: Turned this into a Tailwind class since I was using it in a number of places ✨

resources/js/Admin/shared/FlashMessage/FlashMessage.js
```diff
@@ -0,0 +1,65 @@
...
+  const [flashMessages, setFlashMessages] = useState({ ...messages });
+
+  useEffect(() => {
+    setFlashMessages({ ...messages });
+  }, [messages]);
+
+  return (
+    flashMessages && (
+      <div className="absolute top-0 p-4 right-0 w-1/2">
+        {Object.keys(flashMessages).map((type, index) => {
```
weerd, line 17: The `flash` object is highly unlikely to be populated by messages for all three properties, but it could be possible to have a success or error message along with an info message in some scenario (or if other properties get added in the future). To allow for this I mapped over the object keys and thus allowing for possibility of multiple messages that can then be easily dismissed.

app/Jobs/Imports/ImportEmailSubscriptions.php
```diff
@@ -149,7 +149,7 @@ private function sendPasswordReset($user)
 
         $logParams['type'] = $topicResetType;
 
-        if ($topicResetEnabled !== 'true') {
+        if (!$topicResetEnabled) {
```
weerd, line 152: Changing the values in the [`config/import.php`](https://github.com/DoSomething/northstar/pull/1279/files#diff-9b562cc8788b97ff51621be1f1a8472a703fb7230296eee04aa93a1abef97d78) from strings to proper booleans (which helped with the frontend JS), it was causing an error with this test. But after the update, the logic can now be simplified since just checking truthyness!

config/import.php
```diff
@@ -15,7 +15,7 @@
                 'reset' => [
                     'enabled' => env(
                         'COMMUNITY_SUBSCRIPTION_RESET_ENABLED',
-                        'true',
+                        true,
```
weerd, line 18: Not sure why all of these were string `'true'` instead of boolean `true`, but it was making life more difficult 😖

weerd:
One thing to note that I would like to update is that the "input" components currently require `setData` and `clearErrors` props, which are specific to InertiaJS. I would like to potentially create a couple wrapper helper functions so that these inputs can be used in components talking to InertiaJS Form helpers or just React using `setState()`.

resources/js/Admin/Imports/EmailSubscriptions/Create.js
```diff
@@ -0,0 +1,123 @@
...
+
+        return {
+          description: `New users ${action} automatically receive a ${topic} (${type}) email.`,
+          id: topic,
+        };
+      });
+    }
+
+    setRadioOptions(getRadioOptions());
+  }, [importConfig]);
```
DFurnes, line 33: Is this something we could just derive from the `importConfig` prop in the render process? It looks like, given when this `useEffect` hook runs, we might be able to avoid needing to set/store this in state!

DFurnes (commented):
Really nice stuff!! 🎉

resources/js/Admin/Imports/EmailSubscriptions/Create.js
weerd: Ah ya, I had it like that originally and noticed it was essentially deriving it on _every_ render which didn't seem ideal. 

Essentially what I wanted was something similar to `componentDidMount()`. Based on the React docs, the way to go about this is with `useEffect()` and passing the prop within an array as the second argument; this tells React to just run the useEffect on mount and when that specific prop changes... since it never does it should only run once.

I could maybe add a comment to help clarify this 🤔

resources/js/Admin/shared/Form/RadioOptionsField.js
```diff
@@ -0,0 +1,99 @@
...
+                </small>
+              )}
+            </div>
+          </div>
+        );
+      })}
+
+      {error && (
+        <small className="absolute block mt-2 text-red-500">
+          A subscription topic selection is required.
```
katiecrane, line 65: I noticed in the `TextField` component the error itself is printed out rather than a hardcoded string so wanted to ask about this one in case you missed it - but it might just be like this for now which is fine!

resources/js/Admin/shared/FlashMessage/FlashMessage.js
```diff
@@ -0,0 +1,65 @@
...
+          return (
+            flashMessages[type] && (
+              <div
+                className={classnames(
+                  'mb-6 p-4 pr-16 relative rounded shadow-md text-white',
+                  {
+                    'bg-red-300': type === 'error',
+                    'bg-blurple-100': type === 'info',
+                    'bg-green-300': type === 'success',
```
katiecrane, line 26: I'm having trouble figuring out where this type gets set - when you submit an import the message comes up blurple but I would have expected that to be a green "success" message

katiecrane (approved):
Couple of questions from me but nothing blocking!

resources/js/Admin/shared/Form/RadioOptionsField.js
weerd: Oh ya, one nice thing about this approach is we can decide to customize the default message. So for this one, to make it more readable for the Admin UI, instead of saying the default of "The topic field is required." which sounded weird to me given it is a list of radio button to select from, I decided it was a nicer UX with a custom message.

We could also do this on the Laravel side of things, typically using a `FormRequest` class for this form, etc, but that felt like overkill.

resources/js/Admin/shared/FlashMessage/FlashMessage.js
weerd: Ah! Yep, this is a flash message sent along with the redirect response from the [EmailSubscriptionImportController.php](https://github.com/DoSomething/northstar/pull/1279/files#diff-ee05b614df896a59c5cb5f39be39bdd0b6c049e76715d40949762d5727a62734R98-R101). It's the first param of the `with()` method. Then [Inertia's middleware](https://github.com/DoSomething/northstar/blob/main/app/Http/Middleware/HandleInertiaRequests.php#L41-L47) catches the flash session data and "shares" it to the page as a prop we can pull out. That data is passed to the `<FlashMessage />` component as the `messages` prop. 

Also, now that you mention it, I'm debating if maybe `success` is the better one to use? I used `info` because that's what it was before in Chompy, but maybe it should be changed. Technically this just adds the request to the job queue and doesn't officially import it until the job runs, so maybe `info` is the correct one 🤷‍♂️

resources/js/Admin/shared/Form/RadioOptionsField.js
weerd: I can add a comment if it helps, to clarify use of a custom message here!

katiecrane: Ah, I was more thinking that this component isn't specific to that topic form but is reusable? I totally agree that this message is more helpful than the default though!

resources/js/Admin/shared/FlashMessage/FlashMessage.js
katiecrane: AH thanks for that, I totally missed it on the first pass and then when trying to find it later was searching for "flash." 

Ooooo yeah I guess it successfully added it to the queue but don't want to overstate the success! I don't think there's a wrong choice between the two!

weerd: Part of the logic for this that established how Inertia JS "shares" and makes available the `flash` messages as an object was established in a [prior PR](https://github.com/DoSomething/northstar/pull/1271/files#diff-83c0cdfc438586e614726fc44d098cd76e7605af4f067c21ceaaecae6e2051d1) which is why I think there might have been a disconnect!

resources/js/Admin/shared/Form/RadioOptionsField.js
weerd: Oh snap! That is totally correct! Nice catch @katiecrane 🙌  I'll update it to allow passing a custom error message or use default if none provided!

resources/js/Admin/shared/FlashMessage/FlashMessage.js
katiecrane: 💡 thanks!

katiecrane:
@weerd love it and I really like the rename so the context is more obvious!

resources/js/Admin/Imports/EmailSubscriptions/Create.js
```diff
@@ -0,0 +1,116 @@
...
+      const action = topics[topic].reset.enabled ? 'will' : 'will not';
+
+      const type = topics[topic].reset.type;
+
+      return {
+        description: `New users ${action} automatically receive a ${topic} (${type}) email.`,
+        id: topic,
+      };
+    });
+  }, [importConfig]);
```
weerd, line 26: Thanks to @DFurnes's suggestion for looking into using `useMemo()`! It serves the same idea as I was trying to do with `useEffect()` but more appropriately and in a simpler way that's easier to digest!

For context, I was using `useEffect()` before and passing an array of props, so that it would only update if the `importConfig` prop ever changed, thus avoiding building the radio options object on every render (typing a letter in text field causes a render... and so does each subsequent letter typed); essentially only running the object build once when the component mounted. But `useEffect()` is meant to handle side effects so not actually ideal for this scenario where building the radio option object always returns the same data (no side effects). The setup is also a bit more involved and could be confusing to someone new reviewing the code if they see it's handling a prop that doesn't change.

This is where `useMemo()` helps, because it will run the function and cache the result to reuse it when the function is called instead of recalculating. It will only recalculate and build the radio options array if the `importConfig` prop ever changes, which it doesn't so it only runs once! 🕺

resources/js/Admin/Imports/EmailSubscriptions/Create.js
weerd: ok! [updated here](https://github.com/DoSomething/northstar/pull/1279#discussion_r715066546)!

PR #1283, opened 2021-09-22
Admin Dashboard ✨

weerd:
### What's this PR do?

This pull request updates the DS Admin homepage to start utilizing the new Admin on some other pages and give the entry point a new coat of paint! 🖌️ 

**Admin Dashboard Login page**
Admin Homepage Dashboard (login page)

**Admin Dashboard Logged in as Staff**
Admin Homepage Dashboard (logged in as Staff)

There's lots we could do with the Dashboard to provide a nice overview of things when staff/admin signs in, but for now I just re-purposed what we had in the existing page. More to come! 🔮

👀 

One of the things I really like about Laravel Blade is the convenient directives you can use as quick conditionals to decide whether or not to output some markup/content. We utilized the `@auth` and `@guest` directives on the `home.blade.php` page to show or hide elements based on whether the user is signed in as staff or is a guest to the page.

Now that we're switching to using Inertia JS and React to handle rendering our pages, we don't have access to those directives, so I created a little helper called `getAuthStatus()` to essentially provide the same "feel" when building components in React and needing to check against whether a particular user should be able to "see" something 👁️ .

You can either pass it a `user` object or nothing, in which case it grabs `auth.user` info that Inertia JS has conveniently shared with us and can be nabbed via the `usePage()` hook. In your React component, you can then use destructuring to pull out the items you need to use in the template; they're all just booleans so they're easy to use in the component.

Here are the properties currently available:

- `isAdmin`
- `isAuthenticated`
- `isGuest`
- `isMember`
- `isStaff`

Overall I think this just helps with readability and avoiding easy typos or mistakes! Instead of having to check against a user's role each time to output something:

```js
// Some React component
...
{ auth.user.role === 'staff' && <p>Output some secret stuff here!</p> }
```

You can just use the simpler and less prone to errors helper property!

```js
// Some React component
...
{ isStaff && <p>Output some secret stuff here!</p> }
```

Huzzah! 😌  🎉  

### Relevant tickets

### Checklist


app/Http/Middleware/HandleInertiaRequests.php
```diff
@@ -45,6 +45,14 @@ public function share(Request $request)
...
                 ];
             },
+
+            'auth.user' => function () use ($request) {
+                return $request->user()
+                    ? $request
+                        ->user()
+                        ->only('id', 'first_name', 'email', 'role')
+                    : null;
+            },
```
mendelB, line 55: This is dope! I know we're typically careful about exposing `email` unless explicitly requested by the client -- I suppose here it's just for code purposes? I didn't see where we utilize this property yet though.

resources/js/Admin/Dashboard.js
```diff
@@ -0,0 +1,185 @@
...
+              <a
+                className="font-bold text-blurple-500 hover:text-blurple-200"
+                href="/admin/campaigns"
+              >
+                Campaigns:
+              </a>{' '}
+              <span className="text-gray-500">
+                create/edit campaigns, actions, signups, and reportbacks.
+              </span>
+            </li>
```
mendelB, line 45: This feels like it can be extracted into a `map` or even its own utility component in the future!

mendelB (approved):
🤩 Awesome!

app/Http/Middleware/HandleInertiaRequests.php
weerd: Hmmm, that is a good call. So `email` I'm not using yet, but I could see it being used in the Admin UI at some point. 

So technically, this is specific to the current authenticated user, so "ideally" the person using the browser and thus their own email. Is the concern if user leaves browser open somewhere? I forget what our call has been in Phoenix with something like this 🤔  I could remove it for now too since not using it specifically...

resources/js/Admin/Dashboard.js
weerd: Oh ya, totes! I left it manually added like this for now since not even sure if it's going to stay 😅 , but didn't want the page to be empty! I'd rather in future populate it w/ some quick cached stats on things for staff at a glance. Maybe throw in a chart or two... who doesn't ❤️  a nice chart! 📈

app/Http/Middleware/HandleInertiaRequests.php
mendelB: Ooh nah that's a really good point! We don't do anything special to obscure the user's own email from them in their [profile](
https://github.com/DoSomething/phoenix-next/blob/781709dce81436edd2660d76d16dc6f1ca349a5e/resources/assets/components/pages/AccountPage/Account/AccountQuery.js#L19), so the same should apply here.

resources/js/Admin/Dashboard.js
mendelB: haha heck yeah. Hashtag bring the admin bug to Northstar!

resources/js/helpers/auth.js
```diff
@@ -0,0 +1,24 @@
...
+import { usePage } from '@inertiajs/inertia-react';
+
+/**
+ * Get authentication status details for the provided or current user.
+ *
+ * @param   {Object|null} user
+ * @returns {Object}
+ */
+export const getAuthStatus = (user = null) => {
+  const { auth } = usePage().props;
```
DFurnes, line 11: Should this be called `useAuthStatus` to hint that it has a hook inside it (since hooks have [special rules](https://reactjs.org/docs/hooks-rules.html))?

DFurnes (commented):
Looks so nice!! Left one post-merge suggestion.

resources/js/helpers/auth.js
weerd: Ah ya, nice suggestion! Will update in upcoming PR 👍

PR #1287, opened 2021-10-05
Adding Jest and Testing-Library 🚦

weerd:
### What's this PR do?

This pull request updates the project to include both [Jest](https://jestjs.io/) and [Testing-Library](https://testing-library.com/) packages to help test our JavaScript code and React components!

I've been doing a bunch of work with the new DS Admin UI, creating new components to help build out the UI and felt it was about time to start adding tests for the components so we can feel a bit more confident as we continue to build out the Admin UI and re-use all these components.

I only added a single test file for now,  since this PR is mostly about getting Jest and Testing-Library working correctly! But more test will be coming now that this is all set up 🚥 

😌  will help us feel more relieved if we have tests to backup our code.

### Relevant tickets

References [Pivotal #]().

### Checklist


.gitignore
```diff
@@ -1,4 +1,5 @@
 /build
+/coverage
```
weerd, line 2: I initially committed all these files but felt like a lot, and would result in version changes/updates to the files in this directory every time we added a new test file. If we eventually want to link to the generated `index.html` to showcase our coverage, we can just generate these files somewhere and link to the index.

jest.config.js
```diff
@@ -0,0 +1,194 @@
...
+  setupFilesAfterEnv: ['./jest.setup.js'],
+
+  // The number of seconds after which a test is considered as slow and reported as such in the results.
+  // slowTestThreshold: 5,
+
+  // A list of paths to snapshot serializer modules Jest should use for snapshot testing
+  // snapshotSerializers: [],
+
+  // The test environment that will be used for testing
+  testEnvironment: 'jsdom',
```
weerd, line 140: I meant to ask @DFurnes, I referred a bit to the Phoenix setup. In there we use both `jest-environment-jsdom` and `jest-environment-jsdom-global`, but it didn't seem like something that needed to be included for the simple test created so far. May be it's something we'll need in the future? Seems like it just helps expose `jsdom` to allow messing with it in tests (if I understand correctly?)...

.gitignore
katiecrane: Can you tell me more about what lives in here? It describes how much test coverage we have?

katiecrane (approved):
We love tests!!

jest.config.js
DFurnes: As far as I can remember, we needed `jest-environment-jsdom` to provide that browser-like jsdom environment for tests (which is maybe included in Jest by default now?), and then yeah [`jest-environment-jsdom-global`](https://github.com/simon360/jest-environment-jsdom-global) helped us out in a instances [like this](https://github.com/DoSomething/phoenix-next/blob/e2fe77338ff7f3844a62e813dce54e4b7c7ee73b/resources/assets/components/utilities/DismissableElement/DismissableElement.test.js#L48-L51) where we needed to customize the "browser" environment inside a test.

Definitely not something we need to include here until we run into similar use cases!

weerd: Ya, that sounds good. Thanks for the context @DFurnes 👍

PR #1288, opened 2021-10-07
Adding Storybook to help visualize our components

weerd:
### What's this PR do?

This pull request adds [Storybook](https://storybook.js.org/) to Northstar! This tool will allow us to visualize and build our React components in isolation. It will also provide a nice _visual_ documentation of our components to help our dev/product team easily familiarize themselves with the components available, how they look and function and what options they support. ✨ 

image
Nothing too fancy yet, but this should grow significantly as we add more of our components!

👀 

This is the initial setup; I wanted to get something up and running and add a couple components to test it all out. Eventually I plan on removing the example Storybook components, but felt they would be useful to have for the time being as a quick reference.

Additionally, we could eventually build and expose the storybook as a website so it is publicly accessible. For the time being, as we flesh it out, after installing the NPM dependencies, run:

```shell
$ npm run storybook
```

The Storybook build process will execute and then if everything works as it should, it will pop open a browser and load up a local version of the Storybook app with our library of components (2 for now 😅 )

_Note: When reviewing, you can ignore the files in the root `stories/` directory. Storybook added these and are used for the UI. They have some pages and example components/stories in there. I've chosen to keep our story files along with the directory you find our components in, because it makes more sense to me to have them in the same directory and make them easier to find; like we do with our JS test files!_

### Relevant tickets

References [Pivotal #]().

### Checklist


resources/js/Admin/components/Form/FileUploadField.js
```diff
@@ -95,15 +98,45 @@ const FileUploadField = ({
 };
 
 FileUploadField.propTypes = {
+  /**
+   * What file types should the input accept?
+   */
```
weerd, line 103: One thing I'm digging about Storybook, is that it can pull code comments from the `PropTypes` to use as descriptions in the UI, which is really nice. It forces you to think about how to describe the property better for others to use and can help with understanding when just the prop name alone might not be super clear!

mendelB: ooh this feels like a really helpful approach!

PR #1291, opened 2021-10-22
Admin Search Component 🔍

weerd:
### What's this PR do?

This pull request updates the Admin Club Index page to allow making requests to the backend to search for/filter list of clubs to a specific result based on the search while typing in the search input.

Here's a quick video of the functionality on Clubs and Redirects index pages:

https://user-images.githubusercontent.com/105849/139491987-b8c2ac4f-8dbb-4fbb-8cc0-fd479a2b6e90.mp4

We previously allowed for this in the old interface but did not _debounce_ the calls to GraphQL so as staff typed in the field, new requests would fire off for each keystroke. I don't recall well if this did not matter as much because of GraphQL caching, but for this new Admin interface we're hitting the backend directly so I definitely wanted to do some debouncing.

Took a bit but ended up figuring out a nice solution with a custom hook, thanks to a bit of googling around. It works really nice ✨  and I've also accounted for ensuring the input is disabled when it is fetching so there's no weird user experience where a user could continue typing in the interim while the search request is executing, and when it returns it changes the input text to match what that query was (this all happens super fast and I noticed this behavior while testing it, so made sense to address it).

👀 

I'm still considering a couple changes to see if I can fully encapsulate the behavior inside a single component, but I think that means having a `SearchForm` component... I'll update here if I proceed with that direction.

### Relevant tickets

### Checklist


weerd:
Another thing I realized while building, was that I separated out and unified enough of the logic/structure for the `SearchField` and `FilterField` where they are essentially the same with just a different icon. So I may try and combine them into a single component with a prop you can pass to designate which icon (for type of search it's doing) to show.

I'll leave that for now as a future update along with the idea of providing a `SearchForm` component so all the logic is contained and makes the `Index.js` pages simpler.

katiecrane (approved):
This looks awesome! I love the debounce and how you extracted that so we can use it anywhere.

PR #1293, opened 2021-11-03
Updating CircleCI configuration to specify using Node v14.x

weerd:
### What's this PR do?

This pull request adds "unfinished" Storybook story for the `SearchField` component. For the most part this PR included some minor updates so that we could test CircleCI's build which was failing due to the docker image having it's Node version bumped to v16.x which was causing problems for some of our dependencies.

The main update is that it adds a run step to the CircleCI config to "downgrade" the Node version on the box from 16.x to 14.x and use that as the default! This seems to have fixed the issues with the box error-ing out.

👀 

You can find more background in the associated Pivotal Ticket below.

### Relevant tickets

References [Pivotal #180172620](https://www.pivotaltracker.com/story/show/180172620).

### Checklist


weerd:
Trying to see if this PR will fail CircleCI due to whatever changes they made to the build process and upgrade Node version.

weerd:
Yep, same error with CircleCI on Phoenix regarding Node version. Confirmed!

.circleci/config.yml
```diff
@@ -17,6 +17,26 @@ jobs:
...
+            nvm alias default 14.18.1
+            echo 'export NVM_DIR="$HOME/.nvm"' >> $BASH_ENV
+            echo '[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"' >> $BASH_ENV
+      # Report the versions of different software we use and depend so we can see it at a glance.
+      - run:
+          name: Report software versions
+          command: |
+            php --version
+            node --version
+            npm --version
```
weerd, line 39: Added this to help us easily see at a glance what versions of software the box is using, so we don't need to SSH in and ask it directly!

package.json
```diff
@@ -44,6 +44,10 @@
     "url": "git+https://github.com/DoSomething/northstar.git"
   },
   "author": "David Furnes <dfurnes@dosomething.org>",
+  "contributors": [
+    "Diego Lorenzo <dlorenzo@dosomething.org>",
+    "Katie Crane <kcrane@dosomething.org>"
+  ],
```
weerd, line 50: Figured I'd add us :)

katiecrane (approved):
take that circleci!!!

PR #1295, opened 2021-11-15
Adding new Admin tabbed nav component.

weerd:
### What's this PR do?

This pull request makes a few updates and adds functionality for a new Admin Tabbed Nav component that enables tabbing between "internal" pages for a section of the Admin.

This will end up initially being used on the Campaigns Index page to allow for tabbing between the all campaigns search, open campaigns, or closed campaigns pages.

https://user-images.githubusercontent.com/105849/141847587-c1f26b00-f9cf-46a9-b588-85778234433b.mp4

👀 

There were a couple tricky spots for the logic on this. I think it could be updated in the future as we try this component out on more pages and see how it evolves.

It did provide an opportunity to create a more generic `NavLink` component that I plan to replace the old version with. This newer version is more versatile and can account for both this tabbed nav and the main nav.

### Relevant tickets

### Checklist


weerd:
I meant to note, you pass on object of link data as props to the component when calling it. Here's an example that I use for the Campaign Index page:

```js
// Links for internal tabbed page nav
  const tabbedNavLinks = [
    {
      isDefault: true,
      optionalParam: 'search',
      title: 'Search',
      routeName: 'admin.campaigns.index',
    },
    {
      title: 'Open Campaigns',
      routeName: 'admin.campaigns.index',
      routeParams: { status: 'open' },
    },
    {
      title: 'Closed Campaigns',
      routeName: 'admin.campaigns.index',
      routeParams: { status: 'closed' },
    },
  ];
```

The `isDefault` property is used to indicate that this is the default tab when the index page loads without any query params. The `optionalParam` property is used to indicate a check against a param that is optional, like `?search=` and could have varying levels of values (the query in question).

Since the Search page is the default tab, I needed to be able to set it as the active tab if either the URL route had no params at all (main index page), or if there was a search submitted `?search=recycling`.

resources/js/helpers/url.js
```diff
@@ -1,14 +1,46 @@
...
+  routeName: name,
+  routeParams: params = {},
+  isDefault = false,
+}) => {
+  const windowUrl = new URL(window.location.href, window.location.origin);
+
+  const tabUrl = new URL(route(name, params), window.location.origin);
+
+  if (tabUrl.search) {
+    return windowUrl.href.includes(tabUrl.search);
```
katiecrane, line 22: What is the `search` property that this is accessing? I'm a little unsure of what this is checking for

weerd: Oh! Ya, I could see how this is maybe confusing. The `new URL()` returns an object based on what you provided, and if the string url you passed in contained query params, they get added to a property called `search` on the object it returns.

It would be easier to visualize, so here's an example with no query params:
image

And another with query params:
image

katiecrane: AH got it - so this part is seeing if the query params for the given tab match the actual query params on the current URL?

weerd: Essentially, yep!

katiecrane: Thanks!

katiecrane (approved):
This looks great!

PR #1316, opened 2022-02-24
Admin UI select component tests

weerd:
### What's this PR do?

This pull request adds some Jest tests for the SelectInput component.

👀 

I had a wonky bug with Jest, where it seemed to want to look inside the composer (PHP) `/vendor` directory for module code, and some packages that do work with some JavaScript have their own tests and was causing Jest to throw errors. This wasn't an issue before, but started happening after a few node modules were updated and some PHP composer packages were updated. Ultimately, the best way to resolve this was to update the Jest configuration file and tell Jest to ignore the `/vendor` directory when searching for module code.

### Relevant tickets

### Checklist


jest.config.js
```diff
@@ -87,7 +87,7 @@ module.exports = {
   },
 
   // An array of regexp pattern strings, matched against all module paths before considered 'visible' to the module loader
-  // modulePathIgnorePatterns: [],
+  modulePathIgnorePatterns: ['<rootDir>/vendor'],
```
weerd, line 90: Ignore the any potential module code within the `/vendor` directory!

package.json
```diff
@@ -123,5 +122,8 @@
     "webpack": "^4.46.0",
     "webpack-bundle-analyzer": "^3.6.0",
     "webpack-cli": "^3.3.11"
+  },
+  "devDependencies": {
+    "@testing-library/user-event": "^13.5.0"
```
weerd, line 127: I'm going to start moving dev packages back into the `devDependencies` section. Took it out a while back, but I think it's best to retain this standard, so it's more obvious.

resources/js/components/FormInput/SelectInput.test.js
```diff
@@ -0,0 +1,101 @@
...
+
+beforeEach(() => {
+  clearErrorsCallback = jest.fn();
+
+  setDataCallback = jest.fn();
+});
+
+describe('SelectInput component', () => {
+  /** @test */
+  it('renders a select input element', () => {
```
weerd, line 25: I'm going to switch JS tests to use the `it()` alias instead of `test()` so it all reads a bit nicer as a sentence. Also since I'm already adding `/** @test */` above each test as an helpful visual indicator between tests, it seems weird to have both that and `test(..)` right underneath it.

package.json
katiecrane: I'm into this!!

weerd: Ya, gonna move the rest that should be in there in upcoming PRs. :)

resources/js/components/FormInput/SelectInput.test.js
katiecrane: LOVE the visual indicator, sometimes it is hard to easily tell where the next one starts!

katiecrane (approved):
woohoo!! looks great!

